PR #2021, opened 2014-10-02
ENH: Added censored data analysis module with RobustROS fitter and tests

phobson:
## Preface

I'm not sure that this fits here. I originally had the go-ahead to add it to `lifelines`, but its license got funny and I'd prefer to not get into that mess.
## Overview

In the environment, we can nearly never say that a certain parameter doesn't exist. Instead, we say that it was not detected, and as a result report a detection limit that corresponds to the lowest possible quantifiable value of the instrument and/or measurement method. You can think about this as measuring the diameter of hair with a ruler. You know the hairs have diameters, and you know they're <1 mm, but that's about all you can say.

So trying to compute basic stats about these data sets can be tricky. Luckily Dr. Helsel wrote a couple of text books about how to deal with this. [Here is the one I used to write this code](http://www.amazon.com/Nondetects-Data-Analysis-Statistics-Environmental/dp/0471671738).

Also [here's a overview paper](http://training.fws.gov/resources/course-resources/pesticides/Monitoring/USGS%20Nondetects%20water%20sampling%20bias.pdf) and an [R-package](http://cran.r-project.org/web/packages/NADA/index.html) written by I believe a colleague of Helsel.
## Tests

I got a lot of them. Coverage was 100% on my local machine. A lot of them are redundant, but check things like what should happen when >80% of values are censored, or if all but 1 values are censored, etc.

Most of the tests use known values that come from either text book or papers written by Helsel, or datasets and results that come from the R package.
## Future work

There's a lot more in the R-package that could easily be implemented in python. This is just the start.

### DIFF
--- a/statsmodels/stats/robustros.py
+++ b/statsmodels/stats/robustros.py
@@ -154,6 +154,8 @@ class RobustROSEstimator(object):
         self._result_df = data.copy()
         self._cohn = None
         self._estimated_values = None
+        if not lazy:
+            self.estimate()
 
     @property
     def cohn(self):


statsmodels/censored/tests/test_robustros.py
```diff
@@ -0,0 +1,428 @@
...
+
+    def test_transform_in(self):
+        ntools.assert_equal(self.ros.transform_in, np.log)
+
+    def test_transform_out(self):
+        ntools.assert_equal(self.ros.transform_out, np.exp)
+
+
+
+class test_ROS_Helsel_Arsenic(_base_ROS_Mixin):
```
josef-pkt: pep-8 class names are capitalized and without underline `TestROSHelselArsenic`
base class with tests are usually called `CheckXxxMixin`, e.g.  `CheckROSMixin`

statsmodels/censored/robustros.py
```diff
@@ -0,0 +1,462 @@
...
+    (2005) to estimate the left-censored (non-detect) values of a
+    dataset.
+
+    Parameters
+    ----------
+    data : pandas DataFrame
+        The censored dataset for which the non-detect values need to be
+        estimated.
+
+    result_col : optional string (default='res')
```
josef-pkt: numpy doc standard, no space between items in Parameters and in Returns lists

statsmodels/censored/robustros.py
```diff
@@ -0,0 +1,462 @@
...
+        `True` implies Left-censorship. `False` -> uncensored.
+
+    transform_in : optional function or None (default = np.log)
+        By default, lognormality is assumed and the linear regression
+        is performed on log-transformed data. An arbitrary transformation
+        function can be passed using this parameter. If `None` is passed,
+        the data will remain untransformed.
+
+    transform_out : optional function or None (default = np.exp)
+        This is used after the linear regression to revert the trasformation
```
josef-pkt: typo transformation: missing `n` in trans

statsmodels/censored/robustros.py
```diff
@@ -0,0 +1,462 @@
...
+        performed by `transform_in`. If `None` is passed, the data will remain
+        untransformed.
+
+    Attributes
+    ----------
+    N_obs : int
+        Total number of results in the dataset
+
+    N_cen : int
```
josef-pkt: lower case `n`
we don't use underline in `nobs` number of obserations

statsmodels/censored/robustros.py
```diff
@@ -0,0 +1,462 @@
...
+    statistics of the dataset as a whole.
+
+    '''
+    def __init__(self, data, result_col='res', censorship_col='cen',
+                 transform_in=np.log, transform_out=np.exp):
+
+        # confirm a datatype real quick
+
+        if not isinstance(data, pd.DataFrame):
+            raise ValueError("Input `data` must be a pandas DataFrame")
```
josef-pkt: why does it need to be a DataFrame? Is there extra structure required on the DataFrame, or could we just internally convert if data is an ndarray?

statsmodels/censored/robustros.py
```diff
@@ -0,0 +1,462 @@
...
+            censorship_col=self.censorship_col
+        )
+
+        # create a dataframe of detection limits and their parameters
+        # used in the ROS estimation
+        self.cohn = self._get_cohn_numbers()
+
+    def _get_cohn_numbers(self):
+        '''
+        Computes the Cohn numbers for the delection limits in the dataset.
```
josef-pkt: de`l`ection ?

statsmodels/censored/robustros.py
```diff
@@ -0,0 +1,462 @@
...
+        '''
+
+        def _A(row):
```
josef-pkt: do the A, B, C numbers have an English name or meaning?

statsmodels/censored/robustros.py
```diff
@@ -0,0 +1,462 @@
...
+            # index of results above the lower DL
+            above = self.data[self.result_col] >= row['lower']
+
+            # index of results below the upper DL
+            below = self.data[self.result_col] < row['upper']
+
+            # index of non-detect results
+            detect = self.data[self.censorship_col] == False
+
+            # return the number of results where all condictions are True
```
josef-pkt: typo extra c in condictions

statsmodels/censored/robustros.py
phobson: :heavy_check_mark:

statsmodels/censored/robustros.py
phobson: :heavy_check_mark: (got `N_cen` too )

statsmodels/censored/robustros.py
phobson: :heavy_check_mark:

statsmodels/censored/tests/test_robustros.py
phobson: :heavy_check_mark:

statsmodels/censored/robustros.py
phobson: Excellent point. I think we could easily handle three cases:

 1. `data` is a DataFrame and the column names are specified
 2. `data` is an array and the column indexes are specified
 3. `data` is `None` and 1-D arrays are fed into `result_col` and `censorship_col`

phobson:
@josef-pkt as for the `_A`,  `_B`, and `_C` helpers, I don't think that they have more descriptive names. I'll check the literature tomorrow.

statsmodels/censored/robustros.py
```diff
@@ -0,0 +1,449 @@
...
+    data : pandas DataFrame
+        An expanded version of the original dataset `data` passed the
+        constructor included in the `modeled` column.
+    debug : pandas DataFrame
+        A full version of the `data` DataFrame that includes preliminary
+        quantities.
+
+    Example
+    -------
+    >>> from lifelines.estimation import RobustROS
```
josef-pkt: need to update import in example

phobson: :heavy_check_mark:

statsmodels/censored/robustros.py
```diff
@@ -0,0 +1,449 @@
...
+            data,
+            result_col=self.result_col,
+            censorship_col=self.censorship_col
+        )
+
+        # create a dataframe of detection limits and their parameters
+        # used in the ROS estimation
+        self.cohn = self._get_cohn_numbers()
+
+    def _get_cohn_numbers(self):
```
phobson: @josef-pkt 

Check the literature and the NADA documentation, A, B, C are described as:

```python
# The Cohn Numbers are:
# A_j   = the number of uncensored obs above the jth threshold.
# B_j   = the number of observations (cen & uncen) below the jth threshold.
# C_j   = the number of censored observations at the jth threshold
# P_j   = the probability of exceeding the jth threshold
```

We could probably just document them as such, or rename name.

So we could try: 
`A` :arrow_right: `_uncensored_above`
`B` :arrow_right: `_obs_below`
`C` :arrow_right: `_censored_equal`
`P` :arrow_right: `_exceedance_prob`

Open to suggestions, though.

josef-pkt: I always like informative names, names we where we need to read the definition once and we have a good chance of remembering it a month later.

I'm in favor of the renaming, but your proposed names are missing a "noun",  for example A, B, C are `nobs_xxx`, P is a `prob_xxx`, no leading underscore if they are public.

josef-pkt: or A, B, C could just be `n_xxx`    (given the rest of the name and our n, k convention it should also be clear)

phobson: ok took care of this

phobson:
constructor can take a dataframe:

`RobustROSEstimator(data=df, result='concentration', censorship='qualifier')`

or sequences:

`RobustROSEstimator(data=None, result=concList, censorship=qualArray)`

josef-pkt:
@phobson  I'm trying to figure out whether to merge this now for 0.7 or postpone to 0.8.

I don't see any problems while browsing through the code, but it doesn't look like any of our standard models. Also, we might have to rethink some of the API choices as we get more survival and censored models. But I don't see any changes right now in my brief review that would make it more "standard" in anticipation of what doesn't exist or hasn't been decided yet.

test coverage also looks good.

So back to my earlier suggestion: move it to statsmodels.stats.robustros or similar, and we'll merge for 0.7.
The interfaces and structure in stats have more variation and it will fit in as a more descriptive statistical tool, instead of being already a full model according to our pattern.

Maybe a warning in the docstring that the API might or will likely change might be useful.

Can you rebase and force push?

Also, if you have an example script to quickly try it out, then you could dump it into statsmodels.examples (inside the source tree, not the top level, and it doesn't need to be very clean.)
If you have a notebook, then we can add it also to the top level examples/notebooks which will make it public.

phobson:
@josef-pkt thanks. I'll move and rebase tonight.

phobson:
moved, added an example, and rebased @josef-pkt

statsmodels/stats/tests/test_robustros.py
```diff
@@ -0,0 +1,440 @@
...
+            0.76644378,  0.91850581,  1.10390531,  1.10000000,  2.00000000,
+            2.00000000,  2.40400000,  2.86000000,  3.00000000,  3.00000000,
+            3.70500000,  4.00000000,  5.00000000,  5.96000000,  6.00000000,
+            7.21400000, 16.00000000, 17.71600000, 25.00000000, 51.00000000
+        ])
+        self.known_plot_pos.sort()
+        self.known_modeled.sort()
+
+
+class testROSNoNDs(CheckROSMixin):
```
josef-pkt: class names are still not pep-8, should start with capital letter, and if it's composed of several nouns, each should start with a capital letter.

can you add a `#` comment with a hint to what the name means? mainly as aid in reading and browsing code.

statsmodels/stats/robustros.py
```diff
@@ -0,0 +1,477 @@
...
+                'markeredgecolor': '0.35',
+                'markerfacecolor': 'none',
+                'linestyle': 'none',
+                'label': 'Censored data',
+                'alpha': 0.70
+            }
+            raw_symbols.update(raw_kwds)
+            osm_raw, osr_raw = stats.probplot(
+                self.data[self.result], fit=False
+            )
```
josef-pkt: are the locations of closing braces and brackets because of automatic formatting or is there a style convention somewhere for this?

josef-pkt: Is there a reason for why you use scipy.stats.probplot instead of ours?
Just curious, I don't mind either way.

josef-pkt:
Thanks, rebase looks good. (I don't think squashing hurt much in this case.)

I'll try this out tomorrow or during the weekend so I have at least a rough idea what it is doing before merging. I expect at most small changes.
(The signature of the class is worrying me a bit, but I cannot tell without trying. `data` plus names as arguments is a bit an uncommon pattern but, AFAIR, we do use it in `from_formula`)

It would be useful if you added also a short announcement to the release notes.

statsmodels/stats/robustros.py
phobson: It's the last acceptable format here in PEP8's indentation section: https://www.python.org/dev/peps/pep-0008/#indentation

I'll switch over to statsmodel's probplot. I'm integrating this code directly from another project, so I never bothered with our ProbPlot class before.

josef-pkt:
It's already next week, but I just started to try out the example. I will have to do at least a bit of reading since I don't understand much.

issues so far: 
`fit` cannot be an attribute because we use it everywhere else as the estimation method.
`result` and `censorship` attributes are strings, so I guess it will be better to add post-fix `_name`
`censorship` sounds unusual, I need to check (with survival/duration), but IIRC we use `censoring` as name for the censoring indicator array.

josef-pkt:
(I'm the monkey who tries to write Shakespeare. 
Hit the buttons and see if it works without reading the docs.)

```
>>> ros2 = sm.stats.RobustROSEstimator(data=data, result='res', censorship=data['cen'])
  File "e:\josef\eclipsegworkspace\statsmodels-git\statsmodels-all-new2_py27\statsmodels\statsmodels\stats\robustros.py", line 145, in __init__
    self.ncen = data[data[self.censorship]].shape[0]
  File "C:\Programs\Python27\lib\site-packages\pandas-0.12.0-py2.7-win32.egg\pandas\core\frame.py", line 1997, in __getitem__
    return self._getitem_frame(key)
  File "C:\Programs\Python27\lib\site-packages\pandas-0.12.0-py2.7-win32.egg\pandas\core\frame.py", line 2060, in _getitem_frame
    raise ValueError('Must pass DataFrame with boolean values only')
ValueError: Must pass DataFrame with boolean values only
>>> pandas.__version__
'0.12.0'
```

josef-pkt:
`ros2 = sm.stats.RobustROSEstimator(data=None, result=data['res'], censorship=data['cen'])`
works, but maybe it should be
`ros2 = sm.stats.RobustROSEstimator(result=data['res'], censorship=data['cen'], data=None)`

as far as I can see we only use two columns or two arrays.
Also if there are extra columns in `data`, then they are not preserved in the resulting `data` attribute.
Since it's not vectorized, maybe `data` is an extra not the main argument in the call to RobustROSEstimator.

(In another example I had forgotten to call `estimate` and was surprised not to see the results in the `data` attribute.

josef-pkt:
This is a bit similar in the pattern to the imputation method in MICE.

In terms of usage, like when we want to fill in several columns with censored observations:
maybe there should be a `inplace` and `reindex` (*) option to add the new columns to the existing one, or return just the "imputed" array/Series so it can be added to the DataFrame by the user.
(*) or whatever the pandas version is keep the original order

josef-pkt:
Another thought: I think `estimate` should return `self`.
I think we discussed this somewhere at some point. It is it's own results class, and makes chaining calls possible, for example

`RobustROSEstimator(data=data, result='res', censorship='cen').estimate().data`

josef-pkt:
I was thinking of whether to rename `estimate` to fit like in our models. However, it might be easier to keep `estimate` so we have `fit` available for refactoring if we switch to a returning an instance of results class with additional features.

josef-pkt:
internal to the class:

assigning names / adding explicit attributes would shorten the code in many places

`self.censoring = self.data[self.censorship]`
`self.endog = self.data[self.results]`   # or maybe something else than `endog`

josef-pkt:
'DL' is too short and not informative
alternatives ?

josef-pkt:
confusing:
`self.data` gets assigned all the extra variables in the code, like `self.data['plot_pos']`
however, at the end it goes into `debug` which is the attribute that the user sees.

I haven't checked: preallocate `debug` to avoid having to extend columns (might not make a lot of difference, since we don't add columns so often).

josef-pkt:
code reorganization: move calculations for `self.data['plot_pos']` out of `estimate` into it's own private method.

josef-pkt:
aside: I don't have enough experience with pandas to evaluate the performance difference between the row-wise `apply` approach compared to the vectorized numpy version.
I'm not a `apply` person.

statsmodels/stats/robustros.py
```diff
@@ -0,0 +1,477 @@
...
+        groupcols = ['DLIndex', self.censorship, 'Rank']
+        rankgroups = (
+            self.data[groupcols]
+                .groupby(by=['DLIndex', self.censorship])
+        )
+        self.data['Rank'] = rankgroups.transform(lambda x: x.cumsum())
+
+        # detect/non-detect selectors
+        detect_selector = self.data[self.censorship] == False
+        nondet_selector = self.data[self.censorship] == True
```
josef-pkt: It's more standard (for me) to call this `mask`  `detect_mask` `nondet_mask`
we might want to switch to censored uncensored terminology instead of nondetect detect, but it's mainly used internally.

josef-pkt:
That's it for now. (I might look some more if I have time later today/tonight.)

josef-pkt:
for example
`self.data['modeled'] = self.data.apply(_select_modeled, axis=1)`
is just a np.where call, or should be similar using pandas instead of iterating over rows.

statsmodels/stats/robustros.py
```diff
@@ -0,0 +1,477 @@
...
+
+            # select out the final data
+            self.data['modeled'] = self.data.apply(_select_modeled, axis=1)
+
+        # create the debug attribute as a copy of the self.data attribute
+        self.debug = self.data.copy(deep=True)
+
+        # select out only the necessary columns for data
+        final_cols = ['modeled', self.result, self.censorship]
+        self.data = self.data[final_cols]
```
josef-pkt: above 3 lines:
it would be more efficient to keep the data as debug (large dataframe) and only make a copy of the selected `final_cols` on the new dataframe.

also I would use something like `_data` for the internal work (and keep it instead of debug) and assign self.data at the end `self.data = self._data[final_cols].copy(deep=True)`

statsmodels/stats/robustros.py
```diff
@@ -0,0 +1,477 @@
...
+            self.data['Zprelim'] = stats.norm.ppf(self.data['plot_pos'])
+
+            # fit a line to the logs of the detected data
+            fit = stats.linregress(
+                self.data['Zprelim'][detect_selector],
+                self.transform_in(self.data[self.result][detect_selector])
+            )
+
+            # save the fit params to an attribute
+            self.fit = fit
```
josef-pkt: `self.fit_result` or something like this `self.results_fit` to avoid using attribute `fit`

josef-pkt:
question for plot:
Is there a way to show the uncensored data in a different style (color or symbol) in the same plot, so we see the distinction between artificial (modeled) and original data?

general question:
Does this require that there is fluctuation in the censoring threshold (detection limits) to get overlap between censored and uncensored observations? I only vaguely remember that I read about this before.

josef-pkt:
summary of estimation method afaics:
assume parametric nonlinearly transformed normal model (e.g. log-normal) estimate by fitting plotting positions to sorted uncensored data. Impute based on plotting position.
assumes normal distribution of transformed variable, but would work with any location-scale distribution (with no or fixed shape parameter).

If we extend to box-cox or similar transformation to normality, then we would need to choose a transformation parameter, and we would need some goodness of fit statistics.

also, call to scipy.stats.linregress could be replaced by call to our OLS or RLM, the former wouldn't bring much, the latter would be outlier or distribution robust to some extent.

link to parametric survival models with censoring? that uses full MLE, here use of ranks/plotting positions might bring in some distributional robustness (although we use gaussian plotting position/quantiles)

josef-pkt:
@phobson I think the basic idea fits in quite well. I can think of several extensions to blow it up to a full model and connect it to possible other approaches (that we mostly don't have yet).

For now, i.e. 0.7: I think we should 
- settle on the `__init__` arguments
- fit cannot be an attribute
- return self in estimate
- maybe some renamings

all others, especially internal refactoring could wait for later since it will largely not affect the user interface.
more extensions, and a possible results class returned in a `fit` method can also wait until it's more clear where this is going.

statsmodels/stats/robustros.py
phobson: Switched to `(un)censored_mask`. Good catch.

phobson:
This I did in this last push:

 - added suffix `_name` to the `censorship` and `result` attributes
 - store the passed data into a `raw_data` attribute (see below)
 - allowed users to set the minimum count of non-censored results and max fraction of censored results that trigger the `1/2 * DL` substitution of the censored values
 - spelled out `DL` as `detection_limit` where it made sense
 - moved the plotting position calc into a separate (private) method
 - renamed `fit` to `fit_params`
 - `estimate` method returns `self`

phobson:
Up next: switching to `sm.OLS` for the regression and `sm.ProbPlot` for the plotting and other efficiencies.

Also:
> general question:
>Does this require that there is fluctuation in the censoring threshold (detection limits) to get overlap between censored and uncensored observations? I only vaguely remember that I read about this before.

No overlap is required, though, IME (water and soil contaminant data) it often exists

> it would be more efficient to keep the data as debug (large dataframe) and only make a copy of the selected final_cols on the new dataframe.

> also I would use something like _data for the internal work (and keep it instead of debug) and assign self.data at the end self.data = self._data[final_cols].copy(deep=True)

coming up next.

> for example
> `self.data['modeled'] = self.data.apply(_select_modeled, axis=1)`
> is just a np.where call, or should be similar using pandas instead of iterating over rows.

Not quite. perhaps that function could be named better, but it's selecting from multiple columns based on the value if yet another column. I'm handy with numpy, but I can't think immediately of how to do that without a loop.

#### re: the API/inputs
What do you think should be valid inputs. I personally find the `seaborn` style of passing dataframe + column names or `data=None` and arrays to be pretty convenient and flexible. But there is obviously the rest of `statsmodels` that must be considered.

#### re: separate call the `estimate`
In my mind, this was consistent with the `OLS` convention of creating an object and the calling `.fit()`. In my personal implementation of Robust ROS, the estimation was done upon creating of the object. I'd be happy either way.

#### re: loosing other (non-result, non-censorship columns)
This is intentional. In fact, sometimes I want to rip out all of the dataframe stuff and work only on arrays and only return a single array of estimated results (but that wouldn't work at all in the grand scheme of my workflows). The point is that I want to make it as hard as possible for someone to make the mistake of trying to assign a particular estimated censored value to a particular real-world observation. That's not what ROS is for.

There's probably some stuff that I'm still neglecting. Will try to get to it ASAP.

josef-pkt:
Don't squash commits, so I can look at changes instead of having to read everything again.

phobson:
Roger that.

josef-pkt:
I'm not going to read this again before 0.7. Overall it's fine and you have the warning that there will be possible changes. So, if you think it's ok, then I'll merge it as is (except matplolib see below)
I have to give up looking at new changes or I'll never finish with 0.7, so it would be one of the last few PRs going into 0.7, besides release work and updating of the documentation.

matplotlib is optional for statsmodels, Travis has now one python without matplotlib and it raises an error.
Matplotlib import needs to be try except protected. 
see #2245 as example for making matplotlib optional in other parts.

just one comment for now to the above: Your warning about using the `imputed` values only for descriptive statistics makes the analogy to MICE imputation stronger. There we do imputation essentially internal to the model, and the user mainly get's the estimation results that take the imputation into account.

phobson:
Having had some (well, a lot of) time to think about this and how it would be used. I reorganized some things. 

The `self.debug` atrribute is gone. Now we have `self._raw_data` (a direct copy of the input) and `self._result_df` (all of the intermediate and final values).

The user-facing result is `self.estimated_values`, which is just a property that's  array. This way, there's no confusion/temptation about trying to assign a particular replacement value to an original censored value.

I also added a `lazy` option to the constructor in case the user wanted to delay computation and call `estimate` manually. The fault is that `lazy=False` and the estimation will happen automatically, e.g., `RobustROSEstimator(data=data, result='res', censorship='cen')._result_df` will have everything in it. 

Here's what the plots look like:
screen shot 2015-05-24 at 11 01 22 pm

josef-pkt:
@phobson  I created a new folder `imputation` in #3000, after making up my mind to keep `treatment` separate from `imputation`.

I think you could move this to `imputation` now.
Can I merge this, after rebasing, or did you change your plans for this? 

Related aside: The entire areas of censored, missing, imputation, treatment and so on are still wide open and will be several large long term projects, but it's good to get started with various models before we are able to get to a systematic support in these areas.

phobson:
@josef-pkt 

I think we should **not** merge this. I've created a much more easily (and therefore heavily) unit tested version of this.

The new overall approach is more functional. I would like to create a new PR sometime tonight.

In the meantime if you're really curious, you can review that code, as it'll appear in the PR:

  - [code](https://github.com/Geosyntec/wqio/blob/master/wqio/ros.py)
  - [tests](https://github.com/Geosyntec/wqio/blob/master/wqio/tests/test_ros.py)
  - [docs](http://phobson.github.io/wqio/api/ros.html)

PR #2030, opened 2014-10-06
beta-regression

brentp:
This is an initial go at adding beta-regression starting from `GenericLikelihoodModel` for feedback.

Here is the mailing list discussion: https://groups.google.com/forum/#!topic/pystatsmodels/3qEp-0YpEIc
and gist: https://gist.github.com/brentp/089c7d6d69d78d26437f

Josef suggested adding analytical derivatives and moving out of `miscmodels`. I will require some hand-holding to implement that so I have left it in `miscmodels` for now.

Currently, it has a separate link for the precision parameter (a + b == phi), but the results are reported in the context of the link for the X covariates. I'm not sure how to separate them out in the summary()

The Z matrix for modeling the precision currently must be specified as a matrix. Perhaps it makes sense to allow a z_formula to the `from_formula` method.

I put the csv data for the tests in the `miscmodels/tests` directory. 

The paper for R's `betareg` includes methods for bias-correction and bias-reduction, which I would like to implement eventually.


yoavram:
Hi, I've used the [gist](https://gist.github.com/brentp/089c7d6d69d78d26437f) and it works well, I'm interested in seeing this PR merged, is it still alive?

josef-pkt:
@yoavram  It's stalled, and on the "need to come back to this" waiting list.
I don't remember why this stalled (I often have to get busy with other things and might forget)

https://github.com/statsmodels/statsmodels/compare/master...josef-pkt:betareg_2030_3
has several more commits. It looks like I was trying to improve convergence and add some missing pieces.

It will need updating to current master. AFAIR, there were some ideas in here and in related work that have in the meantime been implemented in a more general way or for other cases.
(Plus I have three additional years of experience with convergence failures, and scipy optimizers have changed.)

yoavram:
Ok makes sense. I’ll use this gist in the meantime.

josef-pkt:
I rebased my branch of this and opened #4238

From a very brief look at a commented out test failure, I think it either needs more work on optimization, or, maybe better, needs a "nicer" testcase that doesn't use identity link.

yoavram:
I think I can provide a test case using simulated data which uses logit link, would that be helpful?

josef-pkt:
@yoavram Yes, that would be helpful.
We need one "nice" test case verified against R or Stata or some other package that shows that the results are correct and pass across versions of scipy and numpy.
"not nice" cases are much more difficult, in GLM we warn now when a link is used that does not constrain the estimator to remain in the support of the distribution parameters or predictions. Some family-link combinations that don't force the support still don't converge in many cases.
Another problem is that bfgs and other scipy optimizers have changed across scipy versions and are not robust in all versions.

mvogelsang:
just checking in to say that myself and plenty of people I know would be immensely grateful for a proper beta regression implementation to be available as part of a proper python package! AFAIK there literally isn't one anywhere else. 
I've been using that gist, but it would be fantastic to see this get merged.

josef-pkt:
looks like a problem from the renaming, this should not be closed

dead because the original branch doesn't exist anymore

bashtage:
This is a bug on GitHub when branches or reps don't exist. Numpy had 8 or
so.

josef-pkt:
I have them in my local clone, I just checked 2030 and I can create a branch from it.
I fetched all `pr` branches

josef-pkt:
This is the only branch that was affected based on the list of recently updated, closed branches

Edwardp17:
Hi, just wondering if this is still open? I'm very interested in seeing this in the statsmodels library, it would be immensely useful

PR #2076, opened 2014-11-02
New PR for MICE

kshedden:
Frank and I decided that I should resubmit a MICE PR from my account due to time limitations on his side.

This PR supercedes #1757.  It is a substantial overhaul of the code, but still contains all the main elements as designed and implemented by Frank in his GSOC project.

Here are two notebooks that illustrate how it can be used:

[Dead links, see below for updated working links]

http://nbviewer.ipython.org/urls/umich.box.com/shared/static/dkinhqyyz374k37tdwdi.ipynb

http://nbviewer.ipython.org/urls/umich.box.com/shared/static/pt8p8tctkl10t32xo97v.ipynb

Feedback welcome and encouraged.

### DIFF
--- a/statsmodels/sandbox/mice/mice.py
+++ b/statsmodels/sandbox/mice/mice.py
@@ -112,7 +112,7 @@ Opening Windows into the Black Box', Journal of Statistical Software,
 """
 
 #TODO: Add reference http://biomet.oxfordjournals.org/content/86/4/948.full.pdf
-#TODO: Change md to mod, mdf to rslt
+
 
 import pandas as pd
 import numpy as np
@@ -134,6 +134,16 @@ class _multivariate_normal(object):
         z = np.random.normal(size=p)
         return self.mean + np.dot(self.cov_sqrt, z)
 
+
+
+_mice_data_example_1 = """\
+>>> imp = mice.MICE_data(data)
+>>> imp.set_imputer('x1', formula='x2 + np.square(x2) + x3')
+>>> for j in range(20):
+>>>     imp.update_all()
+>>>     imp.data.to_csv('data%02d.csv' % j)
+"""
+
 class MICE_data(object):
     """
     Wrap a data set to allow missing data handling for MICE.
@@ -150,6 +160,17 @@ class MICE_data(object):
         MICE_data object is passed as the sole argument to
         `history_func`.
 
+    Examples
+    --------
+    Draw 20 imputations from a data set called `data` and save them in
+    separate files with filename pattern `dataXX.csv`.  The variables
+    other than `x1` are imputed using linear models fit with OLS, with
+    mean structures containing main effects of all other variables in
+    `data`.  The variable named `x1` has a condtional mean structure
+    that includes an additional term for x2^2.
+
+    %(_mice_data_example_1)s
+
     Notes
     -----
     Allowed perturbation methods are 'gaussian' (the model parameters
@@ -160,7 +181,7 @@ class MICE_data(object):
 
     `history_func` can be implemented to have side effects such as
     saving the current imputed data set to disk.
-    """
+    """ % {'_mice_data_example_1': _mice_data_example_1}
 
     def __init__(self, data, perturbation_method='gaussian',
                  history_func=None):
@@ -210,8 +231,8 @@ class MICE_data(object):
         for vname in data.columns:
             self.set_imputer(vname)
 
-        # Variable order for imputations.  Impute variables with
-        # fewest missing values first.
+        # The order in which variables are imputed in each cycle.
+        # Impute variables with the fewest missing values first.
         vnames = data.columns.tolist()
         nmiss = [len(self.ix_miss[v]) for v in vnames]
         nmiss = np.asarray(nmiss)
@@ -767,7 +788,7 @@ class MICE_data(object):
         Perturbs the model's parameters by sampling from the Gaussian
         approximation to the sampling distribution of the parameter
         estimates.  Optionally, the scale parameter is perturbed by
-        sampling from it's asymptotic Chi^2 sampling distribution.
+        sampling from its asymptotic Chi^2 sampling distribution.
         """
 
         endog_obs, exog_obs, exog_miss =\
@@ -924,7 +945,7 @@ class MICE_model(object):
 
     def next(self):
 
-        self.data.update_all(n_iter=self.n_skip)
+        self.data.update_all(n_iter=self.n_skip + 1)
 
         model = self.analysis_class.from_formula(self.analysis_formula,
                                                  self.data.data,
@@ -962,7 +983,7 @@ x4        -0.0253   0.0336  -0.7520 0.4521 -0.0911  0.0406 0.0269
 class MICE(object):
     """
     Use Multiple Imputation with Chained Equations to fit a model when
-    some data are missing.
+    some data values are missing.
 
     Parameters
     ----------
@@ -997,6 +1018,13 @@ class MICE(object):
     def burnin(self, n_burnin):
         """
         Impute a given number of data sets and discard the results.
+
+        Parameters
+        ----------
+        n_burnin : int
+            The number of update cycles to perform.  Each update cycle
+            updates each variable in the data set with one or more
+            missing values.
         """
 
         self.data.update_all(n_iter=n_burnin)
--- a/statsmodels/sandbox/mice/mice.py
+++ b/statsmodels/sandbox/mice/mice.py
@@ -1,4 +1,7 @@
 """
+Overview
+--------
+
 This module implements the Multiple Imputation through Chained
 Equations (MICE) approach to handling missing data. The approach has
 the following steps:
@@ -13,15 +16,13 @@ the same variable.
 variable, regressed against the observed and current imputed values of
 some or all of the remaining variables.  Then impute the missing
 values for the focus variable.  The only currently implemented
-procedure for doing this is the 'predictive mean matching' (pmm)
-procedure.
+procedure for doing the imputation is the 'predictive mean matching'
+(pmm) procedure.
 
-2. Repeat step 1 for all variables.
-
-3. Once all variables have been imputed, fit the analysis model to the
+2. Once all variables have been imputed, fit the analysis model to the
 data set.
 
-4. Repeat steps 1-3 multiple times and combine the results using a
+3. Repeat steps 1-2 multiple times and combine the results using a
 combining rule.
 
 The specific way that each variable is imputed is specified using a
@@ -41,6 +42,7 @@ available.
 
 Terminology
 -----------
+
 The MICE procedure is determined by a family of conditional models.
 There is one conditional model for each variable with missing values.
 A conditional model may be conditioned on all or a subset of the
@@ -58,7 +60,8 @@ conditional model to a bootstrapped version of the data set.
 
 Class structure
 ---------------
-There are three main classes in the module:
+
+There are two main classes in the module:
 
 * 'MICEData' wraps a Pandas dataframe, incorporating information about
   the conditional models for each variable with missing values. It can
@@ -77,6 +80,7 @@ There are three main classes in the module:
 
 Notes
 -----
+
 By default, to conserve memory 'MICEData' saves very little
 information from one iteration to the next.  The data set passed by
 the user is copied on entry, but then is over-written each time new
@@ -87,6 +91,7 @@ intermediate datasets to be saved for future use.
 
 References
 ----------
+
 JL Schafer: 'Multiple Imputation: A Primer', Stat Methods Med Res,
 1999.
 
@@ -101,9 +106,6 @@ A Gelman et al.: 'Multiple Imputation with Diagnostics (mi) in R:
 Opening Windows into the Black Box', Journal of Statistical Software,
 2009.
 """
-
-#TODO: Add reference http://biomet.oxfordjournals.org/content/86/4/948.full.pdf
-
 
 import pandas as pd
 import numpy as np
@@ -129,6 +131,15 @@ _mice_data_example_2 = """
             j += 1
 """
 
+
+class PatsyFormula(object):
+    """
+    A simple wrapper for a string to be interpreted as a Patsy formula.
+    """
+    def __init__(self, formula):
+        self.formula = "0 + " + formula
+
+
 class MICEData(object):
 
     __doc__ = """\
@@ -140,6 +151,12 @@ class MICEData(object):
         The data set.
     perturbation_method : string
         The default perturbation method
+    followed_by : dict-like
+        A map from variable names to lists of variables that 'follow'
+        that variable when imputing.  See notes for details.
+    k_pmm : int
+        The number of nearest neighbors to use during predictive mean
+        matching.  Can also be specified in `fit`.
     history_callback : function
         A function that is called after each complete imputation
         cycle.  The return value is appended to `history`.  The
@@ -167,20 +184,35 @@ class MICEData(object):
     estimated values obtained when fitting a bootstrapped version of
     the data set).
 
+    The `followed_by` dictionary is used is there are auxiliary
+    variables that must be linked to a given variable when doing the
+    imputation.  For example, in duration (survival) models, the
+    censoring status variable will generally 'follow' its
+    corresponding event time variable.  Whichever observed event time
+    is selected by PMM to impute a missing event time, the censoring
+    status for the same case will be used to impute the missing
+    censoring status variable.
+
     `history_callback` can be implemented to have side effects such as
     saving the current imputed data set to disk.
     """ % {'_mice_data_example_1': _mice_data_example_1,
            '_mice_data_example_2': _mice_data_example_2}
 
     def __init__(self, data, perturbation_method='gaussian',
-                 history_callback=None):
+                 followed_by=None, k_pmm=20, history_callback=None):
 
         # Drop observations where all variables are missing.  This
         # also has the effect of copying the data frame.
         self.data = data.dropna(how='all').reset_index(drop=True)
 
+        self.followed_by = followed_by if followed_by is not None else {}
+        self._followers = set([])
+        for x in self.followed_by.values():
+            self._followers |= set(x)
+
         self.history_callback = history_callback
         self.history = []
+        self.predict_kwds = {}
 
         # Assign the same perturbation method for all variables.
         # Can be overriden when calling 'set_imputer'.
@@ -205,8 +237,8 @@ class MICEData(object):
 
         # Map from variable names to init/fit args of the conditional
         # models.
-        self.init_kwds = defaultdict(lambda : {})
-        self.fit_kwds = defaultdict(lambda : {})
+        self.init_kwds = defaultdict(lambda : dict())
+        self.fit_kwds = defaultdict(lambda : dict())
 
         # Map from variable names to the model class.
         self.model_class = {}
@@ -216,22 +248,36 @@ class MICEData(object):
 
         # Set default imputers.
         for vname in data.columns:
-            self.set_imputer(vname)
+            if vname not in self._followers:
+                self.set_imputer(vname)
 
         # The order in which variables are imputed in each cycle.
         # Impute variables with the fewest missing values first.
-        vnames = data.columns.tolist()
+        vnames = [x for x in data.columns if x not in self._followers]
         nmiss = [len(self.ix_miss[v]) for v in vnames]
         nmiss = np.asarray(nmiss)
         ii = np.argsort(nmiss)
         ii = ii[sum(nmiss == 0):]
-        self.cycle_order = [vnames[i] for i in ii]
+        self._cycle_order = [vnames[i] for i in ii]
 
-        # Fill missing values with column-wise mean.  This is the
-        # starting imputation.
-        self.data = self.data.fillna(self.data.mean())
+        self._initial_imputation()
 
-        self.k_pmm = 20
+        self.k_pmm = k_pmm
+
+
+    def _initial_imputation(self):
+        """
+        Use a PMM-like procedure for initial imputed values.
+        """
+
+        def f(vec):
+            vec = vec.dropna()
+            mn = vec.mean()
+            ix = np.argmin(vec - mn)
+            return vec[ix]
+
+        imp = self.data.apply(f, axis=0)
+        self.data = self.data.fillna(imp)
 
 
     def _split_indices(self, vec):
@@ -244,8 +290,8 @@ class MICEData(object):
 
 
     def set_imputer(self, endog_name, formula=None, model_class=None,
-                    init_kwds=None, fit_kwds=None, k_pmm=20,
-                    perturbation_method=None):
+                    init_kwds=None, fit_kwds=None, predict_kwds=None,
+                    k_pmm=20, perturbation_method=None):
         """
         Specify the imputation process for a single variable.
 
@@ -260,18 +306,23 @@ class MICEData(object):
             structure, e.g. use 'x1 + x2' not 'x4 ~ x1 + x2'.
         model_class : statsmodels model
             Conditional model for imputation. Defaults to OLS.
-        init_kwds : Dictionary
+        init_kwds : dit-like
             Keyword arguments passed to the model init method.
-        fit_kwds : Dictionary
+        fit_kwds : dict-like
             Keyword arguments passed to the model fit method.
+        predict_kwds : dict-like
+            Keyword arguments passed to the model predict method.
+        k_pmm : int
+            Determines number of neighboring observations from which
+            to randomly sample when using predictive mean matching.
         perturbation_method : string
             Either 'gaussian' or 'bootstrap'. Determines the method
             for perturbing parameters in the conditional model.  If
             None, uses the default specified in init.
-        k_pmm : int
-            Determines number of neighboring observations from which
-            to randomly sample when using predictive mean matching.
         """
+
+        if endog_name in self._followers:
+            raise ValueError("imputer may not be set for a variable in `followed_by`")
 
         if formula is None:
             main_effects = [x for x in self.data.columns
@@ -292,6 +343,9 @@ class MICEData(object):
 
         if fit_kwds is not None:
             self.fit_kwds[endog_name] = fit_kwds
+
+        if predict_kwds is not None:
+            self.predict_kwds[vname] = predict_kwds
 
         if perturbation_method is not None:
             self.perturbation_method[endog_name] = perturbation_method
@@ -332,7 +386,7 @@ class MICEData(object):
         """
 
         for k in range(n_iter):
-            for vname in self.cycle_order:
+            for vname in self._cycle_order:
                 self.update(vname)
 
         if self.history_callback is not None:
@@ -355,10 +409,19 @@ class MICEData(object):
             Observed values of the variable to be imputed.
         exog_obs : DataFrame
             Current values of the predictors where the variable to be
-            Imputed is observed.
+            imputed is observed.
+        following_obs : DataFrame
+            Current values of the variables that 'follow' the variable
+            to be imputed.
         exog_miss : DataFrame
             Current values of the predictors where the variable to be
             Imputed is missing.
+        init_kwds : dict-like
+            The init keyword arguments for `vname`, processed through Patsy
+            as required.
+        fit_kwds : dict-like
+            The fit keyword arguments for `vname`, processed through Patsy
+            as required.
         """
 
         formula = self.conditional_formula[vname]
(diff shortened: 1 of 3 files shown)


josef-pkt:
just a quick look:
what happened to `get_distribution` attached to the existing models? Are they not needed anymore?

test coverage in this PR is considerably lower than in #1757, looks like just some smoke tests. Can the old tests be ported to this?

josef-pkt:
The notebooks and plots in it look very nice.

kshedden:
At some point we decided to pull back a bit and focus on predictive mean matching.  PMM seems to work quite well and apparently is the favored choice in other packages.  I thought it made sense to focus on other features like diagnostic plots rather than to continue wrestling with get_distribution.

I would be happy to get some feedback on the MICE implementation and can put time into upgrading the branch as needed.  

I will look into improving test coverage.

josef-pkt:
Hopefully @agartland can provide feedback. I can give a brief review after 0.7, but I have also several other PRs that I have kept waiting for too long.

@kshedden  Can you give a brief summary of the difference of this version and Franks #1757 version, especially with respect to what's in 1757 and not in here. Based on your last comment I assume that there are still things in 1757 that are not obsolete and might be ported to the new version at some point.

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1167 @@
...
+  an additional 'analysis model' which is the central model of
+  interest (it may or may not be identical to the conditional model
+  for the same outcome variable).  MICE_model carries out multiple
+  imputation using the conditional models, and fits the analysis model
+  to a subset of these imputed data sets.  It returns the fitted model
+  results for these analysis models.  It is structured as an iterator,
+  so the analysis model results are obtained using `next` or via any
+  standard python iterator pattern.
+
+* 'MICE' takes both a 'MICE_data' object and an analysis model
```
josef-pkt: something like `MICE` is the wrapper class that does the full analysis, ....
IIUC, to get the contrast between the function that holds some pieces and the top class

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1167 @@
...
+
+# Can be replaced with scipy in version >= 0.14
+class _multivariate_normal(object):
+
+    def __init__(self, mean, cov):
+        self.mean = mean
+        self.cov = cov
+        self.cov_sqrt = np.linalg.cholesky(cov)
+
+    def rvs(self):
```
josef-pkt: this is also in numpy.random.multivariate_normal  (which uses svd instead of cholesky)

josef-pkt: but the numpy version won't store the sqrt, and redo it at each call

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1167 @@
...
+
+class MICE_data(object):
+
+    __doc__ = """\
+    Wrap a data set to allow missing data handling for MICE.
+
+    Parameters
+    ----------
+    data : array-like object
+        The data set.
```
josef-pkt: from the use of `dropna` below I assume this requires a dataframe not `array_like`

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1167 @@
...
+    perturbation_method : string
+        The default perturbation method
+    history_func : function
```
josef-pkt: I guess a more standard terminology would be to call this a `call_back` such as in the optimization functions

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1167 @@
...
+        nmiss = np.asarray(nmiss)
+        ii = np.argsort(nmiss)
+        ii = ii[sum(nmiss == 0):]
+        self.cycle_order = [vnames[i] for i in ii]
+
+        # Fill missing values with column-wise mean.  This is the
+        # starting imputation.
+        self.data = self.data.fillna(self.data.mean())
+
+        self.k_pmm = 20
```
josef-pkt: pure initialization might be  better with `None`, let the method handle the default?

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1167 @@
...
+        Specify the imputation process for a single variable.
+
+        Parameters
+        ----------
+        endog_name : string
+            Name of the variable to be imputed.
+        formula : string
+            Conditional formula for imputation. Defaults to a formula
+            with main effects for all other variables in dataset.  The
+            formula should only include an expression for the mean,
```
josef-pkt: drop comma at end ?

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1167 @@
...
+            structure, e.g. use 'x1 + x2' not 'x4 ~ x1 + x2'.
+        model_class : statsmodels model
```
josef-pkt: are "ducks" allowed? which kind of models?
(maybe becomes more obvious later)

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1167 @@
...
+            Conditional model for imputation. Defaults to OLS.
+        init_args : Dictionary
```
josef-pkt: I prefer `init_kwds` as in our internal attribute, it's kwargs not args. ?

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1167 @@
...
+            Keyword arguments passed to the model init method.
+        fit_args : Dictionary
```
josef-pkt: same here `fit_kwds` ?

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1167 @@
...
+
+_mice_data_example_1 = """
+    >>> imp = mice.MICE_data(data)
+    >>> imp.set_imputer('x1', formula='x2 + np.square(x2) + x3')
+    >>> for j in range(20):
+            imp.update_all()
+            imp.data.to_csv('data%02d.csv' % j)
+"""
+
+class MICE_data(object):
```
josef-pkt: general question (maybe obvious later)
Is there a MICE `method`, or currently it's only `pmm`, or it's not relevant here?

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1167 @@
...
+                                         ["white", "darkgrey"])
+            ax.imshow(miss, aspect="auto", interpolation="nearest",
+                      cmap=cmap)
+
+        ax.set_ylabel("Cases")
+        ax.set_xticks(range(len(cols)))
+        ax.set_xticklabels(cols, rotation=90)
+
+        return fig
+
```
josef-pkt: just in general: pep-8:  2 empty lines between methods and functions, IIRC, not yet enforced, but we try to get closer to it

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1167 @@
...
+        ax.set_ylabel(col2_name)
+
+        return fig
+
+    def fit_scatterplot(self, col_name, lowess_args={},
+                        lowess_min_n=40, jitter=None,
+                        plot_points=True, ax=None):
+        """
+        Create a scatterplot between the observed or imputed values of
+        a variable and the corresponding fitted values.
```
josef-pkt: First line of docstring should be a **one**-line summary, otherwise it messes up sphinx doc rendering

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1167 @@
...
+        endog_obs, exog_obs, exog_miss =\
+                   self.get_split_data(vname)
+
+        klass = self.model_class[vname]
+        formula = self.conditional_formula[vname]
+        self.models[vname] = klass.from_formula(formula,
+                                            self.data,
+                                            **self.init_args[vname])
+        self.results[vname] = self.models[vname].fit(
+            **self.fit_args[vname])
```
josef-pkt: question: Do we have to call this repeatedly for the same fit? in that case we would have redundant fit for generating rvs.
I guess, No, in this case where we will have new (imputed) `data` in each round.

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1167 @@
...
+            self.perturb_gaussian(vname)
+        elif self.perturbation_method[vname] == "boot":
+            self.perturb_bootstrap(vname)
+        else:
+            raise ValueError("unknown perturbation method")
+
+    def impute(self, vname):
+        # Wrap this in case we later add additional imputation
+        # methods.
+        self.impute_pmm(vname)
```
josef-pkt: where would the imputation method be specified, when we add alternatives to `pmm`

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1167 @@
...
+        """
+        Update a single variable.  This is a two-step process in which
+        first the parameters are perturbed, then the missing values
+        are re-imputed.
+        """
+
+        self.perturb_params(vname)
+        self.impute(vname)
+
+    def impute_pmm(self, vname):
```
josef-pkt: we don't use enough privatising of method, using leading underscore.
Are there method where it doesn't makes sense for the user to call those?

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1167 @@
...
+        ir = np.random.randint(0, k_pmm, len(pendog_miss))
+
+        # Unwind the indices
+        jj = np.arange(dxi.shape[0])
+        ix = dxi[[jj, ir]]
+        iz = ixm[[jj, ix]]
+
+        imputed_miss = np.array(endog_obs[iz])
+
+        self.store_changes(vname, imputed_miss)
```
josef-pkt: general design question ( I don't have an overview yet)
Can we avoid that many methods work with sideeffects, i.e. changing attributes instead of returning?
I feel more comfortable (safer) if only the main function (like a fit method) changes state.

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1167 @@
...
+        The data set, in the form of a MICE_data object.
+    analysis_formula : string
+        Formula for the analysis model.
+    analysis_class : statsmodels model
+        Model class to be fit to imputed data sets.
+    n_skip : integer
+        Number of imputation cycles to perform before fitting
+        the analysis models.
+    init_args : dict-like
+        Additional parameters for statsmodels model instance.
```
josef-pkt: is this for the `analysis model`

`init_kwds`

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1167 @@
...
+_mice_model_example_1 = """
+    >>> imodel = mice.MICE_model(data, 'y ~ x1 + x2 + x3', sm.OLS)
+    >>> results = []
+    >>> for j in range(100):
+            results.append(imodel.next())
+    >>> params1 = [x.params[1] for x in results]
+    >>> plt.hist(params1)
+"""
+
+class MICE_model(object):
```
josef-pkt: name is not pep-8  `MICEModel`? but is it really a `Model` class?
`MICEIterator`?

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1167 @@
...
+        missing values will be imputed
+
+    Examples
+    --------
+    Simple example using defaults:
+    %(mice_example_1)s
+    """ % {'mice_example_1' : _mice_example_1}
+
+    def __init__(self, model_formula, model_class, data,
+                 init_args={}, fit_args={}):
```
josef-pkt: same here: `init_kwds`, `fit_kwds`
in general it's better to avoid mutable default arguments, i.e. use `init_kwds=None` and set 
```
if init_kwds = None:
    init_kwds = {}
```
(I don't remember the details, just being conservative (defensive programming).)

josef-pkt: or the one-liner, something like
`self.init_kwds = init_kwds if init_kwds is not None else {}`

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1167 @@
...
+        self.init_args = init_args
+        self.fit_args = fit_args
+        self.data = data
+
+        model_chain = MICE_model(data, model_formula,
+                                    model_class, 0,
+                                    init_args, fit_args)
+        self.model_chain = model_chain
+
+    def burnin(self, n_burnin):
```
josef-pkt: private? `_burnin`

josef-pkt:
One reading through most parts:
I didn't see any serious problems (except low test coverage)

a few comments:

several naming questions `kwds`, class name, I didn't see much else

one interface question: I think MICE should get a `fit` method that combines `burnin` `run` and `combine` as a one-stop shopping method and closer to other models.

privatizing names with leading underscore: in general we don't use it enough in statsmodels. In this case (maybe because I don't know the structure so well but also because it has many methods across classes) I think there are methods where it doesn't make any sense for users to call them directly.

triggered partially by the previous comments: What do we do about other methods than `pmm`?
Do we have a rough idea about where we need to "future proof" this when we want to add more imputation methods?

josef-pkt:
semi random thought:
`start_params`: If we have models where fit is time consuming, then we should set start_params to use information from previous iterations to speed up the optimization.

josef-pkt:
question (One of my favorite topics, and a source of "surprises": the stale state. I have no idea yet.)

```
mi = mice.MICE("h39b ~ age + c28 + pcs + mcs37 + b05 + haartadhere", sm.OLS, imp)
mi.burnin(10)
mi.run(20, 5)
result = mi.combine()
mi.run(10, 5)
result2 = mi.combine()
```

What happens to `result`?
Do we need to warn, protect, ...?

josef-pkt:
aside `pmm` predictive mean matching and `psm` propensity score matching look very similar.

datnamer:
@josef-pkt  I'm in need of MICE and I would rather not dip into r. Is this in any sort of  rough usable state?

kshedden:
It works in all the examples we have tried, but admittedly it is not extensively tested.

Expect some renaming/refactoring later in the week when I get to addressing Josef's review.

datnamer:
@kshedden  thanks Ill give a run.

kshedden:
Updated notebooks:

https://umich.box.com/shared/static/5pcv0tbj302pxk7lpbalshlqddmmlk3m.ipynb
https://umich.box.com/shared/static/mv2wvdcicwl4ww1kamp8lkb7fahrnl7m.ipynb
https://umich.box.com/shared/static/yn2cttfg479wm2dmfnwnvopwbswgqkuy.ipynb

donaldbraman:
Just want to thank you for working on getting this up and running -- it's an crucial feature to serious analysis of incomplete data.

josef-pkt:
@kshedden  Can you rebase this? maybe in a few days, after fixing travis, and those in #2679
I'm planning another brief review and then merging it.

kshedden:
Yes, I will take care of this later in the week.

datnamer:
Suhweet

josef-pkt:
TravisCI is working again, although we still have problems with latest numpy, 
newest pandas version is showing some convergence warnings with recent merges, and could possibly also have issues with this PR.

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1287 @@
...
+
+        if isinstance(obj, np.ndarray):
+            return obj
+        elif hasattr(obj, 'predicted_values'):
+            return obj.predicted_values
+        else:
+            raise "cannot obtain predicted values from %s" % obj.__class__
+
+
+    def impute_pmm(self, vname):
```
josef-pkt: we could add here a keyword argument for a selection/aggregation function as a callback.
It can default to the current random selection, but we could use non-uniform randomization or impute the mean of the neighbors for use in propensity score matching.
(not needed for this PR, it;s more for extensions and reusing code for other matching tasks)

josef-pkt:
Another general comment: the plots for checking the match between imputed and observed values look a bit similar to the diagnostics for treatment effect imputation that Stata added in version 14. (I didn't check details yet for that.)

This is still in the sandbox and we will have to move it soon.
I would like to put imputation together with `treatment`, but I'm still not sure `treatment` (used in my treatment effects PR) is a good name for both. Stata has now separate manuals for multiple imputation and treatment effects, but there is a large theoretical and code overlap.

josef-pkt:
TravisCI doesn't run any tests for this, I was checking coveralls and the mice.py module is missing. The number of tests run 4337 on python 3 is the same as in master

the names of the test classes have an unterline, which is not PEP8, but I thought nose would use them anyway. @kshedden  can you rename `Test_MICEData` -> `TestMICEData` and similar to see if it helps.

correction: that's not the reason, the test folder doesn't have an `__init__.py` so it won't be included as code, and it is not decleared as package data.
`statsmodels/sandbox/mice/tests/__init__.py`

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1287 @@
...
+            imp.update_all()
+            imp.data.to_csv('data%02d.csv' % j)
+"""
+
+_mice_data_example_2 = """
+    >>> imp = mice.MICEData(data)
+    >>> j = 0
+    >>> for data in imp:
+            imp.data.to_csv('data%02d.csv' % j)
+            j += 1
```
josef-pkt: the loop variable `data` is not used in the loop. Is this intentional? or should it be `data.to_csv('data%02d.csv' % j)`?

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1287 @@
...
+    other than `x1` are imputed using linear models fit with OLS, with
+    mean structures containing main effects of all other variables in
+    `data`.  The variable named `x1` has a condtional mean structure
+    that includes an additional term for x2^2.
+    %(_mice_data_example_1)s
+
+    Impute using default models, using the MICEData object as an
+    iterator.
+    %(_mice_data_example_2)s
+    Notes
```
josef-pkt: doc build might break without blank line in front of section header

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1287 @@
...
+        # Can be overriden when calling 'set_imputer'.
+        self.perturbation_method = defaultdict(lambda :
+                                               perturbation_method)
+
+        # Map from variable name to indices of observed/missing
+        # values.
+        self.ix_obs = {}
+        self.ix_miss = {}
+        for col in self.data.columns:
+            ix_obs, ix_miss = self._split_indices(self.data[col])
```
josef-pkt: I usually prefer unique names for different things to avoid getting confused
e.g ix_obs_coll, ix_miss_coll   (or often I add an `_i` for one element

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1287 @@
...
+
+    def _initial_imputation(self):
+        """
+        Use a PMM-like procedure for initial imputed values.
+        """
+
+        def f(vec):
+            vec = vec.dropna()
+            mn = vec.mean()
+            ix = np.argmin(vec - mn)
```
josef-pkt: shouldn't this be absolute value?
IIUC, it's the element closest to the mean, in abs norm.

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1287 @@
...
+A Gelman et al.: 'Multiple Imputation with Diagnostics (mi) in R:
+Opening Windows into the Black Box', Journal of Statistical Software,
+2009.
+"""
+
+import pandas as pd
+import numpy as np
+import patsy
+import statsmodels.api as sm
+import statsmodels
```
josef-pkt: it's better to import from specific modules or subdirectories, than from `sm` api
I've only seen OLS so far.

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1287 @@
...
+            main_effects = [x for x in self.data.columns
+                            if x != endog_name]
+            fml = endog_name + " ~ " + " + ".join(main_effects)
+            self.conditional_formula[endog_name] = fml
+        else:
+            fml = endog_name + " ~ " + formula
+            self.conditional_formula[endog_name] = fml
+
+        if model_class is None:
+            self.model_class[endog_name] = sm.OLS
```
josef-pkt: better to use from statsmodels.regression.linear_model, see comment at imports

statsmodels/sandbox/mice/mice.py
josef-pkt: avoids circular imports and doesn't load tsa or others if not needed.

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1287 @@
...
+        ----------
+        col : string
+            Name of variable to be filled in.
+        vals : array
+            Array of imputed values to use in filling in missing values.
+        """
+
+        ix = self.ix_miss[col]
+        if len(ix) > 0:
+            self.data[col].iloc[ix] = vals
```
josef-pkt: should this be a new indexing in pandas `data.ix[col, ix]` or loc or something, but with both slices together?

josef-pkt:
Another general comment, maybe relevant later

The separate call to patsy dmatrices don't use the underlying state in stateful transforms or categorical encoding if one level is missing in a array. We might have to carry around the `design_info` for the full model if we create partial design matrices.
I was just reading through the patsy treatment of the kwds, and didn't see yet whether this uses full arrays or partial arrays, but with separate calls with observed and missing data there could be a problem. I don't expect a problem using full data, i.e. the current code has `self.data` in dmatrix, AFAICS.  
There might be some other limitation to whether the functions used in the formulas are in scope, but that's a special usage.

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1287 @@
...
+
+        endog, exog, init_kwds, fit_kwds = self.get_fitting_data(vname)
+
+        klass = self.model_class[vname]
+        self.models[vname] = klass(endog, exog, **init_kwds)
+        self.results[vname] = self.models[vname].fit(**fit_kwds)
+
+        cov = self.results[vname].cov_params()
+        mu = self.results[vname].params
+        np.linalg.cholesky(cov) #!!!!!!!
```
josef-pkt: what's this?  just a check for positive definite, and an exception if not?

josef-pkt:
random thoughts: two possible additions that might be useful if we want to continue from a stored dataset, for example,

- add a RandomState instance that users can set in `MICEData.__init__` 
- allow to start from an initial dataset, instead of default mean imputation.


- extra: we could `combine` and create Results instance from stored datasets. (sounds like what Stata is doing)

josef-pkt:
I'm stopping here for now. I still need to go over the results class and I skipped the plots.

Looks very good overall, and pretty much ready to merge.

I looked at the Stata manual again for multiple imputation, and there would be enough material to have it's own directory/subpackage, which would make it obsolete to come up with a name for imputation/treatment. 
I think it's better to move it in a new PR after merging this one, so github keeps displaying changes in this PR properly. (github doesn't follow renames and moves.)

statsmodels/sandbox/mice/mice.py
kshedden: I am mixing label-based indexing (for columns) and position-based indexing (for rows).  'col' is a variable name and 'ix' is a numeric position.  I don't know how to do this kind of mixed indexing with a single call to ix or loc.

josef-pkt: I'm just fighting with pandas over contingency table handling, trying to figure out at least some pivoting basics. `df_inj_count_other.ix[:2, 'All']`

ix handles both http://pandas.pydata.org/pandas-docs/stable/indexing.html

`.ix supports mixed integer and label based access. It is primarily label based, but will fall back to integer positional access unless the corresponding axis is of integer type`

AFAIR, we used this for the warnings that we assign to a copy in several places.

kshedden: Would this create a problem if the row axis (which we have no control over) has integer type but is not identical to the positions?  The `ix_miss` values always refer to positions because they are obtained using `numpy.flatnonzero`.

My strategy was to internally use position-based indexing for rows (for performance), but we need to refer to variables (columns) by name because this is how the inputs are supplied.

Is there a downside to using `df[col_name].iloc[ii]`?  I don't see copy warnings on my setup.

josef-pkt: then it's better to keep it the way you have it, I'm no expert on pandas.
Maybe the problem was only if rows where indexed first, which would make sense because columns are always homogeneous (same dtype) and stored in one array while rows are not.

josef-pkt:
missing test coverage looks mostly missing smoke tests for `hist` plot method (*) and `summary` method

(*) plot methods in models or results should have `plot_` prefix in name

I'll try to finish my review tomorrow.

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1295 @@
...
+            The number of burn-in cycles to skip.
+        n_imputations : int
+            The number of data sets to impute
+        """
+
+        # Run without fitting the analysis model
+        self.data.update_all(n_burnin)
+
+        for j in range(n_imputations):
+            result = next(self)
```
josef-pkt: Do we need that the class itself is an iterator?

When Skipper and I took over stats.models we removed all iterators from the classes and converted them to iteration loops inside methods. The main problem is that it is very difficult to keep track what is actually changed and where the side effects are of the iterator.

I haven't looked at a specific example yet, this is just general skepticism to possibly difficult to understand code.

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1295 @@
...
+        """
+
+        # Extract a few things from the models that were fit to
+        # imputed data sets.
+        params_list = []
+        cov_list = []
+        scale_list = []
+        for results in self.results_list:
+            params_list.append(np.asarray(results.params))
+            cov_list.append(np.asarray(results.cov_params()))
```
josef-pkt: We could save here the asarray detour if we use results._results, which is the results without wrapper.

(In cases where we could have wrapped or unwrapped results, we check it `results = getattr(results._results, results` might work)

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1295 @@
...
+            scale_list.append(results.scale)
+        params_list = np.asarray(params_list)
+        scale_list = np.asarray(scale_list)
+
+        # The estimated parameters for the MICE analysis
+        params = params_list.mean(0)
+
+        # The average of the within-imputation covariances
+        cov_within = sum(cov_list) / len(cov_list)
```
josef-pkt: If we don't need `cov_list` otherwise, then we could accumulate `cov_within` inside the loop and save memory.

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1295 @@
...
+        info["Method:"] = "MICE"
+        info["Model:"] = self.model_class.__name__
+        info["Dependent variable:"] = self.endog_names
+        info["Sample size:"] = "%d" % self.model.data.data.shape[0]
+        info["Scale"] = "%.2f" % self.scale
+        info["Num. imputations"] = "%d" % len(self.model.results_list)
+
+        smry.add_dict(info, align='l', float_format=float_format)
+
+        param = summary2.summary_params(self, alpha=alpha)
```
josef-pkt: Does this always use t-distribution for p-values and confidence interval?
I only see the table for the OLS examples in the notebooks.

josef-pkt:
possible bug in summary2 (bug in column label)  label is for `t` distribution, even if it's using normal?

Usecase: How do we handle repeated fitting during model search, evaluating different models? Do we have to start from scratch each time to get replicable random numbers given initial seed?
(extension to this for followup?)

josef-pkt:
I don't see test coverage for pvalues and confint. We have helper test functions to test consistency with normal or t distribution (given that we had some refactoring bugs with inherited pvalues and confint in the past.) 
Given the discussion last year during GSOC, we should also check that the correct df are used if the distribution is t. If there are ambiguities in which df to use, then our choice should be documented.

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1295 @@
...
+            The number of data sets to impute
+        """
+
+        # Run without fitting the analysis model
+        self.data.update_all(n_burnin)
+
+        for j in range(n_imputations):
+            result = next(self)
+            self.results_list.append(result)
+
```
josef-pkt: Storing the entire results can become memory intensive.  I've done the same in adfuller, and user feedback was that the memory consumption goes up a lot with large arrays and many imputations/results. For the combine support we would only need to store params and scale and store or accumulate cov_params.
(storing full results could be optional or users could use a callback.)

statsmodels/sandbox/mice/mice.py
```diff
@@ -0,0 +1,1295 @@
...
+        self.data.update_all(self.n_skip)
+        start_params = None
+        if len(self.results_list) > 0:
+            start_params = self.results_list[-1].params
+        model = self.model_class.from_formula(self.model_formula,
+                                              self.data.data,
+                                              **self.init_kwds)
+        self.fit_kwds.update({"start_params": start_params})
+        result = model.fit(**self.fit_kwds)
```
josef-pkt: another possible extension: 
`fit` from default start_params will be in many cases much slower than good start_params. It would be useful to get good start_parms from previous fits, however we don't have a standardized signature for start_params (not relevant in OLS and similar).
(We reuse start_params in some bootstrap methods but that's within a single model.)

josef-pkt:
Stata manual says they are using t distribution with missingness dependent degrees of freedom if number of imputations is small. (They mention normal distribution for large number of imputation, but without mentioning threshold and in a quick browsing I only see t distribution params tables.)

josef-pkt:
to summarize:

The plot methods need to be renamed to start with `plot_` for API consistency.

I'm not very happy about the iterator in MICE. (vague, because it might not be needed, and makes the class more complex than necessary.) If possible it would be better to change this before merging.

All the other issues can wait for a followup

- use of t distribution looks fine, but we need tests and possibly more code or review of inherited methods for other inference.
- restarting, storage, ..., are usability questions and can be changed over time.
- some code and memory streamlining could be now or later.

I haven't tried to understand all the details.

Overall I think the structure looks very good and flexible, and I expect that we will get many opportunities to build on and extend this. 

Can we add any of your notebooks to the documentation notebook examples?

kshedden:
I think I have addressed most of the critical issues.  When we move it out of the sandbox we can add additional features such as warm starts and saving the random state.

Updated notebooks are here:

https://umich.box.com/shared/static/mv2wvdcicwl4ww1kamp8lkb7fahrnl7m.ipynb
https://umich.box.com/shared/static/5pcv0tbj302pxk7lpbalshlqddmmlk3m.ipynb
https://umich.box.com/shared/static/yn2cttfg479wm2dmfnwnvopwbswgqkuy.ipynb

kshedden:
I set pdf_output = False and added a smoke test for summary.

josef-pkt:
Thanks @kshedden 

I'll merge when travis reports back

Do you have an opinion about directory name and location?

My initial thought was to use one folder for treatment and imputation, but now I think we just use two directories, and we can move this into `statsmodels/imputation` (in the next PR).

kshedden:
Thanks.  I don't have an opinion about directory location.

methodds:
Wow I am looking for a Python MICE implementation since several months, thank you so much for this work! Any idea when this will be included in statsmodels?

kshedden:
It's in.  If you have any questions or suggestions after using it email the
list:

https://groups.google.com/forum/#!forum/pystatsmodels

or open an issue on github.

On Sat, Nov 14, 2015 at 6:08 PM, Carsten Schwemmer <notifications@github.com

datnamer:
Nice!!!

donaldbraman:
THANK YOU!

josef-pkt:
Note: PR #3000  is moving mice to statsmodels.imputation.mice
There is currently no compatibility layer, for those already using it

PR #2090, opened 2014-11-12
Add weights to GEE

kshedden:
This PR adds weights to GEE.  The implementation follows what is done in Stata, as described here:

http://www.stata.com/support/faqs/statistics/weighted-estimation-and-xtgee/

### DIFF
--- a/statsmodels/genmod/cov_struct.py
+++ b/statsmodels/genmod/cov_struct.py
@@ -221,28 +221,39 @@ class Exchangeable(CovStruct):
         endog = self.model.endog_li
 
         nobs = self.model.nobs
-        dim = len(params)
 
         varfunc = self.model.family.variance
 
         cached_means = self.model.cached_means
 
-        residsq_sum, scale, nterm = 0, 0, 0
+        has_weights = self.model.weights is not None
+        weights_li = self.model.weights
+
+        residsq_sum, scale = 0, 0
+        fsum1, fsum2, n_pairs = 0., 0., 0.
         for i in range(self.model.num_group):
 
             expval, _ = cached_means[i]
             stdev = np.sqrt(varfunc(expval))
             resid = (endog[i] - expval) / stdev
 
+            f = weights_li[i] if has_weights else 1.
+
             ngrp = len(resid)
             residsq = np.outer(resid, resid)
-            scale += np.trace(residsq)
-            residsq = np.tril(residsq, -1)
-            residsq_sum += residsq.sum()
-            nterm += 0.5 * ngrp * (ngrp - 1)
+            scale += f * np.trace(residsq)
+            fsum1 += f * len(endog[i])
 
-        scale /= (nobs - dim)
-        self.dep_params = residsq_sum / (scale * (nterm - dim))
+            residsq = np.tril(residsq, -1)
+            residsq_sum += f * residsq.sum()
+            npr = 0.5 * ngrp * (ngrp - 1)
+            fsum2 += f * npr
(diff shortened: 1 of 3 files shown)


statsmodels/genmod/cov_struct.py
```diff
@@ -220,33 +226,45 @@ def update(self, params):
...
-        scale /= (nobs - dim)
-        self.dep_params = residsq_sum / (scale * (nterm - dim))
+        ddof = self.model.ddof_scale
+        scale /= (fsum1 * (nobs - ddof) / float(nobs))
+        residsq_sum /= scale
+        self.dep_params = residsq_sum / (fsum2 * (n_pairs - ddof) / float(n_pairs))
 
     def covariance_matrix(self, expval, index):
         dim = len(expval)
         dp = self.dep_params * np.ones((dim, dim), dtype=np.float64)
```
josef-pkt: as general FYI: the fastest way in numpy to get an array with identical elements is `empty` and `fill`  (there is a new function in recent numpy `full` that I haven't used yet and won't be compatible for us yet.)

josef-pkt:
I don't want to hold this up any longer, but I'm also not trying to figure out how the weights really work.
I'll just rely on unit tests and figuring it out later.

Is it ready to merge?

kshedden:
I think it's ready, I've been using this a lot.  Thanks for the review.

josef-pkt:
Can you rebase and force push? It's rather far behind master. 
Then, I'll hit the merge button.

Otherwise, I'll do the rebase before merging.

kshedden:
Rebased and force pushed.

josef-pkt:
you should fetch from upstream, statsmodels master, before rebase so it rebased to the latest version.

A few commits behind master, but merging anyway. (no related/conflicting commits)

josef-pkt:
maybe of related interest and application

http://support.sas.com/resources/papers/proceedings14/SAS166-2014.pdf

```
Paper SAS166-2014:
Weighted Methods for Analyzing Missing Data with the GEE Procedures

Missing observations caused by dropouts or skipped visits present a problem in studies of longitudinal data. When the analysis is restricted to complete cases and the missing data depend on previous responses, the generalized estimating equation (GEE) approach, which is commonly used when the population-average effect is of primary interest, can lead to biased parameter estimates. The new GEE procedure in SAS/STAT® 13.2 implements a weighted GEE method, which provides consistent parameter estimates when the dropout mechanism is correctly specified. When none of the data are missing, the method is identical to the usual GEE approach, which is available in the GENMOD procedure. This paper reviews the concepts and statistical methods. Examples illustrate how you can apply the GEE procedure to incomplete longitudinal data.
```

kshedden:
Interesting.  I had followed the Stata docs which state that there is no application for subject-specific weights, so in #2090 I required cluster-level weights.  It was easy to generalize and I now have a working branch for subject-level weights that I will PR later after some more testing.

The update formula step 5 on page 3 of this doc is wrong:

http://support.sas.com/resources/papers/proceedings14/SAS166-2014.pdf

The denominator also needs to include the weight.  The correct formula is here:

http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.409.9129&rep=rep1&type=pdf

josef-pkt:
Just in general:
other uses of weights
- endogenous weights that could be imitated by prior weights, see #1777 
- robust estimation with down-weighting of influential outliers.

both would need to be observation specific, in the second if the outlier is an observation and not an individual

PR #2172, opened 2015-01-10
Prediction for functional regression

kshedden:
This is a prediction method tailored to functional regression models in which one or more variables is modeled using b-splines.

The code doesn't produce any graphs of its own, but one natural thing to do with the predictions is to plot them, producing plots like this:

https://umich.box.com/shared/static/g2jx0wi5lo2r8liknpib.pdf

This is implemented as functions that are currently in the sandbox, but could be moved to become a method of the base results class (with the glm version attached to glm results class).

Related to #2151, but tailored to the functional case.

It would be nice to have an option for the pointwise confidence band to be a simultaneous confidence band.

### DIFF
--- a/statsmodels/sandbox/predict_functional.py
+++ b/statsmodels/sandbox/predict_functional.py
@@ -32,7 +32,11 @@ def _make_formula_exog(result, focus_var, summaries, values, num_points):
     model = result.model
     exog = model.data.frame
 
+    if exog[focus_var].dtype is np.dtype('O'):
+        raise ValueError('focus variable may not have object type')
+
     colnames = list(summaries.keys()) + list(values.keys()) + [focus_var]
+    dtypes = [exog[x].dtype for x in colnames]
 
     # Check for variables whose values are not set either through
     # `values` or `summaries`.  Since the model data frame can contain
@@ -47,6 +51,8 @@ def _make_formula_exog(result, focus_var, summaries, values, num_points):
                       % ", ".join(["'%s'" % x for x in unmatched]))
 
     fexog = pd.DataFrame(index=range(num_points), columns=colnames)
+    for d, x in zip(dtypes, colnames):
+        fexog[x] = fexog[x].astype(d)
 
     # The values of the 'focus variable' are a sequence of percentiles
     pctls = np.linspace(0, 100, num_points).tolist()
--- a/statsmodels/sandbox/predict_functional.py
+++ b/statsmodels/sandbox/predict_functional.py
@@ -41,9 +41,9 @@ _predict_functional_doc =\
         A second set of summary functions used to define a contrast.
     values2 : dict-like
         A second set of fixed values used to define a contrast.
-    cvrg_prob : float
-        The coverage probability.
-    method : string
+    alpha : float
+        `1 - alpha` is the coverage probability.
+    ci_method : string
         The method for constructing the confidence band, one of
         'pointwise', 'scheffe', and 'simultaneous'.
     num_points : integer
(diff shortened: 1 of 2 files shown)


kshedden:
For OLS, the simultaneous confidence band is easily obtained using Scheffe's method.  For GLM's it may be much harder but here is one reference:

http://projecteuclid.org/euclid.aos/1016218225

josef-pkt:
Looks good overall, but we are getting different versions for similar functionality.
I put milestone 0.7 but we might not have it ready with a "reasonably stable" API to advertise it as an official function and method for the results classes.

kshedden:
Travis problems solved.  It was due to not specifying dtypes on dataframe creation, and was not an issue on any of my 3 local python installs.  Error messages were not helpful.

kshedden:
I've now added simultaneous confidence bands for both OLS and GLMs.  The simultaneous CB's make the code in this PR tailored for functional regression analyses.  I'm thinking that functional regression has enough specialized techniques that it will be better to have separate predict methods for functional and non-functional regressions.  But I am open to merging this with the other predict if there is a reasonable way to do it.

I'm fine with this staying in the sandbox detached from results classes for 0.7.   I and collaborators/students will be using this routinely once it's merged (need to fix #2178 to make it usable with survival regression).

There are 3 other methods for simultaneous GLM confidence bands in the Sun et al paper.  Some day I may go back and look at the others.  The one method that I implemented is the simplest.

Here are the updated plots created by the test script:

https://umich.box.com/shared/static/g2jx0wi5lo2r8liknpib.pdf

josef-pkt:
>  I'm thinking that functional regression has enough specialized techniques that it will be better to have separate predict methods for functional and non-functional regressions.

From the theoretical side I don't see any difference when using splines, given the knots it's still just a model linear in a finite number parameters (with the same asymptotics if we ignore endogenous knot selection).
Kernel estimators or other fully non-parametric estimators for the functional part would be different.

But I haven't looked at the simultaneous confidence intervals yet.

kshedden:
It's the confidence bands that are specialized to the functional case.
They assume that the points where the predictions are made fall on a
continuous path through the covariate space.  The predictions themselves
are generic.

josef-pkt:
I don't understand the simultaneity issue for the confidence band, but I think I will hold off with reading until I have merged a few more PRs.
and then start with Sun Loader '94 and Loader, Sun 1997 to avoid the complications coming from GLM.
Based on the abstract it still looks like a parametric versus non-parametric issue to me instead of just functional.

statsmodels/sandbox/predict_functional.py
```diff
@@ -0,0 +1,438 @@
...
+        ky = list(ky)
+        if len(ky) > 0:
+            raise ValueError("One or more variable names are contained in both `summaries` and `values`:" +
+                             ", ".join(ky))
+
+    return values, summaries, values2, summaries2
+
+
+def predict_functional(result, focus_var, summaries=None, values=None,
+                       summaries2=None, values2=None, cvrg_prob=0.95,
```
josef-pkt: we use `alpha=0.05` for the significance level of confidence intervals (one of the few greek letters) instead of `cvrg_prob=0.95`

statsmodels/sandbox/predict_functional.py
```diff
@@ -0,0 +1,438 @@
...
+                       method="pointwise", linear=True, num_points=10,
```
josef-pkt: method will also need an indication that it is for confidence interval confint_method or ci_method. I don't know if we use it already somewhere. It will be coming in several functions or methods.

josef-pkt:
@kshedden  Can you rebase and force push? and then I'll merge it as it is.

We can take it out of the sandbox in 0.8 when we have settled on the structure for the various predict functions and methods.

kshedden:
Rebased, waiting for green.  Let me know if you want me to squash commits.

josef-pkt:
about squashing: not really necessary given that you already rebased. You could squash the `travis-only failures` commits, but for the rest I prefer no squashing. (I think the extra information in commits is often useful.)

josef-pkt:
I had to restart one travis-ci job, seems to be build problems most likely not ours.

merging

PR #2180, opened 2015-01-12
Initial commit to fix issue 2178

### LINKED ISSUE #2178: ENH hazard regression prediction with formula
When hazard_regression was fit with a formula, the predict method doesn't accept a new exog data frame to be processed through the formula.  Needs refactoring.

kshedden:
This should close #2178.

### DIFF
--- a/statsmodels/duration/hazard_regression.py
+++ b/statsmodels/duration/hazard_regression.py
@@ -22,6 +22,60 @@ http://www.mwsug.org/proceedings/2006/stats/MWSUG-2006-SD08.pdf
 """
 
 
+_predict_docstring = """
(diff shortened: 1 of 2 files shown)


jseabold:
lgtm

jseabold:
FWIW, in the future, if you add `Closes #2178` to the commit message then it will automatically close the issue, those Josef also wants you to still list the PR # in the description.

kshedden:
Will do.  Thanks for the review.

josef-pkt:
looks good to me also
However, I don't like the inconsistent returns of predict across models, still TODO but this PR doesn't  affect or change it.

around 6 unrelated commits behind master, so still ok.

PR #2187, opened 2015-01-16
CLN: Explicit import for compatibility

bashtage:
Use explicit importing rather than try/except in python compatibility


bashtage:
Just reordered things to make imports explicit.

jseabold:
lgtm

PR #2188, opened 2015-01-18
ENH GEE equivalence class covariance structure enhancements

kshedden:
This PR has two main parts:

1. The equivalence classes can now be specified using a vector of labels, rather than by providing the pairs dictionary directly.  This is not as general, but much easier when it can be used.  For example, if there are two labels 'a' and 'b', the free parameters to be estimated are: cov(a, b), cov(a, a), cov(b, b), var(a) and var(b).  In this case cov(a, a) refers to the covariance between two different observations in the same group that fall into label class 'a', whereas var(a) is the variance.  cov(a, a) will generally be smaller than var(a).  An application where this would be useful is a situation where there are repeated measures on subjects, say of two biomarkers A and B measured at 3 time points, with 2 technical replicates at each time point.  Then there would be a covariance parameter between each A(s) and A(t) (with s and t denoting time), between each B(s) and B(t), and between each A(s) and B(t).  This is 6 + 6 + 9 parameters if I am counting right (there are more parameters for the A/B combinations since the order matters).

2. There is an option to calculate either the covariance or the correlation matrix.  Internally, GEE takes the correlation matrix and converts it to a covariance matrix by multiplying by the corresponding estimates of the standard deviation.  These estimates come from the mean structure (via the family varfunc).  Alternatively, we can estimate variances directly from the data, and pass the covariance matrix to GEE (which then does not rescale it).  Experience to date suggests that the two approaches generally perform similarly.  Intuition suggests that if the variance function is correctly specified (i.e. if the hypothesized mean/variance relationship holds), then it would be a bit more efficient to pass GEE a correlation function rather than trying to estimate the variance parameters directly.  But if there is unmodeled heteroscedasticity it may be picked up in the estimated covariance structure and provide a bit of robustness.  So given all this, I think it makes sense to provide the option to go either way.

Note also the addition of the n_pairs attribute to make it easy to see how much data is available to estimate each variance/covariance parameter.

### DIFF
--- a/statsmodels/genmod/cov_struct.py
+++ b/statsmodels/genmod/cov_struct.py
@@ -983,9 +983,6 @@ class Equivalence(CovStruct):
     level of the `pairs` dictionary to force the corresponding
     covariance to be zero.
 
-    The `pairs` dictionary is modified by the function so the user
-    should pass a copy if it is needed elsewhere.
-
     Examples
     --------
     The following sets up the `pairs` dictionary for a model with two
@@ -1010,7 +1007,9 @@ class Equivalence(CovStruct):
             raise ValueError("Equivalence cov_struct accepts only one of `pairs` and `labels`")
 
         if pairs is not None:
-            self.pairs = pairs
+            import copy
+            self.pairs = copy.deepcopy(pairs)
+
         if labels is not None:
             self.labels = np.asarray(labels)
 
@@ -1090,8 +1089,6 @@ class Equivalence(CovStruct):
                         i1 = i1[jj]
                         i2 = i2[jj]
                         pairs[g_lb][clabel] = (i1, i2)
-
-        self.pairs = pairs
 
 
     def initialize(self, model):
--- a/statsmodels/genmod/cov_struct.py
+++ b/statsmodels/genmod/cov_struct.py
@@ -1007,6 +1007,9 @@ class Equivalence(CovStruct):
         if (pairs is not None) and (labels is not None):
             raise ValueError("Equivalence cov_struct accepts only one of `pairs` and `labels`")
 
+        if self.model.weights is not None:
+            warnings.warn("weights not implemented for equalence cov_struct, using unweighted covariance estimate")
+
         if pairs is not None:
             import copy
             self.pairs = copy.deepcopy(pairs)
--- a/statsmodels/genmod/cov_struct.py
+++ b/statsmodels/genmod/cov_struct.py
@@ -1012,7 +1012,7 @@ class Equivalence(CovStruct):
         if pairs is not None:
             self.pairs = pairs
         if labels is not None:
-            self.labels = labels
+            self.labels = np.asarray(labels)
 
         self.as_cor = as_cor
 
--- a/statsmodels/genmod/cov_struct.py
+++ b/statsmodels/genmod/cov_struct.py
@@ -1,6 +1,7 @@
 from statsmodels.compat.python import iterkeys, itervalues, zip, range
 from statsmodels.stats.correlation_tools import cov_nearest
 import numpy as np
+import pandas as pd
 from scipy import linalg as spl
 from collections import defaultdict
 from statsmodels.tools.sm_exceptions import (ConvergenceWarning,
@@ -139,7 +140,6 @@ class CovStruct(object):
         """
 
         vmat, is_cor = self.covariance_matrix(expval, index)
-
         if is_cor:
             vmat *= np.outer(stdev, stdev)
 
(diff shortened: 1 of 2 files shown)


statsmodels/genmod/cov_struct.py
```diff
@@ -927,9 +927,22 @@ class Equivalence(CovStruct):
...
       Positions where j1 and j2 have the same value are variance
       parameters.
+    labels : array-like
+      An array of labels such that every distinct pair of labels
+      defines an equivalence class.  Either `labels` or `pairs` must
+      be provided.  When the two labels in a pair are equal two
+      equivalence classes are defined: one for the diagonal elements
+      (corresponding to variances) and one for the off-diagonal
+      elements (corresponding to covariances).
+    as_cor : boolean
```
josef-pkt: `as_cor` doesn't sound like a good name (I haven't looked enough to suggest an alternative.)

statsmodels/genmod/cov_struct.py
```diff
@@ -938,6 +951,14 @@ class Equivalence(CovStruct):
     using this covariance structure, missing data should be removed
     before constructing the model.
 
+    If using `labels`, after a model is defined using the covariance
+    structure it is possible to remove a label pair from the second
+    level of the `pairs` dictionary to force the corresponding
+    covariance to be zero.
```
josef-pkt: This is not clear to me (based on reading just through the changes)

statsmodels/genmod/cov_struct.py
```diff
@@ -938,6 +951,14 @@ class Equivalence(CovStruct):
...
+
+    The `pairs` dictionary is modified by the function so the user
+    should pass a copy if it is needed elsewhere.
```
josef-pkt: I need to think about this, in general we avoid changing original arrays in functions.

statsmodels/genmod/cov_struct.py
```diff
@@ -951,36 +972,130 @@ class Equivalence(CovStruct):
...
+        mat.sort(1)
+
+        # Remove repeated rows
+        try:
+            dtype = np.dtype((np.void, mat.dtype.itemsize * mat.shape[1]))
+            bmat = np.ascontiguousarray(mat).view(dtype)
+            _, idx = np.unique(bmat, return_index=True)
+        except TypeError:
+            # workaround for old numpy that can't call unique with complex
+            # dtypes
```
josef-pkt: why us this part necessary? 
(need to check: there was a problem for 1 or 2 numpy versions with return_xxx in np.unique. statsmodels.compat ?)

kshedden: I think it's only an issue because I am applying unique to what is
effectively a very long record constructed by concatenating all the bytes
in one row of an ndarray.  I'm not sure about the details, but there is
some discussion here:

http://stackoverflow.com/questions/16970982/find-unique-rows-in-numpy-array

statsmodels/genmod/cov_struct.py
kshedden: I'll make a copy so this isn't needed.

statsmodels/genmod/cov_struct.py
kshedden: Elements of the covariance matrices are 0 unless they are included in an
equivalence class.  When using labels, you don't have as much control over
what are the free parameters in the covariance structure.  For example, if
there are three label classes a, b and c, then you automatically get a/a,
b/b and c/c equivalence classes for variances, and a/b, a/c, and b/c
equivalence classes for covariances.  If for some reason you want to force
the b/c covariance to be zero, you would just delete the b/c label from all
the dicts.

statsmodels/genmod/cov_struct.py
kshedden: return_cor?
return_type='cor'?

statsmodels/genmod/cov_struct.py
josef-pkt: Thanks that's more understandable.
Is there an easier, more official way for users to do this, other than directly manipulating the pairs dictionary?
For example, is there a zero correlation label? `
or maybe we should have a `zero_label_pairs` keyword for those `b/c` combinations with zero correlation. (sounds more like an additional enhancement that might not be necessary for this PR.)

(BTW: when replying to github comments by email, please remove the quoted original text unless it is a used as a direct reference.)

statsmodels/genmod/cov_struct.py
josef-pkt: since it will be only binary, a boolean would be better?
If it's not a very frequently used option, then we could use `return_correlation` or `return_cov=False`.
One problem is that `cor` is not very common as an abbreviation, ,maybe it should be `corr`. (maybe the only similar case is `nearest_corr` or something like that in correlation_tools.py)

statsmodels/genmod/cov_struct.py
kshedden: I think there are some further convenience functions that can be added in a future PR.  One would be to remove an equivalence class so the parameter is estimated as zero.  Another would be to merge two equivalence classes so that the parameters become equated to each other.

josef-pkt:
@kshedden   I think we should merge this after a rebase, even without another or a more careful review. Since this cov_struct is new in 0.7, we don't have to worry about backwards compatibility, and we can still add improvements later. (I suspect the code can be streamlined and made faster, but I didn't look closely enough.)

We need the rebase because there is also the GEE weights PR merged into master since this branch has branched off. 
Do the weights affect the cov_struct, and does this PR need updating?

kshedden:
Rebased and added warning re weights not being used (same as most other cov structs right now).

josef-pkt:
(I canceled the travis-ci test run for the second to last commit, testing the last commit is enough, and we don't have to wait so long.)

I will merge when travis comes back green.

Thanks.

josef-pkt:
all green now, merging

PR #2204, opened 2015-01-25
Test: Test GLM IRLS vs Newton using all links

kshedden:
This fills some gaps in the GLM test code.  For (almost) every valid link for every family we test the IRLS results versus the Newton's method results (comparing params, bse, llf, and scale).

Mostly things look good.  For some non-canonical links the results are very sensitive to the input data, but I was eventually able to find something that worked, except for the inverse_square link.  This suggests that the calculations in the code are OK.  However we should look into having better support for difficult cases.  

Another issue is that we only get agreement between IRLS and Newton when using the expected information to calculate bse after the Newton fit.  But the default is the observed information.  Thus most users will see a difference when they make the comparison (although the difference seems to only occur for certain, mostly non-canonical links).

This is partly in response to discussion in #1961.  I don't think there are any fundamental problems with the code for the negative binomial family.  exp_lin_pred is just a temporary used to hold something that would otherwise be calculated twice.  It is not assumed to be the mean.

Another thing I noted is that the docstring says that cloglog is a valid link for Negative binomial, but I'm not sure about this, and I was not able to come up with a convergent example.

### DIFF
--- a/statsmodels/genmod/generalized_linear_model.py
+++ b/statsmodels/genmod/generalized_linear_model.py
@@ -732,16 +732,16 @@ class GLM(base.LikelihoodModel):
         using the scipy gradient optimizers.
         """
 
-        if max_start_irls > 0:
+        if (max_start_irls > 0) and (start_params is None):
             irls_rslt = self._fit_irls(start_params=start_params, maxiter=maxiter,
                                        tol=tol, scale=scale, cov_type=cov_type,
                                        cov_kwds=cov_kwds, use_t=use_t, **kwargs)
             start_params = irls_rslt.params
 
-        # TODO: pass more into fit here
         rslt = super(GLM, self).fit(start_params=start_params, tol=tol,
                                     maxiter=maxiter, full_output=full_output,
                                     method=method, disp=disp, **kwargs)
+
         self.mu = self.predict(rslt.params)
         self.scale = self.estimate_scale(self.mu)
 
--- a/statsmodels/genmod/generalized_linear_model.py
+++ b/statsmodels/genmod/generalized_linear_model.py
@@ -631,9 +631,9 @@ class GLM(base.LikelihoodModel):
         else:
             return self.family.fitted(linpred)
 
-    def fit(self, start_params=None, maxiter=100, method='IRLS',
-            tol=1e-8, scale=None, cov_type='nonrobust', cov_kwds=None,
-            use_t=None, full_output=True, disp=False, **kwargs):
+    def fit(self, start_params=None, maxiter=100, method='IRLS', tol=1e-8,
+            scale=None, cov_type='nonrobust', cov_kwds=None, use_t=None,
+            full_output=True, disp=False, max_start_irls=3, **kwargs):
         """
         Fits a generalized linear model for a given family.
 
@@ -673,6 +673,10 @@ class GLM(base.LikelihoodModel):
         disp : bool, optional
             Set to True to print convergence messages.  Not used if method is
             IRLS.
+        max_start_irls : int
+            The number of IRLS iterations used to obtain starting
+            values for gradient optimization.  Only relevant if
+            `method` is set to something other than 'IRLS'.
 
         Notes
         -----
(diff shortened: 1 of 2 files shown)


josef-pkt:
Thanks @kshedden  that's very useful

As far as I remember, Stata also has eim as default for IRLS and oim for MLE optimization. Quickly browsing the Stata documentation indicates this is correct but I didn't see an explicit statement.
Also parts of the sandwiches in IRLS depend on eim instead of oim in Stata, IIUC.

It will be easier on users once the `cov_type = 'oim'` and `cov_type='eim'` are added so we can choose this independently of the optimization algorithm.
Another recommendation would be to run IRLS first and then finish up with optimize MLE, `'newton'`, to get the oim covariance matrix.
Stata is switching between an initial use of eim to the final use of oim in the optimization.
(You had this also in the benchmark in another issue/PR.)

All this is only relevant for non-canonical links where eim and oim differ.

The PR looks good to me, after quickly reading through the changes.

One thing that isn't completely clear to me: Does the choice of IRLS/fisher scoring/eim versus using newton/newton-raphson/oim affect the convergence of the parameters? The optimization, IRLS or optimize, uses the information of second derivatives, oim or eim, but the loglike/deviance and the score should still be the same. I'm not sure how this works out.

Do you have specific cases or examples where the optimization is "sensitive" or difficult, that we could use to investigate what's going on, and how to improve it?

statsmodels/genmod/tests/test_glm.py
```diff
@@ -855,42 +855,85 @@ def test_gradient_irls():
...
+
+           mod_gradient = sm.GLM(endog, exog, family=family_class(link=link))
+           rslt_gradient = mod_gradient.fit(start_params=rslt_irls.params, method="newton")
+
+           assert_allclose(rslt_gradient.params,
+                           rslt_irls.params, rtol=1e-6, atol=1e-6,
+                           err_msg=str(family_class) + " params")
+
+           assert_allclose(rslt_gradient.llf, rslt_irls.llf,
+                           rtol=1e-6, atol=1e-6, err_msg=str(family_class) + " params")
```
josef-pkt: is `" params"` copy paste without changing the name?
Wouldn't the error message would already show `assert_allclose(rslt_gradient.llf, rslt_irls.llf,` with the attribute name?

josef-pkt:
Given that Binomial with 2-D endog (successes, failures) is an outlier in the pattern it would be good to add it also.
I don't see an easy way to add it here, except for adding a `version=1` keyword to `gen_endog` and in the list of cases.

kshedden:
I added a test for 2-d binomial data and cleaned up the error messages.

We don't have good starting values for gradient optimization now, so most of the tests in this PR would fail if we did not start the gradient optimization at the IRLS solution.  That's something to work on.

Also, sometimes for non-canonical links the gradient optimization fails even when we start at the IRLS solution.  I think this is because the clipping for out-of-domain values doesn't work consistently between the gradients and IRLS.  I will make a specific example and open an issue.

josef-pkt:
(without checking the source)

I think what we should do, for at least now, is to use the IRLS as start_params for the gradient based optimization by default, maybe with an extra `max_start_irls` (or similar) keyword.
I don't think we can come up with a much better generic solution, and can still add special case start_params later.

(adding a new GLM method `start_params_optimize` or similar ?)

kshedden:
Good, idea.  We now use IRLS by default to get ascent gradient starting values.  All the test examples run fine using just a single IRLS iteration before switching to gradient optimization.  I have set it to 3 iterations, but perhaps it should be just 1?

Also, since we only need the params from IRLS we should set the cov_type and other options in the irls fit to whatever is cheapest.  I'm not sure what that would be.

statsmodels/genmod/generalized_linear_model.py
```diff
@@ -714,17 +718,26 @@ def fit(self, start_params=None, maxiter=100, method='IRLS',
...
                       disp=True, scale=None, cov_type='nonrobust',
-                      cov_kwds=None, use_t=None, **kwargs):
+                      cov_kwds=None, use_t=None, max_start_irls=3,
+                      **kwargs):
         """
         Fits a generalized linear model for a given family iteratively
         using the scipy gradient optimizers.
         """
 
+        if max_start_irls > 0:
```
josef-pkt: this needs to check for user provided start_params which shouldn't be overwritten
`if start_params is None and max_start_irls > 0:`

kshedden: My thinking was the if there are user-provided start_params then they would
be starting values for the IRLS, and if the user wants it to go straight to
gradient they would set max_start_irls=0.

josef-pkt: I thought about that, but I think it's confusing if the method is "newton" and not "irls-newton".

Conceptually, the simple picture is that we use IRLS to get starting values. If we already have starting values, then we don't need IRLS. 
If I already have starting values that are pretty good, for example from a previous run, then I don't need IRLS and I don't want to have to check other keywords to see whether newton is really just newton in this model.

In this case it's just being picky and I think we could decide on either.
Currently we don't have anything like this, except maybe in tsa. The discrete model use the estimate of a simplified model to get the start_params if they are None. 
But we want to expand hybrid and sequence of optimizers to a general pattern, Skipper has a sequence optimization PR open, so we need to decide on some consistent rules.

My rule would be that, if I specify a method, then the start params are for that method. If I specify a hybrid method, then the start params are for the first optimizer in the hybrid.

kshedden:
I changed the behavior so that if the user provides start_params then the IRLS call is skipped.

josef-pkt:
Travis says "All is well", merging

Thanks Kerby

PR #2240, opened 2015-02-10
CLN: Simplify add_constant and add_trend

bashtage:
Simplified these two functions to handle NumPy arrays and recarrays as
well as pandas Series and DataFrames.  Removed bugs in add_trend.
Added complete test coverage for both functions.

### DIFF
--- a/statsmodels/tools/tools.py
+++ b/statsmodels/tools/tools.py
@@ -253,10 +253,14 @@ def add_constant(data, prepend=True, has_constant='skip'):
 
     Returns
     -------
-    data : array
+    data : array, recarray or DataFrame
         The original values with a constant (column of ones) as the first or
-        last column.
+        last column. Returned value depends on input type.
 
+    Notes
+    -----
+    When the input is recarray or a pandas Series or DataFrame, the added
+    column's name is 'const'.
     """
     is_recarray = _is_recarray(data)
     if _is_using_pandas(data, None) or is_recarray:
@@ -290,9 +294,9 @@ def add_constant(data, prepend=True, has_constant='skip'):
         else:
             data['const'] = const
     else:
-        data = np.column_stack((data, np.ones((data.shape[0], 1))))
-        if prepend:
-            return np.roll(data, 1, 1)
+        order = 1 if prepend else -1
+        data = [np.ones((data.shape[0], 1)), data]
+        data = np.column_stack(data[::order])
 
     if is_recarray:
         data = data.to_records(index=False, convert_datetime64=False)
--- a/statsmodels/tools/tools.py
+++ b/statsmodels/tools/tools.py
@@ -261,8 +261,28 @@ def add_constant(data, prepend=True, has_constant='skip'):
     When the input is recarray or a pandas Series or DataFrame, the added
     column's name is 'const'.
     """
-    from statsmodels.tsa.tsatools import add_trend
-    return add_trend(data, trend='c', prepend=prepend, has_constant=has_constant)
+    if _is_using_pandas(data, None) or _is_recarray(data):
+        from statsmodels.tsa.tsatools import add_trend
+        return add_trend(data, trend='c', prepend=prepend, has_constant=has_constant)
+
+    # Special case for NumPy
+    x = np.asanyarray(data)
+    if x.ndim == 1:
+        x = x[:,None]
+    elif x.ndim > 2:
+        raise ValueError('Only implementd 2-dimensional arrays')
+
+    is_nonzero_const = np.ptp(x, axis=0) == 0
+    is_nonzero_const &= np.all(x != 0.0, axis=0)
+    if is_nonzero_const.any():
+        if has_constant == 'skip':
+            return x
+        elif has_constant == 'raise':
+            raise ValueError("data already contains a constant")
+
+    x = [np.ones(x.shape[0]), x]
+    x = x if prepend else x[::-1]
+    return np.column_stack(x)
 
 
 def isestimable(C, D):
--- a/statsmodels/tools/tools.py
+++ b/statsmodels/tools/tools.py
@@ -9,7 +9,6 @@ from scipy.linalg import svdvals
 from statsmodels.datasets import webuse
 from statsmodels.tools.data import _is_using_pandas, _is_recarray
 from statsmodels.compat.numpy import np_matrix_rank
-from pandas import DataFrame
 
 
 def _make_dictnames(tmp_arr, offset=0):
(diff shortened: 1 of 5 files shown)


bashtage:
Found old issue related to these changes, so closed it too.

statsmodels/tools/tools.py
```diff
@@ -233,107 +233,70 @@ def categorical(data, col=None, dictnames=False, drop=False, ):
...
+        data = np.column_stack((data, np.ones((data.shape[0], 1))))
         if prepend:
-            ones = np.ones((data.shape[0], 1), dtype=[('const', float)])
-            data = nprf.append_fields(ones, data.dtype.names,
-                                      [data[i] for i in data.dtype.names],
-                                      usemask=False, asrecarray=return_rec)
-        else:
-            data = nprf.append_fields(data, 'const', np.ones(data.shape[0]),
-                                      usemask=False, asrecarray=return_rec)
+            return np.roll(data, 1, 1)
```
josef-pkt: I think you should use the explicit `if prepend` paths.
`roll` will need to make a copy, or linalg will need to make a copy when it requires contiguous memory
also `column_stack` creates Fortran ordered arrays which should be better for the Fortran linalg calls, although we never verify this.

statsmodels/tsa/tsatools.py
```diff
@@ -36,54 +37,62 @@ def add_trend(X, trend="c", prepend=False, has_constant='skip'):
...
-        else:
-            X = np.column_stack((trendarr, X))
+        trendarr = trendarr[:, 1]
+
+    if "c" in trend and np.any(np.ptp(np.asanyarray(X), axis=0) == 0):
+        if has_constant == 'raise':
+            raise ValueError("X already contains a constant")
+        elif has_constant == 'add':
+            pass
+        elif has_constant == 'skip' and trend == "ct":
```
josef-pkt: this need to handle also "ctt", i.e. `trend in ["ct", "ctt"]` or  `trend_order > 0`

josef-pkt:
I need to look at it again, but should be possible for 0.7

travis failure in `webuse` test, unrelated to this, and I need to fix that fast
`urllib.error.HTTPError: HTTP Error 503: Service Unavailable`

bashtage:
`webuse` has issues.

josef-pkt:
you mean the failing tests for `webuse` ?
They are disabled in master now. If you rebase, then they won't bother this PR anymore.

josef-pkt:
Maybe this is simplifying too much.

AFAICS from one example that I tried, recarray -> DataFrame -> recarray doesn't roundtrip for string (but I didn't run the code of this PR)

```
>>> dt = np.dtype([('',int),('','<S4'), ('', np.float32), ('',np.float64)])
>>> xr = np.array([(1, 'aaa', 5.4, 1.2),(1, 'aaa', 5.4, 1.2)], dt).view(np.recarray)
>>> xr
rec.array([(1, 'aaa', 5.400000095367432, 1.2),
       (1, 'aaa', 5.400000095367432, 1.2)], 
      dtype=[('f0', '<i4'), ('f1', '|S4'), ('f2', '<f4'), ('f3', '<f8')])
>>> xrp = pd.DataFrame.from_records(xr)
>>> xrp.dtypes
f0      int32
f1     object
f2    float32
f3    float64
dtype: object
>>> xrp.to_records()
rec.array([(0L, 1, 'aaa', 5.400000095367432, 1.2),
       (1L, 1, 'aaa', 5.400000095367432, 1.2)], 
      dtype=[('index', '<i8'), ('f0', '<i4'), ('f1', '|O4'), ('f2', '<f4'), ('f3', '<f8')])
```

is there an option to roundtrip, i.e. not change the dtype?

Also I think the pandas detour will need to make some extra copies because the memory layout doesn't match up if there are different dtypes involved. However, given that these are mainly convenience function I don't think that's a blocker.

**If** we still have users that use recarrays, then changing the dtype will, most likely, break their code.

bashtage:
This might not be popular, but I removed all of `add_constant` and made it a shallow wrapper for `add_trend`.  The code to handle mixed recarrays or DataFrames was somewhat complex, and having 2 copies did not seem like a good idea.  These functions were already extremely similar, and `add_trend` strictly nests `add_constant`.

josef-pkt:
You can short-circuit our most common advertised use case: add_constant to ndarray
i.e. `if trend='c' and type(x) is np.ndarray: return column_stack(..)`

(I won't have time until later to look at it.)

bashtage:
> You can short-circuit our most common advertised use case: add_constant to ndarray
> i.e. if trend='c' and type(x) is np.ndarray: return column_stack(..)

Unfortunately because of `has_constant` more checks would be needed. Almost all of the function is skipped when X is an `array` so I don't think this saves much in terms of time.

josef-pkt:
I'd like to postpone this to 0.8. It has changes that I don't like to rush in just before a release, and I'm still not sure about some of the details, even though it looks good overall. For some changes, I don't know if they are innocent and acceptable or not.

bashtage:
> is there an option to roundtrip, i.e. not change the dtype?

This works as it should in the PR.  `S4` stays `S4`.  Not that anyone should be ever using recarrays in 2016, except when essential to interface with C code (e.g. HDF5).

statsmodels/tsa/tsatools.py
```diff
@@ -238,25 +273,20 @@ def detrend(x, order=1, axis=0):
...
-    x = np.asarray(x)
+    """
+    if x.ndim == 2 and int(axis) == 1:
+        x = x.T
+    elif x.ndim > 2:
+        raise NotImplementedError('x.ndim > 2 is not implemented until it is needed')
+    # could use a polynomial, but this should work also with 2d x, but maybe not yet
     nobs = x.shape[0]
-    if order == 0:
-        return x - np.expand_dims(x.mean(axis), axis)
```
josef-pkt: I would like to keep this. There is no reason to go through linalg to demean.

statsmodels/tools/tools.py
```diff
@@ -233,108 +232,37 @@ def categorical(data, col=None, dictnames=False, drop=False, ):
...
+    -----
+    When the input is recarray or a pandas Series or DataFrame, the added
+    column's name is 'const'.
     """
-    if _is_using_pandas(data, None):
-        # work on a copy
-        return _pandas_add_constant(data.copy(), prepend, has_constant)
-    else:
-        data = np.asarray(data)
-    if not data.dtype.names:
```
josef-pkt: I think we should also handle the plain numpy array case here. 
It's the most common case, and it's a long way to track the code to where it is done in the current changes, when all we need is essentially a np.column_stack, even with a bit of code duplication.

I would also check for constants only if has_constant != 'add'

bashtage:
I added the special case, even though it takes quite a bit of code (to handle exception cases).

statsmodels/tools/tools.py
```diff
@@ -233,108 +232,57 @@ def categorical(data, col=None, dictnames=False, drop=False, ):
...
+        return add_trend(data, trend='c', prepend=prepend, has_constant=has_constant)
+
+    # Special case for NumPy
+    x = np.asanyarray(data)
+    if x.ndim == 1:
+        x = x[:,None]
+    elif x.ndim > 2:
+        raise ValueError('Only implementd 2-dimensional arrays')
+
+    is_nonzero_const = np.ptp(x, axis=0) == 0
```
josef-pkt: this part could be skipped, using `if has_constant != 'add'`

statsmodels/tsa/tsatools.py
```diff
@@ -238,25 +273,25 @@ def detrend(x, order=1, axis=0):
...
+        raise NotImplementedError('x.ndim > 2 is not implemented until it is needed')
 
-
-    '''
-    x = np.asarray(x)
     nobs = x.shape[0]
     if order == 0:
-        return x - np.expand_dims(x.mean(axis), axis)
+        # Special case demean
+        resid = x - x.mean(axis=0)
```
josef-pkt: This ignores axis argument.

josef-pkt: No, it uses axis argument because internally x is transposed if needed

josef-pkt:
also has now merge conflict.

one small speedup, otherwise looks good to me and ready to merge

statsmodels/tsa/tsatools.py
bashtage: Yes

bashtage:
Rebased and squashed a few pep8 style warnings

josef-pkt:
this failed on all travis runs

`TypeError: add_trend() got an unexpected keyword argument 'X'`

bashtage:
Yes - semi-zealous pep8 cleaning missed some keyword useage in testing for
a required argument.

josef-pkt:
this should also close #2656 #2659   supersedes/closes #2665

bashtage:
Green.

josef-pkt:
merged

@bashtage  Thank you

statsmodels/tools/tools.py
```diff
@@ -233,108 +232,57 @@ def categorical(data, col=None, dictnames=False, drop=False, ):
...
+    is_nonzero_const &= np.all(x != 0.0, axis=0)
```
josef-pkt: another small speedup: if ptp == 0, then all values are the same and we only need to check 1 element of x: x[0] != 0

PR #2289, opened 2015-02-22
GEE stationary covariance structure

kshedden:
This is a new covariance structure for GEE.  When grid=True it should be identical to the stationary covariance in Stata.  We also have grid=False, which allows unequal spacings of values.

### DIFF
--- a/statsmodels/genmod/cov_struct.py
+++ b/statsmodels/genmod/cov_struct.py
@@ -616,10 +616,10 @@ class Stationary(CovStruct):
         if self.grid == False:
             return super(Stationary, self).covariance_matrix_solve(expval, index, stdev, rhs)
 
-        from statsmodels.tools.linalg import toeplitz_solve
+        from statsmodels.tools.linalg import stationary_solve
         r = np.zeros(len(expval))
         r[0:self.max_lag] = self.dep_params
-        return [toeplitz_solve(r, x) for x in rhs]
+        return [stationary_solve(r, x) for x in rhs]
 
 
     update.__doc__ = CovStruct.update.__doc__
--- a/statsmodels/genmod/cov_struct.py
+++ b/statsmodels/genmod/cov_struct.py
@@ -487,16 +487,166 @@ class Nested(CovStruct):
         return msg
 
 
+class Stationary(CovStruct):
+    """
+    A stationary covariance structure.
+
+    The correlation between two observations is an arbitrary function
+    of the distance between them.  Distances up to a given maximum
+    value are included in the covariance model.
+
+    Parameters
+    ----------
+    max_lag : float
+        The largest distance that is included in the covariance model.
+    grid : bool
+        If True, the index positions in the data (after dropping missing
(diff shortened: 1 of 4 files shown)


argriffing:
Maybe this is obvious to the statsmodels audience, but I'd be interested to see a few more specific details in the `linalg.toeplitz_solve` docstring.  A [Toeplitz matrix](http://en.wikipedia.org/wiki/Toeplitz_matrix) in linear algebra is not necessarily positive definite, or positive semi-definite or even symmetric.  But when I see the word `correlation` in the docstring, this makes me think that the `linalg.toeplitz_solve` function may have some of these constraints?

kshedden:
This is just the classic Durbin/Levinson algorithm.  I thought it might be in a Statsmodels dependency (or Statsmodels itself) but I couldn't find it.  The matrix is assumed to have unit diagonal, and is symmetric.  The unit diagonal property could easily be dispensed with but that's what I needed.  I believe PSD is required but I may be wrong.

argriffing:
Please please add the constraints to the docstring.  Different variants of the classic Durbin/Levinson algorithm have different constraints.  I wasted a bunch of time trying to contribute an implementation of Wikipedia's [Levinson-Durbin](http://en.wikipedia.org/wiki/Levinson_recursion) definition [here](https://github.com/scipy/scipy/pull/2754) and it was stalled partly because @josef-pkt expected it to mean the same thing as the variant that he had learned, which had different constraints.  There are multiple variants, so please at least document the constraints for your implementation.

josef-pkt:
`levinson_durbin` is in `tsa.stattools`

josef-pkt:
@argriffing   I was arguing that I would like to have more returns, not against the toeplitz_solve or levinson-durbin version itself. In our time series context the toeplitz matrix is always symmetric and positive definite at least in all applications that I know of.

argriffing:
@kshedden @josef-pkt In this thread, I'm only asking that the constraints for your function be documented.  It's great that in your time series context the toeplitz matrix is always symmetric and positive definite, but this is an extra constraint on top of the constraint implied by the toeplitz structure.  I'm not suggesting that anything about your functions be changed, but please if you have a function called `statsmodels.linalg.toeplitz_solve`, please document that it does not work for all [Toeplitz](http://en.wikipedia.org/wiki/Toeplitz_matrix) matrices, only the ones that are symmetric positive definite, if that is the case.  Or only the matrices that are symmetric positive definite and have ones on the diagonal, if that is the constraint.

josef-pkt:
@kshedden  Because I'm busy with vacationing this week, I'm falling behind in checking merge candidates for 0.7 before the scipy sprint.

Can this be merged?
#2307 same question
all three, including #2363, will need a rebase on current master

Because GEE and MixedLM are only one version old, I guess it's better to get the refactoring in as early as possible, but I never checked much of the details in these three PRs.

kshedden:
Yes, I think all three are clean and safe to merge.  I can rebase them.

PR #2297, opened 2015-03-01
Local FDR, estimation of null Z-score distribution

kshedden:
This PR implements two techniques from the paper:

B Efron (2008).  Microarrays, Empirical Bayes, and the Two-Groups Model.  Statistical Science 23:1, 1-22.                                                                               

The ``local_fdr`` function calculates local False Discovery rates from Z-scores.  We could possibly implement this as an option within ``multipletests``, but it's nice to have the option to operate on z-scores rather than p-values, which ``mutlipletests`` does not allow.

The  ``empirical_null`` function estimates a non-standard normal distribution based on a set of Z-scores that consists of an unknown mixture of null and non-null hypotheses.  It uses a likelihood-based version of the "central matching" idea.

See also #1371.

### DIFF
--- a/statsmodels/stats/multitest.py
+++ b/statsmodels/stats/multitest.py
@@ -493,8 +493,7 @@ def local_fdr(zscores, p0=1.0, null_density=None, deg=7,
     zbins = (bins[:-1] + bins[1:]) / 2
 
     # The design matrix at bin centers
-    dmat = [(zbins**k)[:,None] for k in range(0, deg + 1)]
-    dmat = np.concatenate(dmat, axis=1)
+    dmat = np.vander(zbins, deg + 1)
 
     # Use this to get starting values for Poisson regression
     md = OLS(np.log(1 + zhist), dmat).fit()
@@ -503,8 +502,7 @@ def local_fdr(zscores, p0=1.0, null_density=None, deg=7,
     md = GLM(zhist, dmat, family=families.Poisson()).fit(start_params=md.params)
 
     # The design matrix for all Z-scores
-    dmat_full = [(zscores**k)[:,None] for k in range(0, deg + 1)]
-    dmat_full = np.concatenate(dmat_full, axis=1)
+    dmat_full = np.vander(zscores, deg + 1)
 
     # The height of the estimated marginal density of Z-scores,
     # evaluated at every observed Z-score.
--- a/statsmodels/stats/multitest.py
+++ b/statsmodels/stats/multitest.py
@@ -457,7 +457,7 @@ def local_fdr(zscores, p0=1.0, null_density=None, deg=7,
         A vector of Z-scores
     p0 : float
         The assumed proportion of true null hypotheses
-    null_density : function (real to positive real)
+    null_density : function mapping reals to positive reals
         The density of null Z-scores; if None, use standard normal
     deg : integer
         The maximum exponent in the polynomial expansion of the
@@ -475,6 +475,17 @@ def local_fdr(zscores, p0=1.0, null_density=None, deg=7,
     ----------
     B Efron (2008).  Microarrays, Empirical Bayes, and the Two-Groups
     Model.  Statistical Science 23:1, 1-22.
+
+    Examples
+    --------
+    Basic use:
+
+    >>> fdr = local_fdr(zscores)
+
+    Example using an empirical null distribution estimated from the data:
+
+    >>> null = EmpiricalNull(zscores)
+    >>> fdr = local_fdr(zscores, null_density=null.pdf)
     """
 
     from statsmodels.genmod.generalized_linear_model import GLM
--- a/statsmodels/stats/multitest.py
+++ b/statsmodels/stats/multitest.py
@@ -522,8 +522,7 @@ def local_fdr(zscores, p0=1.0, null_density=None, deg=7,
     return fdr
 
 
-def empirical_null(zscores, null_lb=-1, null_ub=1, estimate_mean=True,
-                   estimate_scale=True, estimate_prob=False):
+class EmpiricalNull(object):
     """
     Estimate empirical distribution of null Z-scores.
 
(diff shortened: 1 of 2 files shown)


josef-pkt:
I will look at it in more detail most likely after next week (spring break for elementary schools here).
Based on the description I have seen something similar in nipy a long time ago, maybe not very close in the details.

two questions after a quick browse

Is there anything specific that exploits z-values, or is it just using the normal instead of the uniform distribution as reference distribution?  (p-values as a integral transform)

Would it be better to convert `empirical_null` to a class so we don't need to hide the nested functions, and can possibly get more returns if needed?

kshedden:
Good catch.  It's here:

http://nipy.org/nipy/stable/labs/enn.html

I'll look into it.

statsmodels/stats/multitest.py
```diff
@@ -444,3 +444,197 @@ def fdrcorrection_twostage(pvals, alpha=0.05, method='bky', iter=False,
...
+    bins = np.linspace(minz, maxz, nbins)
+
+    # Bin counts
+    zhist = np.histogram(zscores, bins)[0]
+
+    # Bin centers
+    zbins = (bins[:-1] + bins[1:]) / 2
+
+    # The design matrix at bin centers
+    dmat = [(zbins**k)[:,None] for k in range(0, deg + 1)]
```
josef-pkt: same as 
`np.vander(zbins, deg+?)`  or vectorized `zbins[:, None]**np.arange(deg + 1)`
?

statsmodels/stats/multitest.py
```diff
@@ -444,3 +444,197 @@ def fdrcorrection_twostage(pvals, alpha=0.05, method='bky', iter=False,
...
+
+    # Use this to get starting values for Poisson regression
+    md = OLS(np.log(1 + zhist), dmat).fit()
+
+    # Poisson regression
+    md = GLM(zhist, dmat, family=families.Poisson()).fit(start_params=md.params)
+
+    # The design matrix for all Z-scores
+    dmat_full = [(zscores**k)[:,None] for k in range(0, deg + 1)]
+    dmat_full = np.concatenate(dmat_full, axis=1)
```
josef-pkt: same here np.vander
In general: if there is no special meaning to the units of the polynomial, then using normalized np.polynomial.vander or chebyvander would be numerically more stable. (might not matter here if zscores are standard normal and don't cover huge interval of numbers)

statsmodels/stats/multitest.py
```diff
@@ -444,3 +444,197 @@ def fdrcorrection_twostage(pvals, alpha=0.05, method='bky', iter=False,
...
+        cp = p0 * central_mass
+
+        # Binomial term
+        rval = n_zs0 * np.log(cp) + (n_zs - n_zs0) * np.log(1 - cp)
+
+        # Truncated Gaussian term for null Z-scores
+        zv = (zscores0 - d0) / s0
+        rval += np.sum(-zv**2 / 2) - n_zs0 * np.log(s0)
+        rval -= n_zs0 * np.log(central_mass)
+
```
josef-pkt: just a note: this looks like trimmed estimation based on fixed thresholds (null_lb, null_ub).

How do we choose null_lb, null_ub?

kshedden:
Regarding p-values versus z-scores: the local FDR is basically the ratio of the height of the density of null z-scores to the height of the density of all z-scores.  So it is not natural to reformulate it in terms of p-values.  The  density ratio would be different after transforming to p-values.  Another difference is that positive and negative effects are treated separately in local FDR.

statsmodels/stats/multitest.py
kshedden: I think the idea is to choose a central region that contains very few alternative hypothesis Z-scores.  I don't know whether people have looked closely at the best way to do this central matching.  I can't recall seeing anything other than -1 to 1.

josef-pkt:
I'm looking at PRs that should be merged.

Is this ready for review and merge or are there still changes? Is this implementation ok relative to the one in nipy?  
I only know the basic idea about this and never went through the details.

kshedden:
The model and parameters are the same between our approach and the Nipy approach.  But the estimators seem to be different.  I am following Efron's approach here, which uses maximum likelihood (restricted to the central part of the distribution).  This MLE does not have a closed form, that is why I need to call `minimize`.  The Nipy code is not iterative  It probably uses curvature matching around the mode, or perhaps some type of moment matching.  It' not that clear from the code, but both these approaches have been used in other FDR work.

statsmodels/stats/multitest.py
```diff
@@ -444,3 +444,237 @@ def fdrcorrection_twostage(pvals, alpha=0.05, method='bky', iter=False,
...
+        mean is fixed at zero.
+    estimate_scale : bool
+        If True, estimate the scale of the distribution.  If False, the
+        scale parameter is fixed at 1.
+    estimate_prob : bool
+        If True, estimate the proportion of true null hypotheses (i.e.
+        the proportion of z-scores with expected value zero).  If False,
+        the probability parameter is fixed at 1.
+
+    Returns
```
josef-pkt: section header should be `Attributes`

statsmodels/stats/multitest.py
```diff
@@ -444,3 +444,237 @@ def fdrcorrection_twostage(pvals, alpha=0.05, method='bky', iter=False,
...
+    -------
+    mean0 : float
+        The estimated mean of the empirical null distribution
+    scale0 : float
+        The estimated standard deviation of the empirical null distribution
+    prob0 : float
+        The estimated proportion of true null hypotheses
+    null_density : function (from reals to positive reals)
+        The fitted empirical null density function
```
josef-pkt: AFAICS this is not an attribute but the `pdf` method which will be automatically documented (methods for classes)

statsmodels/stats/multitest.py
```diff
@@ -444,3 +444,237 @@ def fdrcorrection_twostage(pvals, alpha=0.05, method='bky', iter=False,
         return reject, pvalscorr_, ntests - ri, alpha_stages
     else:
         return rej, pvalscorr, ntests - ri, alpha_stages
+
+
+def local_fdr(zscores, p0=1.0, null_density=None, deg=7,
+              nbins=30):
```
josef-pkt: It would be better to replace `p0` with `null_prop` or `null_proportion` (with prefix "null" similar to `null_density`)
An alternative would be to switch to post-fix `"_null"` in all names, but in this case I don't have a real preference.

we don't use "density" much in names, I guess `null_pdf` or `pdf_null` would be better, because we always refer to the function or method by name `pdf`

statsmodels/stats/multitest.py
```diff
@@ -444,3 +444,237 @@ def fdrcorrection_twostage(pvals, alpha=0.05, method='bky', iter=False,
...
+            rval += np.sum(-zv**2 / 2) - n_zs0 * np.log(s0)
+            rval -= n_zs0 * np.log(central_mass)
+
+            return -rval
+
+
+        # Estimate the parameters
+        from scipy.optimize import minimize
+        # starting values are mean = 0, scale = 1, p0 ~ 1
+        mz = minimize(fun, np.r_[0., 0, 3], method="Nelder-Mead")
```
josef-pkt: the minimization uses full params even if some are fixed. It might not be a problem in this simple optimization problem, but should have a unit test (if mean, variance or both are fixed).

statsmodels/stats/multitest.py
```diff
@@ -444,3 +444,237 @@ def fdrcorrection_twostage(pvals, alpha=0.05, method='bky', iter=False,
...
+        If True, estimate the proportion of true null hypotheses (i.e.
+        the proportion of z-scores with expected value zero).  If False,
+        the probability parameter is fixed at 1.
+
+    Returns
+    -------
+    mean0 : float
+        The estimated mean of the empirical null distribution
+    scale0 : float
+        The estimated standard deviation of the empirical null distribution
```
josef-pkt: Do we still have inconsistent `scale` definitions. i.e. variance versus standard deviation?
(scale is squared in models, but without squaring in scipy.stats distributions)

statsmodels/stats/tests/test_multi.py
```diff
@@ -348,3 +351,42 @@ def test_tukeyhsd():
...
+    zs = np.concatenate((z0, z1))
+    emp_null = EmpiricalNull(zs, estimate_prob=True)
+
+    assert_allclose(emp_null.mean0, 0, atol=1e-5, rtol=1e-5)
+    assert_allclose(emp_null.scale0, 1, atol=1e-5, rtol=1e-2)
+    assert_allclose(emp_null.prob0, 0.98, atol=1e-5, rtol=1e-2)
+
+    # smoke tests
+    emp_null.pdf(0.)
+    emp_null.pdf(np.r_[-1, 0, 1])
```
josef-pkt: consistency instead of smoke test, something like
```
from scipy.stats import norm
assert_allequal(emp_null.pdf(np.r_[-1, 0, 1]), norm.pdf(np.r_[-1, 0, 1], loc=emp_null.mean0, scale=emp_null.scale0, rtol=1e-13)
```

josef-pkt:
A few smaller comments, otherwise looks good to merge to me.

I guess we can expand this and share code with similar tasks in future without backwards compatibility problems in the API. The API looks pretty straight forward, but several methods remind me of other things that we don't have yet. 


One question on usage: Is there a recommendation for how to z-score the data before calling these functions. e.g. do we need to use a robust variance estimator?
The main usage question, I guess, is how to choose the bounds relative to the z-scoring scale.

kshedden:
I think I have accounted for most of these comments.  I changed the minimizer to only use the free parameters.

Regarding usage, I think people usually use this approach with a moment-based Z-score.  I suppose it could be used with a robust scale estimate but I've never tried that.

statsmodels/stats/multitest.py
```diff
@@ -444,3 +444,241 @@ def fdrcorrection_twostage(pvals, alpha=0.05, method='bky', iter=False,
...
+
+        # Estimate the parameters
+        from scipy.optimize import minimize
+        # starting values are mean = 0, scale = 1, p0 ~ 1
+        mz = minimize(fun, np.r_[0., 0, 3], method="Nelder-Mead")
+        mean0, sd0, prob0 = xform(mz['x'])
+
+        self.mean0 = mean0
+        self.sd0 = sd0
+        self.prob0 = prob0
```
josef-pkt: I think we should also here switch to an explicit `null`   (0 could just be a numbering, in general)
`mean_null` `sd_null` `prob_null` ?
or maybe without `0` and without `null` since the name already says it's the Null distribution, and we don't need to add any `_null` to pdf for consistency.

statsmodels/stats/multitest.py
```diff
@@ -444,3 +444,241 @@ def fdrcorrection_twostage(pvals, alpha=0.05, method='bky', iter=False,
...
+
+    # The local FDR values
+    fdr = null_proportion * f0 / fz
+
+    fdr = np.clip(fdr, 0, 1)
+
+    return fdr
+
+
+class EmpiricalNull(object):
```
josef-pkt: thinking about this some more, `Empirical` would be misleading for me outside of the context. I would think the analog would be ECDF, and an empirical distribution. But here the distribution is fully parametric.
maybe `NullDistribution`

josef-pkt:
I finished reading the Efron and the Schwartzman et al (NeuroImage) articles.

I still think EmpiricalNull doesn't sound parametric enough, and I think the trailing 0 are not needed in it's attributes.  (EstimatedNull would also sound ok)
(All the deconvolution and outlier/non-null detection comes from the parametric assumption on the Null distribution family.)

The only other issue I have is whether we can extend this to other distributions. Currently names like `zscores` are targeted to the normal distribution as Null distribution. Schwartzman et al. also allow chisquare distribution (t and F by transformation). Efron transforms F to normal distribution to get zscores, I don't know if that is restrictive with respect to assumptions on degrees of freedom.

josef-pkt:
@kshedden   I just saw by chance that you updated this a few days ago.
I don't get a notification of commits, notifications are sent out when a comment is added. So, please add a comment to the PR if there is something I should look at or merge.

I'll check and merge within a few days, or tonight.

kshedden:
Sure, will do.

Kerby

josef-pkt:
merged, Sorry for the delay.

PR #2307, opened 2015-03-07
ENH/BUG Various GEE improvements

kshedden:
* Fix a bug that raised an exception when fitting ordinal and nominal GEE models with formulas

* Add two parameter-free covariance structures for nominal or ordinal data that are independent except between indicators derived from the same observation

* Refactor categorical covariance structures to descend from abstract base class

* Privatize ncut

* Move starting values calculations for categorical models in subclasses

* (Hopefully) accelerate GlobalOddsRatio by using outer minimum to replace two temporary arrays


kshedden:
Also refactored group splitting to avoid looping over rows.

PR #2318, opened 2015-03-12
New kernel_methods module for new KDE implementation

PierreBdR:
This is a new version of the kernel density estimation. The purpose is to provide an implementation that is faster in the case of grid evaluation, and also works on bounded domains.

There is still some work to do, in particular, I need to add tests for the multi-dimensional and discrete densities.


josef-pkt:
How long does it take to just run the nonparametrics tests?
Based on the travis run, it looks like this branch takes 15 minutes longer or more, and we ran into the 50 minutes time out on python 3.3.
This needs to be checked and fixed before merge.

one incompatibility with older numpy, missing function in fft needs to be fixed maybe with a compat function.

do we need named tuples? we have avoided them so far, using results classes instead, 
currently py 2.6 incompatibility in namedtuple code.

PierreBdR:
I noticed all that. For the namedtuple, as I modified them slightly, I am not using the ones from python, and I corrected the incompatibility issues.

As for the slow tests, I am also trying to make them faster. The issue here is to test grid-based functions when FFT/DCT methods are not available ... they are just really slow.

Maybe we can skip them altogether / throw an exception if the users try to use them?

PierreBdR:
Ok, I am now installing python 2.6 with all dependencies on my machine to test properly before I submit more!

josef-pkt:
import ordereddict from statsmodels.compat

josef-pkt:
I haven't bothered with local py 2.6 testing for a while because it's usually just a few items and doesn't happen very often, it takes just a few rounds of travisCI.

PierreBdR:
Ok, I have an issue here: as it happens, there are quite a few functions with a different behaviour / signature between numpy 1.6.2 / scipy 0.11.0 and the more recent ones. For my own case, from numpy the function `meshgrid` has changed, and from scipy the function `scoreatpercentile` has changed.

As both are pure python, should I just add them to the `compat/numpy.py` and `compat/scipy.py` with a `np_` prefix as with the other functions? Although for scipy, there is no such prefix yet ... maybe `scipy_` ?

josef-pkt:
Yes,  we can just add to compat/numpy  (abbreviation for scipy is `sp`, we don't need a separate compat/scipy module yet), and we delete them again when we don't need them anymore.
Can you replace scoreatpercentile by numpy percentile?  I don't remember the details, both numpy and scipy versions have been moving targets recently.

bashtage:
Seems to be a lot of _undertested_ code.

PierreBdR:
@bashtage  Yes, most of the code has been tested in IPython notebooks. I will write the automated tests very soon.

PierreBdR:
I have an issue here: although the tests all seem to pass (except for python 2.6), the process gets killed on the two oldest configurations (e.g. python 2.6 and python 2.7 + numpy 1.7). What does that mean? And what can I do for this? And it's not a timing problem, as the process ran less than 19 min. (less than 13 min. for python 2.6).

bashtage:
Multivariate normal isn't in old. SciPy. Needs compat shim.

josef-pkt:
It started to get killed starting with this commit b656688  `Improved test coverage of grid, fastbin and kernels`

test run before that was green. I don't see anything suspicious because I don't know cython well enough. Also it's possible a test added there triggers something.

You could add verbose to the travis test run temporarily, to see which test gets killed.

bashtage:
The easiest way to test locally is to use Anaconda, which is available on all platforms.  Once you have it installed (use miniconda on Linux or Windows), run

```
conda create -n py27_np16 numpy=1.6.2 scipy pandas statsmodels nose cython
conda remove -n py27_np16 statsmodels
cd ~/git/statsmodels
python setup develop
```

Then you can run tests locally that you wrote using something like

```
nosetests --verbose statsmodels.kernel_methods
```.  

You can change the = to get other versions (like numpy=1.7)

**Comment by bashtage** [comment]
I can't get this to build on Windows with any combinations of Python/NumPy/etc.

**Comment by bashtage** [comment]
Message is:

```
build\lib.win-amd64-2.7\statsmodels\kernel_methods\_cy_kernels.pyd : fatal error LNK1120: 2 unresolved externals
```

**Comment by bashtage** [comment]
You need to use 
```
int _finite( 
   double x 
);
```
https://msdn.microsoft.com/en-us/library/sb8es7a8(v=vs.90).aspx

for Windows.

You will probably need to provide your own ERF (see e.g. http://www.johndcook.com/blog/2009/01/19/stand-alone-error-function-erf/)

**Comment by PierreBdR** [comment]
@bashtage Ah yes, I forgot about this kind of issues. I will had a compilation switches for the `finite` and `erf` functions.

**Comment by PierreBdR** [comment]
@bashtage Sorry, right now I don't have a usable windows machine with python and cython installed. Can you tell me if this is valid in cython for windows?

    from float cimport _finite

Or do I need to define my own `float.pxd` to that purpose?

**Comment by bashtage** [comment]
I've not done anything with conditional paths.  Hopefully @sturlamolden, who has provided tips in the past, could provide some guidance - or you might find some examples from SciPy or NumPy.

**Comment by bashtage** [comment]
Answer appears here:

http://grokbase.com/t/gg/cython-users/128rn1399p/equivalent-of-ifdef-preprocessor

**Comment by PierreBdR** [comment]
@josef-pkt @bashtage I tested everything with python 2.6 and the exact same version of numpy and scipy on my machine. Sadly, I don't have any crash ... Can you tell me how to activate the verbose option of Travis?

**Comment by PierreBdR** [comment]
@bashtage Yes, I found how to do conditional compilation, although now, you can do it in Cython directly as:

    IF condition:
        something
    ELSE:
        something else

My question was about getting the `_finite` function in Cython: is there a standard `float.pxd` defined? Maybe it is in `libc`, maybe somewhere else? Or do I need to define my own `float.pxd` with the definitions I need in it.

**Comment by josef-pkt** [comment]
in `.travis.yml`

add verbose to
`  - echo 'import statsmodels as sm; a=sm.test(label="full"); import sys; sys.exit((len(a.failures)+len(a.errors))>0)' > test.py`

so it is
`  - echo 'import statsmodels as sm; a=sm.test(label="full", verbose=3); import sys; sys.exit((len(a.failures)+len(a.errors))>0)' > test.py`

**Comment by bashtage** [comment]
The Travis exit code of 137 means it is killing your process.  Are you
possibly using a ton of memory?

**Comment by PierreBdR** [comment]
@josef-pkt thanks, I've just made the modification, I'll see what crashes
@bashtage I don't think so! Also, I don't see why it would use much more memory on older versions of numpy/scipy ... unless there is a bug in them.

**Comment by bashtage** [comment]
Some of your tests are using huge amounts of memory -- around 32GiB -- on SciPy 0.11/ Numpy 1.7
untitled

**Comment by bashtage** [comment]
It appears to be 

```
statsmodels.kernel_methods.tests.test_fastbin.TestContinuousBinningnD.test_bad_args(<function fast_bin_nd at 0x30af938>,) ... ok
statsmodels.kernel_methods.tests.test_fastbin.TestContinuousBinningnD.test_bad_args(<function fast_bin_nd at 0x30af938>,
```
Since you appear to be using test generators, it is very hard to tell the specific test that causes the problem.

File: statsmodels/kernel_methods/tests/test_fastbin.py [hunk 1]
```diff
@@ -0,0 +1,247 @@
...
+        fct(self.data, self.bounds, 256, 1., 'B', out)
+
+    @raises(ValueError)
+    def bad_out2(self, fct):
+        out = np.empty(self.weights.shape, dtype=int)
+        fct(self.data, self.bounds, 256, 1., 'B', out)
+
+    def test_bad_args(self):
+        for fct in [linbin.fast_bin, linbin.fast_linbin]:
+            yield self.bad_data, fct
```
bashtage: Add a meaningful name here, as in 


yield self.bad_data, fct, "Bad data"

statsmodels/kernel_methods/tests/test_fastbin.py
```diff
@@ -0,0 +1,247 @@
...
+    def bad_bounds3(self, fct):
+        fct(self.data, 'bad', 256, 1., 'B')
+
+    @raises(ValueError)
+    def bad_out1(self, fct):
+        out = np.empty((1,), dtype=float)
+        fct(self.data, self.bounds, 256, 1., 'B', out)
+
+    @raises(ValueError)
+    def bad_out2(self, fct):
```
bashtage: add third argument here, or just *args

to take name.  This will allow verbose printing to show information to identify test failures

statsmodels/kernel_methods/tests/test_fastbin.py
```diff
@@ -0,0 +1,247 @@
...
+        bounds = self.bounds * 4
+        out = np.empty(self.weights.shape, dtype=int)
+        fct(self.data, bounds, 256, 1., 'BBBB', out)
+
+    def test_bad_args(self):
+        for fct in [linbin.fast_bin_nd, linbin.fast_linbin_nd]:
+            yield self.bad_data, fct
```
bashtage: Same here

PierreBdR:
The odd part is that these tests should be doing absolutely nothing as their arguments are incorrect. But I am working on it already.

PierreBdR:
Ok, I found out why so much RAM was used, and now it should be better!

I also installed python et al. using anaconda on a windows running in VirtualBox. So now the cython should compile, and the tests run under windows too.

PierreBdR:
Good, now it passes, and the coverage is much better! I still have to write tests for the multi-variate case and some of the utilities I wrote

sturlamolden:
I have not followed this discussion so I am not sure what you want me to help with.
(Normally I just create noise.)

josef-pkt:
@sturlamolden   It would be helpful if you could read through and review the cython code in this PR.
The point was that the test run got killed on TravisCI, which however turned out to be a memory problem (unit test used too much memory) and not a cython issue.

sturlamolden:
Ok, now I see.

```erf(x)``` and ```isfinite(x)``` are only available on C99 and C++11. It is not a Windows issue. MSVC does not provide C99. You need a conditional check for the C version. You will get link failures on any platform where you only have an ANSI C compiler.

You can add a C with your ```erf``` and ```isfinite``` implementations to the Cython sources in ```setup.py```. If you are on C99, just make the C preprocessor turn the implementation of erf into nothing. To implement ```isfinite``` I would just call ```npy_isfinite``` in the NumPy core math library. Presumably you have a dependency on NumPy already, so linking with Numpy should not be a problem. 

http://docs.scipy.org/doc/numpy/reference/c-api.coremath.html

If you need add somethibg special for Windows, you can check if ```_WIN32``` is defined in C or define the C file empty otherwise. Or you can check ```os.name``` or ```sys.platform```, and only add the C file to the list of sources if you are on Windows. Both methods will work. (But what you have here is not really a Windows issue.)

**Comment by sturlamolden** [comment]
@josef-pkt ok, I will look at it

**Comment by josef-pkt** [comment]
Can we use `erf`and similar from scipy.special?   I never read andything specific about reusing scipy.special in cython code.

**Comment by sturlamolden** [comment]
For multidimensional KDE it is better to use a kd-tree or a ball-tree than an nd-mesh and FFT. For the latter the memory useage will grow exponentially. See e.g. the KDE code in scikit-learn (sklearn.neighbors.kdtree).

**Comment by sturlamolden** [comment]
The scipy.special code is a mess of Python, C, C++, Cython and Fortran 77. I am not sure about reusing code from that spaghetti in Cython. 

(Obviously you can call it the way you would in Python, unless the Python overhead is prohibitive.)

**Comment by sturlamolden** [comment]
```round(x)``` is also C99 only. It is not a Windows problem. Similar to ```erf``` it will fail on any ANSI C compiler. So don't special case it for Windows but check the C version.

**Comment by bashtage** [comment]
This kind of useful C/Cython/Python information why I pinged you :+1: .  Interesting for me too...

**Comment by sturlamolden** [comment]
Another thing to note is that ```erf```, ```isfinite``` and ```round``` can create silent errors at build-time on Unix and Linux. Because of the way the dynamic linker works, if the symbols are not available at build-time, they will be attempted resolved at load-time. So you will not get link failures is these symbols are unresoved when you build the extension module, but the process might die when you try to load the extension module. This can be rather non-deterministic because another C modules might have put these symbols in the global namespace of the process. If you are lucky they will do what you want, but if you are unlucky they do something totally different. 

On Windows DLLs behave differently, and all symbols must be resolved at build-time. This is much easier to debug and create far less DLL hell. That is why you immediately got an unresolved external error why you tried to build the Cython module with MSVC.

**Comment by PierreBdR** [comment]
@sturlamolden I stopped using the C99 functions you mentioned. For isfinite, I am now using the numpy function. For round, I wrote it quickly in Cython. As for erf, it so happens that the function used by eglibc has been written by Sun Microsystem. It is free to use, distriute, ... as long as we keep the notice at the head of the file. I just adapted the functions to use numpy's types and endianness macros.

As for the memory issue, it was a mistake: I was trying to create a 4D grid with 256 elements on each dimension ... which was unreasonable, in particular as the point was to test error checking on the parameters. I corrected that, and it's all fine.

As for the kd-tree, that sounds like a good idea! But I think that would be for a later version. For now, I will focus on documentation writing.

**Comment by PierreBdR** [comment]
@josef-pkt I have issues generating the documentation. First, classes I specify don't always appear in the documentation, and I don't know under what conditions they appear or not. Same for methods of a class that do appear: some are shown, but not all. And same for attributes.

Is there some documentation on what to do there?

Also, for tutorial, if I understood properly, I am supposed to add an IPython notebook into `statsmodels/examples/notebooks` ?

**Comment by josef-pkt** [comment]
For me rst and sphinx are usually trial and error.
In which classes and methods are there problems?  Check whether there are sphinx warnings or errors related to it.

I'm not sure about the import paths:
`.. currentmodule:: statsmodels.kernel_methods`

you could try referencing specifi modules in the subsections, if those are missing
 `.. currentmodule:: statsmodels.kernel_methods.xxx`

If your notebooks are appropriate for the documentation, then you can move them to examples/notebooks  and add some explanatory text.
If they are only for internal checking and development, then I would leave them where they are for now.

**Comment by PierreBdR** [comment]
@josef-pkt Thanks, I will have a look.

For the notebooks, I was more thinking of writing new ones for documentation purposes! And write them as a set of tutorials.

**Comment by PierreBdR** [comment]
@josef-pkt So I really didn't manage to make the autosummary work. So in the end, I only use autodoc (e.g. autoclass, autofunction, ...). Sadly, this means it looks a bit different from the rest of the documentation. Can you check what I have done and see if this is still acceptable?

**Comment by josef-pkt** [comment]
It works this way, but it would be better to figure out the autosummary and check whether there are problems with it.  It would become too tedious to do this everywhere.

BTW: did you downgrade your ipython?   It shows complete changes and has lower nbformat (3 instead of 4)

**Comment by PierreBdR** [comment]
@josef-pkt Ok, I'll look more into it! For ipython, yes, I downgraded from 3.0 to 2.4.1. The converters to HTML don't work with IPython 3. I haven't had time to look into it at all, so the simplest for now is for me to use an older IPython.

**Comment by PierreBdR** [comment]
@josef-pkt Good news, I figured out what the problem was: you cannot document a class and claim it's in a module it is not really in. Most classes are defined in private modules, with aliases in public modules. So now they will be documented in their real module, but organised by public module with a note at the start saying that all element documented are aliased in a consistent manner.

**Comment by josef-pkt** [comment]
@PierreBdR   We will need to rebase before merging. It would be good if we can keep track of commits like 155edd9e69232aae8eb0c8cba74a6cad352bb123 that should be squashed (to avoid getting temporary additions into the repo)

**Comment by PierreBdR** [comment]
Hey, sorry for the delay in answering: I was away for work, then holliday, then had to catch up on a lot of things. At the same time, I was looking at the bootstrapping questions there: https://github.com/statsmodels/statsmodels/issues/1244#issuecomment-124490837

Otherwise, I just fixed the history ... I removed small "fix up" messages.

**Comment by PierreBdR** [comment]
Ok, first, I can recommand using the new environment for Travis (activated by adding the line 'sudo: false' to the file .travis.yml) as it is, as advertised, faster.

That being said, there is a persistent error that I don't understand (and was there before the switch BTW): with the last testing configuration, the one in which the version of pandas is not specified, the system cannot find the version of pandas. More precisely, the first error is in the dependency check:

    Required Dependencies
    =====================
    cython: 0.23.4 (/home/travis/miniconda/envs/statsmodels-test/lib/python2.7/site-packages/Cython)
    numpy: 1.10.1 (/home/travis/miniconda/envs/statsmodels-test/lib/python2.7/site-packages/numpy)
    scipy: 0.16.0 (/home/travis/miniconda/envs/statsmodels-test/lib/python2.7/site-packages/scipy)
    pandas: Cannot detect version (/home/travis/miniconda/envs/statsmodels-test/lib/python2.7/site-packages/pandas)
        dateutil: 2.4.2 (/home/travis/miniconda/envs/statsmodels-test/lib/python2.7/site-packages/dateutil)
    patsy: 0.4.0 (/home/travis/miniconda/envs/statsmodels-test/lib/python2.7/site-packages/patsy)

What should I do about it?

**Comment by bashtage** [comment]
I believe this is a bug in stats model. Needs to be fixed in a separate PR
and you can rebase.

**Comment by PierreBdR** [comment]
I was trying to prepare a PR for the pandas issues, but I found another, maybe deeper, problem: I need the function "erf" from a Cython extension. As this function is not available for windows (despite being in the C99 and POSIX-2001 standards), I had to add the implementation of Sun Microsystem, which we can use frely as long as the copyright notice at the start of the C file is left untouched.

Now the problem is: by default, .gitignore is configured to refuse to add C files, and the clean target of setup.py will erase any C file in the tree. What should I do there? FYI, I tried, but the implementation of erf uses a lot of tricks that cannot be translated to Cython. Should I try to change the behaviour of the setup.py to only delete C files that correspond to existing pyx files?

**Comment by josef-pkt** [comment]
About `erf`: Can you use the scipy.special function? or is it too slow when we have to go through python? (Which I guess would also rule out releasing the GIL for parallel processing.)
AFAIK, scipy.special doesn't have a cython interface for users yet. The `erf` implementation in scipy will also be C or Fortran.

If we need to start adding pure C files, then we can adjust gitignore and the clean commands, it's a nuisance but not a problem.

Another possibility if there is hope that newer Windows C libraries will have `erf`: It might also be possible to use erf if available and use python scipy.special if not.

I have no idea about the details (C, cython) and leave those completely to you. My only worry is always C bugs, which wouldn't be likely if it's from an established library.

**Comment by sturlamolden** [comment]
```npy_math.h``` should have all of C99 math functions, but ```erf``` is clearly missing.

Sun's implementation is freely available:
https://github.com/pouar/openbsd-src/blob/master/lib/libm/src/s_erf.c

**Comment by bashtage** [comment]
VisualStudio 15, aka MS C++ 14.0, has the erf function, so it should be OK
for Python 3.5 - of course this doesn't help you for the other platforms.

As for the missing ERF, I would just Cythonize this:

http://www.johndcook.com/blog/2009/01/19/stand-alone-error-function-erf/

and then only use if the platform doesn't have erf.It wo't handle edge
cases as well as a good platform implementation, but is probably OK.

You also could just Cythonize

https://github.com/scipy/scipy/blob/81c096001974f0b5efe29ec83b54f725cc681540/scipy/special/cdflib/erf.f

which should be fine.

**Comment by sturlamolden** [comment]
```erf``` in Visual Studio is due to C++11, which is required to have all of C99 math functions.

**Comment by PierreBdR** [comment]
@sturlamolden Yes I know, their function is currently compiled in statsmodels.
@bashtage is there a way to test if the function is available? Or should we rely on configuration variables? We could for example have something like:

    if not windows or if python >= 3.5:
        erf should be available
    else:
        use the pythonized (slower) erf

**Comment by sturlamolden** [comment]
```erf``` is available in C++11 or C99. You need to look at the version of C or C++. It has nothing to do with Windows per se.

**Comment by bashtage** [comment]
@sturlamolden Yeah, but unfortunately it took until now to get a semi-decent compiler as the platform compiler for Windows - plus, there is a stable ABI now so extension land in Windows should be much better in a Python 3.5+ world.

**Comment by PierreBdR** [comment]
Ok, so in the end I cythonized the second version of the erf, e.g.: https://github.com/scipy/scipy/blob/81c096001974f0b5efe29ec83b54f725cc681540/scipy/special/cdflib/erf.f

The first version is too imprecise for our use. My main issue is that I cannot figure out how to test for the version of python at compile time ... only the OS. So the windows version will use the cython erf while other OSes will use the system's erf.

**Comment by bashtage** [comment]
You only need to see if c99 is defined -- you don't need to worry about the
python version.  For example, compiling Python 2.7 on a modern linux distro
should have erf available.

**Comment by PierreBdR** [comment]
@bashtage And how do I check that reliably? All the answers I can find rely on `#ifdef` ... which I don't know any cython equivalent for.

**Comment by PierreBdR** [comment]
Ok, I rebased the PR to the latest master.
Also, I really looked at the problem, and I don't see any way to define a function differently depending on the version of the C compiler or the version of Python ... or anything but the OS really. So I stay with Windows having the local definition of the erf and other OSes the C definition.

**Comment by josef-pkt** [comment]
I think that's fine, especially given that we can expect that the problem will go away eventually.
One thing that numpy does (AFAIK) and we did at some point, is to compile first a test script to get feedback from the compiler. But I don't think it's worth it for this case, and I don't know details just that it is or was done.

**Comment by bashtage** [comment]
> Also, I really looked at the problem, and I don't see any way to define a function differently depending on the version of the C compiler or the version of Python ... or anything but the OS really. So I stay with Windows having the local definition of the erf and other OSes the C definition.

I think you need a C IFDEF sequence in a .h file to define some constant, then you can check this constant in your cython.

http://stackoverflow.com/questions/3826458/cython-conditional-compile-based-on-external-value

The second answer from that stackoverflow post.

**Comment by sturlamolden** [comment]
It might be that C99 with MinGW is broken because it depends on Microsoft's C run-time DLLs. I have not used Windows for a while so I cannot say for sure. 

As for conditional compilation, the setup.py script can check the value of os.name, and assume erf is needed is the value is 'nt'. You can e.g. have erf in an additional source C file and add it to the list of source files if you are on Windows. 

Another option is to always use your of Cython verison of erf, even when using C99 on Linux, because if it is precise enough you gain nothing from using a system CRT erf with unknown precision. If your own erf is good enough on Windows is good enough on Linux as well.

**Comment by sturlamolden** [comment]
Preumably fixing this upstream in NumPy by adding npy_erf to npy_math would not help either, because you likely need backwards compatibility.

**Comment by bashtage** [comment]
> And how do I check that reliably? All the answers I can find rely on #ifdef ... which I don't know any cython equivalent for.

I think you would have something like this in a .h file

Suppose this was called `has_erf.h`
```
#define HAS_C99 0
#if defined(__STDC__)
 #define C89
 #if defined(__STDC_VERSION__)
  #define C90
  #if (__STDC_VERSION__ >= 199409L)
   #define C94
  #endif
  #if (__STDC_VERSION__ >= 199901L)
   #undef HAS_C99
   #define HAS_C99 1
  #endif
 #endif
#endif
```

Then in your Cython, you could

```
cdef extern from "has_erf.h":
    cdef int HAS_C99

if HAS_C99:
    cdef extern from "math.h" nogil:
        double erf(double)
else:
   # your erf function here
```

Or just do what @sturlamolden said :+1:  and always use your own version of `erf`.

**Comment by PierreBdR** [comment]
@bashtage I tried and you cannot define a function within the normal if ... maybe a python function, but not a C function. A normal, lower case 'if' is always evaluated at runtime and not compile time. I think the only option would be indeed what is suggested by @sturlamolden :  to add two alternative files defining erf and change it in the `setup.py` depending on the version of python and os

**Comment by sturlamolden** [comment]
Here is a full translation of ```scipy.special.erf``` from Fortran 77 to ANSI C. It can be called directly from Cython.

```C
#include <math.h>

double erf(double x) 
{
    const double c = .564189583547756E0;
      
    const double a[5] = {.771058495001320E-04, -.133733772997339E-02,
                         .323076579225834E-01, .479137145607681E-01, 
                         .128379167095513E+00};
     
    const double b[3] = {.301048631703895E-02, .538971687740286E-01,
                         .375795757275549E+00};
     
    const double p[8] = {-1.36864857382717E-07, 5.64195517478974E-01,
                          7.21175825088309E+00, 4.31622272220567E+01,
                          1.52989285046940E+02, 3.39320816734344E+02,
                          4.51918953711873E+02, 3.00459261020162E+02};
     
    const double q[8] = {1.00000000000000E+00, 1.27827273196294E+01,
                         7.70001529352295E+01, 2.77585444743988E+02,
                         6.38980264465631E+02, 9.31354094850610E+02,
                         7.90950925327898E+02, 3.00459260956983E+02};
     
    const double r[5] = {2.10144126479064E+00, 2.62370141675169E+01,
                         2.13688200555087E+01, 4.65807828718470E+00,
                         2.82094791773523E-01};
     
    const double s[4] = {9.41537750555460E+01, 1.87114811799590E+02,
                         9.90191814623914E+01, 1.80124575948747E+01};
    
    double y, ax, bot, t, top, x2;

    ax = fabs(x);
    if (ax <= 0.5E0) {
        t = x*x;
        top = ((((a[0]*t+a[1])*t+a[2])*t+a[3])*t+a[4]) + 1.0E0;
        bot = ((b[0]*t+b[1])*t+b[2])*t + 1.0E0;
        y = x*(top/bot);
    }
    else if (ax <= 4.0E0) {
        top = ((((((p[0]*ax+p[1])*ax+p[2])*ax+p[3])*ax+p[4])*ax+p[5])*ax
                 + p[6])*ax + p[7];
        bot = ((((((q[0]*ax+q[1])*ax+q[2])*ax+q[3])*ax+q[4])*ax+q[5])*ax
                 + q[6])*ax + q[7];
        y = 0.5E0 + (0.5E0-exp(-x*x)*top/bot);
        if (x < 0.0E0) y = -y;
    }      
    else if (ax < 5.8E0) {
        x2 = x*x;
        t = 1.0E0/x2;
        top = (((r[0]*t+r[1])*t+r[2])*t+r[3])*t + r[4];
        bot = (((s[0]*t+s[1])*t+s[2])*t+s[3])*t + 1.0E0;
        y = (c-top/(x2*bot))/ax;
        y = 0.5E0 + (0.5E0-exp(-x2)*y);
        if (x < 0.0E0) y = -y;
    }
    else {
        y = (x >= 0 ? 1.0E0 : -1.0E0);
    }
    return y;
}
```

sturlamolden:
You can of course use plain Cython instead of C, but unless things have changed, Cython does not accept arrays of ```const double```. Using arrays of ```double``` (which Cython accepts) leaves the compiler less room for optimization.

PierreBdR:
Ok, so in the end, I went back to a C implementation, but as a static function defined in the header file.
So from the header I can do this:

```
#if !defined(WIN32) || (defined(__STDC_VERSION__) && __STDC_VERSION >= 199901L)
  #include <math.h>
  #define _sm_erf erf
#else
  // implement our own _sm_erf
#endif
```

Is that a satisfying solution?
Advantage: we can use the exact `erf` function used by GCC!

sturlamolden:
There is no such thing as an "exact" erf function. Trancendental functions are always implemented as approximations. Also, erf is not in GCC, it is in libc (glibc on Linux).

PierreBdR:
@sturlamolden I meant the exact same sorry. And you are right, I was sloppy, it's the version in glibc that I meant, not GCC.

sturlamolden:
By the way, the justification for the C code I posted is that it produces the same result as ```scipy.special.erf```. I do not claim that it is more accurate than the versions in libm and glibc. 

(It might not be exactly the same as ```scipy.special.erf``` though, because Fortran is allowed to carry out arithmetic operations in the order which gives the best performance, whereas in C the order of operations is determined by the expression.)

PierreBdR:
Hey, is there anything else that needs doing to this PR to consider merging it?
Just wondering so it doesn't miss the next version ...

josef-pkt:
I put this close to the top of my review and merge queue so we get this in.

I did a rough browsing to get an overview (it's a very large PR), some initial comments:

- Do we need all the setter methods, or can they be moved into the `__init__`? I will need to look at examples to see how this works, but in almost all code we avoided having to set attributes after creating a model instance, mainly to avoid inconsistent attributes. It also provides for an informative `__init__` signature.

- I think we need to privatize by add leading underscore many of the modules that the user shouldn't use.

- postmerge: there are several helper and compatibility functions or modules that need to eventually go to an appropriate location

PierreBdR:
To answer your points:

* I must say I like having both, but it is important to be consistent across statsmodels. So I will change the interface to whatever the rest is doing, just tell me.
* I thought I did it already, but we can have another pass at it.
* Very true

statsmodels/kernel_methods/kde.py
```diff
@@ -0,0 +1,307 @@
...
+
+    @property
+    def method(self):
+        """
+        Method used to estimate the KDE.
+        """
+        return self._method
+
+    @method.setter
+    def method(self, m):
```
josef-pkt: What's the purpose of this construction?
Is this an instance method that prevents the class from being picklable?

I haven't figured out yet where self.method.fit is defined. Just getting started at some details.

josef-pkt: I found the fit, `_kde1d_methods.py` is so large that github doesn't show it in the Files changed

josef-pkt:
(I got delayed again because the kids had several days off from school.)

I finally checked out this branch. It compiled without problems with Winpython on python 3.4. Unit tests finish without problems in 122 seconds. 

I just ran and skimmed the notebooks. The notebooks and the results looks very good. And in spite of cross-validation the notebooks run fast. At least I didn't feel a waiting time.

So, independent of whether I manage a large review, I'd like to get this merged soon, definitely for 0.8. I only looked so far at what fit returns and haven't checked other details. I still think the setter methods and copying of self is dangerous/fragile, and there is maybe a bit too much import magic (that I haven't tried to figure out yet). However, I think we could also merge it and see if this pattern can work out. I didn't try to run my own examples yet.

josef-pkt:
```
mod.numeric_cdf(0)
OverflowError                             Traceback (most recent call last)
<ipython-input-52-bbda47ce0715> in <module>()
----> 1 mod.numeric_cdf(0)

m:\josef_new\eclipse_ws\statsmodels\statsmodels_py34_pr\statsmodels\kernel_methods\kde_utils.py in f(self, z, out)
    291         def f(self, z, out=None):
    292             z, real_out, write_out = _process_trans1d_args(z, out, in_dtype, out_dtype)
--> 293             fct(self, z, out=write_out)
    294             return real_out
    295         return f

m:\josef_new\eclipse_ws\statsmodels\statsmodels_py34_pr\statsmodels\kernel_methods\_kde1d_methods.py in numeric_cdf(self, points, out)
    859         pts = points.ravel()
    860 
--> 861         pts[pts < self.lower] = self.lower
    862         pts[pts > self.upper] = self.upper
    863 

OverflowError: cannot convert float infinity to integer
```

josef-pkt:
first 1D example: Which `method` class do we get?

```
kde, mod, kde.method
Out[59]:
(<statsmodels.kernel_methods.kde.KDE at 0xb2c4a20>,
 <statsmodels.kernel_methods._kde1d_methods.Reflection1D at 0x5988518>,
 <statsmodels.kernel_methods._kde_multivariate.Multivariate at 0xb2d0588>)
```

PierreBdR:
@josef-pkt The purpose of the `fit` method is to return the most efficient object possible for the subsequent evaluation of the KDE, this include two aspects: returning the most efficient method (with the least overhead), and perform some pre-computation if appropriate. And the most efficient method is not always the method object that has been configured by the user. For example, the default method is `Multivariate`. `Multivariate` can handle any number of dimensions ... with a runtime cost (iterating over dimensions, gathering the data and combining them back at the end ...). So if the input vector is 1D only, the fitting will return the equivalent 1D method. The default evaluation method for a continuous dimension is `Reflection1D`, so it will return directly an instance of `Reflection1D`.

For the error with `numeric_cdf`, I will look at it right away!

statsmodels/kernel_methods/kde.py
PierreBdR: For `_kde1d_methods.py`, I can split it so there is one file per method.

For the `method` property, the idea is to provide a unique point of entry to define the KDE problem, and be able to choose the fitting method independently. For example, if you have an existing KDE object `k` with the default fitting method, you may want to try using the `LineageCombination` method, which is a first order approximation for bounded domains. You would do:

    k.method = kde_methods.LineageCombination()
    mod = k.fit()

As for the `fit` method, it is defined in the various classes imported in `kde_methods`. I have put below a more precise description of the role of this `fit` method.

PierreBdR:
I found the issue with `numeric_cdf` ... which happens with integer input! So now I force conversion to float.

josef-pkt:
> Split _kde1d_methods into one file per method

That wouldn't have been necessary, just because github doesn't like it. I switched to browsing the source with eclipse/pydev which is much faster for getting an overview.

I will add more comments when I review today. But don't make any changes immediatly or I have to start again to find my way around the code. It will still take me a bit of time to understand the structure of this.

Question: AFAICS, there are no values cached for the evaluated pdf, cdf, .. or their grid versions, are there?

josef-pkt:
pep-8 two empty lines between functions and methods, makes it easier to browse.

josef-pkt:
to understand attributes/properties:

```
kde = sm.kernel_methods.KDE(obs_dist1)
mod = kde.fit()
kde.adjust = 0.5   # this does not affect `mod`, only used in next `fit`
mod.adjust = 0.25  # this changes how the kde is calculated right away, I haven't found cached "state" yet
```

PierreBdR:
@josef-pkt For the split, the file was definitely too large. Also, this way they can serve an example on how to implement your own method.

For the properties, you are right. One of the main difference is that when you do:

    mod.adjust = 0.25

A run-time check is performed to make sure the value is consistent with the data. The same with the `KDE` object doesn't check anything until `fit` is called.

Otherwise, I am not sure what you mean by cached state, other than all the data is copied in the fitted object (e.g. the fitted object doesn't keep a reference to the original `KDE` object, but rather keep all the data it needs).

josef-pkt:
The original idea in the refactoring that @Padarn  started was to store (cache) the grid values so we have faster access to repeated calculations, for example pdf_values, cdf_values, icdf_values. This is similar to fittedvalues, resid, and similar that are cached in the Results instances of models.
Interpolating would allow fast reuse, for example in things like random number generation.

I haven't seen anything like that in this version. If those things are not cached, then it is less likely that things get out of sync after a property is changed (which was one of my main worries).

PierreBdR:
@josef-pkt Indeed, I do not cache the result of the computation. For the interpolation (which is used for the cross validation), I implemented a `GridInterpolator` object that takes the result of a grid evaluation and has optimized Cython code to compute the interpolation.

statsmodels/kernel_methods/bw_crossvalidation.py
```diff
@@ -0,0 +1,445 @@
...
+
+    Parameters
+    ----------
+    data: tuple of ndarray
+        Data to process.
+    is_sel: tuple of bool
+        Of same length as data, indicate for each array if they are to be filtered or not.
+    npts: int
+        Number of points in the dataset
+    folding: int
```
josef-pkt: folding should be renamed to a more informative `n_folds`.
Need to check naming with other parts and scikit-learn.

josef-pkt:
about speed of cross-validation bandwidth choice: 
10 fold cross-validation as in the example is fast, and we should check if we should use it also in other methods outside of this PR. 100 fold is much slower, and I guess full LOO cross-validation will bring us back to the slow version we have in other places, AFAIR. (I switched to nobs=1000 in the 1-D example)

josef-pkt:
I don't think the name `Method` for the return of fit is intuitive. It's much closer to a Results class, but not quite. I'm trying to think of a more "obvious" name.

statsmodels/kernel_methods/_kde1d_cyclic.py
```diff
@@ -0,0 +1,230 @@
...
+                             "closed or un-bounded domains.")
+
+        exog = self.exog
+        points = points[..., np.newaxis]
+
+        # Make sure points are between the bounds
+        if any(points < self.lower) or any(points > self.upper):
+            points = points - self.lower
+            points %= self.upper - self.lower
+            points += self.lower
```
josef-pkt: this looks like it's shifting the distribution depending on where we evaluate it, or which points are in our current evaluation set.
It's ok if evaluating the kde is translation invariant, which might be a property of cyclic.  (sounds similar to circular distribution)

docs/source/kernel_methods_kernels.rst
```diff
@@ -0,0 +1,60 @@
...
+=========================
+
+Kernel base classes
+-------------------
+
+.. autosummary::
+    :toctree: generated/
+
+    kernels.Kernel1D
+    kernels.KernelnD
```
josef-pkt: the capitalization of "D" in `nD` looks strange in the names

josef-pkt:
I haven't figured out how bounds work but I get nans for out of bounds pdf (I was trying to find out what the consequences are of shifting `points` in for example `Reflection1D.pdf` to be inside the bounds.)

example from kde1d notebook after
```
kde.upper = 2
mod2 = kde.fit()
```

```
print(mod2.pdf([1.]))
print(mod2.pdf([1., 1.99, 2.0]))
print(mod2.pdf([1., 1.99, 2.0, 2.5]))
[ 0.20096123]
[ 0.20096123  1.87008603  1.87358991]
[ nan  nan  nan  nan]
```

josef-pkt:
`botev` is not an informative name. I was wondering what happened to adaptive bandwidth choice but wouldn't have remembered botev.

josef-pkt:
`botev` doesn't look like an adaptive kde, it's just selecting the bandwidth?
It looks like it's overfitting a bit in the bimodal normal case with nobs=1000

statsmodels/kernel_methods/_kde1d_reflection.py
```diff
@@ -0,0 +1,239 @@
...
+                                      normed, self.total_weights, dim=dim)
+
+    def grid_size(self, N=None):
+        """
+        Returns a valid grid size.
+        """
+        if N is None:
+            if self.adjust.shape:
+                return 2 ** 10
+            return 2 ** 16
```
josef-pkt: why is the default so large?  65536  even if I only have 1000 observations, and it looks large for plotting

rgommers:
Looks interesting. It would be very helpful to put a more complete summary of what changes in the API in the PR description at the top. For example: this doesn't touch the existing implementation, but there's a lot of overlap. Is this going to be a replacement, or a second version in a separate submodule? If the latter, when would I use which implementation?

josef-pkt:
@rgommers   It's a second version for now. Pierre and Padarn started with refactoring and replacing the existing KDE but it was too difficult to figure out backwards compatibility issues given that kde is used quite a bit. I think it's easier to start "from scratch" than trying to work around backwards compatibility considerations. 

I think we will label it experimental for one version to see how well the new design works out. (By default I still don't consider API immediately frozen for new code.)
If you just need the fft density of the old version I would stick with it for now. For other things the new code will be faster, it has more in cython, and has other features, like bounded kde. 
Some features might not be available yet in the new code, but I haven't seen enough yet to tell much.  

Also, the old code and kernel evaluation is still needed for kernel regression and similar.

rgommers:
Clear now, thanks Josef.

statsmodels/kernel_methods/_kde1d_cyclic.py
PierreBdR: Indeed it is. Distribution on angles are a good example (although a simple one where the lower bound is 0). If you want to evaluate the distribution on 7/2 pi, you would expect the same as for pi/2. And for the calculation, I need to get the "reduced" form (i.e. the one between lower and upper).

docs/source/kernel_methods_kernels.rst
PierreBdR: It's quite common in geometry though ... you will work on an nD space  with lower case n and upper case D. So unless you feel strongly against this, I would rather keep it.

statsmodels/kernel_methods/_kde1d_reflection.py
PierreBdR: I chose 2**16 because it's was fast enough on most computers I use, it's not too much memory, and accuracy depends on the number of slots used for binned methods. That being said, the main reason why `grid_size` is a method and not a function is to allow for more complex behaviour if we need. For example, we could test the number of samples to return something related to it.

josef-pkt:
some `**kwargs` or `**kwords` swallow incorrect keywords. If possible, then it should raise an exception, e.g. bandwidth `botev`

josef-pkt:
I didn't find an example for a kde in transformation space. unit tests seems to cover only parts of it, not a full KDE.

The following seems to work, by rough visual inspection

```
from statsmodels.kernel_methods import kde_methods
meth = kde_methods.Transform1D(kde_methods.LogTransform)
kdem = sm.kernel_methods.KDE(np.exp(obs_dist1 * 0.75), method=meth)
modm = kdem.fit()
```

Given how much functionality and how many options are in here, I think we need more notebooks. It's difficult to figure out otherwise how to use it, since it's not a very common set of features, at least not to me.
I'm working based on the first univariate example, and just keep trying out new options with it.
The notebooks in the test directory are a bit outdated, and at least some parts are redundant with the main notebooks. Other parts are needed, like bootstrap.

josef-pkt:
changes for test modules don't show on github: plain assert should not be used, either numpy.testing.assert_ for generic truth testing, or better a specific assert. e.g. assert_array_less (IIRC)
for example in `test_crossvalidation.py`
The inequality asserts don't assert much. If there are no reference numbers, then we should use regression tests at a few digits to detect future changes

josef-pkt:
strange bootstrap confidence intervals, they seem to be shifted in the example
https://gist.github.com/josef-pkt/4554834feeec7dc49143
I just copied the code that worked for the 2-component mixture and don't see any mistake in the notebook itself.

PierreBdR:
@josef-pkt The shift you observe is because the support grids are not the same. Replace the two last plots with:

```
ax.plot(bs_grid, bs_values[:, 0, 0], lw=2, color='blue')
ax.plot(bs_grid, bs_values[:, 0, 1], lw=2, color='blue')
```

And if should be a lot better.

josef-pkt:
@PierreBdR  Yes, using `bs_grid` makes it look good.

josef-pkt:
another question:  
I was trying to see what the effect of the bandwith on the cdf is. However, I get exactly the same plot for oversmoothed default bandwidth as for correctly smoothed crossvalidated bandwidth. However, the grid points/support differ. 
The code seems to scale the data correctly by the bandwidth, so I don't know what's going on.

```
In [45]:
kde0 = sm.kernel_methods.KDE(obs_dist1)
mod0 = kde0.fit()
support0, cdf0 = mod.cdf_grid()
mod0.bandwidth, mod0.adjust
Out[45]:
(0.462019776989512, array(1.0))

In [59]:
xs, ys = mod2.cdf_grid()
print(mod2.bandwidth)
f = plt.figure(figsize=(12, 8))
ax = f.add_subplot(111)
ax.plot(xs, ys)
ax.plot(support0, cdf0, lw=2, color='black');
ax.set_title('CDF')
0.21368414685764883
```

updated in Gist, at lines `In [36]` and 37

PierreBdR:
@josef-pkt I think that's because you meant to use `mod0.cdf_grid()` and not `mod.cdf_grid()`, mod and mod2 have almost the same effective bandwidth ...

josef-pkt:
Thanks again, I think I haven't had enough coffee yet.
It looks now as expected.
I'm a bit surprised that undersmoothing compared to the crossvalidated bandwidth has little effect on the cdf. It just makes the cdf a bit wigglier. So, in this example crossvalidated bandwidth looks already good. (Theoretically we can undersmooth the cdf compared to the pdf.)

PierreBdR:
@josef-pkt You can definitely undersmooth it without much consequences. In the case of your dataset, the "raw" CDF is almost identical to the CDF estimated by `mod2`.

josef-pkt:
I largely fixed the multivariate notebook, 
https://gist.github.com/josef-pkt/261eff0338fef2499e81
For one part, with `_cy_fast_linbin.fast_linbin_nd`, I didn't know what it was supposed to be doing.

- standard cross-validation with grid is very slow, 6 minutes. 
- cross-validation without specifying grid is very fast but gives a bandwidth that is too small.
- crossvalidation with CV_IMSE looks reasonably fast and bandwidth looks fine, bandwidth is a bit smaller than the slow default (likelihood) crossvalidation.

general question for multivariate that I haven't tried to figure out yet: Suppose we have several continuous variables, Do we get a product kernel or the multivariate Gaussian kernel? Is there an option?

josef-pkt:
right now my introspection skills are failing:

I'm trying to figure out what crossvalidation is using
after setting a crossvalidation bandwidth and fitting
```
k4.bandwidth = bandwidths.crossvalidation(bandwidths.CV_IMSE, folding=10)
est4 = k4.fit()
```
all the attached things to k4 are classes and est4 just has the number for the bandwidth.
Are the instances that are used for crossvalidation attached or available somewhere in these objects? Can I set the starting values `func.init_bandwidth` somewhere?

And notebooks are weird, I never know what has been executed.
Now it's an instance

```
k4.bandwidth.func.init_bandwidth, k4.bandwidth.func, type(k4.bandwidth.func)
Out[45]:
(array([ 0.35826709,  0.08600239,  0.1       ]),
 <statsmodels.kernel_methods.bw_crossvalidation.CV_IMSE at 0x1b70c9b0>,
 statsmodels.kernel_methods.bw_crossvalidation.CV_IMSE)
```

and now it's a class and unbound property again

```
k4.bandwidth.func.init_bandwidth, k4.bandwidth.func, type(k4.bandwidth.func)
Out[57]:
(<property at 0x86b0368>,
 statsmodels.kernel_methods.bw_crossvalidation.CV_IMSE,
 type)
```

And I don't remember which cells I evaluated to turn it into an instance.

Confusing.
???

(maybe I better switch back to spyder, at least it's linear instead of going in circles)

josef-pkt:
As motivation for the above: 
I was trying to see why crossval default takes 6 minutes, and CV_IMSE takes 3 to 4 seconds. The functions don't look so different, so my only guess is that the likelihood is difficult to optimize, which might improve with better starting values.
But I didn't see how to get any information about the optimization out of it, and how to influence it.

josef-pkt:
Is there a reason not to put the bandwidth choice into the `fit` method?  
It would be less confusing to me to use it as an argument to get a kde with different bandwidth than to have to set it as a attribute that is used as default.

Overall I have a hard time to figure out where things are set since they are not in the fit or `__init__` methods. For example, if I set the bandwidth on the KDE instance, how does it get set for the final `method` class?

Similar what's the reason for doing the copy and shallow_copy instead of creating new instances with `__init__` args? It looks kind of flexible, but I don't see what is actually `copied`.

josef-pkt:
about the plots:

KDEnD.ipynb looks very nice, after fixing some outdated parts.
Can we keep the color scale fixed in the different plots? For example, I just realized that the oversmoothed KDE has a different scale on the color bar but similar colors in the plot. The oversmoothing would be more visible if the colors correspond to the same values across plots. (not in diff plots which have different units)

josef-pkt:
To follow up on what Ralf @rgommers  asked:
We should have a list of new features, like boundary handling and fft based multivariate KDE and ...
I think this would be good both in the release announcement and more emphasized in the rst docs so we get this advertised.

Also I think we need some comparison or translation with corresponding code for users to make the transition. And it would be better to keep some of the old names, e.g. var type for mixed multivariate kde is lower case in the old and upper case in the new API..

PierreBdR:
Ok, I couldn't figure out how to reply to each comment separately, so here we go:

1 - the Cython modules really shouldn't be used directly, instead the users should always use the Python `fast_linbin` module. It handles default arguments, and basic checks for argument validity.
2 - For the initial value problem, it is possible to specify the starting point by setting `initial_method` to a value, like this: ```k.bandwidth = bandwidths.crossvalidation(imse_arg=dict(use_grid=True, grid_size=256, folding=5), initial_method=[1,1,.1])```. Maybe `initial_method` is not the right name though ...
3 - For the speed issue, I have it also ... I will investigate, as I don't think it used to be that slow!
4 - Regarding the instances for the cross-validation, in the end, no they are created as temporary variables and destroyed before returning. And that's because I don't see where we could put these objects.
5 - About putting the bandwidth evaluation method as argument to the `fit` method ... I am not too keen on that. The `KDE` object is really a configuration object. As such, it makes more sense for the `fit` method to have no argument. Also, if we put the bandwidth in fit, why not put the rest too?
6 - About the color scale, we can definitely fix the color scale. I will work on that.

At last, I agree about a list of new features and a conversion cheat sheet with equivalent features on either modules.

josef-pkt:
> 5 - About putting the bandwidth evaluation method as argument to the fit method ... I am not too keen on that. The KDE object is really a configuration object. As such, it makes more sense for the fit method to have no argument. Also, if we put the bandwidth in fit, why not put the rest too?

That's the main point that is bothering me in this. "Configuration object" is a good description, and I didn't think about it explicitly in this way, but it describes it well.
bottomline: What I was thinking about and trying to figure out is whether various options belong as arguments to `fit` or to `__init__`
So far statsmodels has no setter methods in the public API (AFAIR), they are used in some internal parts. The main reason for most models is that we want to avoid stale state, which is not a problem here right now, as far as I have seen, because there is no "state" except for configuration (and `exog`). The second problem which seems to be relevant here is that the setter methods are not *self documenting*. I, and most likely users, have a hard time figuring out where to provide the options. In most functions, classes and methods the signature shows what are the "configuration options". In this case it is difficult to figure out which setter I have to use, or where I have to set an option to get the desired result.

Most likely related is that I think we need a class that holds/caches the results of the estimation. I thought we can tack it on at a later point #2715, but I'm not so sure anymore, that wouldn't be useful  to think about the design from the beginning.

The complexity of the internal code reminds me a bit of @ChadFulton's inital merged version of the statespace models that used triple inheritance. In that case it was difficult to figure out which inherited class was doing or storing which part. I think after switching to delegation with separate instances, the improved modularity made the code much easier to understand and work with. (But I didn't actually work on the code, so this is just a impression from the outside.)

PierreBdR:
@josef-pkt The advantage of configuration objects is that the user can set items one by one. As the KDE can be quite complex to create, with intermediate objects needed to specify all there is to specify, I really feel this is a good use case for such a technique. It also allows to create a KDE, and then change a single property, for example trying a different bandwidth selection method, or change the estimation method itself, without having to respecify everything.

If documentation is the main issue, I can add explicitly the main arguments, mainly bandwidth selection and estimation method, to the init function like `exog` is right now.

As for code complexity, the only one here is in the `fit` function that can return any object it wants, and indeed returns the object that is the most efficient to solve the current problem. I will work more on the documentation for that.

josef-pkt:
From my workflow while trying out different examples, I thought the bandwidth would fit in `fit`, i.e. use the same kde instance and try out different bandwidths. (The only other thing that maybe should be fixed in `__init__` are the transforms because they create a new exog. But I only saw a few comments about it and haven't looked at how the transformed exog is stored.)
In my interpretation the "configuration object" `kde` is setting the defaults, but when I was trying out different bandwidths (for example), then I wasn't really interested in changing the defaults, e.g. change `adjust` or bandwidth method to see how it affects the plot. e.g. multiply scott or silverman by a sequence of factors.

Another advantage for using arguments is that they don't need a separate line and can be used in chaining.
`OLS(x,  y, ...).fit(...).predict(x2)`
or longer
`OLS(x,  y, ...).fit(...).get_prediction(x2).summary_frame()`
This is useful in the models if we only need a specific result and don't need to hold on to any results instances.

What we use in a few cases or examples is to use a dictionary, or similar, to hold the `configuration`, which could look something like:

```
default_options = dict(bandwidth=....,  ...)
KDE(y, **default_options)
```

(most likely I'm back to this for Wednesday and Thursday)

PierreBdR:
Ok, maybe I should explain my workflow then. I tend to work a lot with bounded or cyclic domains. But often, whether I am close to the bounds or not is dependent on the exact experiment. So I will often start by estimating my KDE, and if I get non-zero density outside the domain, I will use a bounded method. Moreover, in the case of simply bounded domains, I will often try a few methods, even if just to see if they differ in their result.

I will also point out that with the current implementation, one-liners are still possible:

    from statsmodels.kernel_methods import kde, kde_methods
    xs, ys = kde.KDE(data, method=kde_methods.LinearCombination, lower=0).fit().grid()

Now, we can offer an option to redefine some of the attributes in the fit method too, so we can do something like:

   k = kde.KDE(data, lower=0);
   x1, y1 = k.fit(method=kde_methods.Renormalization).grid()
   x2, y2 = k.fit(method=kde_methods.LineageCombination).grid()
   x3, y3 = k.fit().grid()  # use the default method

Again, for documentation purposes, I can explicitly enumerate all the possible arguments instead of relying on `**kwargs` magic.

josef-pkt:
@PierreBdR   Sorry for the delay, I'm still stuck in `stats`.

Would you agree to explicit setter methods, e.g.  `kde.set_bandwidth(....)`?
I started to write a message to the mailing list because I thought also of other places where setters for defaults might be useful. I was thinking about how matplotlib is handling attributes, and they use `set_xxx` methods. 
The main advantage from my perspective is it's documentation value. In almost all cases users are not allowed to set any attributes in statsmodels (and would most likely mess up the results if they do). If we have explicit setter methods, then there will be a clear distinction between what attributes can be changed and which cannot.

PierreBdR:
@josef-pkt No worries, I am also being quite busy on my own side. Although my current project makes me wish I had taken the time to implement conditional probabilities estimation ...

No problem with that. So here is what I suggest I do, tell me if you agree:

1. Remove the `**kwargs` from `__init__` and instead have an explicit list of acceptable arguments.
2. Add explicit setters methods for each property
3. Add a `**kwargs` argument to the `fit` method to allow setting properties while calling `fit`

josef-pkt:
Sounds good, that should take care of having a `self-documenting` class.
Do the kwargs go in `__init__`, `fit` or both?

One alternative I thought about is to name KDE explicitly to indicate it's a dispatch class, e.g. `KDEDispatch`, `KDEConfiguration` or something nicer. Then it would also be clear that it's not a regular `Model` class.
(But it would raise the question which are the model classes.)

About conditional densities: Did you try the current Multivariate KDE implementation?

PierreBdR:
@josef-pkt I would remove kwargs from `init` and add it to `fit`.

For the name, I like the idea, although finding a good name might be tricky. Other options could be: `KDEParams`, `KDESet`.

The model classes are the ones availables in `kde_methods` in that sense. They are the ones implementing the actual model.

As for conditional: not yet, it arose recently, and I will have a look!

PierreBdR:
@josef-pkt Sorry for the delay: I have been very busy at work. But I've found the time to update the module and include the changes we talked about. So now:

* all attributes that can be set have a setter method (e.g. the `method` attribute has a corresponding `set_method` method)
* bin and axis types are now lower case instead of upper case
* `KDE.__init__` now only has explicit arguments
* `KDE.fit` has `**kwargs` which accepts the same arguments than `__init__`
* I added tests for these

josef-pkt:
Thanks, I won't be able to look much at it until beginning of January.
Are there still any open issues now?

PierreBdR:
@josef-pkt The main issue remaining would be the possible renaming of the `KDE` class.

josef-pkt:
@PierreBdR   Can you rebase this?  I'd like to merge this after at most another brief look.

In case there are still issues open, we add an experimental-API warning to the documentation.

I don't think I will get back into this topic enough for a careful review anytime soon. And the last review was already pretty good. So, I don't think I should hold this up even longer than I already did.

PierreBdR:
Hey, sorry for the long time without update. In the past few years I was really busy (new job, new family member, ...). But I rebased the PR to the new master, removed any python2-support code, updated the tests to use pytest and updated the jupyter example notebooks. Tell me what you think.

PierreBdR:
BTW, I remember that, originally, there was a problem with the "erf" function that wasn't present on windows. I saw in a comment on a parallel thread that it may have been solved. However, I do not have the environment necessary to check. Note that the tests will fail without it, so if someone can run them on windows, that would

bashtage:
Seems to have quite a few failures related to `AttributeError: module 'statsmodels.kernel_methods.tests.kde_utils' has no attribute 'Grid'`

PierreBdR:
Indeed, sorry, I corrected the tests.

PierreBdR:
Ok, so I still have some work to do on the tests, but if you have comments on the main part of the code, please tell me.

PierreBdR:
Ok, so I have re-written all the tests with pytest in a way that simpler, and (IMO) more "as intended" by pytest. I also increased the coverage, and merged the latest head.

I also changed the organisation of the modules to avoid having modules including many others for no reason beyond grouping of methods, and updated the example notebooks to reflect this change.

I corrected a bug in the handling of non-continuous variables, where the point-wise estimation was actually incorrect.

I would like to know if this can now be integrated into the next version of statsmodels, or if there are other things you would like me to do.

josef-pkt:
I think something is wrong with the last commit.
We don't merge master into a feature branch.

I guess the best would be to rebase on master and push it to a new branch and PR.
maybe without the last merge commit.

PierreBdR:
Arg .. I know I made that mistake, but I was sure I corrected it before pushing :/ Clearly, I was mistaken. Oh well ... I'll do what you suggest.

PierreBdR:
Here is the new pull request: https://github.com/statsmodels/statsmodels/pull/6567

josef-pkt:
a bit related: I worked on some cases for univariate kernel regression, local polynomials 

#3492 AFAIR this uses a similar binning idea as fft-KDE to get a faster implementation of kernel or local polynomial regression
#3380 related but different, nonparametric estimation based on difference filters (for variance)

PR #2385, opened 2015-05-05
ENH/WIP Generic elastic net code

kshedden:
This is a generic elastic net implementation that will work with any model that implements log-likelihood, score, and Hessian.  It replaces the previous elastic net code that was duplicated between OLS regression and proportional hazards regression.  The new code is also used to add regularization to GLM.  

Notes:

* To make the generic code work, some minor changes to ``GLMResults`` were needed: there is no generic way to attach ``mu`` to results, so instead we use ``fittedvalues`` (same thing) instead of ``mu`` in the results.

* Changes to ``GLM``: moved some code from `fit` into ``__init__`` so that ``loglike``, etc. can be called without calling `fit` first.

* Changes to ``test_glm`` : the "hacked" version of Poisson with offsets was incompatible with the refactored GLMResults, so instead I created a proper set of R results for the Poisson offset fit to cpunish.

* Added a fake ``scale=1`` parameter to ``PHRegResults`` -- the results need to follow a generic signature to work with elastic net.

* I changed the signatures of ``fit_regularized``, moving some named arguments into kwargs.  I'm not sure if it breaks compatibility.  I think we should do this so that we can implement various types of regularized fitting according to the 'method' argument, and each method will have its own set of keywords as appropriate.

* I fixed an apparent bug in ``loglike`` for ``GLM`` (arguments in wrong order).  Loglike is not used much since we usually do IRLS.  Surprisingly it passed the tests with the arguments in the wrong order.

* The OLS ``fit_regularized`` can be run with the loss function being either the residual sum of squares, or the profile log likelihood.  Only the former is comparable to R.

* Test agreement for OLS is very good.  For Cox and GLM it is OK.  Generally our results give better results on the fit function so I think R glmnet is making some shortcuts.

* This could be refactored to work with mixins as in #2374.


josef-pkt:
Good to see this, I will start to look at it in a few days.

aside about the mixins: The main reason for it right now was that I'm using the standard scipy optimizer via the main `fit` method and we don't have an option yet to specify which objective function is used (It's always the `loglike` method.) This is not necessary with a specific, non-scipy fitting method.

A second reason to combine this with the Mixin version is if we want to share and unify the `Results` for penalized likelihood. I haven't gotten that far yet, but my target was an M-estimator sandwich cov_params for the non-zero params as default, which will also need the penalized hessian. Maybe that could also be just a common penalized Results class.

Another aside: I just saw that SAS switches optimizer in an unrelated model/procedure between bfgs, quasi-newton and steepest descent (IIRC) depending on the number of parameters that need to be estimated. We might also recommend different versions of penalized estimation depending on the scale of the data once we get the different versions implemented.

josef-pkt:
As FYI:
My other target for generic penalization with scipy optimizers, is to use non-orthogonal L2 penalization with the target for penalization splines and GAM, which in the simplest case will penalize the difference between consecutive parameters.
(I haven't quite figured out yet how to use feature selection regularization for knot selection. as alternative to penalization splines.)

statsmodels/base/elastic_net.py
```diff
@@ -0,0 +1,295 @@
...
+         L1_wt=1., start_params=None, cnvrg_tol=1e-7, zero_tol=1e-8,
+         return_object=False, loglike_kwds=None, score_kwds=None,
+         hess_kwds=None):
+    """
+    Return an elastic net regularized fit to a regression model.
+
+    Parameters
+    ----------
+    model : model object
+        A statsmodels object implementing ``log-like``, ``score``, and
```
josef-pkt: spelling method is `loglike` no dash

statsmodels/genmod/generalized_linear_model.py
```diff
@@ -845,6 +861,77 @@ def _fit_irls(self, start_params=None, maxiter=100, tol=1e-8,
...
+                kwargs[ky] = defaults[ky]
+
+        result = elastic_net.fit(self, method=method,
+                                 alpha=alpha,
+                                 start_params=start_params,
+                                 return_object=True,
+                                 **kwargs)
+
+        self.mu = self.predict(result.params)
+        self.scale = self.estimate_scale(self.mu)
```
josef-pkt: why are those attached to the model?

statsmodels/base/elastic_net.py
```diff
@@ -0,0 +1,295 @@
...
+        klass = rslt.__class__
+
+    # Not all models have a scale
+    if hasattr(rslt, 'scale'):
+        scale = rslt.scale
+    else:
+        scale = 1.
+
+    # Assuming a standard signature for creating results classes.
+    refit = klass(model, params, cov, scale=scale)
```
josef-pkt: this will need more cleanup and refactoring in several models. AFAIR, we also attach some attributes directly to the results. It's one part where we are not very **generic** yet.

(It also doesn't support `cov_type`)

statsmodels/base/elastic_net.py
```diff
@@ -0,0 +1,295 @@
...
+    # Fit the reduced model to get standard errors and other
+    # post-estimation results.
+    ii = np.flatnonzero(params)
+    cov = np.zeros((k_exog, k_exog))
+    init_args = dict([(k, getattr(model, k)) for k in model._init_keys])
+    if len(ii) > 0:
+        model1 = model.__class__(model.endog, model.exog[:, ii],
+                               **init_args)
+        rslt = model1.fit()
+        cov[np.ix_(ii, ii)] = rslt.normalized_cov_params
```
josef-pkt: `normalized_cov_params` might be on the way out as **generic** always available attribute. Future of it is still open and unclear. (so far the only deviation from the pattern was in the tsa statespace models.)

My new pattern is to use and emphasise either `cov_params_default` for the full covariance or `hessian` for the case when we calculate sandwiches.

statsmodels/base/elastic_net.py
josef-pkt: There is no equivalent to `_init_keys` for creating results instances yet.

statsmodels/base/elastic_net.py
```diff
@@ -0,0 +1,295 @@
...
+    else:
+        # Hack: no variables were selected but we need to run fit in
+        # order to get the correct results class.  So just fit a model
+        # with one variable.
+        model1 = model.__class__(model.endog, model.exog[:, 0], **init_args)
```
josef-pkt: this assumes constant is the first exog?

josef-pkt:
I did a quick skimming and spot reading. I didn't read enough yet to understand all the bullet items in the PR description.

I think it looks very good overall. Changes to existing code will need another backwards compatibility review, but I didn't see any problems.

`_offset` in linear model is only for internal use. **If** we want to make it an option for the user, then we would need more invasive changes.

We still need to do more work in unifying internal APIs across models. My guess is that we still run into problems recreating results instances. 
(One option I was thinking about was to use `model.fit(start_params=my_params. maxiter=0 , ...)` to just create a new results instance. but that's mostly in the idea stage.)

As a quick check whether the most important results are available in a results instance, I usually use a smoke test for `summary`, and maybe `t_test(np.eye(len(params)))`

About the returned results (i didn't go through all the details yet): 
Since you are refitting the model with the reduced number of exog, the `cov` (hessian) will be evaluated at the unpenalized parameters, not at the reported, penalized parameters. 
Maybe this could be fixed with `model.fit(start_params=my_params. maxiter=0 , ...)`. ?
As I mentioned before, in my current view, I think we should return the sandwich (Fan and Li) that corresponds to the equivalent M-estimator. this might require the penalized hessian, and score_obs to call the sandwich_covariance code, or calculate directly if we only want the simple sandwich.

I will go over some details again and run some examples in a few days, after getting some other things out of the way.

josef-pkt:
For performance: In the long term, I would like to refactor and add code so we can avoid creating new model and results instances repeatedly in loops, if that turns out to be costly. see #2287 

It won't be immediately relevant for this. It's better to rely on generic code in the first round, but eventually we could add some private methods to the models for computational shortcuts when we don't need all the extra features that we have added to the models and results.

josef-pkt:
Another still open issue for the results instance that I just remembered 

PR for current discrete `fit_regularized` added nan handling for the cov_params (used for t_test IIRC).
My recent trend was to consider dropped variables as equivalent to imposing a zero constraint for the implementation of inferential results and cov_params usages.

For my PenalizedMixin version I was experimenting with returning `fit_constrained` with zero constraints and penalized likelihood, as "finishing" fit.

statsmodels/base/elastic_net.py
kshedden: I just need to know the concrete type of the result class.  It doesn't matter which exog we use.  I will set maxiter=0 so we don't waste time fitting something that will just get trashed.

josef-pkt:
@kshedden  I started to work on a first version for creating results instances from a auxiliary regression in #2374

I think we can largely skip creating the correct results instance in this PR, and try to develop it in a setting that is easier to unit test across models. I ran into several problems in #2374 that require more refactoring or a different approach.
(current discrete `fit_regularized` doesn't refit a reduced model and adds model specific results creation and avoided thereby to have to find a generic solution.)

We should merge the consistency improvements in the models, GLM, linear models, ..., soon(ish), so we have a common basis to write the various generic extensions

(I will soon go back to maintenance to get 0.7 out of master, so we can merge several of the waiting PRs. It's starting to get annoying to have code spread over related but distinct branches and PRs.)

statsmodels/genmod/families/varfuncs.py
```diff
@@ -45,14 +45,11 @@ def __call__(self, mu):
...
 
-
     def deriv(self, mu):
         """
         Derivative of the variance function v'(mu)
         """
-        from statsmodels.tools.numdiff import approx_fprime_cs
-        # TODO: diag workaround proplem with numdiff for 1d
-        return np.diag(approx_fprime_cs(mu, self))
+        return np.zeros_like(mu)
```
josef-pkt: There is a problem now with inheriting from `VarianceFunction`
We need to decide whether it is a base class that can be used for inheritance or a default constant variance class.

I think the only example for user defined variance function is in my risk ratio experiments with Poisson that has almost Binomial variance.

josef-pkt: I think we will be better off introducing a separate `Constant` variance class. 

But, AFAIK there is currently no usecase anymore for inheritance from VarianceFunction inside statsmodels.

statsmodels/genmod/generalized_linear_model.py
```diff
@@ -214,14 +214,40 @@ def __init__(self, endog, exog, family=None, offset=None, exposure=None,
...
         # register kwds for __init__, offset and exposure are added by super
         self._init_keys.append('family')
 
+        self.nobs = len(endog)
+
+        # Construct a combined offset/exposure term.  Note that
+        # exposure has already been logged if present.
+        offset_exposure = 0.
+        if hasattr(self, 'offset'):
+            offset_exposure = self.offset
```
josef-pkt: general comment: those `if hasattr`  can often be replaced by `getattr` with default
`offset_exposure = getattr(self, 'offset', 0)` should work in this

statsmodels/genmod/generalized_linear_model.py
```diff
@@ -281,7 +309,9 @@ def loglike(self, params, scale=None):
         expval = self.family.link.inverse(lin_pred)
         if scale is None:
             scale = self.estimate_scale(expval)
-        return self.family.loglike(expval, self.endog, scale)
+        llf = self.family.loglike(self.endog, expval, scale)
```
josef-pkt: as reminder: bugfix for signature

statsmodels/genmod/generalized_linear_model.py
```diff
@@ -809,12 +823,12 @@ def _fit_gradient(self, start_params=None, method="newton",
                                     maxiter=maxiter, full_output=full_output,
                                     method=method, disp=disp, **kwargs)
 
-        self.mu = self.predict(rslt.params)
-        self.scale = self.estimate_scale(self.mu)
+        mu = self.predict(rslt.params)
+        scale = self.estimate_scale(mu)
```
josef-pkt: This implies that the model doesn't have mu and scale attributes anymore.

josef-pkt: But it looks like irls and regularized still attach mu and scale to the model
I guess we need consistency across methods (even if it's currently not used).

josef-pkt: `fit_constrained` doesn't attach them either.

statsmodels/regression/linear_model.py
```diff
@@ -773,6 +668,189 @@ def whiten(self, Y):
...
+            return -sdr / scale
+
+
+    def _setup_score_hess(self):
+        y = self.wendog
+        if hasattr(self, 'offset'):
+            y = y - self.offset
+        self._wendog_xprod = np.sum(y * y)
+        self._wexog_xprod = np.dot(self.wexog.T, self.wexog)
+        self._wexog_x_wendog = np.dot(self.wexog.T, y)
```
josef-pkt: in terms of names, I like the `t` for transpose in the middle  xtx, xty (for internal)
`_wexog_t_wendog` ?

josef-pkt:
The scale handling in the regression linear model looks fine, I don't expect any problems with this way, and it makes it more usable for other cases.

I'll try to read roughly through the elastic net code itself in the next few days.
The only issue that I saw so far is that I want to get away from `alpha` for penalization weights `pen_weights` or `penal_weights`.

statsmodels/base/elastic_net.py
```diff
@@ -0,0 +1,326 @@
...
+        penalty weight for each coefficient.
+    L1_wt : scalar
+        The fraction of the penalty given to the L1 penalty term.
+        Must be between 0 and 1 (inclusive).  If 0, the fit is
+        a ridge fit, if 1 it is a lasso fit.
+    start_params : array-like
+        Starting values for `params`.
+    cnvrg_tol : scalar
+        If `params` changes by less than this amount (in sup-norm)
+        in once iteration cycle, the algorithm terminates with
```
josef-pkt: typo: once -> one

statsmodels/base/elastic_net.py
```diff
@@ -0,0 +1,326 @@
...
+        convergence.
+    zero_tol : scalar
+        Any estimated coefficient smaller than this value is
+        replaced with zero.
```
josef-pkt: extension for later:  the version in `discrete` checks this from the gradient and first order condition of the optimization

statsmodels/base/elastic_net.py
```diff
@@ -0,0 +1,326 @@
...
+    refit : bool
+        If True, the model is refit using only the variables that have
+        non-zero coefficients in the regularized fit.  The refitted
+        model is not regularized.
```
josef-pkt: to check: I don't know yet or don't remember how cov_params is calculated

statsmodels/base/elastic_net.py
```diff
@@ -0,0 +1,326 @@
...
+    def nphess(params, model):
+        nobs = model.nobs
+        pen_hess = alpha[k] * (1 - L1_wt)
+        return -model.hessian(np.r_[params], **hess_kwds)[0,0] / nobs + pen_hess
+
+    return nploglike, npscore, nphess
+
+
+
+def fit(model, method="coord_descent", maxiter=100, alpha=0.,
```
josef-pkt: I think the name should be `fit_xxx` in these extensions for easier import, e.g.  `fit_elasticnet`

statsmodels/base/elastic_net.py
```diff
@@ -0,0 +1,326 @@
...
+        Keyword arguments for the Hessian function.
+
+    Returns
+    -------
+    A results object.
+
+    Notes
+    -----
+    The ``elastic net`` penalty is a combination of L1 and L2
+    penalties.
```
josef-pkt: I would mention elementwise, in contrast to other penalization like Ridge/GAM where we penalize the params non-orthogonally. (I'm not sure how to phrase it.)

statsmodels/base/elastic_net.py
```diff
@@ -0,0 +1,326 @@
...
+                params[k] = 0.
+
+        # Check for convergence
+        pchange = np.max(np.abs(params - params_save))
+        if pchange < cnvrg_tol:
+            converged = True
+            break
+
+    # Set approximate zero coefficients to be exactly zero
+    params *= np.abs(params) >= zero_tol
```
josef-pkt: maybe clearer with assignment
`params[np.abs(params) < zero_tol] = 0`

josef-pkt:
I didn't try to understand all of the 1-d optimization step, I don't remember the math.

I guess the results class needs future work or enhancments. I haven't tried an example yet.

statsmodels/duration/hazard_regression.py
```diff
@@ -438,46 +441,36 @@ def fit(self, groups=None, **args):
...
-            in once iteration cycle, the algorithm terminates with
-            convergence.
-        zero_tol : scalar
-            Any estimated coefficient smaller than this value is
-            replaced with zero.
+        refit : bool
+            If True, the model is refit using only the variables that
+            have non-zero coefficients in the regularized fit.  The
+            refitted model is not regularized.
+
```
josef-pkt: kwargs will need documentation or link to it.

josef-pkt:
unit test coverage looks good for `params` but not clear to me for other results.

statsmodels/genmod/tests/test_glm.py
```diff
@@ -1620,6 +1621,52 @@ def testTweediePowerEstimate():
...
+                def plf(params):
+                    llf = model.loglike(params) / len(endog)
+                    llf = llf - alpha * ((1 - L1_wt)*np.sum(params**2) / 2 + L1_wt*np.sum(np.abs(params)))
+                    return llf
+
+                # Confirm that we are doing better than glmnet.
+                from numpy.testing import assert_equal
+                llf_r = plf(params)
+                llf_sm = plf(sm_result.params)
+                assert_equal(np.sign(llf_sm - llf_r), 1)
```
josef-pkt: numpy.testing also has a assert less

josef-pkt:
I went briefly through the latest changes and spot checked a few of the overall changes (for GLM)

looks fine to me. I guess we will need some followups until this is settled but I think we should merge it  soon.

ready to merge?

kshedden:
Thanks for the review.  Let me look it through it one more time then we can
go ahead.  Leland will need this as the basis for his GSOC project.

kshedden:
@josef-pkt OK,  went through this again, I think we should merge it.

josef-pkt:
Good, merging

Thank, to a penalized future

Issue #2416, opened 2015-05-19
ENH: Unobserved components


ChadFulton:
# Unobserved components

The first model considered in the state space models GSoC 2015 project is the univariate unobserved components models. This PR lays out the general structure and the different variations that will be allowed. Any discussion or thoughts would be very helpful, and is encouraged.

The basic unobserved components (or structural time series) model can be written (see Durbin and Koopman 2012, Chapter 3 for notation and additional details):

![uc_model](https://cloud.githubusercontent.com/assets/280864/7713168/e49fc286-fe2a-11e4-873a-e5d6b91dc6de.png)

where different specifications for the different individual components can support a range of models.

## Trend

The trend component is a dynamic extension of a regression model that includes an intercept and linear time-trend. It can be written:

![trend](https://cloud.githubusercontent.com/assets/280864/7713527/2e24bd4c-fe2d-11e4-888f-1da3b69b19bc.png)

where the level is a generalization of the intercept term that can dynamically vary across time, and the trend is a generalization of the time-trend such that the slope can dynamically vary across time.

For both elements (level and trend), we can consider models in which:

- The element is included vs excluded (if the trend is included, there must also be a level included).
- The element is deterministic vs stochastic (i.e. whether or not the variance on the error term is confined to be zero or not)

The only additional parameters to be estimated via MLE are the variances of any included stochastic components.

This leads to the following specifications:

|                                                                      | Level | Trend | Stochastic Level | Stochastic Trend |
|----------------------------------------------------------------------|-------|-------|------------------|------------------|
| Constant                                                             | ✓     |       |                  |                  |
| Local Level <br /> (random walk)                                     | ✓     |       | ✓                |                  |
| Deterministic trend                                                  | ✓     | ✓     |                  |                  |
| Local level with deterministic trend <br /> (random walk with drift) | ✓     | ✓     | ✓                |                  |
| Local linear trend                                                   | ✓     | ✓     | ✓                | ✓                |
| Smooth trend <br /> (integrated random walk)                         | ✓     | ✓     |                  | ✓                |

## Seasonal

The seasonal component is written as:

![seasonal](https://cloud.githubusercontent.com/assets/280864/7713902/ac50a648-fe2f-11e4-98c9-1afdeee8b98a.png)

The periodicity (number of seasons) is `s`, and the defining character is that (without the error term), the seasonal components sum to zero across one complete cycle. The inclusion of an error term allows the seasonal effects to vary over time.

The variants of this model are:

[...]

PR #2432, opened 2015-06-02
ENH: Statespace: Add unobserved components.

ChadFulton:
PR corresponding to #2416.

This is a pretty good first draft, but it needs work in the following areas:

- [x] Example notebooks
- [x] Improve starting parameter estimation
- [x] Unit tests
- [x] Code cleanup / style improvements
- [x] Improve model specification if possible (e.g. allow string names for the trend component as described in the table in #2416)
- [x] Add results class with result options specific to the UC model framework (i.e. syntactic sugar for getting the level, trend, etc. components from the `filtered_states` matrix)
- [x] In Summary: change overall title to "Unobserved Components Results" so that the "Model:" line can be more specific (e.g. Deterministic level + Cycle, etc)

Here's a notebook showing an example usage (the Nile models considered there are from http://www.jstatsoft.org/v41/i02/paper):

http://nbviewer.ipython.org/gist/ChadFulton/9ec4aaa13841bbdbe4df/


josef-pkt:
Notebook looks very nice. I haven't looked at the PR itself.

Is it possible to add an outlier dummy for the year 1913?
General question: Why is there little persistence in the filtered level? Break in level is abrupt, and the 1913 outlier doesn't seem to distort neighboring years.

josef-pkt:
Nevermind, I shouldn't just look at pretty plots and read some of the code also.

ChadFulton:
Question: in 1502e11 I added specification of the level / trend component via a string name; for example the deterministic trend model (see #2416) can be specified as:

```python
mod = UnobservedComponents(endog, irregular=True, level=True, trend=True)
```

or as:

```python
mod = UnobservedComponents(endog, 'dtrend')
```

The question I have is that I used the string specifications that Stata uses. Stata didn't come up with the names themselves, which are pretty standard in the UCM literature, but they did come up with these specific abbreviations.

There are other (maybe better) abbreviations (e.g. I personally like "local_level" more than "llevel"). So can we / should we use the Stata names, or should we use our own (possibly more verbose) names?

ChadFulton:
> There are other (maybe better) abbreviations (e.g. I personally like "local_level" more than "llevel"). So can we / should we use the Stata names, or should we use our own (possibly more verbose) names?

Oh, or we could support both.

ChadFulton:
> Oh, or we could support both.

I think I'll move forward with supporting them both, since there are a couple of specifications for which Stata does not have a name.

I plan to primarily use more verbose names ('llevel' -> 'local_level') but then also accept Stata's names.

josef-pkt:
In general I also prefer names that are not easy to misspell. With abbreviations I often have to guess how the word is abbreviated.

question (I haven't looked at the details yet): Does this stay with trend is constant or linear trend or can this be extended to quadratic or higher order polynomial trend?
That's a question whether you want `local_constant` and `local_trend` as separate keywords, or similar to ARIMA, `local_trend='c'` or 'ct'.

ChadFulton:
> question (I haven't looked at the details yet): Does this stay with trend is constant or linear trend or can this be extended to quadratic or higher order polynomial trend?

Trend has a bit of a different meaning here, so the "level" refers to a possibly time-varying intercept, and the "trend" refers to a possibly time-varying slope, on time. If I think of the intercept as a "degree 1 trend" and the slope as a "degree 2 trend" then it is possible to have higher-order trends (and in fact R's package KFAS allows this), but I've never seen it in practice, and it isn't referenced in the standard texts on UC models.

ChadFulton:
If you look at the trend component from #2416, the degree 3 trend would be allowing the degree 2 trend trend (beta) to have a new component in it (lets call it gamma) that evolves according to a random walk. And higher order trends are added in the same way.

josef-pkt:
about #2416  in the table you have (local) trend and stochastic trend separately, but the 2 equations look to me like a stochastic trend.  stochastic drift that is a random walk.

which is the deterministic trend and which is the stochastic trend ?
And how is a mean specified?

if beta = 0, then the mean mu is also zero
if beta = constant (sigma_beta = 0), then mu is a random walk with constant drift
if beta = random, then mu is a random walk with random walk drift

???

my initial analogy with local trend that I thought about was local polynomial regression (like in lowess or similar kernel regression)

ChadFulton:
> about #2416 in the table you have (local) trend and stochastic trend separately, but the 2 equations look to me like a stochastic trend. stochastic drift that is a random walk.

The distinction is whether the stochastic part is allowed or not. So, for example, if you had a trend but not a stochastic trend, that would mean that zeta_{t+1} = 0 for all t, or another way to put it is that sigma_{zeta}^2 = 0.

> And how is a mean specified?

If beta = 0, then mu is not necessarily zero, although it is constant. This is how a mean component can be specified.

For example, take a look at the notebook I provided for the Kalman smoother: http://nbviewer.ipython.org/gist/ChadFulton/c3bc720c2fa31c564844/. If you look at the "Deterministic level + stochastic cycle" model, the smoothed level component is non-zero (although here there's also a regression effect, but nonetheless the mu term is non-zero).

---------

Stata has a nice page in its ucm manual showing the formulas for several of the specifications. See http://www.stata.com/manuals13/tsucm.pdf, page 20

josef-pkt:
Ok, I guess I understand the stochastic distinction
`stochastic_level=False, stochastic_trend=False` specifies whether the sigma is 0 or estimated, if the corresponding level or trend is True.
About the keywords: if I want stochastic_trend=True, do I have to set, level, trend also to true, and also stochastic_level to true if I also want sigma_eta to be nonzero? 4 `True`s?

about the mean: if beta = 0, then the mean is equal to the initial condition.
 Is the intitial value of mu, mu_0, estimated jointly with the other parameters?

ChadFulton:
> Ok, I guess I understand the stochastic distinction
stochastic_level=False, stochastic_trend=False specifies whether the sigma is 0 or estimated, if the corresponding level or trend is True.

Yes, that's right. There may be a better way to set this up? But maybe people will mostly just use the keywords anyway.

> About the keywords: if I want stochastic_trend=True, do I have to set, level, trend also to true, and also stochastic_level to true if I also want sigma_eta to be nonzero? 4 Trues?

Right now:

1. If you set stochastic_trend=True, that currently has no effect unless trend=True (and otherwise it "fails" silently)
2. If you set trend=True with level=False, then it issues a warning "Trend component specified without level component; deterministic level component added.") and sets level=True and stochastic_level=False.
3. However, you do not need to have stochastic_level=True in order to have a trend or a stochastic trend.

This behavior is not quite consistent, so maybe instead of (1), above, we should have:

1(b). If you set stochastic_trend=True, then it issues a warning "Stochastic trend component specified without trend component; trend component added" and sets trend=True

> about the mean: if beta = 0, then the mean is equal to the initial condition.
Is the intitial value of mu, mu_0, estimated jointly with the other parameters?

If beta=0, then it is just an intercept estimated by recursive OLS. If level=True and trend=True but neither are stochastic, then it is an intercept and the slope of a time trend, both estimated by recursive OLS. The initialization is diffuse (mean=0, variance = 1e6) which allows it to pretty quickly pick out the OLS estimate (e.g. the red line settles down relatively quickly in the graph in the notebook I mentioned above).

You could also estimate the intercept (or even the slope) parameter by MLE (which is what we do in SARIMAX, which allows arbitrary time trends), but that is not typically done in the UC models because the stochastic extensions allow time-varying-parameters.

I suppose if you wanted arbitrary (non-time-varying) time-trends, you could add them as regressors in the `exog` argument.

ChadFulton:
I've started adding the results class. Here's a sample of what I currently have the summary looking like:

                                   Unobserved Components Results                                
    ============================================================================================
    Dep. Variable:                     US monetary base   No. Observations:                  241
    Model:             local linear deterministic trend   Log Likelihood                 786.765
                              + damped stochastic cycle   AIC                          -1563.531
    Date:                              Fri, 10 Jul 2015   BIC                          -1546.107
    Time:                                      12:00:36   HQIC                         -1556.511
    Sample:                                  01-01-1948                                         
                                           - 01-01-2008                                         
    Covariance Type:                                opg                                         
    ====================================================================================
                           coef    std err          z      P>|z|      [95.0% Conf. Int.]
    ------------------------------------------------------------------------------------
    sigma2.irregular  1.196e-12   6.26e-06   1.91e-07      1.000     -1.23e-05  1.23e-05
    sigma2.level      7.201e-05   1.66e-05      4.349      0.000      3.96e-05     0.000
    sigma2.cycle      5.057e-16   1.94e-05    2.6e-11      1.000     -3.81e-05  3.81e-05
    frequency.cycle      0.1309      0.352      0.372      0.710        -0.559     0.821
    damping.cycle        0.7866      0.157      5.021      0.000         0.480     1.094
    ====================================================================================

    Warnings:
    [1] Covariance matrix calculated using the outer product of gradients.

Here I allow multiple lines for the "Model:" specification, because the string specification is potentially very long. Does this seem like a reasonable approach?

ChadFulton:
The last commit (tentatively) adds a convenience function for plotting the components of the model. For example, the code:

```python
y = sm.datasets.nile.load_pandas().data
y.index = pd.date_range('1871', '1970', freq='AS')

mod = structural.UnobservedComponents(y['volume'], 'rwalk', cycle=True, damped_cycle=True, stochastic_cycle=True)
res = mod.fit(method='powell')
print res.summary()
res.plot_components(figsize=(13,5));
```

produces the following summary table:

                                Unobserved Components Results                            
    =====================================================================================
    Dep. Variable:                        volume   No. Observations:                  100
    Model:                           random walk   Log Likelihood                -618.646
                       + damped stochastic cycle   AIC                           1245.292
    Date:                       Tue, 14 Jul 2015   BIC                           1255.713
    Time:                               15:12:07   HQIC                          1249.510
    Sample:                           01-01-1871                                         
                                    - 01-01-1970                                         
    Covariance Type:                         opg                                         
                          coef    std err          z      P>|z|      [95.0% Conf. Int.]
    -----------------------------------------------------------------------------------
    sigma2.level      676.0566    483.558      1.398      0.162      -271.699  1623.812
    sigma2.cycle     1.702e+04   3053.499      5.574      0.000       1.1e+04   2.3e+04
    frequency.cycle     0.5236      1.286      0.407      0.684        -1.997     3.044
    damping.cycle       0.2257      0.147      1.533      0.125        -0.063     0.514

and the following graph:

plot_components

ChadFulton:
I'm still not sure about when adding convenience plotting functions is appropriate, so we can remove this if it is not in the right place.

ChadFulton:
Failure due to time-out on coverage run.

ChadFulton:
With the last four commits, I believe I have exhausted everything I had planned to do with this model. A decision needs to be made on `plot_components`, but other than that I "*think*" it is pretty well ready to go.

ChadFulton:
There is one other thing I wanted to check on here, actually: right now, as with all state space models, the default optimization method is l_bfgs. The unobserved components models don't always have the best starting parameters though (I think that `start_params` are actually pretty good a lot of the time even for complicated models with seasonals and cycles, but they're still not nearly generically as good relative to the optimum as something like ARMA).

For that reason, I have found that doing a first round with powell and then a second round with l_bfgs, starting from powell's parameters, does the best job of optimizing.

Is this something I should set up in the `fit` method, either as a default or an option? Or should I just make a note in the Notes section that sometimes powell can be helpful?

josef-pkt:
about the two method optimization:

We have plans to add this more generally, Skipper has a PR to allow sequences in the optimization method.
My experience so far using `"nm"` as starting optimizer: nm is pretty slow and needs many iterations. In a large number of cases, either newton or bfgs work well, and using automatically nm for those cases would make it much slower.  In those cases, I recommend that users use it if the default fails.

In GLM with the scipy optimizers, we use by default a few steps of `IRLS` that is controlled by a fit keyword `max_start_irls=3`, which produces in most cases good staring values for the optimizer. 
(IRLS is less general and we need to set the number of iterations with it to zero for penalized estimation.)

> Is this something I should set up in the fit method, either as a default or an option? Or should I just make a note in the Notes section that sometimes powell can be helpful?

That depends on the tradeoffs between speed in nice cases and how often a preliminary more robust optimizer is necessary.
How many iterations with `powell` do you usually need?  (I didn't like powell because a few years ago it didn't work very well on Windows, but I haven't tried or used it much since then.)

You could, for example, add a `max_start_powell` option in fit.

It would also be useful to add a link to a difficult example case in #1649
Stata often uses a two methods optimization, and we still need better defaults in several cases, but so far we have mostly "anecdotal" evidence.

ChadFulton:
I'm not quite sure about what to think. But here's some results from UC models decomposing output, money, and prices into level, trend, cycle, and seasonal. I don't know what conclusion can be drawn from this yet.

**Model 1 (Output)**

Start params llf: -103.65

        llf          method converged  sigma2.irregular  sigma2.level  sigma2.trend  sigma2.seasonal  sigma2.cycle  frequency.cycle  damping.cycle
    1 755.6           lbfgs      True         1.709e-06     8.378e-07     5.068e-06         1.53e-08     3.407e-05           0.5236          0.841
    2 455.1              nm      True          0.000374     1.704e-05     5.252e-08        0.0002535     0.0004102           0.4954         0.8465
    3   757          powell      True         3.092e-11     9.305e-06     3.053e-06        1.579e-08     3.639e-05           0.4539         0.8698
    4 758.6    basinhopping      True         3.013e-18     8.121e-05     1.434e-08        1.117e-08     1.854e-16           0.7086         0.9781
    5 756.5            bfgs     False         3.183e-12     5.378e-05     3.213e-06        1.456e-08     4.999e-06           0.6346         0.9243
    6   757  powell -> lbfgs      True         3.178e-10      1.56e-05     2.798e-06        1.455e-08     3.199e-05           0.4539         0.8698

**Model 2 (Money)**

Start params llf: -246.54

        llf           method converged  sigma2.irregular  sigma2.level  sigma2.trend  sigma2.seasonal  sigma2.cycle  frequency.cycle  damping.cycle
    1 840.3            lbfgs     False         1.653e-13     1.969e-05     3.628e-06        2.827e-11     5.212e-06           0.5226         0.8709
    2 831.2               nm     False         9.756e-08     1.515e-08     1.272e-06        3.978e-10      2.62e-05           0.2614         0.9996
    3 841.6           powell      True         4.991e-10      7.25e-06     1.805e-06         4.15e-19     1.876e-05            0.268         0.9005
    4 846.8     basinhopping      True         1.452e-15     2.499e-05     5.573e-06        7.887e-17     3.033e-14            1.047         0.3946
    5 840.4             bfgs     False         8.601e-13      1.64e-05       4.1e-06        6.455e-14     6.613e-06           0.5698         0.8252
    6 841.7  powell -> lbfgs      True         1.339e-09     2.349e-06     1.749e-06        2.893e-11      2.26e-05            0.268         0.9005

**Model 3 (Prices)**

Start params llf: -134.12

        llf           method converged  sigma2.irregular  sigma2.level  sigma2.trend  sigma2.seasonal  sigma2.cycle  frequency.cycle  damping.cycle
    1 982.1            lbfgs      True         7.556e-08     6.181e-06     2.569e-06        1.047e-08     1.024e-09           0.5237         0.9075
    2 929.5               nm      True         8.147e-08     2.597e-05     1.205e-06        1.071e-09     2.095e-12           0.3541          0.849
    3 985.7           powell      True         7.054e-07      4.17e-06     2.968e-06        7.709e-09     3.871e-11           0.6127         0.9236
    4 985.8     basinhopping      True         3.752e-07     4.886e-06     2.898e-06        8.301e-09     1.192e-15           0.6134         0.9233
    5 985.8             bfgs     False         3.547e-07      4.93e-06     2.894e-06        8.283e-09     8.338e-12           0.6138         0.9234
    6 985.7  powell -> lbfgs      True         7.052e-07     4.169e-06     2.968e-06        7.683e-09     3.851e-11           0.6127         0.9236

ChadFulton:
A couple of notes for future reference:

- The maxiter was set to 10000 and all non-convergence was due to non-changing gradient rather than running into maxiter.
- Nelder-mead tended to not perform particularly well, despite reporting convergence
- I guess powell + lbfgs isn't really much better than just powell
- Powell required 6, 5, and 10 iterations, respectively
- Compared the lbfgs step in powell + lbfgs requires 512, 368, 528 function evals, compared to 560, 1152, 528 when lbfgs is used alone
- If we regard basinhopping's result as a reference (and it was always the best result), then the other methods tend to do pretty well at getting close to that likelihood, but sometimes the optimizing parameters can have different implications.

ChadFulton:
Oh, and bfgs tends to do really well, but reports convergence failure anyway.

ChadFulton:
Newton-CG also does reasonably well, typically doing better than Nelder-Mead but worse than the others.

ChadFulton:
For the output series, here's a graph of the time plots of the estimated trend and cycle components from the various methods.

trendcycle

Interestingly, while lbfgs, powell, and bfgs find a very similar time plot (and have very similar maximized likelihoods), they have strikingly different properties from the actual optimal parameters found by basinhopping.

ChadFulton:
For the other two models, the components are very similar for all methods (except nm, which can be a bit different).

ChadFulton:
I still don't have any answers to the above questions, however I do not think I'll do anything in this PR. It is good to be merged whenever there has been appropriate review.

ChadFulton:
Rebased.

josef-pkt:
The Stata documentation also mentions and warns that there will be convergence problems with the uc models if there is too little variation so that a variance estimate tries to go to zero.
They recommend that the user drops components to get convergence.

Aside: Given the many options or variations in these models some (semi-)automatic model selection would be useful. I didn't see anything in the Stata documentation about it.

One general question to these statespace models.
variance=0 is on the boundary of the parameter space. Stata uses one-sided test for the p-value AFAICS. 
How are the summary statistics pvalue and confidence intervals for variance estimates currently defined in the statespace models. Are these the generic two sided versions?

related What's the transformation to keep the variance estimate from going negative? square?

AFAIR, mixed models don't currently don't show p-values for variance estimates.

ChadFulton:
> The Stata documentation also mentions and warns that there will be convergence problems with the uc models if there is too little variation so that a variance estimate tries to go to zero.
They recommend that the user drops components to get convergence.

This is an interesting idea. Possibly if a variance estimate is below some threshold I should add a warning to the summary output?

> Aside: Given the many options or variations in these models some (semi-)automatic model selection would be useful. I didn't see anything in the Stata documentation about it.

That would be useful, although I haven't seen anything on it. In the past I've used some information criteria methods, selection, but that can be slow and I don't know how IC performs given that these models can be stationary or non-stationary depending on the options selected.

> One general question to these statespace models.
variance=0 is on the boundary of the parameter space. Stata uses one-sided test for the p-value AFAICS. 
How are the summary statistics pvalue and confidence intervals for variance estimates currently defined in the statespace models. Are these the generic two sided versions?

Yes, Stata uses a one-sided test and we still use the generic two-sided versions. How hard is it to change the summary behavior?

As you mention, though, the variance parameters are tested against the boundary of the space, which can lead to non-standard distributions. For example in a local level model (which generally corresponds to a restricted ARIMA(0,1,1) model), if the variance of the random walk goes to zero then that corresponds to a non-invertible ARIMA(0,1,1) model.

> related What's the transformation to keep the variance estimate from going negative? square?

Yes, the transformation used is squaring.

josef-pkt:
> Possibly if a variance estimate is below some threshold I should add a warning to the summary output?

I would just add a note to the fit or class docstring for now. I'm not sure whether we can tell in an application when a small variance is too small. The trigger might be more in convergence problems than a threshold on the parameter values.
You could also add a comment about your experiments with different optimizers. e.g. basinhoppin might help if the optimization problem is difficult.

> Yes, Stata uses a one-sided test and we still use the generic two-sided versions. How hard is it to change the summary behavior?

The main thing you will need to find and add is which parameters could be on the boundary, i.e. a mask or index of parameters that are variances. Then we can decide whether we set the corresponding p-values to np.nan or replace them by one sided p-values. If Stata does the same, then we could just divide the `pvalues` for the variances by two.
(However, AFAIK dividing by two is only correct if there is only a single parameter that could be on the boundary. We might need a comment like Stata about what we use.)
I'm not sure yet how we can fix the confidence intervals and prevent invalid t_tests. That could be a new follow-up issue.

I guess the above will apply to all statespace mlemodels with variance parameters.

model selection will be for the future, I have no idea for these cases, especially with parameters on the boundaries. Maybe model selection should go more by prediction/forecast accuracy than on parameters. ICs might still be a good proxy for that.

josef-pkt:
I didn't see (check) that you have rebased the branch.

Please add a comment, so it sends out a notification.

merging now

Thanks @ChadFulton

PR #2516, opened 2015-07-10
Generalized Linear Mixed Model

gjhunt:
Poisson and Binomial with cannonical links working. Non cannonical links not working yet. Gaussian not working.


statsmodels/regression/tests/test_mixed_glm.py
```diff
@@ -0,0 +1,197 @@
...
+        # Get the log-likelihood function, and its gradient and Hessian.
+        fungrad = model._gen_joint_like_score(fe_params, cov_re, scale)
+        fun1 = lambda x : fungrad(x)[0]
+        grad = lambda x : fungrad(x)[1]
+        fun2 = lambda x : model._joint_like_hess(x, fe_params, cov_re, scale)[0]
+        hess = lambda x : model._joint_like_hess(x, fe_params, cov_re, scale)[1]
+
+        # Check that the gradients vanish at the mode.
+        mp = model._get_map(fe_params, cov_re, scale, "BFGS")
+        fp1 = approx_fprime(mp.ravel(), fun1)
```
josef-pkt: Using centered=True option will give higher accuracy at the minimum or maximum

statsmodels/regression/mixed_glm.py
```diff
@@ -0,0 +1,1420 @@
...
+        return mp
+
+    def fit(self, start_params=None, niter_sa=0,
+            do_cg=True, fe_pen=None, cov_pen=None, free=None,
+            full_output=False, method='Nelder-Mead', **kwargs):
+        """
+        """
+
+        rslt = super(MixedGLM, self).fit(start_params=start_params,
+                                        skip_hessian=True, method='nm',
```
josef-pkt: Use method=method here and define default method='nm'in the signature

josef-pkt:
Can you add an example to statsmodels/examples (inside the source tree), so it's faster to try out an example. Also if you use an R script to generate reference numbers than you can commit it inside the test directory.

josef-pkt:
Are there famous examples with public domain or non-restrictively licensed datasets that we can add for  docs and unit tests, e.g. Binomial and Poisson?

josef-pkt:
@gjhunt  Do you have a list of references on which the implementation is based upon?

We need references especially for the Laplace approximation both for review and maintenance internally, and also for users so they can check or learn what this actually does.

statsmodels/regression/tests/test_mixed_glm.py
gjhunt: Thanks. Added.

statsmodels/regression/mixed_glm.py
gjhunt: Added.

gjhunt:
@josef-pkt @kshedden 

I'll look at finding some examples. I don't know if we can use the data that is used in the LME4 examples or vignettes? Those might be ideal but I'm not sure what the restrictions on using them are. 

Honestly the reference I've been using for the implementation is conversations with Kerby. I've been looking at some lecture notes form Michael Jordan for the Laplace approximation (http://www.cs.berkeley.edu/~jordan/courses/260-spring10/lectures/lecture16.pdf). I can look at getting some better references if we need them. I'll probably have to ask Kerby for what he views as good references.

kshedden:
Some references:

http://www2.compute.dtu.dk/~hmad/GLM/slides/lect12.pdf

http://www.cs.berkeley.edu/~jordan/courses/260-spring10/lectures/lecture16.pdf

Also many references cited here:

http://stats.stackexchange.com/questions/19164/reference-for-generalized-linear-mixed-models-using-laplace-approximation

kshedden:
A document in progress attempting to derive the exact Laplace approximation for both canonical and non-canonical links:

https://umich.box.com/shared/static/cc38q61ipjd6mow2o8rzm5zgxstf6od4.pdf

statsmodels/regression/tests/test_mixed_glm.py
```diff
@@ -0,0 +1,197 @@
...
+    model = MixedGLM(endog, exog, groups=groups, family=sm.families.Poisson())
+    par = MixedGLMParams.from_components(fe_params=np.asarray([1.01505, 0.9464392]),
+                                         cov_re=np.asarray([[0.1195]]))
+    logLik = model.loglike(par)
+    np.testing.assert_allclose(logLik, -85.9951, rtol=1e-5)
+
+
+# Not working for non-canonical GLM links yet. 
+
+# def test_laplace_loglike_gamma():
```
josef-pkt: you can disable tests by changing the name instead of commenting out, that's less interuptive for the change history.
I often just add an underline like `t_est_laplace_loglike_gamma` or drop the "t" in "test", then nose will not identify as a unit tests

gjhunt: Cool. I'll do that for future commits.

gjhunt:
Updated. I got a correctly implemented gradient and hessian for the Laplace approximation sub-problem working for non-canonical links. It is noticeably faster than using the (albeit very accurate) numerical approximations. I need to now add some more tests for the log-like and add some tests showing the .fit() function. Things that might need to be fiddled with are the two optimization problems (1) for Laplace and (2) for .fit(). They can be temperamental. Although I'm highly confident that the things we are passing to them are correctly now. The current tests seem to pass well in the canonical cases however they still aren't matching quite perfectly for the non-canon cases. Its possible (probable?) that glmer is using an approximation to the Laplace approximation so I'll see if I can find anything about that. Something like that might explain the discrepancy. I'm fairly confident in these results so I'm starting to think that glmer is doing something different than a straight-up Laplace approximation. Strangely the log-like seems to be accurate for non-canon cases when the number of groups is small but diverges when this gets large. I'll look into that more.

statsmodels/genmod/families/family.py
```diff
@@ -516,16 +516,16 @@ def loglike(self, endog, mu, scale=1.):
...
         """
-        if isinstance(self.link, L.Power) and self.link.power == 1:
-            # This is just the loglikelihood for classical OLS
-            nobs2 = endog.shape[0]/2.
-            SSR = np.sum((endog-self.fitted(mu))**2, axis=0)
-            llf = -np.log(SSR) * nobs2
-            llf -= (1+np.log(np.pi/nobs2))*nobs2
-            return llf
-        else:
+        # if isinstance(self.link, L.Power) and self.link.power == 1:
```
gjhunt: So this is really a hack right now because MixedGLM assumes the log-likelihood to be in the normal form. We should probably support Gaussian case though? In which case this needs to be changed if it hasn't already in an updated version? I think I'm on .7 right now.

josef-pkt: Kerby has prepared PR #2353 
I wanted to postpone it for 0.8 because I'm not sure about the API and impact.
I see if I can figure out #2353 in the next few days with examples and unit tests. #2353 will be merged before MixedLM, so this can always be adjusted later.

kshedden:
For the binomial variance function it should be easy to use the analytic derivative.

kshedden:
Do you have a derivation for your derivative and Hessian?  I think what I derived earlier is the derivative and Hessian with respect to beta.  We need the derivative and Hessian with respect to the random effects, right?

kshedden:
We should retain the earlier simple canonical calculation and use it when applicable in the Laplace calculations.  You can create a private variable _canonical_link for each family (e.g. _canonical_link = statsmodels.genmod.families.links.logit for Binomial family), then set a variable to true/false in the GLM class:

self._canonical =  isinstance(self.family.link, sm.family._canonical_link).  

If self._canonical is True you would use the old code that is simple and fast.

It would also be good to come up with a test where you use the non-canonical code in a setting where the link is actually canonical, then compare the results using the old and new code.  You could do that by creating a version of the link class that is not recognized as being the canonical link.

gjhunt:
Here is my derivation for basically everything I've implemented up to this point:

https://drive.google.com/file/d/0Bw80pOADr9y6OXRKM2RfT0Q0cHc

Right now its hand-written. I can TeX it up if people want a more read-able version.  What you derived was with respect to beta (for a GLM?), Kerby, but it plays a similar role to the random effect in GLMMs. This derivation was done independently of yours but I think it matches, generally, although its not done in precisely the same manner. The notation might need a bit of work and I had to guess at the form of the product rule for derivatives of vector functions (because I couldn't find any sources for this) but it seems to reduce to the correct form in the canonical case and matches approx_hess/approx_fprime when implemented so I'm highly confident in the math here. 

I think there are a couple things we should add. It would probably be pretty easy to work out some of the simplified analytic derivatives in the links.py and varfunc.py classes. Those might help speed things up a bit. Right now the derivatives of the variance functions are mostly approximated with approx_hess which is accurate but slow. There are also still some cases where the first and second derivatives of the inverse link functions could be simplified in links.py. 

I think adding a field in the Family class for canonical links is a good idea. I think I'll have to add a link because the canonical link for the Gamma Family doesn't exist yet (it's negative-reciprocal?). Or I'll just throw an error. Using the simplified canonical calculations is a good idea because it will be simpler and faster (even though its already faster in the canon case). I can probably hack it to use the non-canon calculations but the canonical link for some tests to make sure that the non-canonical calculations still go through when using the canonical link. 

Currently I'm calculating the joint log-likelihood, its gradient with respect to the RE and it's Hessian wrt the RE all at the same time because they use so much of the same overlapping information. Since I'm already calculating the hessian I wonder if it makes sense to pass that along to the optimizer for the Laplace method. A newton-like optimizer will probably use that information, no? I'd assume it can't hurt. Otherwise (and maybe I'll do this anyways) I can always add some flags into the _joint_like_grad_hess function to only actually calculate/return the likelihood/gradient/hessian as desired.

josef-pkt:
To the last point if score and hessian are defined then `super().fit()` will pick them up automatically if the optimizer `method` is one that uses score or hessian (first or second derivatives)
(scipy has two more Newton methods in recent versions, that are not connected yet in our wrapper of scipy optimizers.)

josef-pkt:
it looks like all but the last two travisci failures are connected to the definition of loglike (my guess is related to the scale estimate)

gjhunt:
@josef-pkt I'm talking about the grad/hess for the Laplace optimization sub-problem not the .fit(). To calculate the loglike to be optimized over the fe_params and cov_re in .fit() we need to solve an optimization problem (in order to approximate the integral integrating out the random effects we need the first two derivs wrt the REs of the joint loglike of the parameters (fe_params and cov_re) and the REs). This is what I've been fighting with to get to work with the non-canonical case -- the derivs of the joint loglike wrt the REs. I'm talking about the optimization here:

https://github.com/gjhunt/statsmodels/blob/glmm/statsmodels/regression/mixed_glm.py#L968

gjhunt:
The failures (I think) were due to me hacking around with the Gaussian loglike. I've hacked around with it better so they don't fail. It still will have to be changed in production.

kshedden:
Around line 889 it would be cleaner to have something like this (also slightly faster because there are no temporary matrices):

resid = self.endog_li[k] - mean
vec = scale * resid * deriv_link_inv / var_mean
d[k, :] = np.dot(exog_re.T, vec) - s[:, k]

kshedden:
We should have all the canonical links.  I think for gamma the reciprocal link is equivalent to using the -1/x as the link.  See links.py line 359 for the reciprocal function (called inverse_power). 

You definitely don't want to throw an exception in something that is just there to improve speed.  If you aren't sure that you have the canonical link, then fall back to the more complicated code.

kshedden:
Here is an update to my derivation of the gradient/Hessian for the Laplace approximation:

https://umich.box.com/shared/static/cc38q61ipjd6mow2o8rzm5zgxstf6od4.pdf

I still need to review some of the details in the Hessian calculation.  But the general form is z' * D * z, where z is exog_re, and D is a diagonal matrix.  If this is correct, then it shouldn't be necessary to have the implicit loops that are currently on lines 905 and 918.  In numpy, if d is the diagonal of d, then

np.dot(z.T, d[:, None] * z)

is a fairly efficient way to get this result.

gjhunt:
So the two links that I think aren't canonical are for Gamma and InverseGaussian. AFAIK those canonical links are -1/x and -1/x^2, respectively, (c.f. https://en.wikipedia.org/wiki/Generalized_linear_model#Link_function) the two closest existing links are inverse_power and inverse_squared but those implement 1/x and 1/x^2 and are missing the negative sign. I'm not sure how big of an issue it is. However its super easy to implement negative versions so I've done that.

gjhunt:
I suppose I can test the negative versions against the non-negative versions of these links to see if there is any difference. I think it might do something like flip the sign of the estimated coefficients?

gjhunt:
Right now I have a batch of tests updated. Lots more need to be added and some optimization needs to be done in the code. I wanted to get some tests written before I fiddled around with optimizing the code because I was afraid I was going to break something. 

Right now the problems from these tests look like this:
(0) Some weird error for Gaussian getting thrown in model.py?
(1) Some problems with precision using Binomial family and cloglog/probit links. The numerically approximated hessian doesn't match the analytically derived one. Honestly I trust the analytical hessian more. 
(2) Some problems with NaNs in the gradient of InverseGaussian. Probably related to the domain of the data I'm giving it and the link function I'm using?
(3) Loglike value for non-canonical links doesn't match GLMER. Probalby GLMER is using some other approximation? I'm thinking we'll have to probably look at fitting these GLMMS with some other package or software to get a benchmark for Laplace-based loglike values. I'll need to look into this more. 

Next I'll be figuring out if there are any other easy links I passed over testing in the current Families. Eventually all of the links should be tested but the data needs to conform to the link-family domains correctly and I'm not going to cook up a bunch of special data for that quite yet. 

After that I'll write some tests for the Laplace optimization sub-problem. My intuition is that this will reveal some design issues that will need to be addressed because the scipy.optimize.minimize loves to fail for complicated links. 

After those are implemented I'll directly write some log-like tests for families and links and then some .fit() tests. Point (3) above will be pertinent here.

josef-pkt:
@gjhunt   I would put issues like (1) and (2) lower on the priority because most likely those are not GLMMixed specific. 
I haven't checked your code, but we should leave some time to get the results to work, We will need extra model specific results, unless those are already copied from MixedLM.
If not all family-link combinations work correctly, then we could leave figuring out those for the end.

To (1)  I wrote unit tests to compare GLM with the discrete models. In most cases the precision of the agreement was very high. However, the Hessian of the Binomial with Probit link differs from the discrete Probit at a few decimals (I don't remember the exact precision.) From looking at how those are calculated I didn't see any bug, and I thought it's most likely numerical precision problems that I didn't try to dig into.

To (2) IIRC, Kerby reported a similar problem with Gamma, because we don't enforce strictly larger than zero. My guess is that we will need to separate negative power link (1/x, 1/x**alpha with alpha >0) from the positive power links, which don't need to enforce x>0.

To (3) Do your parameter estimates agree?  If those agree, then you would have still the same loglike function except for optimization irrelevant transformations.
In the past we had sometimes problems when other statistics packages didn't use the full loglikelihood and dropped, for example, constant terms. Sometimes it was possible to reverse engineer the correction term. Also, depending on the pattern of what can be verified, some things will be correct by implication because it uses the same code that is tested in several other combinations. (I thought R usually only has deviance instead of `llf`for GLM models).
Stata has mixed GLM with a Laplace options (IIRC), that we can also compare to.

gjhunt:
Added an example in statsmodels/examples. The file structure may be wrong but the example is there and works well. 

The main thing I added was some efficiency enhancements to _gen_joint_like_grad_hess and _get_map in mixed_glm and .fit(). The code is a bit more complicated now but I think speed-wise its a fair bit better. 

I still could add some more real data examples. I'm planning on adding some examples of .fit(), some for _get_map and some to test canonical links using the non-canonical link code as Kerby suggested. I still want to also add some examples involving situations where self.k_re is bigger than 1 because that's a largely untested area right now as most of these examples are just random intercepts. However I don't quite understand how to mimic this behavior in R code so I may need some help with this or I'll have to spend some time looking into it myself. 

For small data sets fitting the MixedGLM seems to work pretty well. However if the data set gets too big our statsmodels implementation has problems fitting non-canonical links. Fitting with canonical links seem to work okay in larger data sets. I'm talking about like 1500 total observation being a problem for non-canonical links. 

As far as I can tell basically the Laplace optimization subproblem seems to fail a lot for "large" data sets and non-canonical links. This seems to depend a fair bit on what the current parameters of the model. So I've tried to restart the .fit() with a different starting parameter guess if the Laplace optimization has too much trouble. This works great for small data sets but for the larger data sets even this doesn't seem to alleviate the problem. Either way fitting seems to get exponentially harder for larger data sets especially so for non-canonical links. Partially I think .fit() has a harder time optimizing because its a larger data set and more complicated function and the Laplace step seems to balk at that difficulty. I'm by no means an expert in these optimization algorithms. I've looked at these for a bit and definitely welcome any help here.

gjhunt:
It's possible that implementing the score and information for the log-like will help solve some of these convergence issues however those calculations involve their own numerical integration and potentially another round of similar issues so I'm not entirely sure what will happen. Implementing the gradient and hessian definitely helped for Laplace so I'm cautiously optimistic something similar will happen for the score/information mtx.

josef-pkt:
`clean` in `InversePower` looks good to me. I don't think we can use it directly for the existing families, because of the change in sign. We might need some deprecation.

about the optimization problems: do they occur only further away from an optimal solution or also with starting values that are close to the optimal solution?

@gjhunt  I don't think it will be possible to solve all problems before we should merge this. It would be pretty good already if it works for some standard use cases and has verified test coverage for those. Open issues can be improved in future PRs if necessary. But it is important to document the limitations and differences to the reference implementation in R. You could open issues for things that remain unresolved after the GSOC, and open for example FAQ issues for a summary about differences to R's implementation. 
(As example, #2252 lists open issues for last years statespace GSOC when I merged it, which have been largely resolved by now.)

gjhunt:
My intuition is that the optimization problems are better the closer they are to the optimal solutions. This is largely intuition though. I'll have to play a bit more with initialization to make any stronger statements. I need to take a look at which types of parameters tend to cause problems  -- its probable that we can come up with something a little smarter than uniform (0,1) random initialization. 

I think you're right about solving these issues before the merge, however. The fitting seems to work decently well for standard cases as long as the data sets aren't too large. Hence I think we have some code that is ready to merge. I think I'll try and focus on the remaining test coverage and collecting and documenting existing issues and limitations.

josef-pkt:
About optimization:

AFAICS you are currently using random start params.
I think you should use standard GLM params as starting values for the exog, fixed effects parameters.
GLM should provide consistent estimator for the fixed effects params even under misspecified variance  so using those as start_params for MixedGLM fit should be pretty good.
For now I would attach the auxiliary GLM model, and we decide later whether to keep it or not, or to make attaching  it optional. My guess is that we could get specification tests directly from the comparison between MixedGLM and GLM fits.

I expect that this would solve problems that arise from evalutating the model at values far from the optimum

josef-pkt:
@gjhunt  When you said that there are optimization problems with larger datasets, do you have larger datasets with many groups or many observations in each group?   (wide or long panel)

statsmodels/regression/tests/test_mixed_glm.py
```diff
@@ -0,0 +1,350 @@
...
+
+    sqrt_link = sm.families.links.sqrt
+    model = MixedGLM(endog, exog, groups=groups, family=sm.families.Poisson(link=sqrt_link))
+    par = MixedGLMParams.from_components(fe_params=np.asarray([0.5840612,0.5596115]),
+                                         cov_re=np.asarray([[1.2337**2]]))
+    logLik = model.loglike(par)
+    np.testing.assert_allclose(logLik, -118.5216, rtol=1e-4)
+
+def test_laplace_loglike_gamma_log():
+    """
```
josef-pkt: The docstring mess up the failure report in nose. The failing test shows mainly `Python` for the test function (header), see TravisCI failing test.

In general we use code comments `#`. but here I don't know what the best way would be to avoid having every line start with `#`. (If there is a python expression before, then `"""` would start a string instead of a docstring. Assign string to code?)

statsmodels/regression/tests/test_mixed_glm.py
```diff
@@ -0,0 +1,350 @@
...
+            fp1 = approx_fprime(ref.ravel(), fun1, centered=True)
+            fp2 = grad(ref.ravel())
+            np.testing.assert_allclose(fp1, fp2, atol=ATOL, rtol=RTOL)
+
+            # Check the Hessian
+            he1 = approx_hess(ref.ravel(), fun1)
+            he1ld = np.linalg.slogdet(he1)[1]
+            he2 = approx_fprime(ref.ravel(), grad, centered=True)
+            he2ld = np.linalg.slogdet(he2)[1]
+            he3 = hess(ref.ravel())
```
josef-pkt: you could combine this with `grad` to calculate fp2 and he3 at the same time (small speedup in avoiding duplicate calculation)
Also, `rev` is already `ravel`ed AFAICS

gjhunt:
Re optimization:
You're saying that we should try fitting a (non-mixed) GLM with the same endog, exog and fixed-effects structure and use those fitted params as the starting values?

As far as larger data sets go I was working specifically with the toenail data set mentioned previously: 
http://rpubs.com/bbolker/toenail_laplace
http://www.stata-press.com/data/r12/toenail.dta

It's decently large with maybe total observations of 1900 and 300 groups -- so something like 6 or 7 obs per group. The R fit is

m_glmer <- glmer(outcome~treatment+visit+(1|patient),toenail,
            family=binomial))

It fits pretty quickly for a binomial-logistic model in statsmodels (maybe a few minutes) however if you switch to a cloglog link then fitting gets a lot tricker. If you only work with, say, the first 500 observations then fitting it becomes possible but I was having a lot of trouble getting the MixedGLM.fit not to fail too many attempt trying to fit it over several hours.

josef-pkt:
> Re optimization:
You're saying that we should try fitting a (non-mixed) GLM with the same endog, exog and fixed-effects structure and use those fitted params as the starting values?

Yes, we did something similar with MixedLM to switch to use OLS starting values and it is a considerable speedup. 
You still need additional start_params for the random effects distribution. I guess something like uncorrelated random effects with small positive variances might work reasonably well.
It might also be possible to get a preliminary estimate for the random effects distribution, but I guess it might not be much more efficient than starting MixedGLM from GLM.

To get a rough comparison with R: how long does fitting the Logit MixedGLM take in R? relative to your "few minutes". 
I would expect more that it should fit in a few seconds. (But optimization can wait, reasonably good convergence is more important.)
Also, I think GEE parameter estimates for fixed effects would also be (very) close to the optimal solution for MixedGLM, but would take considerable more time than just GLM for getting start_params.

Also, maybe you have a reference to confirm that the GLM params are consistent under MixedGLM assumption, which is what I understand. MixedGLM will be more efficient (smaller cov_params), but my guess is that in large sample it won't make a huge difference.

statsmodels/regression/mixed_glm.py
```diff
@@ -0,0 +1,1514 @@
...
+        # Convert to the final parameterization (i.e. undo the square
+        # root transform of the covariance matrix, and the profiling
+        # over the error variance).
+        fe_params = params.get_fe_params()
+        cov_re_unscaled = params.get_cov_re()
+        scale = self._scale
+        cov_re = scale * cov_re_unscaled
+
+        # TODO: Fix pcov
+        pcov = 1
```
josef-pkt: Getting cov_params to work is high priority.
Can we just use the Hessian? 
Is loglike the log likelihood as function of all parameters, so we could just use derivatives?

statsmodels/regression/tests/test_mixed_glm.py
```diff
@@ -0,0 +1,350 @@
...
+        grad = lambda x : fungradhess(x)[1]
+        hess = lambda x : fungradhess(x)[2]
+
+        for k in range(10):
+
+            RTOL = 1e-2
+            ATOL = 1e-2
+
+            # An arbitrary state for the random effects
+            ref = np.random.normal(size=(model.n_groups, model.k_re)).ravel()
```
josef-pkt: Based on the first test error: fun is nan
are these parameters all valid?  My impression (just getting started so I haven't figured it out yet) is that the unpacked params are not arbitrary.

statsmodels/regression/tests/test_mixed_glm.py
```diff
@@ -0,0 +1,350 @@
...
+
+        self.like_hess_grad_helper(model, fe_params, cov_re, scale)
+
+    def test_links_derivs2(self):
+        for lnk in self.test_links:
+            try:
+                self.like_hess_grad2(self.family, lnk)
+            except Exception as e:
+                e.args += ("Link type: " + str(lnk),)
+                raise
```
josef-pkt: we get a nicer failure message if instead of using `try ... except` we just add the link to the err_msg in the asserts

statsmodels/regression/mixed_glm.py
```diff
@@ -0,0 +1,1514 @@
...
+
+        return (self._gen_joint_like_grad_hess(fe_params, cov_re, scale, fn, grad, hess))(ref)
+
+    
+    def _gen_joint_like_grad_hess(self, fe_params, cov_re, scale, fn = True, grad = True, hess = True):
+        """
+        Function, gradient and hessian of the joint log likelihood.
+
+        The log-likelihood is for the data and random effects, viewed as a
+        function of the random effects.
```
josef-pkt: It would be helpful to specify in a notes section or here which derivatives these are, they are with respect to random effects keeping fixed effects as constant, are they?

statsmodels/regression/mixed_glm.py
```diff
@@ -0,0 +1,1514 @@
...
+        fn: bool, optional
+            If True the returned function calculates the loglikelihood value.
+            Else the returned function always reports zero for the loglike.
+            If not specified defaults to True
+        grad: bool, optional
+            If True the returned function calculates the gradient value.
+            Else the returned function always reports a matrix of zeroes
+            for the gradient.
+            If not specified defaults to True
+        hess: bool, optional
```
josef-pkt: call it `hess_logdet` or something else, I got confused because I thought `hess` is hessian.

statsmodels/regression/mixed_glm.py
```diff
@@ -0,0 +1,1514 @@
...
+        tuple
+            the value of the negative joint log likelihood,
+            its gradient and the log-determinant of it's hessian
+            evaluated at the given state, respectively, in a tuple.
+        """
+
+        return (self._gen_joint_like_grad_hess(fe_params, cov_re, scale, fn, grad, hess))(ref)
+
+    
+    def _gen_joint_like_grad_hess(self, fe_params, cov_re, scale, fn = True, grad = True, hess = True):
```
josef-pkt: what does `joint` stand for?    If it refers to all groups, then I would drop it, assuming by default loglike grad hess refer to the entire sample.

josef-pkt: I guess I figured out that joint (in Madsen et all) stands for likelihood including the individual random parameters. (non-marginal)

josef-pkt:
The Madsen et al lecture slides helped to get a basic overview over the nested optimization structure.

@gjhunt  Do you have a unit test for the simple cases that compares to R, params, ...?

josef-pkt:
A good reference comparing Laplace approximation with Adaptive Gaussian quadrature

Pinheiro, José C., and Edward C. Chao. "Efficient Laplacian and adaptive Gaussian quadrature algorithms for multilevel generalized linear mixed models." Journal of Computational and Graphical Statistics 15.1 (2006).
https://scholar.google.ca/scholar?cluster=8679236785542571963&hl=en&as_sdt=0,5&sciodt=0,5

statsmodels/regression/mixed_glm.py
gjhunt: Yes. Joint meaning the fe_params, cov_re and the random effects that need to be integrated out.

statsmodels/regression/mixed_glm.py
gjhunt: Yup. They are derivatives wrt the random effects keeping fixed effects and the random-effect covariance structure constant.

gjhunt: I'll try to make the comments a bit clearer.

statsmodels/regression/mixed_glm.py
gjhunt: Yup. Will change in next revision.

statsmodels/regression/tests/test_mixed_glm.py
gjhunt: This makes more sense. I'll rewrite it this way.

statsmodels/regression/tests/test_mixed_glm.py
gjhunt: 'ref' are the random effects which are assumed to have a normal dist so there should be no problem with any value.

statsmodels/regression/tests/test_mixed_glm.py
gjhunt: I'll just make the docstrings a variable 'code' as suggested.

statsmodels/regression/tests/test_mixed_glm.py
josef-pkt: Then you need to track down why the fun1 value is nan.   (I didn't check which tests case this was where started the debugger.)
There could still be numerical over- or underflow in specific cases, even if theoretically there is no problem, and when there might not be a problem at the estimated values.

gjhunt: Yes. I'll add it to the list of things that need to be fixed.

statsmodels/regression/mixed_glm.py
gjhunt: Yea, I don't really understand a lot of the stuff going on in fit. Not sure I follow your comment.

gjhunt:
So I added a GLM fit which initializes the start_params. I'm not sure it fixes or problems. I just added some fit tests comparing some canonical cases to R. As you can see the negative binomial test fails to fit even with the new GLM initialization.

josef-pkt:
Viewed from the other side: Does this mean Gaussian, Binomial and Poisson work correctly?

Those three are the important ones that are reasonably easy to optimize. The others might have messy optimization problems that need specific attention (after the GSOC).

gjhunt:
The short answer is yes. For canonical links those three fit correctly and quickly.

josef-pkt:
> For canonical links those three fit correctly and quickly.

Very good.
How well do non-canonical links work for Gaussian and Binomial? Which if any work?

I just remembered: R GLM doesn't have `cloglog`, R has a misnamed `loglog` which we don't have yet (open issue IIRC).

gjhunt:
So for small data sets it can definitely fit some non-canonical links for those. Its a little more complicated than "which links work?". Things start to get difficult if the data gets too large. So for some links they work for small data sets and not for larger ones. Its hard to tell the accuracy because I don't have much to compare it against since R uses some approximation for non-canonical links. 

I added a Gaussian fit example which unfortunately isn't quite working yet. Not clear to me why the basic Gaussian case should be failing.

josef-pkt:
General question: Does R also have adaptive quadrature as a numerical integration option?
I was looking at an article that uses 7 point adaptive gaussian quadrature as base reference for the accuracy of Laplace approximation and penalization approximation.

josef-pkt:
What I suggest given that we are running out of time for GSOC right now:

Focus on the 3 cases that work: Gaussian, Binomial and Poisson with canonical link, and get the missing results (like bse/cov_params) and test coverage compared to R for those cases.
Then we have the basic functionality working and unit tested, and we can then slowly expand on the parts that don't work properly or where we don't have an exact equivalent in R for reference.

gjhunt:
That sounds like a good plan. I'll try to finish these up.

gjhunt:
What's the thoughts on leveraging some parallel computing for this stuff? the multiprocessing library is part of python's standard library. For larger data sets we might be able to justify the overhead. There are some embarrassingly parallel problems here e.g. computing the means for each point or even looping over groups. We could take advantage of multi-core processors this way. This wouldn't solve convergence problems but could give a speed-up.

josef-pkt:
@gjhunt  Yes, that will be a possibility to add. We have an optional dependency on `joblib` for parallel processing.  (I guess it would also be a good candidate for out of core processing because of the embarrassingly parallel structure.)

However, I wouldn't add it in the first round when we still need to fix some problems, because it would make that more difficult.

gjhunt:
Yup. I think its best to try and get these convergence issues fixed.

jona-sassenhagen:
Is this still being worked on? It would be amazing to have a fully-featured LMM in statsmodels.

kshedden:
Yes, I'm hoping to get this usable within a month or so.  This is GLMM, our
LMM is already somewhat fully-featured (see MixedLM).

jona-sassenhagen:
Cool.

The features I'm most missing would indeed be generalized linear mixed models plus crossed random effects. Both are huge right now in linguistics and psychology.

kshedden:
We have decent support for crossed linear mixed models now.  If you try it
out and have trouble file an issue or post to the statsmodels list and we
will try to help out.

Mixed GLM is not really a "feature", it is an entirely different code base
from the ground up.  But we should be able to build up to something usable
over time.

jona-sassenhagen:
>We have decent support for crossed linear mixed models now

Cool!

mmagnuski:
@kshedden 
> We have decent support for crossed linear mixed models now.

It should be corrected in the documentation then, I was not using statsmodels LMM before because I needed crossed random effects and [the docs say](http://www.statsmodels.org/dev/mixed_linear.html):

> Some limitations of the current implementation are that it does not support structure more complex on the residual errors (they are always homoscedastic), and **it does not support crossed random effects**.

kshedden:
@mmagnuski I will change the docs.  It remains unclear how powerful our implementation is when it comes to crossed effects.  We do not use sparse Cholesky like R and some other packages.  We do have some other tricks to help with performance.

pzivich:
Is GLMM implementation still in progress?

kshedden:
We have a new PR #4189 for this using a Bayesian approach.  If you can try
it out and make comments that would be great.

josef-pkt:
This pull request **introduces 14 alerts** when merging 76b63d053ae51b808f4950158ca1023a26d44887 into b5cb3de385b225ba57af6021d1a4936e76bfdcad - [view on lgtm.com](https://lgtm.com/projects/g/statsmodels/statsmodels/rev/pr-3020e2450cc0bd3407888ea52945258ff65043f4)

**new alerts:**

* 10 for Module\-level cyclic import
* 1 for Wrong name for an argument in a call
* 1 for Missing call to \_\_init\_\_ during object initialization
* 1 for Redundant assignment

---

emjun:
It looks like the Bayesian implementation of GLMMs is a great start but incomplete. 

What is the latest effort for supporting more families in GLMMs? 

I am just digging into this part of the code base, but would a reasonable implementation approach be to port/reuse some of the family (link and variance) calculations in the fit method for [MixedLM](https://github.com/statsmodels/statsmodels/blob/master/statsmodels/regression/mixed_linear_model.py) to "prep" the model for fitting in [GLMM](https://github.com/statsmodels/statsmodels/blob/master/statsmodels/genmod/generalized_linear_model.py) ?

[Apologies if this seems like a naive question, really just trying to see how I can contribute] Conceptually, a GLMM can be thought of as a GLM with mixed effects, so it seems like we should be able to do some "matrix expansion" to include the random intercepts and slopes before using the GLM fitting procedure. I know that we might need a new fitting algo (which it seems the [Bayesian implementation](https://github.com/statsmodels/statsmodels/pull/4189) provides), but is there any reason we can't do this in GLM class? Am I missing something about the complexity of incorporating mixed effects in the the current GLM class implementation?

PR #2603, opened 2015-08-28
surv object being overwritten bug

joestox:
the PHSurvivalTime.surv object was getting overwritten when the number of strata is greater than 1. This commit fixes the bug. I did not include a test. I'm not that comfortable writing them yet.

### DIFF
--- a/statsmodels/duration/hazard_regression.py
+++ b/statsmodels/duration/hazard_regression.py
@@ -1222,8 +1222,8 @@ class PHReg(model.LikelihoodModel):
                 xp0 -= e_linpred[ix].sum()
 
             cumhaz = np.cumsum(h0) - h0
-            surv = np.exp(-cumhaz)
-            rslt.append([uft, cumhaz, surv])
+            current_strata_surv = np.exp(-cumhaz)
+            rslt.append([uft, cumhaz, current_strata_surv])
 
         return rslt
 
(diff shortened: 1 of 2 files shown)


josef-pkt:
@joestox Thanks

fix looks correct to me.
@kshedden 

(We have currently a problem with one machine on TravisCI which are unrelated to this.)

kshedden:
It looks reasonable.  Apparently I had a test for this that is commented
out in master, I can't remember why, or maybe it was inadvertent.  Can you
try uncommenting line 83 in test_phreg.py and see if the tests pass after
making your change?

joestox:
The test case runs successfully after uncommenting. Had to change baseline_hazard on line 83 to baseline_cumulative_hazard

kshedden:
This looks good to me.  Thanks @joestox !

josef-pkt:
good, then I recommend copying the line also to the last case in these sequence which uses strata, i.e. after line 104. And add a code comment `# smoke test` before the line to indicate that we are not comparing numbers yet.
(could also be done with amend last commit and force push.)

joestox:
ok, done. checked it out and it runs fine.

josef-pkt:
@joestox Thank you, I'll merge when TravisCI reports no related failures.

josef-pkt:
merged, (sorry for the delay, I got distracted)

joestox:
Awesome! Thanks

PR #2647, opened 2015-10-08
FIX: PEP-0440 version checks for numpy

saketkc:
See #2645


saketkc:
The build time did not improve. Also a random failure with python 2.7

rgommers:
There's a lot of strange failures on TravisCI here. The ``setup.py`` change should be good to go in I think. Maybe separate that from the CI changes?

bashtage:
Is this still needed?

PR #2805, opened 2016-02-09
ENH: GLM Weights

thequackdaddy:
Again, I'm a newbie at this, and I haven't documented this much... So I appreciate your forbearance as you work with me. 

So a few notes:
* I pass all the genmod tests other than the perfect separation test. The perfect separation test essentially gives perfect separation, but it it fits. I changed the perfect separation test to something really, really obvious and passed the test that way. 
* As far as I can tell, I am only failing `statsmodels.base.tests.test_shrink_pickle.TestRemoveDataPickleGLM.test_remove_data_pickle` test. I don't understand this test, so I'm not sure what to look for. If you could help me look into that, I'd appreciate it. 

You can compare the results from R with my results for Gaussian, Gamma, and Poisson using the code below:

R:

    library(MASS)
    data(Insurance)

    Insurance$Group <- factor(Insurance$Group, ordered = FALSE)
    Insurance$Age <- factor(Insurance$Age, ordered = FALSE)
    Insurance$Group <- relevel(Insurance$Group, "1-1.5l")
    Insurance$Age <- relevel(Insurance$Age, "25-29")
    model <- glm('Claims ~ 1 + C(District)  + Group + Age', 
                 data=Insurance, 
                 family=poisson(), 
                 weights = unlist(Insurance['Holders']))
    summary(model)
    model <- glm('Claims ~ 1 + C(District)  + Group + Age', 
                 data=Insurance, 
                 family=gaussian(), 
                 weights = unlist(Insurance['Holders']))
    summary(model)
    Insurance$Claims[Insurance$Claims ==0 ] <- 0.001
    model <- glm('Claims ~ 1 + C(District)  + Group + Age', 
                 data=Insurance, 
                 family=Gamma, 
                 weights = unlist(Insurance['Holders']))
    summary(model)

Python Code: 

    import pandas as pd
    from patsy import dmatrices
    import statsmodels.api as sm
    import numpy as np
    
    data_link = 'https://vincentarelbundock.github.io/Rdatasets/csv/MASS/Insurance.csv'
    data = pd.read_csv(data_link, index_col=0)
    y, x = dmatrices('Claims ~ 1 + C(District) + Group + Age', data)
    sm.GLM(endog=y, exog=x, family=sm.families.Poisson(), data_weights=data['Holders']).fit().summary()
    sm.GLM(endog=y, exog=x, family=sm.families.Gaussian(), data_weights=data['Holders']).fit().summary()
    y[y == 0] = 0.001
    sm.GLM(endog=y, exog=x, family=sm.families.Gamma(), data_weights=data['Holders']).fit().summary()


josef-pkt:
Looks good overall, There are a few places where I don't see how the changes work, e.g. Binomial, or where it's not clear to me yet how it's supposed to work.

In terms of unit tests: 
Besides checking against R, it would be good to have a simple model that can be calculated with weights and  with expansion to full number of observations. (That's how I tested freq_weights in stats). This makes it easier to verify all the extra results that we have.

e.g. df_resid will need adjustment and it will be used in the scale/Pearson estimates. It will also be used for F/Wald and t-tests when using t and f distribution

thequackdaddy:
Thanks for your review. I'll get cracking on the unit tests. 

I'm a little confused with your comment "calculated with weights and will full number of observations" are you suggesting that I build a dataset and assign each record a frequency of "n" (with different n's for each record), then create a dataset with n-many records for each observation in the original dataset and every record now with weight 1?

I'm not opposed but I seem to recall this might not result in the same deviance for every record. However, the parameter estimated should be the same.

josef-pkt:
> are you suggesting that I build a dataset and assign each record a frequency of "n" (with different n's for each record), then create a dataset with n-many records for each observation in the original dataset and every record now with weight 1

Yes, I used np.repeat for this here
https://github.com/statsmodels/statsmodels/blob/master/statsmodels/stats/weightstats.py#L507
https://github.com/statsmodels/statsmodels/blob/master/statsmodels/stats/tests/test_weightstats.py#L277

Observation/record specific statistics like residuals will not be the same (without extra weighting), but the aggregate statistics, deviance, llf, ..., and the parameter estimates, standard errors, ..., should be the same in the weighted and the "repeated" version.

There are a few `xxx_obs` methods that will need to be adjusted somewhere, or need to be checked and tested. e.g. score_factor already has data_weights (I didn't remember). 
https://github.com/statsmodels/statsmodels/blob/master/statsmodels/genmod/generalized_linear_model.py#L358
Score and Hessian are used in `_fit_gradient`

Those things will be easier to find when we have a unit test that compares two equivalent models. The `repeat` version with weights=1 is already verified. so we don't need to verify all our extra results to another package.
(Test driven development where we iterate between writing unit tests and fixing code makes this more reliable and less painful.)

As an example
https://github.com/statsmodels/statsmodels/pull/2774/files#diff-7724ec719d725a7d4b88ce01e355edc2R27
where I compared OLS with OLS that has already some (fixed) effects absorbed or partialled out.
The list of attributes was created based on `dir(results_instance)` and checking the list.
`res1` and `res2` are the two results instances that are supposed to be the same except where there is a inherent, expected difference in definitions/interpretation.

statsmodels/genmod/families/family.py
```diff
@@ -374,7 +374,9 @@ def loglike(self, endog, mu, scale=1.):
         llf = scale * sum(-mu + endog*log(mu) - gammaln(endog+1))
         where gammaln is the log gamma function
         """
-        return scale * np.sum(-mu + endog*np.log(mu)-special.gammaln(endog+1))
+        loglike = np.sum(np.log(data_weights) - mu + endog * np.log(mu) -
+                         special.gammaln(endog+1))
```
josef-pkt: When you change lines like this, you could add spaces around all operator except power `**`  `endog + 1`

josef-pkt:
fit_irls also has `data_weights` already in master (implemented for Binomila). So there might not be many changes left.

thequackdaddy:
Thanks. I think I figured out the pickle thing. Needed to make sure that `freq_weights` (I also renamed `data_weights` to `freq_weights`) is removed before the pickling. I also redid the algebra to add spaces more consistently in the family module (when there is sufficient space). 

My TODO is really to write some tests.

thequackdaddy:
Hello,

Well as I anticipated, I've run into some roadblocks that I need some better minds for.

First, I realized that my math making the log-likelihood functions was completely wrong. So, AIC/Loglikelihood tests and all deviants aren't working in what I've committed. I know that. 

Next, and most critically, I'm not able to match what R is doing. In fact, R doesn't match itself in some cases on a duplicated dataset vs. the weighted dataset. I was thinking we weren't expecting that. Right now, I'm only really looking at Poisson, Gaussian, and Gamma with their canonical links. 

#Poisson 

Looks great. Matching R, and duplicated data matching weighted data for the following: Parameters, parameter covariances, loglikihood, deviance, and AIC. 

#Gaussian

Things fall apart here. First off, R doesn't match itself when you use duplicated data. Try this code, for example. 

    library(MASS)
    data(Insurance)

    Insurance$Group <- factor(Insurance$Group, ordered = FALSE)
    Insurance$Age <- factor(Insurance$Age, ordered = FALSE)
    Insurance$Group <- relevel(Insurance$Group, "1-1.5l")
    Insurance$Age <- relevel(Insurance$Age, "25-29")

    # Test Gaussian

    model <- glm('Claims ~ 1 + C(District)  + Group + Age', 
                 data=Insurance, 
                 family=gaussian(), 
                 weights = unlist(Insurance['Holders']))
    summary(model)
    logLik(model)

    InsuranceBig<-Insurance[unlist(lapply(seq_along(Insurance$Holders), function(x) rep(x, Insurance$Holders[x]))),]
    modelbig <- glm('Claims ~ 1 + C(District)  + Group + Age', 
                 data=InsuranceBig, 
                 family=gaussian())
    summary(modelbig)
    logLik(modelbig)# Test Gaussian

You can see R gives the same parameter estimates and deviance, but much different covariance, scale, and loglikelihood. My weighted version is unable to match pretty much anything here--but I don't really know what I should be shooting for. (Note, that I realized I can get the same covariances from R if I alter the code to use `cov_params`. However, the IRLS currently uses `normalized_cov_params`. I'm not fully sure what the difference between the 2 are. https://github.com/statsmodels/statsmodels/blob/master/statsmodels/genmod/generalized_linear_model.py#L885

#Gamma

Haven't really tested this at all. But I wanted to point out that R has the same loglikelihoods and deviance, but differing scale and covariance parameters between the weighted table and the duplicated table. 

So if a stronger mind wants to think about this and dream up the correct tests, I would appreciate it.

josef-pkt:
The df.resid in R is wrong for the interpretation as caseweights. Running your model and looking at the summaries called summ and summbig:

```
> summbig$cov.scaled[1:5]
[1]  1.2419349 -0.2208420 -0.2000068 -0.1999720 -0.2234051
> summ$cov.scaled[1:5]
[1] 536.99885 -95.48961 -86.48071 -86.46568 -96.59787
> summ$cov.scaled[1:5] / summbig$cov.scaled[1:5]
[1] 432.3889 432.3889 432.3889 432.3889 432.3889
> modelbig$df.residual / model$df.residual
[1] 432.3889
```

So this is either buggy in R, or they use a different definition for the weights.

Poisson and Binomial always have scale=1 (except quasi mle case).
So those two should match independent of df_resid.
scale in normal and gamma or dispersion in quasi mle  depends on the correct df_resid

The target for us is to get the same results in repeat and in freq_weight form, because those are supposed to be the same with the interpretation as frequency weights.

Are you getting the same params in both cases, identical to R?
deviance should also be the same.

If Poisson match, and Gaussian doesn't, then the problem needs to be in the scale calculation, either it doesn't use the weights, or it doesn't use the correct df_resid.

You could push your changes and we squash them before merging. (git is good for cleaning up an experimental PR.)

josef-pkt:
@thequackdaddy   To makesure: Do you need frequency weights for your application? i.e. each weight represents a number of observations with identical exog and endog.

e.g. about different kinds of weights:
https://groups.google.com/d/msg/pystatsmodels/0Bu1aAKE3rM/chnCuQxrCQAJ
(frequency weights are the most important because the other ones could be obtained by weight normalization in most use cases.)

josef-pkt:
(
aside to the application with insurance data https://stat.ethz.ch/R-manual/R-devel/library/MASS/html/Insurance.html

AFAIU, claims is the aggregate number of claims for all `Holders`. In this case claims is already scaled up to be relative to the total number of holders for that unique exog. In this case only the prediction needs to be scaled up, which is available through offset and exposure (in log link models). 
Frequency or the data weights scale up both endog and exog, i.e. the loglikelihood contribution for each line/observation.

It can still serve as a test case, but AFAICS, the interpretation as frequency weights is not appropriate substantively.
)

josef-pkt:
from R's glm documentation
"Non-NULL weights can be used to indicate that different observations have different dispersions (with the values in weights being inversely proportional to the dispersions); or equivalently, when the elements of weights are positive integers w_i, that each response y_i is the mean of w_i unit-weight observations"

so these are weights for the variance and not frequency weights.

If I see this correctly, then we could get this through frequency weights if we multiply the weights by nobs / weights.sum()  so that total weights is nobs. Stata does this for most weights, even if we don't do it the same way in WLS.
(I'm never sure whether the weights are supposed to be squared or not, i.e. weights = 1 / var or weights = 1 / std.)

So we can still implement frequency weights internally and then adjust in some initialization to get the right interpretation.

thequackdaddy:
I've committed my changes. Have a look. 

To be clear, what I want to model is the number of insurance claims per year. (The dataset I've given you is total number of claims... I know... I was too lazy to divide). The datasets are generally 1 record per insurance record and a weight of the proportion of a year the policy was in effect. (Policies can frequently only have a portion of year as weight because of mid-term endorsements and cancellations.) So this is a frequency calculation (number of claims per year is the endog) and the proportion of a year is the weight. 

I had written a long dialogue about this, but then you posted this info about R. Let me read through it again and get back to you. At the least, I'd like to continue working on this out of curiosity. 

I'm going to be away for a few days soon. So feel free to edit this or write tests. I started writing tests with nose, but then gave up after everything I wrote failed. 

At the very least, I appreciate your help with this. Its made me understand software development much better.

josef-pkt:
Write whatever helps to understand the case.

We need to have frequency weights for many usecases and to reuse them for other weights. Large parts of other weights and interpretation is still unclear to me, they should all be the same for the parameter estimate, I think. However, differ in the interpretation for inference, as you have seen with R. Given that there is also the possibility of offset and exposure and the possibility to manipulate the variance function, it's not clear to me what different options are appropriate for different use cases.

You can also commit failing tests, or put them in a script or a notebook. The latter is what I often do when I'm still in the experimental stage for new methods. For frequency weights, we could just work through the failures and get the details to work correctly.

thequackdaddy:
Josef, 

Oh my. I've been overthinking and hadn't fully grasped what you are writing. 

I hadn't grasped the difference between R's weight and SAS's weight until just now. Take this...

https://support.sas.com/documentation/cdl/en/statug/63033/HTML/default/viewer.htm#statug_genmod_sect030.htm
Now that I think about it, that isn't a frequency weight at all, but rather the same thing as R's variance adjustment weight.... its in the denominator of the variance function! 

I feel dumb now. 

I think I shouldn't be calling the weights in `family` `freq_weight` but rather something else. 

I think I'm seeing the light... Though that could be the beer talking.

josef-pkt:
Count models like Poisson and Binomial have an inherent additive structure to model the "sample size" of an observation either through total number of trials in Binomial or exposure in Poisson. 
The normal distribution is similar for the sum. 
In all those cases the sum of identically and independently distribute random variables is in the same distribution with the variance equal to n times the individual variance.

If the observation is an average, then the variance is inverse proportional to n, the group size.

So, the scaling (weight) and exposure depends on the units of endog.

If we only have discrete explanatory variables, then we can just combine records that are identical and assign frequency weights, In additive distributions like Binomial and Poisson, those things are identical, except that we have to watch out how we count the number of observations, i.e. rescale weights, nobs and df_resid in a way that is appropriate to the data.
(Additionally, for distributions in the linear exponential family which is currently implemented in GLM, the scale factor or weights are just multiplicative in the estimating equations, so for the parameter estimates everything is the same.)

IIRC, the SPSS documentation starts out with two weights and just multiplies them together.

Related aside: WLS regression model uses weights in the inverse variance interpretation. In terms of unit tests GLM Gaussian with variance weights should be the same as WLS, except for possible small sample corrections.

thequackdaddy:
I went ahead and wrote what I think some appropriate tests would be. Could you look them over to make sure they make sense and you agree? 

Right now, I'm only testing the frequency weighted small dataset vs. the duplicated one. There's lots of fails, as I expect. If these tests make sense, I'll see if I can get these to equal. 

Thanks.

josef-pkt:
Tests look good. I prefer now assert_allclose to assert_almost_equal because it allows better control over the tolerance with atol and rtol which is an advantage if the values vary widely.
The older code in statsmodels doesn't use assert_allclose because the code was written before that assert was added to numpy.

A tip: I often rename test temporarily so that nose doesn't run them, e.g.
TestWtdGlmBinomial - > T_estWtdGlmBinomial
nose checks for `test` as part of the name.

Another "how I work":
If there are so many failures, I find it easier and faster to write an example and some asserts in a script, because then it's easier to look at other attributes that might indicate where the problem is.
I usually switch to `nosetests --pdb --pdb-failures <path_to_module or to test class>` on the command line when I just have a few failures or errors left.

josef-pkt:
some comments on R in different use cases, and the double meaning of weights in Binomial
https://stat.ethz.ch/pipermail/r-help/2010-April/235528.html   [R] Weights in binomial glm
http://stats.stackexchange.com/questions/141412/r-glm-function-with-family-binomial-and-weight-specification
http://stats.stackexchange.com/questions/121763/r-large-glm-with-sample-weights

thequackdaddy:
Ok statistics question. 

I'm struggling a little with calculating the scale parameter but I think I have a solution. I want to make sure its appropriate. 

The default scale parameter is the Chi-Squared Statistic (Pearson) divided by the degrees of freedom of the residual. In python:

    ((np.power(resid, 2) / self.family.variance(mu)).sum() / self.df_resid)

I think the solution is really to take a weighted average of the scale. In [The Practioner's Gude] (an actuarial reference) (https://www.casact.org/pubs/dpp/dpp04/04dpp1.pdf) page 19 defines the scale parameter as the following:

    1 / (n - p) * \sigma (\omega_i * (Y_i - \mu_i)) / V(\mu_i)

I'm thinking of implementing that as... 

    1 / (self.freq_weights.sum() - self.df_model - 1) * (self.freq_weights * np.power(resid, 2) / self.family.variance(mu)).sum()

So when `freq_weights` is less than the number of parameters, this model will likely not work correctly. 

Thoughts?

josef-pkt:
looks fine, AFAICS.

In WLS we use `wresid` for the weighted residuals which I think  `\omega_i * (Y_i - \mu_i)` in the reference.

You could start in your way, and later we can check which intermediate results we want to store for reuse.  for example I thought of using `wnobs` attribute for freq_weights.sum().

thequackdaddy:
Ok, we're begging to cook with gas here here. Poisson, Gamma, Gaussian, and Inverse Gaussian are looking great. Able to match params, covariance, loglikelihoods. Looks really promising. 

But we have a few problems. 

* bic Bayesian information criterion - I'm not sure how to adjust this formula for weights just yet. Haven't thought about it much. 

* t/p values - In a few of the families, (InverseGaussian comes to mind), the t-values match fine. However, the problem tries to calculate p_valus and craps out here. If you look at the test, you get a bunch of zeroes and then a few non-zeroes. If I switched to absolute tolerance, this would work. Haven't done that yet... I recall that some families t-tests are more important while others, p tests are. So I'm thinking there's a reason for this behavior. It might actually be right. 
* Gaussian loglike - This is broken too. There's some code in it that I don't fully understand. I've pasted below so if someone can explain, I'd appreciate it. The real question is, why must there be a special the identity link function? 

    if isinstance(self.link, L.Power) and self.link.power == 1:
        # This is just the loglikelihood for classical OLS
        nobs2 = np.sum(freq_weights, axis=0) / 2.
        SSR = np.sum((endog-self.fitted(mu))**2, axis=0)
        llf = -np.log(SSR) * nobs2
        llf -= (1+np.log(np.pi/nobs2))*nobs2
        return llf
    else:
        return -0.5 * np.sum(freq_weights * ((endog - mu)**2 / scale +
                             np.log(scale / freq_weights) +
                             np.log(2 * np.pi)))

* Binomial - I really tried to make this work before, but it is now completely broken. I think I need to read some of Josef's comments on it. 
* Negative Binomial - Same as binomial. I broke it when I was following the SAS documentation. I think I need to just re-approach it. 

So tons of progress today! 

I will be away for the next few days. So, don't expect much progress from me then. 

Thanks.

kshedden:
I don't remember the whole story behind this, but there are two versions of
the likelihood for the Gaussian GLM.  One is with the scale parameter
profiled out (a.k.a. "concentrated log-likelihood"), the other is the more
traditional likelihood that is comparable to the likelihoods for other GLM
families (where the scale is not profiled out).  This has given me some
trouble in the past.

josef-pkt:
sounds great.

I'll try to go over the details in the next few days. Comments for what I suspect:

- for gaussian loglike we have a correction for this in another PR. The problem is that it uses the concentrated likelihood, which will be changed. in the current version the `SSR` would have to use the weights or weighted residuals.

- for binomial: I guess we run into the problem with two different interfaces for the same thing: If we use freq_weights, then the endog should be scaled for an individual observation (to get proportion or mean). If we use the current pattern, then the data is given by the total number of successes and failures and implied total number of trials. I guess. (Related confusion also in the binomial questions for R on stackoverflow)

- t/p values: no idea yet. Those should be correct by inheritance if they are correct for the other distributions. There will, most likely, still be some results for specific options that need to be adjusted to support the freq_weights in all parts or for all cases. One difference in the results comes from whether a canonical link is used or not. inverse gaussian is still fragile in the optimization in some cases.

statsmodels/genmod/families/family.py
```diff
@@ -703,36 +721,31 @@ def resid_anscombe(self, endog, mu):
...
     """
     Binomial exponential family distribution.
-
     Parameters
     ----------
     link : a link instance, optional
         The default link for the Binomial family is the logit link.
         Available links are logit, probit, cauchy, log, and cloglog.
         See statsmodels.family.links for more information.
-
```
josef-pkt: There are several empty lines that got removed here. This will break the doc build and needs to be fixed again.

statsmodels/genmod/families/family.py
```diff
@@ -798,42 +806,35 @@ def deviance(self, endog, mu, scale=1.):
...
         --------
         deviance : float
             The deviance function as defined below
-
         Notes
         -----
         If the endogenous variable is binary:
-
         `deviance` = -2*sum(I_one * log(mu) + (I_zero)*log(1-mu))
-
```
josef-pkt: don't remove empty lines. 
Do you have an automatic reformatting setting in the editor?

thequackdaddy: Oh shoot. Not sure why that happened. I presume I can just copy-paste it back from the original. The math was the hard part. Writing code is easy.

Is that a setting in spyder? That's what I've historically used. It's buggy at times but is really easy to use.

statsmodels/genmod/generalized_linear_model.py
```diff
@@ -574,10 +584,13 @@ def estimate_scale(self, mu):
...
-                return ((np.power(resid, 2) / self.family.variance(mu)).sum()
-                        / self.df_resid)
+                return ((self.freq_weights * (np.power(resid, 2) /
+                         self.family.variance(mu))).sum() /
+                        (self.freq_weights.sum() - self.df_model - 1))
             elif self.scaletype.lower() == 'dev':
-                return self.family.deviance(self.endog, mu)/self.df_resid
+                return (self.family.deviance(self.endog, mu,
+                                             self.freq_weights) /
+                        self.freq_weights.sum() - self.df_model - 1)
```
josef-pkt: it looks like a missing opening parenthesis in `(self.freq_weights.sum() - self.df_model - 1)`

we can change df_resid to `self.freq_weights.sum() - self.df_model - 1` then these cases with df_resid don't have to be changed, and it's a single point for verifying the definition

statsmodels/genmod/generalized_linear_model.py
```diff
@@ -749,21 +762,21 @@ def fit(self, start_params=None, maxiter=100, method='IRLS', tol=1e-8,
...
+        if isinstance(self.family, families.Binomial):
+            tmp = self.family.initialize(self.endog)
+            self.endog = tmp[0]
+            self.n = tmp[1]
+        if self.freq_weights is None:
+            self.freq_weights = np.ones((self.endog.shape[0]))
+        if np.shape(self.freq_weights) == () and self.freq_weights > 1:
+            self.freq_weights = (self.freq_weights *
+                                 np.ones((self.endog.shape[0])))
+        self.freq_weights = self.n * self.freq_weights
```
josef-pkt: Given that self.n is already the correct length array, this works also if freq_weights is scalar. So previous if looks redundant. Either n or freq_weights can be initialized to scalar 1.

josef-pkt: Binomial handling here looks correct to me.

statsmodels/genmod/generalized_linear_model.py
```diff
@@ -1051,7 +1064,8 @@ def __init__(self, model, params, normalized_cov_params, scale,
         self._endog = model.endog
         self.nobs = model.endog.shape[0]
         self.mu = model.mu
-        self._data_weights = model.data_weights
+        # Divide by n for binom
+        self._freq_weights = model.freq_weights / model.n
```
josef-pkt: This looks fragile/dangerous to me. We are mixing up the two meanings of weights in binomial.
My guess would be that this shouldn't be divided by `model.n`

statsmodels/genmod/generalized_linear_model.py
```diff
@@ -1084,31 +1098,33 @@ def __init__(self, model, params, normalized_cov_params, scale,
...
     @cache_readonly
     def resid_working(self):
         val = (self.resid_response / self.family.link.deriv(self.mu))
-        val *= self._data_weights
+        val *= self._freq_weights
         return val
 
     @cache_readonly
     def resid_anscombe(self):
+        # TODO: Data weights?
```
josef-pkt: This depends a bit on how we **define** the residuals, per unit/observation or with weights
In linear model WLS, we distinguish between `resid` and `wresid`

statsmodels/genmod/tests/test_glm.py
```diff
@@ -1021,6 +1022,200 @@ def test_gradient_irls():
...
+        res2_fitted = self.res2.predict(self.res1.model.exog)
+        assert_allclose(self.res1.fittedvalues, res2_fitted, 1e-6)
+
+    decimal_tpvalues = DECIMAL_4
+
+    def test_tpvalues(self):
+        # test comparing tvalues and pvalues with normal implementation
+        # make sure they use normal distribution (inherited in results class)
+        assert_allclose(self.res1.tvalues, self.res2.tvalues, 1e-6)
+        assert_allclose(self.res1.pvalues, self.res2.pvalues, 1e-6)
```
josef-pkt: This needs an atol because some values are very small. (problem with using only relative tolerance if the value is essentially zero)
And in general it's better to use the keyword `rtol` instead of relying on positional interpretation

`assert_allclose(self.res1.pvalues, self.res2.pvalues, rtol=1e-6, atol=1e-20)`

i.e. for example atol=1e-20: if the pvalue is smaller than 1e-20, then we don't care how close to zero.

josef-pkt:
About the problem with Gaussian: Switching to log link would avoid the code path for the concentrated loglikelihood, which needs to be fixed separately from this.

workflow: We usually avoid merging master into a feature branch. When necessary or if needed before merging, then we rebase on master.

thequackdaddy:
OK I'm back. 

I've re-approached the binomial problem and I think I have that working. I now am passing all tests. (There's a few fails, but I think those are issues are beyond me.)

I haven't thought about using resid/wresid. Do you want me to get that going? I'm not as familiar with WLS class--just haven't looked at it obviously. 

I also would like to check the documentation piece. I haven't done anything with that and am not familiar with Sphinx. How do I check that the documentation is building correctly? 

And is there anything else I need to do? 

Thanks for your help. This has been very educational for me.

josef-pkt:
Looks very good. I need to look at the details to see what's left to do.

One extra test, now that the basics are running well:
Switch to a robust covariance with one of tests. For example copy `TestWtdGlmPoisson` and add cov_type as argument to the fit calls:

` ....fit(cov_type='HC0')`

There is some chance that it fails, then I might have to dig into the sandwich code. If we don't fix it in this PR, then we need to add to the doc string that it doesn't work yet.

docbuild currently only works with python 2, but fix is almost finished (the doc build on python 3 needs an updated numpydoc)
`make html` in the docs directory is running the sphinx docs generation. The windows version doesn't try to build the examples. The Linux version is building the examples, AFAIR. Notebook example handling also only works on python 2 until the PR to update this is merged.

git question: Do you know how to squash commits with interactive rebase?
I think we should squash some of the earlier commits that are without informative commit messages, but I can do this also if needed.

josef-pkt:
The advantage of having unit tests is that we don't need to verify the math everywhere by hand, if the tests pass.
The changes to Binomial with keeping the separate `_n` around look correct to me, but I don't remember the details of what I was thinking about this a few days ago.

josef-pkt:
The main docstring to which freq_weights has to be added is here after family before extra_params
https://github.com/thequackdaddy/statsmodels/blob/glm_weights/statsmodels/genmod/generalized_linear_model.py#L69
docstrings for `__init__` of a class are in the class docstring itself. We keep the `__init__` docstring empty because it doesn't get included in the documentation.

I guess the freq_weights parameter needs to be copied to every method that has it now as argument.
The main class docstring for freq_weights should be a bit more informative, mentioning that the number of observations is taken to be the sum of freq_weights. Other methods just need to mention that these are frequency weights.

thequackdaddy:
I will look into squashing commits. Not familiar with git other than what some "git for dummies" type posts I've seen.

josef-pkt:
One more test: The default method is IRLS. For some cases using a standard optimizer is better, and some things might only be supported for scipy optimizers.

you can copy a test and add a method argument `....fit(method='newton')`
or if that doesn't work on the example use `method='bfgs'`

josef-pkt:
about git: if you need to try out something, then it's better to create a temporary branch that is a copy of the working branch. 
`git -i rebase` starts interactive rebase and pops up an editor with a list of commits that can be squashed or deleted.

The merge commit 286446f3f1c40174c39ae536152b65065d28335d should be deleted.

But if you don't want to fight with (figure out) git right now, then I can do it.

thequackdaddy:
Ok git squishing: Ok. How about you do it this time and I'll figure it out in the future. Not really sure what finished product looks like with this.

thequackdaddy:
Well I changed my mind on git and decided to try it. I think I got it mostly fixed, but I didn't have the finished product just right. I `git rebase -i` and squashed all the commits into one and reverted `286446f`. However, its not doing what I thought. 

Reading through git's documentation (which is excellent, but I'm just new to it) I think I might need to do a `git rebase master` on my machine to get it all synced up. Right now, it says there's a conflict (I think having to do with reverting `286446f`. 

So before I break anything else, can someone kindly show me the light here? I didn't realize all the power of git until now...

josef-pkt:
Don't experiment on this branch. We need it, and cleaning up experiment might require more "git magic".

"git is very powerful - it offers many opportunities to shoot your feet" 
It's also powerful enough to recover from every mistake, using more power.

I'll look at it a bit later today.

thequackdaddy:
Sounds fine. Sorry if I broke anything!

Good news is that there's only 4 files that have been changed. I'll let you sort it out and not risk my ignorance.

I will await your further orders before making any more changes or commits.

josef-pkt:
Based on the network graph https://github.com/statsmodels/statsmodels/network:
Your `first attempt` looks like it has squashed all commits correctly.
But then you merged the full unsquashed branch back into it, I guess from the remote tracking branch.

Instead of the last merge, you could have just force pushed on this remote branch. (Or, to be safe, push it to a new branch and open a new PR.)

Aside: It looks like, based on the commit message, you reverted the previous commit that merged into master. That's not needed because in interactive rebase we can skip commits by just deleting the from the rebase pop up text file.

Also by squashing the PR I meant squashing mostly the first few commits. In general I prefer for statsmodels to keep most of the commits separate because most of the time they provide additional information about the changes, discussion and design.

Since this branch still has the original commits, it just takes another interactive rebase to remove the last two commits, I think. I will try it out later today. (Right now I have too many checkouts with open work in progress branches.)

Aside: I don't remember some of the git commands (syntax and details) because for much of the standard work I just use git gui.

thequackdaddy:
Sounds good. There's only 4 files that have changed, so making a new PR with these changes is easy enough for me. 

I'll let you try to handle. 

Again, my apologies.

josef-pkt:
apropos git: see my recent comment here https://github.com/statsmodels/statsmodels/pull/2749#issuecomment-172860111
I still work with temporary branches and push to new PRs whenever I need to do something with git that I'm not used to.

josef-pkt:
I rebased locally and trying things out now.

I just realized that this is using rdatasets in the unit tests.
We assume that we do not have internet access during the test runs, for example Ubuntu build and testing does not allow it.

Running nosetest on a test case takes very long, no idea why. The time it prints is only 0.009s for TestWtdGlmPoisson.
There is a large amount of duplicated repeated work in the tests because they all start with (down)loading the dataset.

josef-pkt:
An observation on convergence, see also #2803: 
The "repeated" model is hitting maxiter 100. The params converge relatively fast but the deviance still fluctuates at around the 10th significant digit, which is the 5th decimal in this case.
The freq_weights version converges in 9 iterations.

I don't know where the differences come from, the optimization should be the same, except for floating point issues. The starting values, first params in history, differ a bit. First deviance differs by a large amount, but "repeated" is smaller.

I'm looking at the HC0 case, but that does not differ in the optimization part.
The bse for `cov_type='HC0'` fails, does not agree between the two versions.

josef-pkt:
`fit(method='newton')` seems to work well in the Poisson test case, the tolerance for bse and tpvalues needs to be increased a bit.  mle_retvals shows only 1 iterations, which sounds a bit low.

josef-pkt:
some timing on the Poisson example, on my computer:

loading rdataset is 25s 
loading from cache about 1.2 to 2 seconds
fit with weights 0.02s
fit big with irls: 1.3s
fit big with newton: 14s
fit big with bfgs: 7.5s

big exog shape is (23359, 10)
fweight exog shape is (64, 10)

thequackdaddy:
I was thinking that the `_fit_gradient` was slower in general, but haven't done any formal test.

A question... there is a variable `max_start_irls` that is supposed to limit the number of IRLS iterations used to get initial parameters for `_fit_gradient`. However, the only test I see is that `max_start_irls` is checked to be greater than zero, then it runs the irls. [But I don't see how that variable limits the number of IRLS iterations:](https://github.com/statsmodels/statsmodels/blob/master/statsmodels/genmod/generalized_linear_model.py#L802)

I don't fully understand how the optimizer is supposed to work, so perhaps there's a deeper bug in there I don't understand.

josef-pkt:
looks like a good catch.
I think maxiter in the call to _fit_irls is supposed to be max_start_irls.
Kerby did some experiments when he added _fit_gradient and in many cases the gradient optimizers were faster, with good starting values, He tried out differen max_start_irls and a small number like 3 looked like a good choice.

That could then be a reason that newton shows number of iterations equal to 1 if it's already optimized, but then I don't know where the time is spent with `newton`

josef-pkt:
to summ up a bit what I looked at

- squashing all commits into one looks good because there are still later commits that revert previous changes
- unit test need a refactor 
  - no internet, add the data to the repo
  - avoid some of the repeated calculations, e.g. load and create data once.
  - I think we can shrink the replication, I went the opposite way to check timing and multiplied the "Holders" by two, instead we could divide by something like 10
- fit_gradient passes the tests (adding `method='newton'`) but is slow at full or double repeat
- sandwiches `cov_type='HC0'` failed the unit tests, but I haven't looked at it yet.
- some streamlining in the code would be useful, store `freq_weights.sum()`. Maybe avoid the full weight=ones(...) if there are no weights. That might help in the optimization loop.
   (I'm not sure if numpy optimizes on x * 1 as no-op.) 
   (I was browsing julia GLM issues, and they also had the debate whether to special case the no weights case.)

I haven't tried out yet the code after fixing `_max_start_irls`.

josef-pkt:
switching to maxiter=_max_start_irls with newton leaves the time it takes unchanged at 25 seconds, but now res2.mle_retvals shows 3 iterations.

josef-pkt:
another comparison
using exposure takes about half the time of using weights, 0.01 seconds instead of 0.02 In Poisson with log-link (using the non-repeated data)
params and bse are the same

josef-pkt:
freq_weights are stored as pandas Series, it should get an asarray to have numpy arrays
I think we don't attach alias like `_freq_weights` to results instances anymore.

I'm still trying to figure out the robust covariances, and I don't understand exposure anymore.
Different weights, exposure, offset and similar are equivalent, except when they are not.

thequackdaddy:
Gee whiz, Josef. Didn't realize how much bug-checking you were going to be doing on this. I'm impressed.

I hope my PR's haven't made you lose any sleep. Godspeed. 

Let me know if you want me to handle. However, I think you might be working on a different code base and my git skills are awful. I will await your orders.

josef-pkt:
Because I'm working on top of your branch after rebase, I will push my branch to a new PR most likely later today. Then there will still be work left.

I have more work here because there are now a lot of extra things available for GLM. Some we can postpone, but I have to figure it out sooner or later anyway. freq_weights in GLM is the start to add it to other models, and this way you did all the model specific work.

thequackdaddy:
This is now part of #2835.

Issue #958, opened 2013-07-10
Add SETAR model


ChadFulton:
- See https://gist.github.com/ChadFulton/5962854 for a very first draft.
- See http://www.ssc.wisc.edu/~bhansen/progs/joes_99.html for a link to the implementation paper; Hansen (1999)
- See https://github.com/statsmodels/statsmodels/wiki/SETAR-Model for a longer set of notes / problems with the implementation. Written in a wiki to hopefully prevent cluttering up this issue.

Branch (empty so far) is at https://github.com/ChadFulton/statsmodels/tree/setar

PR #2815, opened 2016-02-14
ENH: Add self exciting threshold autoregression (SETAR) model

ChadFulton:
This is the code from three years ago, see #958 for more details.

The main issue to fix here is unit tests, which do not cover everything in the model.


ChadFulton:
Also, it uses multiple inheritance and has to modify `linear_model.py` to do it. Probably would be better not to do that.

Issue #2817, opened 2016-02-16
BUG: Logit margeff summary TypeError


TomAugspurger:
Really the only example from my notebook cleanup that looks like a real bug.

``` python
import numpy as np
import statsmodels.api as sm

spector_data = sm.datasets.spector.load()
spector_data.exog = sm.add_constant(spector_data.exog, prepend=False)

logit_mod = sm.Logit(spector_data.endog, spector_data.exog)
logit_res = logit_mod.fit(disp=0)

margeff = logit_res.get_margeff()
print(margeff.summary())
```

raises

``` python
TypeError                                 Traceback (most recent call last)
<ipython-input-1-210bf23ad669> in <module>()
      9 
     10 margeff = logit_res.get_margeff()
---> 11 print(margeff.summary())

/Users/tom.augspurger/Envs/dev/lib/python3.4/site-packages/statsmodels/statsmodels/discrete/discrete_margins.py in summary(self, alpha)
    520         _, const_idx = _get_const_index(model.exog)
    521         if const_idx is not None:
--> 522             exog_names.pop(const_idx)
    523 
    524         J = int(getattr(model, "J", 1))

TypeError: only integer scalar arrays can be converted to a scalar index
```

josef-pkt:
I don't really understand this `_get_const_index(model.exog)` just uses np.where

However, this code is outdated since const_idx is now attache to the model or model.data

Same code is in GEE (copied and adjusted from discrete)

no smoketests for `summary` method in test suite. 
discrete margins are only tested on Logit, AFAICS

(I haven't installed numpy 1.11.0.bxxx yet because I either don't remember how to set up a virtualenv, or the python built in virtualenv in 3.4 doesn't work properly, or it's because pip hates me back)

josef-pkt:
the problem is that const_idx is a 1d array with one element, which cannot be used for indexing or popping a python list.

The newer code in base data squeezes: `np.where(...)[0].squeeze()`
AFAIK, the model const_idx is used quite often and should have proper functional unit test coverage

theaverageguy:
@kshedden Fixed?

kshedden:
I was not planning to fix this in Logit, I have not worked on that code at
all.

I have a pending PR (#2833) that fixes the same bug in GEE.  There are a
few more things that I need to deal with there before it can be merged
though.

josef-pkt:
looks like we missed this one, I had thought it was incorporated somewhere

[...]

PR #2833, opened 2016-02-20
const_idx handling in GEE margins and pep8

kshedden:
This PR addresses comments about const_idx handling made in #2817.

Also, pep8 fixes to GEE code and tests.

### DIFF
--- a/statsmodels/genmod/cov_struct.py
+++ b/statsmodels/genmod/cov_struct.py
@@ -4,8 +4,7 @@ import numpy as np
 import pandas as pd
 from scipy import linalg as spl
 from collections import defaultdict
-from statsmodels.tools.sm_exceptions import (ConvergenceWarning,
-                                             IterationLimitWarning)
+from statsmodels.tools.sm_exceptions import ConvergenceWarning
 import warnings
 
 """
@@ -14,6 +13,7 @@ docs:
 
 http://www.stata.com/manuals13/xtxtgee.pdf
 """
+
 
 class CovStruct(object):
     """
@@ -42,7 +42,6 @@ class CovStruct(object):
 
         # Method for projecting the covariance matrix if it not SPD.
         self.cov_nearest_method = cov_nearest_method
-
 
     def initialize(self, model):
         """
@@ -162,9 +161,10 @@ class CovStruct(object):
         self.cov_adjust.append(cov_adjust)
 
         # Last resort if we still can't factor the covariance matrix.
-        if success == False:
-            warnings.warn("Unable to condition covariance matrix to an SPD matrix using cov_nearest",
-                          ConvergenceWarning)
+        if not success:
+            warnings.warn(
+                "Unable to condition covariance matrix to an SPD "
+                "matrix using cov_nearest", ConvergenceWarning)
             vmat = np.diag(np.diag(vmat))
             vco = spl.cho_factor(vmat)
 
@@ -193,7 +193,7 @@ class Independence(CovStruct):
         return np.eye(dim, dtype=np.float64), True
 
     def covariance_matrix_solve(self, expval, index, stdev, rhs):
-        v = stdev**2
+        v = stdev ** 2
         rslt = []
         for x in rhs:
             if x.ndim == 1:
@@ -207,7 +207,8 @@ class Independence(CovStruct):
     covariance_matrix_solve.__doc__ = CovStruct.covariance_matrix_solve.__doc__
 
     def summary(self):
-        return "Observations within a cluster are modeled as being independent."
+        return ("Observations within a cluster are modeled "
+                "as being independent.")
 
 
 class Exchangeable(CovStruct):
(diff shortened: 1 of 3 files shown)


kshedden:
Also I modified the covariance computation logic so that the bias reduced robust covariance is not calculated unless requested (it involves n_obs x n_obs calculations so is expensive for large clusters)

kshedden:
@josef-pkt ready for review

josef-pkt:
I'll go over it tonight. thank for the ping

statsmodels/genmod/cov_struct.py
```diff
@@ -583,11 +582,10 @@ def update_nogrid(self, params):
         dep_params /= dn
         self.dep_params = dep_params[1:] / dep_params[0]
 
-
     def covariance_matrix(self, endog_expval, index):
 
         if self.grid:
-            return self.covariance_matrix_grid(endog_expal, index)
+            return self.covariance_matrix_grid(endog_expval, index)
```
josef-pkt: this looks like a bug fix, which might mean insufficient test coverage.
?

kshedden: Yes, flake8 picked this up.  We don't use `covariance_matrix` internally anymore, instead we do everything through `covariance_matrix_solve`, which has decent test coverage.  But `covariance_matrix` is useful for people who want to see the actual fitted covariance matrices.  I will add some tests.

josef-pkt:
changes look fine, (I already read through them a few days ago when I wasn't logged into github)

about pep-8: did you use autoformatting? 
Most of it looks good. We still follow in general the rule not to put spaces around power `**`.
The latest pep-8 version is more agnostic about putting spaces around operators, but so far we stick to the old rules.

One observation, not affected by the changes here but I just saw it:
`_handle_constraint(self, mean_params, bcov)`  uses the naive `bcov`
Shouldn't this use whatever cov_type has been chosen as the default?

statsmodels/genmod/cov_struct.py
kshedden: I added a regression test for the place where the bug was fixed.  We don't have a generic way to run through all of the covariance structures in a test right now, so it's not easy to have simple consistency checks like this.  I will add better, more systematic testing of GEE in a future PR.

kshedden:
I now have it pulling the correct covariance matrix in `handle_constraint`. Good catch.

IIRC you now have generic code to handle constraints, so GEE-specific code should all go away in future refactoring.

statsmodels/genmod/generalized_estimating_equations.py
```diff
@@ -1122,26 +1127,42 @@ def fit(self, maxiter=60, ctol=1e-6, start_params=None,
...
             if mean_params is None:
                 warnings.warn("Unable to estimate constrained GEE "
                               "parameters.", ConvergenceWarning)
                 return None
+            if cov_type == "robust":
+                bcov = cov
+            elif cov_type == "naive":
+                ncov = cov
+            elif cov_type == "bias_reduced":
+                bc_cov = cov
```
josef-pkt: given that bcov and ncov are also attached to the results instance, I think they also need to be transformed to the unconstrained parameter space, don't they?

`bcov = self.constraint.unpack_cov(bcov)`

I'll add a general question in main comment

josef-pkt:
question to the score test:
AFAICS, it only uses `cmat` and not the actual covariance that is provided as an argument.
Do you know how to do the robust score_test?
I worked on score tests with robust covariance matrices in a PR, but I'm not sure yet how the pieces fit together. And I didn't find a reference to test against in R or in Stata.
For GLM I only have a nonrobust score test attached to the model.

(Now that I understand a lot more than when GEE was initially merged, I should work my way through this code again.)

statsmodels/genmod/generalized_estimating_equations.py
josef-pkt: @kshedden  Do you still want to look at this or do we merge as is?
(I'm trying to work my way through some of the PR backlog.)

kshedden: I will look at this again and ping you before we merge.

kshedden:
@josef-pkt I think this is ready to merge now.  I fixed the bug you caught about reparameterizing cov_params

statsmodels/genmod/tests/test_gee.py
```diff
@@ -580,6 +580,28 @@ def test_scoretest(self):
...
+        va = Independence()
+        for cov_type in "robust", "naive", "bias_reduced":
+            model = GEE(endog, exog, group, family=family,
+                        cov_struct=va, constraint=(L, R))
+            result = model.fit(cov_type=cov_type)
+            result.standard_errors(cov_type=cov_type)
+            assert_allclose(result.cov_robust.shape, np.r_[4, 4])
+            assert_allclose(result.cov_naive.shape, np.r_[4, 4])
+            if cov_type == "bias_reduced":
+                assert_allclose(result.cov_robust_bc.shape, np.r_[4, 4])
```
josef-pkt: integers can be tested with `assert_equal`  (there is no floating point noise)

josef-pkt:
I only spot checked the last commit.  Looks good to me, merging

PR #2854, opened 2016-03-12
Fix plot_components issue if endog without dates

Leoyzen:
There's no 'resid' in plot_components function, I think it will be the endog length.


ChadFulton:
Thanks for this PR! You're right, that should be the behavior.

PR #2923, opened 2016-05-03
Don't merge

kshedden:
Leland is going to integrate this into his PR for issue #2894, I just want him to be able to see the diffs and test results


kshedden:
@lbybee note that I removed a test that made no sense and was failing.  You will need to do the same.  It's generally bad practice to remove tests but this one really doesn't make sense.

PR #2972, opened 2016-05-28
PERF: Speed improvements for elastic net

kshedden:
Avoid using SVD to check for constant and exog rank in internal elastic net iterations where this is not needed.

Cuts run time in half for larger problems.

Also minor refactor in base.model to avoid needlessly calculating the rank of the same matrix twice.

See issue #2971

### DIFF
--- a/statsmodels/base/elastic_net.py
+++ b/statsmodels/base/elastic_net.py
@@ -147,7 +147,7 @@ def fit_elasticnet(model, method="coord_descent", maxiter=100, alpha=0.,
         params = start_params.copy()
 
     converged = False
-    btol = 1e-8
+    btol = 1e-4
     params_zero = np.zeros(len(params), dtype=bool)
 
     init_args = dict([(k, getattr(model, k)) for k in model._init_keys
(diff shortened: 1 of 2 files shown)


kshedden:
Based on profiling code below, after this PR is adopted 75% of the remaining time is spent in `dot` so there isn't a lot or Python overhead remaining.

There are 660 calls to `dot`, so the 10x gap relative to R elastic net is either call overhead, or calling `dot` too many times.

```
import cProfile
import pstats
import statsmodels.api as sm
import numpy as np

n = 100000
p = 30

xmat = np.random.normal(size=(n, p))
ey = xmat[:, 0:10].sum(1)
y = ey + np.random.normal(size=n)

model = sm.OLS(y, xmat)
result = model.fit_regularized(alpha=0.01)
cProfile.run("result = model.fit_regularized(alpha=0.01)", "restats")

ps = pstats.Stats('restats')
ps.strip_dirs().sort_stats(-1).print_stats()
ps.sort_stats('cumulative').print_stats(20)
```

kshedden:
Another possible performance enhancement would be to allow the loglike, grad, and hessian, or the loglike and grad to be computed jointly, since they often involve redundantly recalculating the same intermediate quantities (e.g. in OLS)

bashtage:
This could be a good test case for using Numba in statsmodels -- they have recently updated Numba to include core linear algebra routines (dot, inv, svd) so that code that uses these can be JIT compiled.

josef-pkt:
We could try numba for statsmodels 0.9.
Given that this code is generic across different models, we would need to get numba to work with the full models which I guess will not be easy or possible. 
So, I guess the best way to get started is to try is to write a simplified model that only uses instructions that numba understands, and then expand from there once we see how well it works.

kshedden:
MixedLM would be another candidate for this.

Kerby

josef-pkt:
merged, mainly removes one redundant calculation

PR #2979, opened 2016-06-01
TST: Inherit TestCase to enable runs with nosetests

bashtage:
Inherit class-based tests from TestCase to allow
statsmodels test suite to be run with nosetest.
This allow various plug-ins to be used to measure
issues with tests, such as test time.


bashtage:
This is a small set of changes to a few tests that allow nose to directly test statsmodels.  This change allows improved timing of tests using, e.g.

```
nosetests --with-congestion statsmodels
```

josef-pkt:
Do nose plugins really need that we inherit from TestCase?

I can merge this as an experiment. But, I'd don't like starting to inherit from TestCase everywhere, that would take most of the fun out of using nose.

(Given the uncertain future of `nose`, we might have to eventually think about alternatives but for now it's working fine.)

bashtage:
They do not.  The issue here was the these classes implemented some
`unittest.TestCase`-like features (setUp) which confused nose.  The
simplest fix was to just make these proper TestCase subclasses.

josef-pkt:
setup_class  is a standard nose feature that we use in many places. There might have been just some incorrect usage of it. I don't see what's wrong with quantile regression test, but the kernel tests in the old code are missing the classmethod decorators.

josef-pkt:
Ok, then I think the solution that is more consistent with the rest of statsmodels, is to fix the setup class methods to be nose compatible, instead of inheriting from TestCase.

josef-pkt:
I'm still puzzled that the setUp methods are even called. I never used those.
From what I can figure out that the `setUp` method in the kernel tests are setup for each method call not for the class. So nose still runs them, but for each test_method separately.

@bashtage  Is this correct?

bashtage:
Yes, `setup` is run before each test in a class, while `setup_class` is a class method that is run once.

I restored the use of class methods here in the latest change.

josef-pkt:
to clarify again: Now inheriting from `TestCase` is essentially redundant and doesn't change how nose treats the tests?

merging

josef-pkt:
(closed by accident, merged)

Thanks @bashtage

bashtage:
It was just to ensure nose could find some methods (teardown) that it wanted - even though these are just `pass`.

PR #2992, opened 2016-06-07
ENH: Add freq_weights to Kaplan Meier estimation

kshedden:
This is a relatively simple extension to support frequency weights in Kaplan Meier estimation of survival curves.

Some relevant background:

http://www.inside-r.org/r-doc/survival/survfit.formula

http://www.ssc.ca/survey/documents/SSC2003_J_Lawless.pdf

### DIFF
--- a/statsmodels/duration/tests/test_survfunc.py
+++ b/statsmodels/duration/tests/test_survfunc.py
@@ -238,3 +238,50 @@ def test_plot_km():
 
     if pdf_output:
         pdf.close()
+
+
+def test_weights1():
+    """
+    tm = c(1, 3, 5, 6, 7, 8, 8, 9, 3, 4, 1, 3, 2)
+    st = c(1, 1, 0, 1, 1, 0, 0, 1, 0, 0, 1, 1, 0)
+    wt = c(1, 2, 3, 2, 3, 1, 2, 1, 1, 2, 2, 3, 1)
+    library(survival)
+    sf = survfit(Surv(tm, st) ~ 1, weights=wt, err='tsiatis')
+    """
+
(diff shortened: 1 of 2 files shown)


statsmodels/duration/tests/test_survfunc.py
```diff
@@ -238,3 +238,50 @@ def test_plot_km():
 
     if pdf_output:
         pdf.close()
+
+
+def test_weights1():
+    """
```
josef-pkt: No docstring for test function, use `#` for code comment instead of triple quotes.
(nosetests uses docstring if available as "title")

josef-pkt:
I haven't looked at the actual calculation, so mainly a question about other results.

`n_risk` needs unit tests, I didn't see any other nobs related attribute, like degrees of freedom that would be affected by freq_weights.

Does the postestimation code, or things like the hypothesis tests work with freq_weights, will it be added if needed, or do we add a warning for now to what things are not yet implemented for freq_weights?

(related aside: GLM with freq_weights has full unit tests for the standard results, but still some problems with some robust standard errors like incorrect small sample correction and I haven't check GLM.score_test yet, and is possibly or likely still wrong if freq_weights are used.)

kshedden:
@josef-pkt I updated according to your comments.  The standard errors account for the weights, and the other inference procedures should be OK as they only depend on the standard errors being correct.

Log rank testing does not support frequency weights, but that is in a different function without a `freq_weights` argument, so no confusion there I think.

josef-pkt:
looks fine to me. merge?

kshedden:
Yes, great.

Kerby

josef-pkt:
merged, thanks @kshedden

PR #3003, opened 2016-06-12
DOC: Verify doctests

bashtage:
Verify doc tests execute.  Final result is not verified due to
precision issues across platforms/BLAS implementation

### DIFF
--- a/statsmodels/base/model.py
+++ b/statsmodels/base/model.py
@@ -1159,8 +1159,12 @@ class LikelihoodModelResults(Results):
 
         >>> T_test = results.t_test(r)
         >>> print(T_test)
-        <T contrast: effect=-1829.2025687192481, sd=455.39079425193762,
-        t=-4.0167754636411717, p=0.0015163772380899498, df_denom=9>
+                                     Test for Constraints
+        ==============================================================================
+                         coef    std err          t      P>|t|      [0.025      0.975]
+        ------------------------------------------------------------------------------
+        c0         -1829.2026    455.391     -4.017      0.003   -2859.368    -799.037
+        ==============================================================================
         >>> T_test.effect
         -1829.2025687192481
         >>> T_test.sd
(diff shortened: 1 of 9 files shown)


bashtage:
Does some initial clean up by runnign doc tests to see if examples actually work.  Biggest issue is missing data.

In particular `datasets.bc` is referenced but not available.  Another examples uses formulas and `Days` and `Weight` as variables but this doesn't match any dataset. 

GEE doesn't load data in the examples and so it isn't possible to tell if these work correctly.

josef-pkt:
The problem is that requiring doctests to pass makes examples really ugly, so we never decided to do that (nor numpy and scipy if my information is still up to date).

I haven't tried running doctest over our docstrings in several years. The trend is that we put more examples into notebooks which are easier to run.

bashtage:
The issue is that lots of doctests are simply wrong - not just not working,
because things have moved.  Wrong doctests end up in docs, which makes docs
wrong.

So far having working doctests doesn't really make things ugly.  Maybe a
line or two more.

bashtage:
One can always say "See Example Notebooks" rather than putting in an example that may become broken.

josef-pkt:
Related: 
Debian is checking whether building the documentation is reproducable, and statsmodels and many other python packages fail in this. Some of the problems are datetimes in `summary`, floating point noise, indeterministic order in python dictionaries, cases where an `id` of a variable is printed (eg. `repr`) and so on.

bashtage:
I agree that it isn't reproducible.  One other obvious reason is that
different BLAS implementation will produce different numbers.

I think this is OK to not be reproducible.  This said, it is easy enough to
check for errors that are not just mismatches in output.

Really need a flag to tell doctest to not care about output so that it can
be run in a smoke test mode.

josef-pkt:
Does doctest show errors separate or distinguishable from failure so we could use doctest for smoke testing?  I never thought of restricting it to this.

Changes look good (with minor formatting quibbles that can be ignored), and as far as I can tell this shouldn't cause (m)any merge conflicts with open PRs.

If your are finished with this, then I'll merge

josef-pkt:
merging, the only potential merge conflict is that we have several PRs for survfunc, but in different parts of the modul, AFAIK

@bashtage Thank you

PR #3019, opened 2016-06-14
ENH: add function ROS implementation to imputation module

phobson:
Supercedes #2021 

## Overview
In the environment, we can nearly never say that a certain parameter doesn't exist. Instead, we say that it was not detected, and as a result report a detection limit that corresponds to the lowest possible quantifiable value of the instrument and/or measurement method. You can think about this as measuring the diameter of hair with a ruler. You know the hairs have diameters, and you know they're <1 mm, but that's about all you can say.

So trying to compute basic stats about these data sets can be tricky. Luckily Dr. Helsel wrote a couple of text books about how to deal with this. [Here is the one I used to write this code](http://www.amazon.com/Nondetects-Data-Analysis-Statistics-Environmental/dp/0471671738).

Also [here's a overview paper](http://training.fws.gov/resources/course-resources/pesticides/Monitoring/USGS%20Nondetects%20water%20sampling%20bias.pdf) and an [R-package](http://cran.r-project.org/web/packages/NADA/index.html) written by I believe a colleague of Helsel.

## Tests
I got a lot of them. Coverage was 100% on my local machine. A lot of them are redundant, but check things like what should happen when >80% of values are censored, or if all but 1 values are censored, etc.

Most of the tests use known values that come from either text book or papers written by Helsel, or datasets and results that come from the R package.

## Future work
There's a lot more in the R-package that could easily be implemented in python. This is just the start.

### DIFF
--- a/statsmodels/imputation/ros.py
+++ b/statsmodels/imputation/ros.py
@@ -461,10 +461,10 @@ def _do_ros(df, result, censorship, transform_in, transform_out):
     return modeled
 
 
-def ROS(result, censorship, df=None, min_uncensored=2,
-        max_fraction_censored=0.8, substitution_fraction=0.5,
-        transform_in=numpy.log, transform_out=numpy.exp,
-        as_array=True):
+def impute_ros(result, censorship, df=None, min_uncensored=2,
+           max_fraction_censored=0.8, substitution_fraction=0.5,
+           transform_in=numpy.log, transform_out=numpy.exp,
+           as_array=True):
     """
     Impute censored dataset using Regression on Order Statistics (ROS)
     or simple substitution if insufficient uncensored data exists.
--- a/statsmodels/imputation/ros.py
+++ b/statsmodels/imputation/ros.py
@@ -11,7 +11,7 @@ Company: Geosyntec Consultants (Portland, OR)
 2016-06-14
 
 """
-
+from __future__ import division
 import warnings
 
 import numpy
--- a/statsmodels/imputation/ros.py
+++ b/statsmodels/imputation/ros.py
@@ -1,0 +1,558 @@
+"""
+Implementation of Regression on Order Statistics for imputing left-
+censored (non-detect data)
+
+Method described in *Nondetects and Data Analysis* by Dennis R.
+Helsel (John Wiley, 2005) to estimate the left-censored (non-detect)
+values of a dataset.
+
+Author: Paul M. Hobson
+Company: Geosyntec Consultants (Portland, OR)
+2016-06-14
+
+"""
+
+import warnings
+
+import numpy
+from scipy import stats
+import pandas
+
+
+def _ros_sort(df, result, censorship, warn=False):
+    """
+    This function prepares a dataframe for ROS. It sorts ascending with
+    left-censored observations first. Censored results larger than the
+    maximum uncensored results are removed from the dataframe.
+
+    Parameters
+    ----------
+    df : pandas.DataFrame
+
+    result : str
+        Name of the column in the dataframe that contains observed
+        values. Censored values should be set to the detection (upper)
+        limit.
+
+    censorship : str
+        Name of the column in the dataframe that indicates that a
+        result is left-censored. (i.e., True -> censored,
+        False -> uncensored)
+
+    Returns
+    ------
+    sorted_df : pandas.DataFrame
+        The sorted dataframe with all columns dropped except the
+        result and censorship columns.
+
+    """
+
+    # separate uncensored data from censored data
+    censored = df[df[censorship]].sort_values(by=result)
+    uncensored = df[~df[censorship]].sort_values(by=result)
+
+    if censored[result].max() > uncensored[result].max():
+        censored = censored[censored[result] <= uncensored[result].max()]
+
+        if warn:
+            msg = ("Dropping censored results greater than "
+                   "the max uncensored result.")
+            warnings.warn(msg)
+
+    return censored.append(uncensored)[[result, censorship]].reset_index(drop=True)
+
+
+def cohn_numbers(df, result, censorship):
+    """
+    Computes the Cohn numbers for the detection limits in the dataset.
+
+    The Cohn Numbers are:
+
+        - :math:`A_j =` the number of uncensored obs above the jth
+          threshold.
+        - :math:`B_j =` the number of observations (cen & uncen) below
+          the jth threshold.
+        - :math:`C_j =` the number of censored observations at the jth
+          threshold.
+        - :math:`\mathrm{PE}_j =` the probability of exceeding the jth
+          threshold
+        - :math:`\mathrm{DL}_j =` the unique, sorted detection limits
+        - :math:`\mathrm{DL}_{j+1} = \mathrm{DL}_j` shifted down a
+          single index (row)
+
+    Parameters
+    ----------
+    dataframe : pandas.DataFrame
+
+    result : str
+        Name of the column in the dataframe that contains observed
+        values. Censored values should be set to the detection (upper)
+        limit.
+
+    censorship : str
+        Name of the column in the dataframe that indicates that a
+        result is left-censored. (i.e., True -> censored,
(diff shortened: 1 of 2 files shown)


statsmodels/imputation/ros.py
```diff
@@ -0,0 +1,558 @@
...
+          .assign(rank=lambda df: _ros_group_rank(df, 'det_limit_index', censorship))
+          .assign(plot_pos=lambda df: plotting_positions(df, censorship, cohn))
+          .assign(Zprelim=lambda df: stats.norm.ppf(df['plot_pos']))
+          .pipe(_ros_estimate, result, censorship, transform_in, transform_out)
+    )
+
+    return modeled
+
+
+def ROS(result, censorship, df=None, min_uncensored=2,
```
josef-pkt: name `ROS` (non-pep-8 capital letters) is misleading for function
maybe something like `impute_ros` ?

statsmodels/imputation/ros.py
```diff
@@ -0,0 +1,558 @@
...
+    ND_plotpos = plot_pos[df[censorship]]
+    ND_plotpos.values.sort()
+    plot_pos.loc[df[censorship]] = ND_plotpos
+
+    return plot_pos
+
+
+def _ros_estimate(df, result, censorship, transform_in, transform_out):
+    """ Computed the estimated censored from the best-fit line of a
+    probability plot of the uncensored values.
```
josef-pkt: should be one line  (compute is redundant)
maybe "estimate censored observation from probability plot"

(just realized it's private, but still ...)

statsmodels/imputation/ros.py
```diff
@@ -0,0 +1,558 @@
...
+    return modeled
+
+
+def ROS(result, censorship, df=None, min_uncensored=2,
+        max_fraction_censored=0.8, substitution_fraction=0.5,
+        transform_in=numpy.log, transform_out=numpy.exp,
+        as_array=True):
+    """
+    Impute censored dataset using Regression on Order Statistics (ROS)
+    or simple substitution if insufficient uncensored data exists.
```
josef-pkt: one line summary
first line looks good

there can be more lines of description after the header/summary line

statsmodels/imputation/ros.py
```diff
@@ -0,0 +1,558 @@
...
+
+    Method described in *Nondetects and Data Analysis* by Dennis R.
+    Helsel (John Wiley, 2005) to estimate the left-censored (non-detect)
+    values of a dataset.
+
+    Parameters
+    ----------
+    result : str or array-like
+        Label of the column or the float array of censored results
```
josef-pkt: `results` here is a bit confusing/ambiguous
Are these other results that are used as inputs here, or does this refer to the results of this function?

josef-pkt:
overall looks fine to me, based on brief skimming and spot checking.

As I mentioned in #2021 I thinks it's useful and can be merged as is. But I expect that eventually this are will hopefully get a more consistent pattern where we can tie in different models and functions.
(e.g. this still leaves it open that we create a ROS class again, if needed.)

statsmodels/imputation/ros.py
phobson: how would you feel about just `impute`?

a brief comparison:

  - `statsmodels.imputation.ros.ROS` (we know this doesn't work)
  - `statsmodels.imputation.ros.impute_ros` (maybe too redundant)
  - `statsmodels.imputation.ros.ros` (repetitive, but has a precedent e.g., `glob.glob`, `datetime.datetime`)
  - `statsmodels.imputation.ros.impute` (`<package>.<module>.<description>.<verb>` -- i like it)

I agree the first choice was the wrong one. Perfectly happy with your suggestion. My preference is for one of the last two.

statsmodels/imputation/ros.py
phobson: a global replacement of "results" to either "values" or "observations" would make sense here, IMO.

(FWIW, in my domain, "results" maps to "laboratory results", but I 100% agree that it's awkward here)

statsmodels/imputation/ros.py
josef-pkt: redundant part of names depends on where we add it in the api.
If we add it at a higher level api directly, then `ros` might not be part of the path. If it's always accessed through `ros.xxx`, then an extra ros would not be very informative.

another problem: on Windows the file paths are not case sensitive, so sphinx cannot create html files for functions or classes that differ only in case. (we have a few problem cases like that in genmod links)

phobson: understood. `impute_ros` it is.

bashtage:
Old Pandas troubles.

bashtage:
Master is quite broken now.  Only will pass on pandas 0.18+ I think.

josef-pkt:
AFAICS is mostly in the unit tests, so in the worst case we add skip to them. 
main problems, AFAIKT are pipe and assign

bashtage:
Actually not that bad, assign was introduced in 0.16.

phobson:
The ROS implementation -- not just the tests -- uses `assign` and `pipe`.

bashtage:
> AFAICS is mostly in the unit tests, so in the worst case we add skip to them. 
> main problems, AFAIKT are pipe and assign

https://github.com/statsmodels/statsmodels/blob/63f162e2a51b766ec61ed7ed673dbe9840a88ea4/statsmodels/imputation/ros.py#L186

phobson:
and here's where I use pipe:
https://github.com/statsmodels/statsmodels/blob/63f162e2a51b766ec61ed7ed673dbe9840a88ea4/statsmodels/imputation/ros.py#L472

phobson:
I can revise to not use those methods (`assign` and `pipe`), but I don't know how much of a priority it is to maintain compatibility with (what is now) a fairly old version of pandas.

In either case, it's hard to imagine that this PR is the place to have the discussion and make decisions about the supported versions of pandas. So I should probably make those revisions anyway.

bashtage:
You never know...I'm a big fan of dropping ancient versions (testing against pandas 0.12 and NumPy 1.6.2?!?).  For pandas this much bc really hurts - NumPy and SciPy are more conservative about API changes.

phobson:
I'll hold off until we hear back from Josef. 

In either case, he and discussed upthread about moving about from the `result` variable to something better like `observations`. That should still happen, IMO.

josef-pkt:
I thought pipe is mainly for interactive use (I don't have a functional R/Hadleyverse background and to me it looks "unnatural")

pipe is too recent, but I think `assign` will be ok if it's 0.16. We can revert some compatibility fixes after 0.8 is out. However, I don't see much reason in dropping compatibility just before a release if we managed reasonably well so far.

phobson:
My impression is that `pipe` is the official API for "extending dataframe functionality" without having to subclass `DataFrame`.

In either case, I can revise to remove the use `pipe` and `assign` tonight and take of `"result"` vs `"observations"` at the same time.

Do you want to revert the merge so that I can continue to work off of this PR or should I start a new PR off of master?

josef-pkt:
@phobson  If you manage compatibility with older versions in the main code, then that would be better. If you have a clean commit for very old pandas, then it should be possible to just revert it in 0.9.

I was rushing too much, hoping to get more PRs out of the way, but that doesn't always work.

josef-pkt:
@phobson 
I think you can make a new PR. I haven't figured out any use for revert yet (since it just commits the reverse changes), except in an emergency.
I don't have any PRs that need to be merged right away, I have some PRs to finish and rebase, but those can be rebased again after I merge your fix.

phobson:
Sounds good. Feel free to be a squeaky wheel if you don't need a new PR before tomorrow morning west coast time.

josef-pkt:
> My impression is that pipe is the official API for "extending dataframe functionality" without having to subclass DataFrame.

My impression was that it's purely for extending method chaining, which are equivalent to a sequence of statements without chaining.

josef-pkt:
@phobson  Ping Ping
Did you have time to work on this?

PR #3023, opened 2016-06-15
TST: Switch appveyor 64bit

bashtage:
Switched Python 2.7 to 64 bit which mostly elimiates noise which seems to be driven by small integer types.  

@ChadFulton Also found an incorrect practice of using `isinstance(*, int)` which should usually be `isinstance(*, (int, long))` for older Python.  Fixed all instances where I though it could be hit, even it accidentally.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,7 +6,7 @@ os: Visual Studio 2015
 environment:
   matrix:
     - PY_MAJOR_VER: 2
-      PYTHON_ARCH: "x86"
+      PYTHON_ARCH: "x86_64"
     - PY_MAJOR_VER: 3
       PYTHON_ARCH: "x86_64"
 
--- a/statsmodels/datasets/utils.py
+++ b/statsmodels/datasets/utils.py
@@ -1,6 +1,6 @@
 from statsmodels.compat.python import (range, StringIO, urlopen,
                                        HTTPError, URLError, lrange,
-                                       cPickle, urljoin, BytesIO)
+                                       cPickle, urljoin, BytesIO, long)
 import sys
 import shutil
 from os import environ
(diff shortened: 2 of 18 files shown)


ChadFulton:
Great, thanks for catching that!

josef-pkt:
I assume this is good, I didn't try to check this.

merge conflict needs to be resolved, then we merge

josef-pkt:
PR title is not really appropriate anymore, since main changes are in second commit.
I will open a new issue for Windows 32 bit, since there might be more problems hiding.

bashtage:
These all his tsatools and so there might be some merge conflicts after you
merge one or two.  Let me know and I can rebase.  It is all imports so easy.

bashtage:
In fairness to the title, these fixes only are needed to pass tests on Python 2.7/Windows 64

josef-pkt:
> In fairness to the title, these fixes only are needed to pass tests on Python 2.7/Windows 64

I thought it was a 32bit problem. What kind of errors where they? (I didn't find an appveyor run with it.)
I have python 2.7 on windows 64 bit and don't remember any failures, but I haven't tested with it in a while.

bashtage:
Win32 has problems due to multiplication of fixed precision integers, as far as I could tell. These lead to small precision errors - no outright failures.

Wind64 doesn't have these but does often return long on Python 2.7 rather than integer.  This mostly happens when inteerfacing with cython since the number of bits in a type can differ between Win32 and Win64, and so Cython uses different converters on these which can produce a long on Win64 and an int (even for very small numbers like 40, which were causing the failures).

Win32 is worth watching, but it is probably hard to replicate the failures.

josef-pkt:
I will merge #2240 before this (when it's green), so rebase can wait a bit.

josef-pkt:
I will merge after rebase and green Travis

josef-pkt:
appveyor 2.7 is also green, py3 will still run for a while

merging

@bashtage Thank you

PR #3029, opened 2016-06-17
BUG: Add CumIncidenceRight to the duration API

kshedden:
This should have been part of #3016


josef-pkt:
I guess it's not important enough to go into the top level api, is it?

looks good to merge

kshedden:
I think we can leave it out of the top level api.

PR #3048, opened 2016-06-20
PERF: Implement a minimal WLS for use in iterative estimation

### LINKED ISSUE #2287: PERF: speed up iteratively reweighted least squares

bashtage:
Implement minimal WLS that computes the bare minimum needed
for GLM and RLM model estimation.

closes #2287

### DIFF
--- a/statsmodels/regression/_tools.py
+++ b/statsmodels/regression/_tools.py
@@ -1,9 +1,7 @@
 from collections import namedtuple
 import numpy as np
+from statsmodels.tools.tools import Bunch
 
-_MinimalWLSResults = namedtuple('_MinimalWLSResults',
-                                ['params', 'resid',
-                                 'model', 'fittedvalues', 'scale'])
 _MinimalWLSModel = namedtuple('_MinimalWLSModel', ['weights'])
 
 
@@ -93,13 +91,11 @@ class _MinimalWLS(object):
         else:
             params, _, _, _ = np.linalg.lstsq(self.wexog, self.wendog)
 
-        model = _MinimalWLSModel(weights=self.weights)
         fitted_values = self.exog.dot(params)
         resid = self.endog - fitted_values
         wresid = self.wendog - self.wexog.dot(params)
         df_resid = self.wexog.shape[0] - self.wexog.shape[1]
         scale = np.dot(wresid, wresid) / df_resid
 
-        return _MinimalWLSResults(params=params, fittedvalues=fitted_values,
-                                  resid=resid,
-                                  model=model, scale=scale)
+        return Bunch(params=params, fittedvalues=fitted_values, resid=resid,
+                     model=self, scale=scale)
--- a/statsmodels/regression/_tools.py
+++ b/statsmodels/regression/_tools.py
@@ -2,7 +2,7 @@ from collections import namedtuple
 import numpy as np
 
 _MinimalWLSResults = namedtuple('_MinimalWLSResults',
-                                ['params', 'normalized_cov_params', 'resid',
+                                ['params', 'resid',
                                  'model', 'fittedvalues', 'scale'])
 _MinimalWLSModel = namedtuple('_MinimalWLSModel', ['weights'])
 
@@ -50,14 +50,12 @@ class _MinimalWLS(object):
         else:
             self.wexog = w_half[:, None] * exog
 
-    def fit(self, cov=True, method='pinv'):
+    def fit(self, method='pinv'):
         """
         Minimal implementation of WLS optimized for performance.
 
         Parameters
         ----------
-        cov : bool, optional
-            Flag indicating whether to the compute normalized covariance
         method : str, optional
             Method to use to estimate parameters.  "pinv", "qr" or "lstsq"
 
@@ -75,7 +73,6 @@ class _MinimalWLS(object):
               * params : Estimated parameters
               * fittedvalues : Fit values using original data
               * resid : Residuals using original data
-              * normalized_cov_params
               * model : namedtuple with one field, weights
               * scale : scale computed using weighted residuals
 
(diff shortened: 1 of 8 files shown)


josef-pkt:
I'm adding 0.9, too late for 0.8

josef-pkt:
@bashtage  Thanks, Do you know what the speed improvement is with this?

bashtage:
About a factor of 3 -- pinv is easily the fastest of pinv, qr and lstsq.

josef-pkt:
#2203   most likely we should move this to the special models in regression
(doesn't exist yet, but I have two PRs with two different special_model modules, for sparse OLS and for absorb)

statsmodels/regression/_tools.py
```diff
@@ -0,0 +1,118 @@
...
+        -----
+        Does not perform and checks on the input data
+
+        See Also
+        --------
+        statsmodels.regression.linear_model.WLS
+        """
+        normalized_cov_params = None
+        if method == 'pinv':
+            pinv_wexog = np.linalg.pinv(self.wexog)
```
thequackdaddy: Is there a way that we could `pinv` on `self.exog` only once? Seems like this will run ever iteration and the result shoudl be same every time, if I understand correctly.

thequackdaddy: On second thought, this is `wexog`, and the weight should change every iteration as `mu` updates. Never mind.

bashtage: Yes, the weight changes and so it isn't possible to improve that step.

One source of slowness is the sheer number of statistics that need to be
computed for the update dictionary - residuals, whitened residuals, scale.
For moderate datasets these impact speeds.  For very large datasets, pinv
dominates.

bashtage:
I tested with a very large dataset - 30,000 observation.

Using `pinv` took around 18s, `lstsq` took 12s and `qr` took 24.  Using the full `WLS` was virtually identical to `pinv`, around 19s.

bashtage:
I think @kshedden made some improvements to remove obvious inefficiencies and so there isn't alot, other than using lstsq which is a bit faster.

The testing was with GLM using the notebook example using `np.tile` to repeat the data 100 times.

bashtage:
In a large RLM with 1000000 obs master takes 10s for fit and this lstsq takes 5s.  Using lstsq is definitely helpful, but it isn't clear whether a dedicated class/function is better than adding lstsq as a fit option to WLS.

josef-pkt:
you switched to numpy lstsq which I think also uses svd, so any sizable differences would be surprising, or it uses a better setting for svd.

One possibility at the cost of loosing precision would be to use the normal equations directly which might be faster for large nobs but not large k_vars.

@thequackdaddy is working with a few million observations in GLM.

josef-pkt:
(I'm intermittently loosing internet now, fortunately after the rc release.)

are you doing the timing just for pinv/lstsq or including cov_params?

bashtage:
Numpy uses `dgesv` while I think SVD uses `dgesdd` in this case.

bashtage:
> (I'm intermittently loosing internet now, fortunately after the rc release.)
> 
> are you doing the timing just for pinv/lstsq or including cov_params?

The times reported are just for the `fit` step. The parts that aren't covered by `_MinimalWLS` have a fixed cost across the two implementations.

josef-pkt:
within the `fit` step, you have the optional cov=True in the fit method.
lstsq needs another matrix inverse for the cov, while pinv only needs the dot product. If lstsq is faster in spite of the inv, then it would be even faster if cov=False

bashtage:
It is, but only a small amount since the matrix inverse is O(K^3) while the
dot product is O(N^2) and when N is large, the second will be much bigger.

thequackdaddy:
Hello,

Wanted to say nice work to @bashtage . This seems to be much faster for me. 

Using current statsmodels:

https://gist.github.com/thequackdaddy/0fefc83175cc3d70c98bccc7f2468679

Using @bashtage 's minimal WLS implementation.

https://gist.github.com/thequackdaddy/1f594e3dcf5fcb853489f4a31617a537

It goes from 50.9s with the current implementation to 29.6s with the change. 

For notes, this is built using a laptop with 16 GB of RAM and Intel i7-4710MQ processor, 2.5 GHz with 4 physical cores, 8 virtual cores. (Honestly still confused about difference between physical/virtual cores, but I understand that this means that MKL can run up to 4 threads, only.)

So nice work.

josef-pkt:
@thequackdaddy  thanks for the feedback

If the main differences are between pinv and lstsq, then we should add lstsq as a fit option to the linear_models so we can compare WLS with lstsq with the MinimalWLS.

statsmodels/regression/_tools.py
```diff
@@ -0,0 +1,105 @@
+from collections import namedtuple
+import numpy as np
+
+_MinimalWLSResults = namedtuple('_MinimalWLSResults',
```
josef-pkt: It is better to make this into simple Dummy/Holder/Bunch classes.
tuples are not forward compatible if we want to change what is returned.

statsmodels/regression/_tools.py
```diff
@@ -0,0 +1,105 @@
...
+        if method == 'pinv':
+            pinv_wexog = np.linalg.pinv(self.wexog)
+            params = pinv_wexog.dot(self.wendog)
+        elif method == 'qr':
+            Q, R = np.linalg.qr(self.wexog)
+            params = np.linalg.solve(R, np.dot(Q.T, self.wendog))
+        else:
+            params, _, _, _ = np.linalg.lstsq(self.wexog, self.wendog)
+
+        model = _MinimalWLSModel(weights=self.weights)
```
josef-pkt: What's the purpose of adding this Model class?
Why not just use self, which is already a model class?

josef-pkt:
I think we should merge this soon.

Some future ideas:
I think we make this swappable. We can use the same pattern for WLS that have a different structure, e.g. ridge penalized or sparse matrices. (something similar for structured Ridge/L2 penalization is in the GAM/penalized spline PR (*). a draft version of SparseOLS is in #2711)
For this we will need to eventually fix a internal API for the MinimalWLSXxx classes and their use in IRLS.

This PR is fully backwards compatible (except for maybe some near singular corner cases), we don't have to settle on a new API yet.


(*) checking this: it uses data augmentation to implement the penalty, instead of solving the Theil/Ridge moment conditions directly
https://github.com/statsmodels/statsmodels/pull/2744/files#diff-c83444952c95fc7854675f85758d9006R329

bashtage:
Named tuple is not a tuple.  It is a Python provided collection that is
similar to a bunch without requiring a user provided class.  It uses
attribute access.   Future can easily add attributes as needed.

josef-pkt:
but it allows tuple unpacking and slicing/indexing, which breaks when we want to change the number of returns
a, b, c = results

(that's a problem that scipy, especially scipy.stats, has for making changes)

josef-pkt:
sorry last comment was for a different PR, I must have switched browser windows by accident.

josef-pkt:
in it goes

@bashtage Thank you

Issue #1063, opened 2013-08-26
Remove version compatibility code


jseabold:
We can remove some old functions that are no longer needed, etc.. A lot of the time there are source comments so they're easy to find with grep. For example, #127.

Will need to do similar for pandas workarounds 0.7.x -> 0.8.x, mainly in tsa.

jbrockmendel:
#5036 is about all that’s left of this, unless you count code copy/pasted from scipy.

bashtage:
Appears to be closed in master.

PR #3049, opened 2016-06-20
CLN: Remove old unreachable code

bashtage:
Remove unreachable code for ancient versions of NumPy, etc.
Cleanup direct version tests to use compat

### DIFF
--- a/statsmodels/stats/tests/test_power.py
+++ b/statsmodels/stats/tests/test_power.py
@@ -11,26 +11,26 @@ Author: Josef Perktold
 """
 
 import copy
+import warnings
+import nose
+from distutils.version import LooseVersion
 
 import numpy as np
 from numpy.testing import (assert_almost_equal, assert_allclose, assert_raises,
-                           assert_equal, assert_warns)
-
+                           assert_equal, assert_warns, dec)
+import scipy
 
 import statsmodels.stats.power as smp
-import warnings
-#from .test_weightstats import CheckPowerMixin
 from statsmodels.stats.tests.test_weightstats import Holder
 
-# for testing plots
-import nose
-from numpy.testing import dec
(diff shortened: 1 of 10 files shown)


bashtage:
Mostly driven by #1063

statsmodels/stats/tests/test_power.py
```diff
@@ -727,28 +727,6 @@ def test_power_solver_warn():
     nip.start_bqexp['nobs1'] = {'upp': 50, 'low': -20}
     val = nip.solve_power(0.1, nobs1=None, alpha=0.01, power=pow_, ratio=1,
                           alternative='larger')
-    import scipy
-    if scipy.__version__ < '0.10':
```
josef-pkt: AFAIR, this shouldn't be deleted, the problem was with 0.12 which failed compared to older versions.
This needs checking if it works with recent scipy.
i.e. need to check changes in scipy optimize.

IIRC

bashtage: Passes tests on 0.12.  0.10 isn't supported.

josef-pkt: I don't know if I make myself clear:
Tests currently pass because those tests are not enabled.
We need to remove the `if scipy.__version__ ...` condition, unindent the following code an check whether those tests no pass on recent scipy.
https://github.com/bashtage/statsmodels/commit/f33ebfb9f599a53bce507847e9a2a3712dab80cd

There are several backup approaches to solve for the root, and different scipy versions succeeded or failed in different cases with different optimize solvers/root finders. 
But this code check has never been updated since I wrote it for scipy 0.9

josef-pkt:
Looks good, still merging to 0.8

@bashtage  Thanks

PR #3055, opened 2016-06-21
Distributed estimation

lbybee:
Ok, I wanted to make a PR for this, there are still a couple of things that I need to fix but things are pretty close to done and I'm happy with the current approach.  

The key part of the current approach is a function `distributed_estimation`.  This function works by taking a generator for endog and exog, `endog_generator` and `exog_generator`, as well as a series of functions and key word arguments to be run on each machine and then used to recombine the results.  The generator approach allows for a variety of use cases and can handle a lot of the ideas discussed in the initial proposal.  For each data set yielded by the generators a model is initialized using `model_class` and `init_kwds` and then the function `estimation_method` is applied to the model along with the key words `fit_kwds` and `estimation_kwds`.  Finally, the results are recombined from each data set using `join_method`.  

Currently, this defaults to the distributed regularized approach discussed here:

http://arxiv.org/pdf/1503.04337v3.pdf

but the way I've set things up means that the user should be able to apply any number of procedures here.

The current todo list:
- [x] Fix hess_obs
- [x] Fix joblib fit
- [x] Add dask fit
- [x] Add WLS/GLS for debiased regularized fit
- [x] Add likelihood result
- [x] Add data tests

Let me know if there are any comments/questions/criticisms, as I've said, it certainly isn't complete but I wanted to get this out there so I can start integrating any changes as I finish it up.


kshedden:
I would consider having a single generator that yields both endog and exog.  There are several reasons for this:

* Danger that they become desynchronized and no obvious way to verify that they are synchronized

* Should be more efficient since in most cases endog and exog are obtained from the same source and we may end up creating arrays or reading files twice if there are separate generators

* No obvious use case for having separate generators

* More analogous to what Patsy does with dmatrices

kshedden:
Not sure what @josef-pkt thinks but perhaps `fit_distributed` would be a better top-level name than `distributed_estimation` (for the method or function, the module name is fine)

josef-pkt:
(I haven't looked at pr yet)
One question is how to support generated exog, e.g. Polynomials, sparse dummy sets.  Sparse exog would be created on the fly and not stored on disk.
In other context i was thinking of adding an exog-extra)

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,410 @@
...
+    return beta_tilde
+#    return beta_tilde, beta_mn, theta_hat, grad_mn, gamma_hat_l, theta_hat.dot(grad_mn)
+
+
+def distributed_estimation(endog_generator, exog_generator, partitions,
+                           model_class=None, init_kwds=None, fit_kwds=None,
+                           estimation_method=None, estimation_kwds=None,
+                           join_method=None, join_kwds=None):
+    """
+    This functions handles a general approach to distributed estimation,
```
kshedden: First line of docstring should be one line.  Then blank line, then can have additional text that is longer than one line.

statsmodels/duration/hazard_regression.py
```diff
@@ -446,7 +442,7 @@ def fit(self, groups=None, **args):
 
         return results
 
-
+
```
kshedden: You can configure your editor to strip trailing whitespace on save.  We want to avoid the history being affected by this.

statsmodels/duration/hazard_regression.py
```diff
@@ -1172,19 +1168,14 @@ def baseline_cumulative_hazard_function(self, params):
...
 
     def predict(self, params, exog=None, cov_params=None, endog=None,
-                strata=None, offset=None, pred_type="lhr"):
-        # docstring attached below
+                strata=None, offset=None, pred_type="lhr", 
+                return_object=False, **kwargs):
 
         pred_type = pred_type.lower()
-        if pred_type not in ["lhr", "hr", "surv", "cumhaz"]:
+        if pred_type not in ["lhr", "hr", "surv", "cumhaz", "qsurv"]:
```
kshedden: Is this part of the distributed estimation PR?  It looks like you may have gotten your two PRs crossed.

statsmodels/duration/tests/test_phreg.py
```diff
@@ -374,6 +403,24 @@ def plf(params):
                 llf_sm = plf(sm_result.params)
                 assert_equal(np.sign(llf_sm - llf_r), 1)
 
+    # Simulation based test of survival quantiles.
```
kshedden: I don't think this belongs in the PR.  It is already merged to master.  May need to rebase and cherry pick unwanted commits.

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,410 @@
...
+        grad_l.append(r[1])
+        gamma_hat_l.extend(r[2])
+        tau_hat_l.extend(r[3])
+
+    p = len(gamma_hat_l)
+
+    beta_mn = np.zeros(p)
+    for beta_hat in beta_hat_l:
+        beta_mn += beta_hat
+    beta_mn = beta_mn / partitions
```
kshedden: Slightly more efficient: `beta_mn /= partitions`

statsmodels/duration/hazard_regression.py
lbybee: Yeah, I do think they got crossed, I'm not sure what the best way is to fix that.

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,410 @@
...
+
+    grad_mn = np.zeros(p)
+    for grad in grad_l:
+        grad_mn += grad
+    grad_mn = -grad_mn / partitions
```
kshedden: Slightly more efficient: `grad_mn *= -1 / partitions`

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,410 @@
...
+    model_results_l : list
+        A list of tuples each one containing the beta_hat, grad,
+        gamma_hat and tau_hat values for each partition.
+    partitions : scalar
+        The number of partitions that the data will be split into.
+    threshold : scalar
+        The threshold at which the coefficients will be cut.
+    """
+
+    # TODO currently the way we extract p is roundabout, should be
```
kshedden: I think what you are doing is fine (with p = len(model_results_l))

statsmodels/duration/hazard_regression.py
kshedden: What I usually do is something like this:

First make sure you have pushed all your work to github

Then:

git checkout master
git reset --hard HEAD
git fetch upstream
git merge upstream/master
git push origin master -f
[login]
git checkout distributed estimation
git rebase master
[hopefully no merge conflicts]
git push origin distributed_estimation -f
[login]

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,410 @@
...
+        The number of partitions that the data will be split into.
+    model_class : statsmodels model class
+        The model class which will be used for estimation. If None
+        this defaults to OLS.
+    init_kwds : dict-like or None
+        Keywords needed for initializing the model, in addition to
+        endog and exog.
+    fit_kwds : dict-like or None
+        Keywords needed for the model fitting.
+    estimation_method : function or None
```
kshedden: The name sounds like a string describing the method type.  I would use `estimator` or `estimator_func` if it is for a function argument.  

Also, function arguments are OK, but usually we use classes with methods rather than a lot of functional programming.

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,410 @@
...
+        The function that performs the estimation for each partition.
+        If None this defaults to _est_regularized_distributed.
+    estimation_kwds : dict-like or None
+        Keywords to be passed to estimation_method.
+    join_method : function or None
```
kshedden: Again, maybe `joiner`, `joiner_func`, or `join_func`

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,410 @@
...
+"""
+distributed regularized estimation.
+
+Routines for fitting regression models using a distributed
+approach outlined in 
+
+Jason D. Lee, Qiang Liu, Yuekai Sun and Jonathan E. Taylor.
+"Communication-Efficient Sparse Regression: A One-Shot Approach." 
+arXiv:1503.04337. 2015.
+"""
```
kshedden: If there is a URL please add it here

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,410 @@
+from statsmodels.base.elastic_net import fit_elasticnet
+from statsmodels.base.model import Results
+from statsmodels.tools.decorators import cache_readonly
+from statsmodels.regression.linear_model import OLS
```
kshedden: Importing specific models (OLS) suggests that this is not fully generic?

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,410 @@
...
+distributed regularized estimation.
+
+Routines for fitting regression models using a distributed
+approach outlined in 
+
+Jason D. Lee, Qiang Liu, Yuekai Sun and Jonathan E. Taylor.
+"Communication-Efficient Sparse Regression: A One-Shot Approach." 
+arXiv:1503.04337. 2015.
+"""
+
```
kshedden: Add something like this:

The main use case for distributed estimation is to fit models to data that are too large to fit in memory, due to large number of cases and/or variables.  The distributed estimation here does not currently support multi-node or multi-core processing -- the data subsets must be fit sequentially on a single machine.

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,410 @@
...
+
```
kshedden: Also add the reference to Han Liu's paper and a brief note indicating how the wpproaches differ.

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,410 @@
...
+def _gen_grad(mod, beta_hat, n, alpha, L1_wt, score_kwds):
```
kshedden: We try to avoid using greek letters since they don't have a lot of meaning in themselves.  It's OK for an internal function but try to minimize.  For example, we usually use "params" instead of "beta".  We are currently using "alpha" as a regularization parameter for legacy reasons but may likely rename it in the future.

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,410 @@
...
+    -----
+
+    gamma_hat_i = arg min 1/(2n) ||X_beta,i - X_beta,-i gamma||_2^2
+                          + alpha ||gamma||_1
+    """
+   
+    ind = range(pi) + range(pi + 1, p)
+
+    # TODO use elastic net optimization routine directly
+    tmod = OLS(X_beta[:, pi], X_beta[:, ind])
```
lbybee: @kshedden, I thought it might make more sense to respond to your line note about importing OLS here to show why that is done.  Essentially we are importing OLS because we need to perform nodewise regression to form the approximate inverse covariance matrix.  So everything should still be fully general.

statsmodels/base/distributed_estimation.py
lbybee: Currently everything here comes from the Taylor/Sun paper, I did use the Han Liu paper to confirm that I'm thinking about things correctly but there is nothing in the code that is distinct, do you think it is worth including the citation in that case?

statsmodels/base/distributed_estimation.py
lbybee: Also, the model_class for distributed_estimation currently defaults to OLS, though the handling of this could be changed to only import OLS if model_class was None in the case where gamma calls the elastic_net code directly.

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,400 @@
...
+
+    Parameters
+    ----------
+    gamma_hat_l : list
+        A list of array-like object where each object corresponds to
+        the gamma hat values for the corresponding variable, should
+        be length p.
+    tau_hat_l : list
+        A list of scalars where each scalar corresponds to the tau hat
+        value for the corresponding variable, should be length p.
```
josef-pkt: those are pretty tautological descriptions if we don't know what the greek letter stands for.

A glossary for greek translation would be useful until those have descriptive names.

statsmodels/base/tests/test_distributed_estimation.py
```diff
@@ -0,0 +1,178 @@
+import numpy as np
+from numpy.testing import assert_equal, assert_
+from statsmodels.regression.linear_model import OLS
+from statsmodels.genmod.generalized_linear_model import GLM
+from statsmodels.genmod.families import Binomial
+from statsmodels.base.distributed_estimation import _gen_grad, _gen_wdesign_mat, _gen_gamma_hat, _gen_tau_hat, _gen_theta_hat, _est_regularized_distributed, _join_debiased, distributed_estimation
+
+
+def exogGen(exog, partitions):
```
josef-pkt: pep-8 undeline instead of camel case

josef-pkt:
the unit tests are checking mainly the shape of the returns.
Are there any verified examples or reference packages to compare against?

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,400 @@
...
+
+        model = model_class(endog, exog, **init_kwds)
+
+        # TODO possibly fit_kwds should be handled within
+        # estimation_kwds to make more general?
+        results = estimation_method(model, mnum, partitions, fit_kwds,
+                                    **estimation_kwds)
+        model_results_l.append(results)
+
+        mnum += 1
```
josef-pkt: I like that it's all separated out in functions, that makes the pattern here easier to understand.
However, AFAICS this is not yet distributed computation, it's out of core where the data needs to come to the central function.
My guess is that this loop could work in parallel using joblib. We have it as optional dependency, but I haven't looked at it in a few years.

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,400 @@
...
+"""
+
+
+def _gen_grad(mod, beta_hat, n, alpha, L1_wt, score_kwds):
+    """generates the log-likelihood gradient for the debiasing
+
+    Parameters
+    ----------
+    mod : statsmodels model class
+        The model for the current machine.
```
kshedden: Can the model be different on each machine?

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,400 @@
...
+    beta_hat : array-like
+        The estimated coefficients for the current machine.
+    n : scalar
```
kshedden: I think we usually use `n_obs`

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,400 @@
...
+
+
+def _gen_grad(mod, beta_hat, n, alpha, L1_wt, score_kwds):
+    """generates the log-likelihood gradient for the debiasing
+
+    Parameters
+    ----------
+    mod : statsmodels model class
+        The model for the current machine.
+    beta_hat : array-like
```
kshedden: `params` would be better than `beta_hat`

statsmodels/regression/linear_model.py
```diff
@@ -750,8 +750,37 @@ def hessian(self, params, scale=None):
...
+            Whether to return the diag of the 3d mat
+
+        Returns
+        -------
+        hessian_obs : ndarray, 2d
+            The second derivative of the loglikelihood function evaluated at
+            params for each observation.
+        """
+
+        if diag:
```
kshedden: I don't think we want to actually create the diagonal matrix here.  Just return the diagonal itself (as a vector), then use it as if it were a diagonal matrix.  The full diagonal matrix ix nxn, too big and wasteful to create.

statsmodels/base/tests/test_distributed_estimation.py
```diff
@@ -0,0 +1,178 @@
...
+    for i in range(2):
+        res = _est_regularized_distributed(mod, i, 2, fit_kwds={"alpha": 0.5})
+        res_l.append(res)
+    joined = _join_debiased(res_l, 2)
+    assert_equal(joined.shape, beta.shape)
+
+
+def test_distributed_estimation():
+
+    X = np.random.normal(size=(50, 3))
```
kshedden: Use np.random.seed(34341) [the number is arbitrary] so the results are deterministic.

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,400 @@
...
+        Keyword arguments for the hessian function.
+
+    Returns
+    -------
+    An array-like object, updated design matrix, same dimension
+    as mod.exog
+    """
+
+    # TODO need to handle duration and other linear model classes
+    return np.sqrt(mod.hessian_obs(np.r_[beta_hat], **hess_kwds)).dot(mod.exog)
```
kshedden: If you get `hessian_obs` to return only the diagonal vector, something like this should work:

```
hess = mod.hessian_obs(beta_hat, **hess_kwds)
rhess = np.sqrt(hess)
return rhess[:, None] * mod.exog
```

Also, I'm not sure why you need the `np.r_[beta_hat]`, you might not need it, or just use `np.asarray(beta_hat)`.

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,400 @@
...
+    Notes
+    -----
+
+    gamma_hat_i = arg min 1/(2n) ||X_beta,i - X_beta,-i gamma||_2^2
+                          + alpha ||gamma||_1
+    """
+
+    # TODO this is not a very elegant solution, we can't concat
+    # range in Py 3 because it is now a generator.
+    ind = list(range(pi)) + list(range(pi + 1, p))
```
kshedden: I think you can use:

```
ind = list(range(p))
ind.pop(pi)
```

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,400 @@
...
+
+    Parameters
+    ----------
+    X_beta : array-like
+        The weighted design matrix for the current machine
+    pi : scalar
+        Index of the current variable
+    p : scalar
+        Number of variables
+    n : scalar
```
kshedden: n is not used

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,400 @@
...
+    """generates the gamma hat values for the pith variable, used to
+    estimate theta hat.
+
+    Parameters
+    ----------
+    X_beta : array-like
+        The weighted design matrix for the current machine
+    pi : scalar
+        Index of the current variable
+    p : scalar
```
kshedden: Can you just get this from X_beta?

```
p = X_beta.shape[1]
```

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,400 @@
...
+    ind = list(range(pi)) + list(range(pi + 1, p))
+
+    # TODO use elastic net optimization routine directly
+    tmod = OLS(X_beta[:, pi], X_beta[:, ind])
+
+    # TODO this is currently going to fail with some alpha
+    # values, the issue is that may have an array for of
+    # alphas for beta hat and we want each individual element
+    # here
+    gamma_hat = tmod.fit_regularized(alpha=alpha).params
```
kshedden: Maybe something like this will help:

```
if not np.isscalar(alpha):
    alpha = alpha[ind]
```

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,400 @@
...
+
+    beta_tilde = beta_mn + theta_hat.dot(grad_mn)
+
+    beta_tilde[np.abs(beta_tilde) < threshold] = 0
+
+    return beta_tilde
+#    return beta_tilde, beta_mn, theta_hat, grad_mn, gamma_hat_l, theta_hat.dot(grad_mn)
+
+
+def distributed_estimation(data_generator, partitions,
```
kshedden: You might want to start thinking about a `from_formula` version of this.

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,431 @@
...
+
+Currently the primary usecase is for out of memory computation,
+this approach currently does not support mutli-core processing
+and processing must be handled sequentially.
+
+
+There are several variables that are taken from the source paper
+for which the interpretation may not be directly clear from the
+code, these are mostly used to help form the estimate of the
+approximate inverse covariance matrix as part of the
```
kshedden: Inverse covariance matrix of what?

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,431 @@
...
+
+    Returns
+    -------
+    An array-like object of the same dimension as params
+
+    Notes
+    -----
+    In general:
+
+    nabla l_k(params)
```
kshedden: I would use "gradient" instead of "nabla" and clarify what the "k" denotes.  Also, the log-likelihood whould not include the 1/n_obs factor.

lbybee:
Ok, I believe I've incorporated most of the comments so far.  A couple of notes on the bigger ones/things still to do:

- @josef-pkt I don't believe there are any good reference packages to compare this to since it is relatively new but I can put together some simulated results and adds some tests based on that.  

- @kshedden and @josef-pkt I'm still thinking about the best way to handle formulas and generated exog.  Currently, I think the best approach is to include this in the generator, but we do get into problems with the exog not necessarily lining up for different machines.  An example of this would be fixed effects where certain groups only exist on certain machines.  Something would need to be done with the join_func to account for this.  This somewhat translates into my next point.

- @kshedden Currently, the model should be the same for each machine, in that the `model_class` and `init_kwds` are the same for each, but it is possible that the structure of the data can differ somewhat based on the generator.  There isn't much stopping this framework from handling different models for different machines, this does get into two sub-questions that @josef-pkt might weigh in on as well.

1. How much do we want this to be a "distributed computing" framework vs a "distributed fitting" framework?  Currently, `fit_distributed` is abstract enough that the user can do a lot by changing the `estimation_func`, `data_generator`, and `join_func` that they use.

2. How many use-cases do we want to include, prebuilt?  Currently we just support the debiased regularized approach but including more of these may make it more user friendly since the user won't have to build it themselves.  Obviously, this can be built out over time as desired but it is something I've been thinking about.

- @josef-pkt Since it looks like the general framework makes sense, I'm hoping to have something to handle true distributed computing sometime this week that won't break what I have so far.

josef-pkt:
about unit tests: I think a set of unit test cases could be build by comparing to the non-distributed version of a model. This would also give a check on how much precision is lost compare to the centralized version.

about combining data, endog, exog and others: I agree that they need to be produced by the same generator to avoid syncronisation problems. I think whether various arrays should be put together or kept separate should depend on computational efficiency. For example, column_stacking arrays and then taking them apart again is most likely inefficient because it will have to create intermediate arrays and copy the data. (E.g. for WLS/IRLS I was thinking of having weights calculated in a decentralized array but it would be given the other data, i.e. produced on the fly but coordinated or syncronized with the other arrays.

My guess is that fit_distributed will have to return a basic LikelihoodModel so we get Wald inference, t_test, ... from the joined results. I don't know what else will be available, predict?
Once the basic features work, I think switching to classes, where the methods could largely call the standalone functions, will be necessary or useful. I.e. what results do we or does the user get out of this.

I think that supporting the regular fit besided fit_regularized will be an important use case.

statsmodels/base/distributed_estimation.py
lbybee: In the OLS case it is an approximation of (1 / n X^T X)^{-1}, I've added a note of this in the code.  It is used as part of the debiasing procedure:

beta_tilde = 1 / K sum_k  beta_hat_k + theta_hat 1 / K sum_k nabla l_k(beta_hat_k) / N_k

josef-pkt:
BTW: `nobs` is one word without underline, most other n_xxx or k_xxx have underlines.

datnamer:
This PR is probably too far in for this suggestion to matter, but I wonder [Dask Distributed](https://github.com/dask/distributed) library can be of any assistance. 

It has custom distributed workflows, fault tolerance  and linear algerba among other features.

lbybee:
There has been some talk about using dask offline, I looked into it some earlier and it is appealing but I wanted to handle the general framework and sequential procedure first.  I'm probably going to spend some more time with it when I start working on the non-sequential procedure later this week.

kshedden:
We should be able to get compatible exogs from a formula across various
data subsets by using the Patsy API.  We can put that aside for now and get
the array interface working, but it would be nice to come back to during
the GSOC period.

kshedden:
We should aim to support true distributed fitting via Dask or joblib.  But there is also a valuable use-case for this approach in fitting large models to large datasets, but still processing the subsets sequentially.  This is a useful thing to do when the data cannot be fit in memory, but the model can be fit fast enough that it is still practical to fit all the subsets on the same machine in sequence.  I think many use cases of OLS, GLM, and PHReg should fit into this category (the fitting is as fast or faster than reading the data).

kshedden:
@lbybee maybe I'm missing something, but I don't see the usecase for different models on different splits of the data.  The way I think about this, the goal is to fit one model to one dataset, but it is impractical to do so, so as a workaround we split the data into subsets.  But at the end of the day, all we want to do is fit that one model.

datnamer:
@kshedden i definitely agree the single machine but larger than memory usecase is a common one.

Dask can still help with any caching, parallelism, and data tranfer from disk for the single machine usecase.

lbybee:
@kshedden I should clarify this some, the intended use-case has only one model for every machine.  I don't really have a use-case in mind where this isn't true.

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,436 @@
...
+    # TODO given that we already have an example where generators should
+    # produce more than just exog and endog (partition for variables)
+    # this should probably be handled differently
+    for endog, exog in data_generator:
+
+        model = model_class(endog, exog, **init_kwds)
+
+        # TODO possibly fit_kwds should be handled within
+        # estimation_kwds to make more general?
+        results = estimation_func(model, mnum, partitions, fit_kwds,
```
kshedden: I would use `fit_func` and `fit_kwds` for consistency

kshedden:
I think I was just confused with the docstring for mod that says "The model
class for the current machine".  Here we are dealing with a class instance,
not the class itself.  But I think that the code itself is fine.  Can you
revise the docstring to something like "Model class instance for the
current machine"?

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,436 @@
...
+    return grad
+
+
+def _gen_wdesign_mat(mod, params, hess_kwds):
+    """generates the weighted design matrix necessary to generate
+    the approximate inverse covariance matrix
+
+    Parameters
+    ----------
+    mod : statsmodels model class
```
kshedden: statsmodels model class instance

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,436 @@
...
+    An array-like object, updated design matrix, same dimension
+    as mod.exog
+    """
+
+    # TODO need to handle duration and other linear model classes
+    rhess = np.sqrt(mod.hessian_obs(np.array(params), **hess_kwds))
+    return rhess[:, None] * mod.exog
+
+
+def _gen_gamma_hat(X_beta, pi, alpha):
```
kshedden: I would rename this to something like "_nodewise_row" to better reflect what it is doing.

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,436 @@
...
+
+    ind = list(range(p))
+    ind.pop(pi)
+
+    d = np.linalg.norm(X_beta[:, pi] - X_beta[:, ind].dot(gamma_hat))**2
+    d = np.sqrt(d / n + alpha * np.linalg.norm(gamma_hat, 1))
+    return d
+
+
+def _gen_theta_hat(gamma_hat_l, tau_hat_l, p):
```
kshedden: I think in most cases where you have a function name including "gen" it should be changed to "get" or "calc".  "gen" implies a Python generator to me, which these are not.

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,436 @@
...
+        Keyword arguments for the hessian function.
+
+    Returns
+    -------
+    An array-like object, updated design matrix, same dimension
+    as mod.exog
+    """
+
+    # TODO need to handle duration and other linear model classes
+    rhess = np.sqrt(mod.hessian_obs(np.array(params), **hess_kwds))
```
kshedden: np.asarray is better than np.array here

statsmodels/base/distributed_estimation.py
kshedden: I would change X_beta to exog

kshedden: I would change `pi` to `ix` or `idx` (so as not co confuse with the number pi)

lbybee: I'm not sure if `exog` makes the most sense as a replacement for `X_beta` since it isn't necessarily the same as the model `exog`.  Something like `wexog` or `weighted_exog` might make more sense as a replacement.  I do agree with getting away from the greek letters though, and I like `nodewise_row`, I'll be making these changes with the next round of commits.

statsmodels/base/distributed_estimation.py
lbybee: The one issue we run into here is that `estimation_func` has its own key word arguments `estimation_kwds` as well as `fit_kwds`, which are then passed in ideally to whatever fit method is called within `estimation_func`.  I'll admit I don't have a use-case in mind for `estimation_kwds` but given that `estimation_func` can possibly do more than just make a call to `fit` I thought it made sense to allow for the possibility of additional key word arguments.

statsmodels/base/distributed_estimation.py
kshedden: Yes, wexog is good, we have used it elsewhere.

statsmodels/base/distributed_estimation.py
kshedden: OK for now.  Eventually we will have to make this more class-based.  We
will just want to be able to do something like:

result = OLS(y, x).fit_distributed(...)

etc.

datnamer:
What class of models can this estimate? Just OLS?

lbybee:
@datnamer Currently the regularized portion (which follows from http://arxiv.org/abs/1503.04337) can handle any model class that has a `hessian_obs` method.  Right now that is just OLS and GLM but these are going to be expanded.  That is currently the only limitation on model class.

josef-pkt:
@lbybee  `hessian_obs` is a misnomer, right now it's the same as hessian_factor, hessian_obs would also have the x dot x' part, but so far the full hessian_obs with shape (nobs, k_vars, k_vars) wasn't needed, and requires quite a bit of memory which is wasted if we don't need it.
(I haven't read the paper so I don't have any background on the application.)

lbybee:
@josef-pkt, yeah this was something @kshedden and I discussed earlier when I was first implementing this.  I think `hessian_factor` could just be called instead and implemented for the other model classes that don't have it but there was a reason that we wanted to use `hessian_obs` instead which I don't completely remember.  Kerby may be able to weigh in more.

josef-pkt:
hessian_factor is currently targeted for one parameter distributions/families with a linear X b central part. For other models, I think score and hessian factor will need more than one column, e.g. NegativeBinomial in discrete not in GLM. The advantage of having the "factor" for score and hessian is that they can be applied to new variables, e.g. in variable addition tests or score tests. 
As I said the full hessian_obs looks to memory intensive to me given the use cases that we had so far. If we need something similar to a full hessian_obs, then it should most likely get a different name if it is not the Hessian for each observation.

kshedden:
Just a note -- the Taylor, Sun et al. method apparently is not theoretically supported for use with Cox models, which are not M-estimators.  I haven't looked carefully to see how critical of a problem this will be.  GLM's should be OK.  We could still do the simple averaging, but perhaps not the debiasing.  @lbybee it might be worth looking again at Han Liu's paper on this topic to see if his approach is more general.

kshedden:
The pattern we might be moving toward is to have a new class called 'DistributedModel' with a signature like this:

`DistributedModel(model_class, data_generator, n_chunk=1, init_kwds=None, fit_kwds=None, parallel_method=None)`

Calling `fit` on an instance of this class would run all the distributed calculations and return a results instance appropriate to the model type, e.g. `OLSResults` or `GLMResults`

So a semi-complete example would be:

```
model = DistributedModel(sm.OLS, my_gen, 10, parallel_method="joblib")
results = model.fit()
```

Comments welcome.

kshedden:
Thinking ahead to how this might be done with joblib and other parallel toolkits:

First, we add a method to `DistributedModel` that handles everything that needs to happen to one data chunk in isolation:

```
def _fit_chunk(self, endog, exog):
    model = self.model_class(endog, exog, **self.init_kwds)
    results = model.fit_regularized(self.fit_kwds, **self.fit_kwds)
    nrow = self.nodewise_row(...)
    return results,  nrow
```

Then, the `fit` method of `DistributedModel` has something like this:

if parallel_method == "none":
    chunk_results = (self.fit_chunk(endog, exog) for endog, exog in self.data_generator)
elif parallel_method == 'joblib`:
    try:
        import joblib
    except ImportError
        raise ValueError("joblib not available")
    chunk_rslts = joblib.Parallel(n_jobs=2)(joblib.delayed(self.fit_chunk)(endog, exog) for endog, exog
           in self.data_generator)
elif:
    ...other methods

(Note: I haven't tested any of this, just proposing the pattern)

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,436 @@
...
+    model_results_l : list
+        A list of tuples each one containing the beta_hat, grad,
+        gamma_hat and tau_hat values for each partition.
+    partitions : scalar
+        The number of partitions that the data will be split into.
+    threshold : scalar
+        The threshold at which the coefficients will be cut.
+    """
+
+    params_l = []
```
kshedden: Since params_l and grad_l are only used to create params_mn and grad_mn, it's better to create the means directly rather than having a list as an intermediate.

lbybee:
Ok, this most recent build is guaranteed to fail but I wanted to put it up here so that we can discuss.  Several things,

First, I've rewritten the code such that it is similar to what @kshedden suggests.  It isn't identical, in part because I currently have it set up such that parallel_method is an argument for fit_distributed and not an attribute of the class.  
Second, I've gone through and made a large number of small changes to function/variable names and docstring to hopefully add some clarity.
Third, I've added the basics for joblib.  This last point is where I think, currently, the most discussion would be helpful.  

There are two issues that come up,
- What is the best way to handle the optional dependency for joblib with the tests?
- joblib.delayed fails with a type error `can't pickle instancemethod objects` because we are passing a class method in.  I need to think about this some more to decide on the best way to handle it.

I'm also going to add a todo list to the base comment to make it clear what still needs to be done.

lbybee:
Also, should note that I've added support for standard fit as well the regularized fit.  Currently, I don't use the debiasing procedure for the standard fit and just do simple coefficient averaging.

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,661 @@
...
+    approx_inv_cov = _calc_approx_inv_cov(nodewise_row_l, nodewise_weight_l, p)
+
+    debiased_params = params_mn + approx_inv_cov.dot(grad_mn)
+
+    debiased_params[np.abs(debiased_params) < threshold] = 0
+
+    return debiased_params
+
+
+class DistributedModel():
```
kshedden: Should inherit from object.

kshedden: We might be able to come up with a better name, but I can't think of one right now.  The issue as I see it is that the model is not distributed, it is the fitting that is distributed.  We could consider things like "DistributedFitter" or "DistributedEstimator"

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,661 @@
...
+    __doc__ = """
+    Distributed model class
+
+    Parameters
+    ----------
+    data_generator : generator
+        A generator that produces a sequence of tuples where the first
+        element in the tuple corresponds to an endog array and the
+        element corresponds to an exog array.
+    partitions : scalar
```
kshedden: Do we need `partitions`?  The generator will stop when all the data have been read, that should be sufficient.

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,661 @@
...
+        The model class which will be used for estimation. If None
+        this defaults to OLS.
+    init_kwds : dict-like or None
+        Keywords needed for initializing the model, in addition to
+        endog and exog.
+    fit_kwds : dict-like or None
+        Keywords needed for the model fitting.
+    estimation_method : function or None
+        The method that performs the estimation for each partition.
+        If None this defaults to _est_regularized_distributed.
```
kshedden: _est_regularized_distributed -> _est_regularized_debiased

statsmodels/base/distributed_estimation.py
lbybee: Each machine needs to know how many partitions there are so that we can split up the variables for the node-wise regression.  So not in general, but for the debiased regularized case we need something like this.  This might be one argument for allowing for a more general class of generators, because we could split the variables up in the generator instead of handling them internally.

josef-pkt:
about joblib: we have a wrapper in statsmodels\tools\parallel.py which might have to be compared to mne-python to see if there are updates.
the only direct usage I find with a grep/filesearch is in statsmodels\nonparametric\_kernel_base.py
(I thought it's used also in time series bootstrap, but search doesn't find a "parallel" there.)

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,661 @@
...
+    for idx in range(p):
+        ind = list(range(p))
+        ind.pop(idx)
+        approx_inv_cov[idx,ind] = nodewise_row_l[idx]
+        approx_inv_cov[idx,:] *= (- 1. / nodewise_weight_l[idx]**2)
+
+    return approx_inv_cov
+
+
+def _est_regularized_debiased(mod, mnum, partitions, fit_kwds=None,
```
kshedden: I think this can be moved to being a method of DistributedModel

kshedden:
For joblib, would it work to wrap the instance method in a regular function, like this?

```
f = lambda x, y, z: self.fit_partition(x, y, z)
f = delayed(f)
results_l = par(f(pnum, endog, exog) for pnum, (endog, exog)
                    in enumerate(self.data_generator))
```

lbybee:
Using a anonymous function was my first thought as well but unfortunately joblib fails with that too.  It is very particular about what you feed in, see further discussions,

https://github.com/joblib/joblib/issues/114

http://stackoverflow.com/questions/20628623/how-to-overcome-joblibs-typeerror-cant-pickle-instancemethod-objects-on-cla

What I've done as a workaround is define a new function `_helper_fit_partition`, outside of `DistirbutedModel` which takes a `DistributedModel` as an argument as well as `pnum`, `endog` and `exog`.

lbybee:
I will note, I'm not super happy with this fix, so I'm going to keep looking to see if I can find a better solution, but this should at least fill in for now.

kshedden:
@lbybee I think your approach to calling joblib is fine and quite natural.  So assuming it works, no need to change it in my view.  I will suggest some minor refactoring in line notes.

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,698 @@
...
+        -------
+        join_method result.  For the default, _join_debiased, it returns a
+        p length array.
+        """
+
+        if parallel_method == "sequential":
+            return self.fit_dist_sequential()
+
+        elif parallel_method == "joblib":
+            try:
```
kshedden: No need for this `try` block because the statsmodels parallel tool will do the check again.

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,698 @@
...
+        """
+
+        results_l = []
+
+        for pnum, (endog, exog) in enumerate(self.data_generator):
+
+            results = _helper_fit_partition(self, pnum, endog, exog)
+            results_l.append(results)
+
+        return self.join_method(results_l, self.partitions,
```
kshedden: This will be the last line of every `fit_dist_xxx` function, so better to move it into `fit_distributed` to avoid needless copying.

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,698 @@
...
+        from statsmodels.tools.parallel import parallel_func
+
+        par, f, n_jobs = parallel_func(_helper_fit_partition, self.partitions)
+
+#        par = Parallel(n_jobs=self.partitions)
+#        f = delayed(_helper_fit_partition)
+        results_l = par(f(self, pnum, endog, exog)
+                        for pnum, (endog, exog)
+                        in enumerate(self.data_generator))
+        return self.join_method(results_l, self.partitions,
```
kshedden: Move to `fit_distributed` to avoid repeated code.

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,698 @@
...
+                                **self.join_kwds)
+
+
+    def fit_dist_dask(self):
```
kshedden: I think we can avoid having this here.  Instead, now (or later) we can extend `statsmodels.tools.parallel` to use dask (if it is available).

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,698 @@
...
+    if not np.isscalar(alpha):
+        alpha = alpha[ind]
+
+    ind = list(range(p))
+    ind.pop(idx)
+
+    tmod = OLS(wexog[:, idx], wexog[:, ind])
+
+    nodewise_row = tmod.fit_regularized(alpha=alpha).params
+
```
kshedden: You already have `ind` here, so instead of creating it again at line 327, you can just do something line this:

```
prow = tmod.fit_regularized(alpha=alpha).params
ndrow = np.zeros(p)
ndrow[ind] = prow
```

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,698 @@
...
+    nwr = nodewise_row
+    nww = nodewise_weight
+
+    approx_inv_cov_j = - 1 / nww_j [nwr_j,1,...,1,...nwr_j,p]
+    """
+
+    approx_inv_cov = np.eye(p)
+    for idx in range(p):
+        ind = list(range(p))
```
kshedden: See above for suggestion to avoid recalculation of `ind`.

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,698 @@
...
+        ind.pop(idx)
+        approx_inv_cov[idx,ind] = nodewise_row_l[idx]
+        approx_inv_cov[idx,:] *= (- 1. / nodewise_weight_l[idx]**2)
```
kshedden: This can be moved outside the loop and vectorizes:

```
approx_inv_cov *= -1 / nodewise_weight_l[:, None]
```

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,698 @@
...
+    """
+
+    if fit_kwds is None:
+        raise ValueError("_est_unregularized_naive currently " +
+                         "requires that fit_kwds not be None.")
+
+    return mod.fit(**fit_kwds).params
+
+
+def _join_naive(params_l, partitions, threshold=0):
```
kshedden: partions = len(params_l) so not needed as an argument

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,698 @@
...
+
+    ind = list(range(p))
+    ind.pop(idx)
+
+    d = np.linalg.norm(wexog[:, idx] - wexog[:, ind].dot(nodewise_row))**2
+    d = np.sqrt(d / n + alpha * np.linalg.norm(nodewise_row, 1))
+    return d
+
+
+def _calc_approx_inv_cov(nodewise_row_l, nodewise_weight_l, p):
```
kshedden: If the value of `p` can be deduced from other args better not to include it as an argument.

kshedden:
@lbybee  Josef is right, what we are calling `hessian_obs` is equivalent to `hessian_factor` as defined here:
https://github.com/statsmodels/statsmodels/blob/master/statsmodels/genmod/generalized_linear_model.py#L470

So we should rename `hessian_obs` to `hessian_factor`.  I'm not sure if in the end we will want to include a method stub for `hessian_factor` in the base model or not, but if we do, its docstring should include the documentation linked above.

statsmodels/base/distributed_estimation.py
kshedden: Nodewise regression is of independent value, but the way it is setup here is too tailored for this application.  We have a file called "correlation_tools.py":

https://github.com/statsmodels/statsmodels/blob/master/statsmodels/stats/correlation_tools.py

I suggest you extract the nodewise regression as a general purpose procedure and move it to that module.  Then you can call it as needed here.  There are R packages implementing nodewise regression, you could write unit tests against them.

statsmodels/base/distributed_estimation.py
lbybee: To make sure I understand what you are thinking here, since we use `ind` in multiple parts of the nodewise regression do you think we should generate it externally and pass it into each function?

statsmodels/stats/nodewise_regression.py
```diff
@@ -0,0 +1,124 @@
+from statsmodels.regression.linear_model import OLS
+import numpy as np
+
+
```
kshedden: I think this should be a class called NodewiseCovariance

kshedden: On second thought, the class should be called RegularizedCovariance, and the fit method should be called "fit_nodewise"

statsmodels/base/distributed_estimation.py
kshedden: It appears to me that if you expand out to full width here, then you would not need to create ind down below (so each ind is created once instead of twice)

statsmodels/stats/nodewise_regression.py
kshedden: Tests needed.  

Also I would rename the file as regularized_covariance.py

lbybee:
I still need to integrate some of @kshedden 's changes and there is a bug that was introduced with the refactoring that I'm fixing but I wanted to push this most recent commit to show one approach to handling dask.distributed.  I spent some time with this and it looks like joblib actually has support for different parallel backends including distributed, yarn and ipyparallel.  What I've done is restructure things to allow the user to provide a parallel backend for joblib, if they would like, when they go to fit.   Currently, I'm thinking this is the cleanest approach and the most flexible.  I'm actually quite excited about the joblib backends and want to do some more testing with this but I think it is a promising approach.  The only cost is that it ties this more closely to joblib but I think that is acceptable, certainly for now.

For a little more on the joblib backends:

https://github.com/dask/distributed/issues/124

http://www.slideshare.net/AlexandreAbadie/pydata-joblib-abadie/19

https://github.com/joblib/joblib/pull/306

statsmodels/stats/regularized_covariance.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+class RegularizedInvCovariance(object):
+    __doc__ = """
+    Class for estimating regularized inverse covariance with
+    nodewise regression
+
+    Parameters
+    ----------
+    wexog : array-like
```
kshedden: I don't think `wexog` is meaningful here (what does the `w` refer to)?  I would chance it to `exog`.

statsmodels/stats/regularized_covariance.py
```diff
@@ -0,0 +1,179 @@
...
+        A weighted design matrix for covariance
+    alpha : scalar
+        Regularizing constant
```
kshedden: alpha should be an argument to `fit_nodewise`, not to the class init method.

statsmodels/stats/regularized_covariance.py
```diff
@@ -0,0 +1,179 @@
...
+            nodewise_weight = _calc_nodewise_weight(self.wexog, nodewise_row,
+                                                    idx, self.alpha)
+            nodewise_weight_l.append(nodewise_weight)
+
+        nodewise_row_l = np.array(nodewise_row_l)
+        nodewise_weight_l = np.array(nodewise_weight_l)
+
+        approx_inv_cov = _calc_approx_inv_cov(nodewise_row_l, nodewise_weight_l)
+
+        return approx_inv_cov
```
kshedden: I think the more consistent way to do this is to not return anything, just attach the result to the class (call it something shorter and private, like `_cov_inv`).

Then have an accessor function

```
@cache_readonly
def cov_inv(self):
    return self._cov_inv
```

statsmodels/stats/regularized_covariance.py
```diff
@@ -0,0 +1,179 @@
...
+    for idx in range(p):
+        ind = list(range(p))
+        ind.pop(idx)
+        approx_inv_cov[idx,ind] = nodewise_row_l[idx]
+    approx_inv_cov *= -1 / nodewise_weight_l[:, None]**2
+
+    return approx_inv_cov
+
+
+class RegularizedInvCovariance(object):
```
kshedden: Needs tests.

When I run this on iid data (result should be I), I get -I.  Is there a missing minus sign somewhere?

statsmodels/stats/regularized_covariance.py
```diff
@@ -0,0 +1,179 @@
...
+    __doc__ = """
```
kshedden: Also, when I use alpha=0 should I get the inverse of the sample covariance matrix?  I do not (not even close).

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,545 @@
...
+        else:
+            self.join_method = join_method
+
+        if join_kwds is None:
+            self.join_kwds = {}
+        else:
+            self.join_kwds = join_kwds
+
+
+    def fit_distributed(self, parallel_method="sequential",
```
kshedden: Since the class name already has "Distributed" this can just be "fit" for brevity.

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,545 @@
...
+
+        else:
+            raise ValueError("parallel_method: %s is currently not supported"
+                             % parallel_method)
+
+        return self.join_method(results_l, self.partitions,
+                                **self.join_kwds)
+
+
+    def fit_dist_sequential(self):
```
kshedden: "fit_sequential" ("distributed is redundant with class name)

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,545 @@
...
+
+        for pnum, (endog, exog) in enumerate(self.data_generator):
+
+            results = _helper_fit_partition(self, pnum, endog, exog)
+            results_l.append(results)
+
+        return results_l
+
+
+    def fit_dist_joblib(self, parallel_backend):
```
kshedden: "fit_joblib"

kshedden:
@lbybee I tried to put together a small example, see below, but the parameter estimates are all zero... am I doing something wrong?

https://gist.github.com/kshedden/10b6da262f411e1bc8f208cb999df129

lbybee:
@kshedden yeah, I believe this is the bug that resulted from refactoring that I mentioned in my earlier comment, that's currently what I'm focused on and will have some additional tests as well soon.

kshedden:
Leland -- Any updates on this?  We have about one month to go...

Kerby

lbybee:
Sorry, should have pushed some of this earlier, everything is working on my end at this point, essentially one of the issues was that the sign for the diagonal was wrong.  Let me know if you still encounter issues, there should also be some tests now.  I'm finishing up some examples today which will go up on the blog.  

One thing I wanted to discuss was the likelihood results idea.  I played around with this some but I'm not sure if it makes sense with the current set up, since there isn't a model object to work with.  @josef-pkt you can maybe weigh in here if you have any thoughts.

Also, let me know if there are any other feature requests at this point.

lbybee:
There is actually one issue with the regularized_covariance, pushed the wrong version, fixing now.

lbybee:
This should be fixed now

josef-pkt:
about the likelihood results idea

Our models have two main use cases, inference and predict. inference is mostly available in the LikelihoodModel results, the basic predict method for given parameters and exog is in the model.
So, it seems to me that using simplified/stripped-down model and results instances would be the simplest way to satisfy those use cases.

I didn't keep up with the details here, and don't have a specific suggestion. If you have a draft or specific issues, then I can look into it. (GenericLikelihoodModelResults might be a good example for what needs to be defined besides params and cov_params. e.g. df_resid is required for t_test. The model instance could be mostly "fake", i.e. no real data, just to have access to the predict method.)

lbybee:
Ok, I think I can put something like that together, I'm also still finishing up the examples, but I think I can get both together this week.

kshedden:
This procedure gives point estimates with convergence guarantees, but no
p-values or other inference tools.  The estimation is not properly
likelihood based, so many of the components of a LikelihoodResults class
(AIC, LR test, ...) would be inapplicable here.  Basically all we can give
people is point estimates, residuals, fitted values, and prediction.

The same authors (Taylor, Sun, et al) have papers about doing inference
with elastic net fits.  If we implement those ideas, we would also be able
to get inference here.  But that won't happen right away.

I think best option is to return a RegularizedResults instance.

(sorry if this contradicts what I advised before)

Kerby

josef-pkt:
@kshedden   Isn't the inference issue for regularized estimates separate from the distributed computation issue?
Does standard inference apply (approximately) if we use "distributed" with non regularized fit, or with L2 fit?

kshedden:
This approach to distributed estimation is heavily dependent on
regularization, and all the theoretical guarantees that the procedure
should work do not apply without regularization.  But since the
unregularized approach is basically a free side-effect of the real
procedure, there is no reason not to include it as an option.  And yes, in
the unregularized case it would be easy to get standard errors and I
suppose we should provide those when applicable.  But we still will lack
most of the other components of a proper likelihood fit.  So rather than
returning an instance of the results class corresponding the underlying
model (OLSResults, GLMResults, etc.), should we just provide a
RegularizedResults instance, and add standard errors when appropriate?

Kerby

lbybee:
One possibility would be to determine what results class is returned based on the join method/as part of the join method.  What I'm imagining is the _join_debiased could return RegularizedResults and other join methods could return a more standard likelihood result.  Obviously, that makes the distributed results less clear in general but it allows for more flexibility with use cases/estimation methods.

lbybee:
I think it also fits somewhat better with the overall approach I have taken here.

josef-pkt:
I think it would be good to return different results instances depending on the assumptions that we can make and have access some basic properties of the underlying model (e.g. family, and so on in GLM and to `predict`)

Getting the infrastructure for distributed estimation for the "side-effect" of unregularized estimation would be one of the main use cases IMO. For example, distributed, out-of-memory or parallel GLM.

kshedden:
You can look at either MICE or Mediation to see how to create a bare
results class of the appropriate type and populate it with estimates and
standard errors.

When regularization is involved, I still think that the RegularizedResults
class should be returned, but if no regularization, you can return an
OLSResults or GLMResults.  This will give people access to some
model-specific features like predict options.  But it will also give people
methods and results that are not applicable (e.g. AIC and loglike), so you
should probably set those to NaN, but there is no way to know generically
what all those results fields are.

Kerby

josef-pkt:
@kshedden We might still want to return a simplified DistributedResults class that just subclasses LikelihoodModelResults but attach a specific model instance (e.g. OLS or GLM)
This way we only get the Wald inference and the wrapper for predict, without getting exceptions or incorrect numbers in all the things that are not available.

kshedden:
@josef-pkt Sure (should probably do that with MICE and mediation too).

@lbybee -- if you have trouble figuring out how to do this let us know.
I'll try to look over everything during the weekend.

Kerby

lbybee:
To explain how the current setup works for the results, what I've done is add a DistributedResults class which subclasses LikelihoodModelResults.  Currently it only has a method for predict.  The way I have this set up a dummy model is generated as part of fit method for DistributedModel and the results instance is setup using the dummy model.  I also have it set up so that the user can specify a results_class as well as results_kwds to be returned.  This is handles the RegularizedResults vs other LikelihoodModelResults.  Currently the default is RegularizedResults since that fits with all the other defaults.

lbybee:
I'm going to fully link up the likelihood code later this week if everything looks good.  I also think I can add a couple of notebooks to go with the examples on the blog.  Besides that there is just WLS/GLS debiased elements and the formula handling.  For the duration debiased elements I'm not sure if we want to handle that now or leave it for later.  Any thoughts on the formula handling would be appreciated.

lbybee:
Ok, I'm happy with how this looks.  I've made some changes that should be discussed but I think everything that needs to be done is done at this point. 

Changes:

- data_generator was moved from the model init to fit.  The reasoning for this is that since the generator is one-use, it doesn't make sense to have it tied to a DistributedModel instance that might be reused.  
- I've added something called init_kwds_generator to fit as well.  It defaults to None but if provided it will generate additional init_kwds dependent on the data partition being used.  We discussed this idea some earlier and originally decided against it but I added it back while working on adding GLS/WLS because I think that is a use case.  The way this is set up it could suffer from the sync issues that prompted the joining of the endog and exog generators but given how this is something of an edge case option I don't think it is much of an concern.
- On that topic though, while I have added a hessian_factor for both GLS and WLS I forgot (until after I had the code written up) that GLS and WLS currently don't support fit_regularized.  Regardless, the hessian_factor is there now as well as some commented out tests (which can be removed if desired).  

It is also worth noting that I removed the formula handling from the todo list.  My thinking on this is that it is an issue that should be handled in the data_generator and since we aren't providing any of those in statsmodels I don't think it makes sense to have that here.  Adding more to this could be something for down the road though.

Again, I'm happy with where this is, certainly for GSoC.  There probably is more that we could down the line and I'm planning to keep playing with this set of code but I think this PR has everything that is needed to get distributed_estimation off the ground.

Let me know if there are any thoughts/changes desired.

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,677 @@
...
+    are ultimately used to form approx_inv_cov.
+
+    approx_inv_cov
+
+    This is the estimate of the approximate inverse covariance
+    matrix.  This is used to debiase the coefficient average
+    along with the average gradient.  For the OLS case,
+    approx_inv_cov is an approximation for
+
+        (1 / n X^T X)^{-1}
```
kshedden: Is this correct?  I think n * (X^T X)^{-1} is what you want?

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,677 @@
...
+    gradient l_k(params)
+
+    where k corresponds to the index of the partition
+
+    For the simple linear case:
+
+    X^T(y - X^T params)
+    """
+
+    grad = -mod.score(np.r_[params], **score_kwds)
```
kshedden: np.asarray is preferable to np.r_ (I think np.r_ always makes a copy, even if not needed).  You shouldn't need the copy here, but if you do I think params.copy() is more clear.  If the only goal is to be sure that an ndarray is going into the score function, np.asarray will do that.

statsmodels/base/distributed_estimation.py
```diff
@@ -0,0 +1,677 @@
...
+
+    Notes
+    -----
+    In general:
+
+    gradient l_k(params)
+
+    where k corresponds to the index of the partition
+
+    For the simple linear case:
```
kshedden: It's not really "simple linear" if there are more than 1 covariates, better to use "OLS" I think.

lbybee:
@kshedden @josef-pkt 

I believe that this should be ready to be merged whenever the pandas issue is solved since that is where the failures for the tests are coming from.

statsmodels/stats/regularized_covariance.py
lbybee: This issue has been fixed.  I believe it was a problem with a missing minus sign.

statsmodels/stats/regularized_covariance.py
lbybee: This has also been fixed.

kshedden:
@lbybee @josef-pkt 

This looks good, I will merge in a day or two if nobody objects.  Once it's merged I will do some more testing and we can do a quick follow-up PR if needed.

josef-pkt:
based on a brief look

- unit tests look "skinny": Is it possible to verify that numbers are "close" to what they should be? Eg. distributed should get results that are close to the non-distributed results. Otherwise, we need at least regression test to prevent future refactoring bugs, including unintended dead code
- the branch looks far behind master and will need a rebase before merging
- git history doesn't look so good, 79 commits is a bit large and it seems to have changed files early on like PHReg that are not in the final change set. Some squashing would be very helpful here. 
- there are only a few places where it touches existing models directly, so merging does not look "dangerous"

I guess discussion and changes in this PR might be helpful to understand the development of the code, so I would suggest opening a new PR with a cleaned up history and rebased branch.

lbybee:
@josef-pkt In regards to the tests, there are a number that do comparisons between distributed and non-distributed results, though they are further down in the testing file.

I'm taking care of the commit cleanup now.

PR #3070, opened 2016-06-23
[WIP] [GSoC 2016] VECM #3013

yogabonito:
The VECM framework developed in this year's GSoC aims to add the support for non-stationary VAR processes to statsmodels.
Currently, the following topics are implemented (see also this [Jupyter-notebook](https://gist.github.com/yogabonito/5461b26bed335cad6907aa4e613acb99#file-gsoc-vecm-ipynb)):
- Parameter estimation,
- forecasting,
- testing for Granger-causality,
- testing for instantaneous causality (not on my initial todo-list and may be moved to vector_ar.py to extend the VAR framework's functionality),
- impulse-response analysis, and
- lag order selection.

Some of them work without _and_ with deterministic terms in the model. Some functions, however, make use of the existing VAR framework which lacks support for some deterministic terms, such as seasonal terms. As a consequence much of the VECM-functionality relying on the VAR framework is not working with deterministic terms yet. This limitation will be overcome with the refactoring of the VAR-related code.

Future development will also include the implementation of the following VECM-related functions:
- Testing for the rank of cointegration,
- checking for residual autocorrelation,
- testing for non-normality (I have already begun working on this), and
- tests for structural change.

It's been a great summer and I am looking forward to working on these open issues!

---

_Original comment:_
Estimation of parameters and tests have been added.
A failing test (related to the parameter matrix for the deterministic terms) indicates that deterministic seasonal terms are not implemented correctly yet.
@josef-pkt, @bashtage, can you have a look at it?


bashtage:
@yogabonito Thanks for this - looks very promising.  Will give detailed feedback later today

statsmodels/tsa/vecm/vecm.py
```diff
@@ -0,0 +1,254 @@
...
+        
+        Delta_x = np.zeros((diff_lags*K, T))
+        for j in range(Delta_x.shape[1]):
+            Delta_x[:, j] = Delta_y_all[:, j+p-2:None if j-1<0 else j-1:-1].T.reshape(K*(p-1))
+        if 'c' in deterministic_terms:
+            Delta_x = np.vstack((Delta_x, np.ones(T)))
+        if 's' in deterministic_terms: # TODO: How many seasons??
+            seasons = np.zeros((3, Delta_x.shape[1])) # 3 = 4-1 (4=number_of_seasons)
+            for i in range(seasons.shape[0]):
+                seasons[i, i::4] = i+1
```
josef-pkt: AFAICS, this should be assigning 1, i.e.
`seasons[i, i::4] = 1`

This should fix the test failure for non-seasonal parameters, even though it might not be the same coding that JMulti is using, and it still leaves periodicity hardcoded at 4.

josef-pkt:
general
limit line length to <80, space around operator to make it easier to read
use temporary variables: both for readability, and to avoid repeated calculation of dot products (I saw one case when I was reading earlier, but cannot find it right now.)

we will drop all use of numpy.matrix and stick with arrays, even though we have to spell out `dot` right now (eventually it will be `@`)

add comment with equation or page number in Lutkepohl or whatever reference is used. VAR is doing that, and it makes it easier to compare notes during development.

renaming: most can wait, the main one right now is `Delta_xxx` to `delta_xxx` lower case
(right now we can use names that are close to Lutkepohl but eventually we want "informative" names. e.g. part of VAR cleanup #815 #3028 )

Given that the basic estimation is working, I think you should use a basic, stub (?) 

I read through the main code a bit later.

statsmodels/tsa/vecm/vecm.py
```diff
@@ -0,0 +1,254 @@
...
+import numpy as np
+import scipy
+
+# VECM class: for known or unknown VECM
+
+class VECM(tsbase.TimeSeriesModel):
+    r"""
+    Fit VECM process and do lag order selection
+
+    .. math:: y_t = A_1 y_{t-1} + \ldots + A_p y_{t-p} + u_t
```
bashtage: Math in notes, one liner to start class

statsmodels/tsa/vecm/vecm.py
```diff
@@ -0,0 +1,254 @@
...
+    Parameters
+    ----------
+    endog : array-like
+        2-d endogenous response variable. The independent variable.
+    dates : array-like
+        must match number of rows of endog
+
+    References
+    ----------
+    Lutkepohl (2005) New Introduction to Multiple Time Series Analysis
```
bashtage: Use RST markup for a bibliography entry.  Look in SM for examples.  Start with [1]

statsmodels/tsa/vecm/vecm.py
```diff
@@ -0,0 +1,254 @@
...
+    """#TODO: docstring + implementation
+    
+    # TODO: implementation
+    def __init__(self, endog, dates=None, freq=None, missing='none'):
+        super(VECM, self).__init__(endog, None, dates, freq, missing=missing)
+        if self.endog.ndim == 1:
```
bashtage: also to check shape[1] > 1 to avoid T by 1 arrays

statsmodels/tsa/vecm/vecm.py
```diff
@@ -0,0 +1,254 @@
...
+            raise ValueError("Only gave one variable to VECM")
+        self.y = self.endog #TODO delete this line if y not necessary
+        self.neqs = self.endog.shape[1]
+
+    def predict(self, params, start=None, end=None, lags=1, trend='c'):
+        """
+        Returns in-sample predictions or forecasts
+        """#TODO: docstring + implementation
```
bashtage: Comment after end of Docstring.  Also need Parameters, Returns, etc.

statsmodels/tsa/vecm/vecm.py
```diff
@@ -0,0 +1,254 @@
...
+            12 * (nobs/100.)**(1./4), see select_order function ##### ##### ##### #####
+        method : {'ls', 'egls', 'ml'}
+            Estimation method to use.
+        ic : {'aic', 'fpe', 'hqic', 'bic', None}
+            Information criterion to use for VECM order selection.
+            aic : Akaike
+            fpe : Final prediction error
+            hqic : Hannan-Quinn
+            bic : Bayesian a.k.a. Schwarz
+        deterministic terms, str {"", "c", "lt", "s"}
```
bashtage: : here.  Missing a _. Probably just call it `deterministic`

statsmodels/tsa/vecm/vecm.py
```diff
@@ -0,0 +1,254 @@
...
+            "" - no deterministic terms
```
bashtage: Maybe use `None` for none.

statsmodels/tsa/vecm/vecm.py
```diff
@@ -0,0 +1,254 @@
...
+            with intercept)
+
+        Notes
+        -----
+        Lutkepohl pp. 146-153
+
+        Returns
+        -------
+        est : VARResults
+        """#TODO: docstring + implementation
```
bashtage: Same here as above

statsmodels/tsa/vecm/vecm.py
```diff
@@ -0,0 +1,254 @@
...
+            else:
+                diff_lags = max_diff_lags
+        else:
+            selections = self.select_order(max_diff_lags=max_diff_lags, verbose=verbose)
+            if ic not in selections:
+                raise Exception("%s not recognized, must be among %s"
+                                % (ic, sorted(selections)))
+            diff_lags = selections[ic]
+            if verbose:
+                print('Using %d based on %s criterion' %  (diff_lags, ic))
```
bashtage: print is fine now, but must be removed eventually

statsmodels/tsa/vecm/vecm.py
```diff
@@ -0,0 +1,254 @@
...
+        elif method == 'egls':
+            if coint_rank is None:
+                coint_rank = 1
+            return self._estimate_vecm_egls(diff_lags, deterministic_terms, coint_rank)
+        elif method == 'ml':
+            if coint_rank is None:
+                coint_rank = 1
+            return self._estimate_vecm_ml(diff_lags, deterministic_terms, coint_rank)
+        else:
+            raise Exception("%s not recognized, must be among %s"
```
bashtage: ValueException. Never use Exception directly

statsmodels/tsa/vecm/vecm.py
```diff
@@ -0,0 +1,254 @@
...
+                seasons[i, i::4] = i+1
+            Delta_x = np.vstack((Delta_x, seasons))
+        
+        return y_all, y_1_T, Delta_y_all, Delta_y_1_T, y_minus1, Delta_x
+
+    def _ls_Pi_Gamma(self, Delta_y_1_T, y_minus1, Delta_x, deterministic_terms):
+        mat1 = np.hstack((Delta_y_1_T.dot(y_minus1.T), Delta_y_1_T.dot(Delta_x.T)))
+
+        B = y_minus1.dot(Delta_x.T)
+        mat2 = np.linalg.inv(np.vstack( (np.hstack( (y_minus1.dot(y_minus1.T),   B                    ) ),
```
bashtage: Need to follow PEP spacing guidelines (eventually)

statsmodels/tsa/vecm/vecm.py
```diff
@@ -0,0 +1,254 @@
...
+        print(Pi_hat)
+        print("Gamma:")
+        print(Gamma_hat)
+        print("Sigma_u")
+        print(Sigma_u_hat)
+        
+        return {"Pi_hat": Pi_hat, "Gamma_hat": Gamma_hat, "Sigma_u_hat": Sigma_u_hat}        
+
+    def _M_and_R_matrices(self, T, Delta_x, Delta_y_1_T, y_minus1):
+        M = np.identity(T) - Delta_x.T.dot(np.linalg.inv(Delta_x.dot(Delta_x.T))).dot(Delta_x)
```
bashtage: I prefer to import inv from numpy.linalg since it improves readability of matrix math.  your choice

statsmodels/tsa/vecm/vecm.py
```diff
@@ -0,0 +1,254 @@
...
+                coint_rank = 1
+            return self._estimate_vecm_egls(diff_lags, deterministic_terms, coint_rank)
+        elif method == 'ml':
+            if coint_rank is None:
+                coint_rank = 1
+            return self._estimate_vecm_ml(diff_lags, deterministic_terms, coint_rank)
+        else:
+            raise Exception("%s not recognized, must be among %s"
+                            % (method, ("ls", "egls", "ml")))
+
```
bashtage: No return here?

statsmodels/tsa/vecm/vecm.py
```diff
@@ -0,0 +1,254 @@
...
+        m_alpha_tilde = S01 * m_beta_tilde * (m_beta_tilde.T * S11 * m_beta_tilde).I
+
+        m_Delta_y_1_T = np.matrix(Delta_y_1_T)
+        m_y_minus1 = np.matrix(y_minus1)
+        m_Delta_x = np.matrix(Delta_x)
+        m_Gamma_tilde = (m_Delta_y_1_T - m_alpha_tilde*m_beta_tilde.T*m_y_minus1) * m_Delta_x.T * (m_Delta_x * m_Delta_x.T).I
+        m_temp = (m_Delta_y_1_T - m_alpha_tilde*m_beta_tilde.T*m_y_minus1 - m_Gamma_tilde*m_Delta_x)
+        m_Sigma_u_tilde = m_temp * m_temp.T / T
+        
+        return {"alpha":np.array(m_alpha_tilde), "beta":np.array(m_beta_tilde),
```
bashtage: Not sure about need to return dict.  Standard is to return tuple.  Mostly a personal preference.

bashtage:
A promising start.  Testing still needs quite a bit of work - don't have to hit 100% but >90 is important (e.g., often painful to test all raise statements)

statsmodels/tsa/vecm/vecm.py
yogabonito: Thanks! I have tried assigning 1 now but the test is still failing.

yogabonito: Thanks! I have tried assigning 1 now but the test is still failing.

statsmodels/tsa/vecm/vecm.py
yogabonito: Thanks! I will add _ for now and maybe rename it later.

statsmodels/tsa/vecm/vecm.py
yogabonito: Thank you! I have now replaced Exception with ValueError.

statsmodels/tsa/vecm/vecm.py
yogabonito: The are three different returns depending on which method for estimation shall be used (LS, EGLS, or ML).

yogabonito:
Thank you for all your valuable hints! I will do my best to make the code look accordingly.
@josef-pkt Apart from avoiding .dot() numpy matrices seemed to have another advantage for me. In my calculations I needed the square root of a matrix and didn't know how to achieve this with arrays. (To me it looks like I can take arrays only to the power of an integer and not 1/2.)
@bashtage I think that one reason for the decreased coverage is also that my tests only cover the ML method for estimation. (The JMulTi results look as if they were calculated this way. I will look for a way to change the calculation method in JMulTi.)

bashtage:
You can use svd to compute square root. If x=usv then square root x =
us^(0.5)v where s^0.5 is element by element.

statsmodels/tsa/vecm/vecm.py
josef-pkt: In ARMA and similar we call it "nc" for no constant.

(based on johansen cointegration tests, we might need 5 or 6 versions, depending on whether the constant, deterministic part is in the cointegrating vector or in the difference equation.)

josef-pkt:
@yogabonito  after you make a round of changes you need to push them to your branch on github, so we can see what the latest version is. (You mentioned code changes, but we cannot see them.)

statsmodels/tsa/vecm/vecm.py
josef-pkt: At the end of refactoring, there should be almost no `inv` left. I usually prefer the full path for the few occasions where we need them.

note M will have to go away completely because it's nobs x nobs, and be replace with a multivariate residual function. 
I think after unit test coverage is sufficiently high and tests pass, we can streamline the linear algebra from the textbook version to computationally more efficient versions.

josef-pkt: e.g. https://github.com/statsmodels/statsmodels/blob/master/statsmodels/stats/multivariate_tools.py#L26
is what I used for similar cases for one-time projection and residual/orthogonal projection.  (P and M)

josef-pkt:
cross-link #453 for johansen contains MLE for VECM in a similar way, but doesn't treat the `deterministic` in the same way.

yogabonito:
### Related issues
The issue corresponding to this PR is #3013 which includes a todo list.
Other Issues that might be related to this PR are:
#1581 (maybe helpful with not correctly implemented seasonal terms),
#942, #1101, #1451, #2340, #2344 (Granger causality test)
#448, #2315, #3009, #3024, (Johansen cointegration test)
#3025 (lag order selection)
#2710, #2933 (sparse matrices)
Also, the code for the orthogonal complement of a matrix may be helpful (#3039).

josef-pkt:
#3028 also belongs to that list. I only managed to get partway through reviewing those changes.

Optimally, VECM should be able to reuse large parts of VAR/vector_ar for the postestimation results, and for results conditional on the cointegrating vectors, however VAR is still missing the exog option. But there is also VARMAX in statespace, which can handle VARX but I've never used it so far and skipped most details of it in the review.

josef-pkt:
@yogabonito  you need to start with a results class. fit should return a results class that holds all the estimation results and has methods to connect to other results that we can get through VAR.

one refactoring in the VECM class: The pattern over time has changed and VAR is "out-of-date". `fit` should not have any arguments that change the underlying model. that is, the number of lags should be specified in `__init__`, `ic` and select order option needs to be added in a separate method and dropped from the `fit` method.

(I start to worry that we are too far behind schedule to finish up even the basic VECM in GSOC.)

statsmodels/tsa/vecm/vecm.py
```diff
@@ -0,0 +1,292 @@
...
+        elif method == "ml":
+            if coint_rank is None:
+                coint_rank = 1
+            return self._estimate_vecm_ml(diff_lags, deterministic,
+                                          coint_rank)
+        else:
+            raise ValueError("%s not recognized, must be among %s"
+                             % (method, ("ls", "egls", "ml")))
+
+    def _est_matrices(self, diff_lags, deterministic):
```
josef-pkt: I think this might have to become `initialize` called by `__init__`

statsmodels/tsa/vecm/vecm.py
```diff
@@ -0,0 +1,292 @@
...
+            seasons = np.zeros((3, delta_x.shape[1]))  # 3 = 4-1 (4=#seasons)
+            for i in range(seasons.shape[0]):
+                seasons[i, i::4] = 1
+            #seasons = seasons[:, ::-1]
+            #seasons = np.hstack((seasons[:, 3:4], seasons[:, :-1]))
+            #seasons = np.hstack((seasons[:, 2:4], seasons[:, :-2]))
+            seasons = np.hstack((seasons[:, 1:4], seasons[:, :-3]))
+            # seasons[1] = -seasons[1]
+            delta_x = vstack((delta_x,
+                              seasons))
```
josef-pkt: the `if` above need to be refactored, I think constant/trends should not be separately specified, i.e. use `elif` for different constant and trend specifications (Stata has 5).
repeated `vstack` will make several copies of the data, It's better to keep them in a list and then use one np.column_stack

josef-pkt:
@yogabonito   What constant/trend options does JMulti have? 
Stata has 5 corresponding to johansen's cointegration test, where the main difference is whether the constant or trend is in the cointegrating equation or in the short run diff equation, and whether they are implied given the link between differences and levels.

josef-pkt:
The results instance needs `llf`, the log-likelihood evaluated at the estimated parameters.
I'm not sure yet what kind of loglike should be added to the model, because there are several variants depending on what is concentrated out and which parameters are explicitly used.

josef-pkt:
I think we should use a separate `seasonal` argument instead of deterministic which combines trend and seasonal. Internally they can be combined to deterministic or similar.
Eventually we will also need an  `exog` argument, but I haven't gone through that literature yet, and I haven't seen it in Lutkepohl.

In terms of actual code it might be useful to distinguish deterministic in the short run (diff) equations from deterministic in the cointegration equations.

yogabonito:
Thank you Josef, for your input!
I have found out the following when I was trying to fix the estimation for seasonal components:
- After my latest commit the estimation results (constant+seasonal) match those presented in Lut p. 304. There, the constant term is outside of the coint. relation. The PR's code treats a constant the same way.
- The test case still fails because the estimation for the constant term doesn't match the JMulTi output.
- The JMulTi output equals that of the R package urca where the constant term is considered to be in the coint. relation. So it looks like JMulTi is treating constant terms and linear trends within the coint. relation. I haven't found a corresponding option neither in JMulTi nor in the package urca.

So I will now add an option for the constant to be in the cointegration relation and introduce a Results class.

josef-pkt:
about (weakly) exogenous variables: Lutkepohl has sections 10.2.4 and 10.3.2 for models exogenous variables. 
Adding stationary variables to the short-run/diff equation should be trivial. Additional exog in the cointegrating equations will create some more difficulty when estimating alpha, exogenity requires that alpha_x = 0, i.e. no feedback on the exog. Lutkepohl has some comments on rank of alpha and beta in this case which could be full rank, but I didn't try to figure out the details.

@yogabonito  I don't think we need to support exog in the first version, but in the estimation itself we can just combine the exog with the deterministic and include them in the same way, i.e. one part/array that holds both, or just added to the regressor matrix in the two parts, short-run and coint equations.

It looks like post-estimation and inference will have to distinguish between the different components and it will be more work to keep the effects separate. 
(the MacKinnon et al tables of critical values for johansen's cointegrating rank test include the case with exogenous variables, citing Pesaran for the theory and approach.)

statsmodels/tsa/vecm/vecm.py
```diff
@@ -0,0 +1,557 @@
...
+    y = endog_tot
+    K = y.shape[0]
+    y_1_T = y[:, p:]
+    T = y_1_T.shape[1]
+    delta_y = np.diff(y)
+    delta_y_1_T = delta_y[:, p-1:]
+    y_min1 = y[:, p-1:-1]
+    if "lt" in deterministic:
+        y_min1 = vstack((y_min1,
+                         np.arange(T)))  # p. 299
```
josef-pkt: I think it's better to use two deterministic options in the code (but maybe not in the general API), one for the cointegrating relationship here, i.e. y_min1 with no extra deterministic with constant or with trend. That's D_co in Lutkepohl
http://www.jmulti.de/download/help/vecm.pdf  From section 3.3, I assume the deterministic options refer to D_co, I don't see what specified what's in the difference equation, i.e. `D`

statsmodels/tsa/vecm/vecm.py
```diff
@@ -0,0 +1,557 @@
...
+    # p. 286:
+    delta_x = np.zeros((diff_lags*K, T))
+    for j in range(delta_x.shape[1]):
+        delta_x[:, j] = (delta_y[:, j+p-2:None if j-1<0 else j-1:-1]
+                         .T.reshape(K*(p-1)))
+    # p. 299, p. 303:
+    if "c" in deterministic:
+        delta_x = vstack((delta_x,
+                          np.ones(T)))
```
josef-pkt: this adds the second deterministic to the diff equation

josef-pkt: I think it's better to move the translation from deterministic options ("nc", "lt", ..) to the two deterministic arrays to a separate function and use here just premade arrays deterministic_co and deterministic_diff.
deterministic_diff can include the seasonal component additional to any possible constant or trend.

(This way all definition for user options, whether jmulti verstion or Stata version can be isolated to a single function.)

statsmodels/tsa/vecm/vecm.py
```diff
@@ -0,0 +1,557 @@
...
+from statsmodels.tsa.tsatools import duplication_matrix, vec, vech, unvec
+
+import statsmodels.tsa.base.tsa_model as tsbase
+
+def mat_sqrt(_2darray):
+    u_, s_, v_ = svd(_2darray, full_matrices=False)
+    s_ = np.sqrt(s_)
+    return chain_dot(u_, np.diag(s_), v_)
+
+def _endog_matrices(endog_tot, diff_lags, deterministic):
```
josef-pkt: add brief docstring to all functions
eventually these will need full docstrings, but can be expanded as the functions or methods and their signature settles down.

statsmodels/tsa/vector_ar/var_model.py
```diff
@@ -743,6 +743,16 @@ def forecast_interval(self, y, steps, alpha=0.05):
 
         return point_forecast, forc_lower, forc_upper
 
+    def to_vecm(self):
```
josef-pkt: we also need the reverse transformation `vecm_to_var`

josef-pkt:
AFAICS, fixing the deterministic option is the main problem right now, see my inline comments
Internally we should use two deterministic arrays, but users will get only a limited set of options which have full support, like the 5 cases in Stata.
(My mistake in johansen's cointegration test was to add more options for users than what is covered by the tables of critical values or p-values for the test.)

From your last comment, it sounds like jmulti supports a different constant from the one in the book. Both would be covered with the extended set of options in Stata, AFAICS.

statsmodels/tsa/vector_ar/var_model.py
yogabonito: That's already implemented in vecm.py (method var_repr of the VARResults class).

yogabonito:
> AFAICS, fixing the deterministic option is the main problem right now, see my inline comments
Internally we should use two deterministic arrays, but users will get only a limited set of options which have full support, like the 5 cases in Stata.
(My mistake in johansen's cointegration test was to add more options for users than what is covered by the tables of critical values or p-values for the test.)

> From your last comment, it sounds like jmulti supports a different constant from the one in the book. Both would be covered with the extended set of options in Stata, AFAICS.
Seems like I have just had a major breakthrough concerning the constant term inside the cointegration relation (reproduced estimated parameters from JMulTi (not pushed yet)). I still need to have a look at the other combinations of deterministic terms but I am positive that I can commit and push tomorrow something that works.

josef-pkt:
after reading a bit more in jmulti documentation
The `vecm.pdf` file has separate options for deterministic in diff and "EC" term in section 3.2 figure 6. I (center column and right panel). I haven't used jmulti and don't know which combinations are allowed to be chosen.
the `initanal.pdf`  shows that only three cases of deterministic constant and trend are supported by jmulti for the johansen cointegration test. 

also: http://www.statsmodels.org/devel/examples/notebooks/generated/statespace_varmax.html#Example-1:-VAR  has the VARX example implemented with statespace VARMAX. It looks like it has all the parts needed to run the diff VAR with fixed cointegrating equations `beta * Y_{t-1}` as exog, and for running the transformed VARX model in levels.

about exog:
the general model in equation 1 of vecm.pdf has exog only in the short term diff equation but not in the cointegrating equations.

yogabonito:
Unfortunately I have bad news. Yesterday I ran the tests for a constant deterministic term with [new code](https://gist.github.com/yogabonito/eb0f6e33ceca5470a436abfcd39921c5#file-constant_term-py). The results looked good but running the tests with seasonal deterministic terms again yielded the same wrong result for the constant parameter as before.
I have now downloaded three papers from Johansen but I have no access to his often cited book (cited also [here](http://de.mathworks.com/help/econ/introduction-to-cointegration-analysis.html) showing that Matlab works with the same options as Stata). I hope that I will have this implemented very soon.

josef-pkt:
make some smaller commits and push more often, so I can see where you are and what you are doing.
I don't see what's difficult in this part. The way you have it, you can just try the 3 or 5 combinations of trend and constant in deterministic_diff and in deterministic_coeq. 
Once that works you can add the seasonal dummies to deterministic_diff.

(I'm in Austria until Sunday, then I will be back in Montreal.)

yogabonito:
Good news today: the bug with seasonal dummies and a few other things are fixed now, all tests are passing (see commit). I hope that now the code will start to grow faster.
@josef-pkt, I can take the train to see you if you have time for me this Friday afternoon or evening.

josef-pkt:
> Good news today: the bug with seasonal dummies and a few other things are fixed now, all tests are passing (see commit). I hope that now the code will start to grow faster.

That's great. Can you briefly summarize which versions or combinations of constant and trend in coeq and diff are now unit tested. We will need a list or table of cases also for the documentation, similar to Stata or jmulti.


> I can take the train to see you if you have time for me this Friday afternoon or evening.

I doubt it's worth the train trip. I could arrange for a few hours, but the general plan for me and my familiy is either swimming pool or mountains depending on the weather.


the deterministic choice will also need to be a `__init__` option and not an argument in `fit`.
I will try to go over the latest changes tomorrow.

statsmodels/tsa/vecm/vecm.py
```diff
@@ -0,0 +1,699 @@
...
+        A tuple of five ndarrays as well as eigenvalues and -vecotrs of a
+        certain (matrix) product of some of the returned ndarrays.
+    """
+    T = y_min1.shape[1]
+    r0, r1 = _r_matrices(T, delta_x, delta_y_1_T, y_min1)
+    # p. 294: optimizable: e.g. r0.dot(r1.T) == r1.dot(r0.T).T ==> s01==s10.T
+    s00, s01, s10, s11 = (Ri.dot(Rj.T)/T for Ri in (r0, r1) for Rj in (r0, r1))
+    s11_ = inv(mat_sqrt(s11))
+    # p. 295:
+    eig = np.linalg.eig(chain_dot(s11_, s10, inv(s00), s01, s11_))
```
josef-pkt: `s11_.dot(s10)` is calculated twice, reuse by creating a temporary variable.

statsmodels/tsa/vecm/vecm.py
```diff
@@ -0,0 +1,699 @@
...
+    y = endog_tot
+    K = y.shape[0]
+    y_1_T = y[:, p:]
+    T = y_1_T.shape[1]
+    delta_y = np.diff(y)
+    delta_y_1_T = delta_y[:, p-1:]
+    y_min1 = y[:, p-1:-1]
+    # if "co" in deterministic and "cc" in deterministic:
+    #     raise ValueError("Both 'co' and 'cc' as deterministic terms given. " +
+    #                      "Please choose one of the two.")
```
josef-pkt: In my opinion we shouldn't raise any exceptions for these options, but we can add warnings if a combination is not fully supported or doesn't make much sense in general. 
Supported/unit tested combinations need to be listed in the docstrings

statsmodels/tsa/vecm/vecm.py
```diff
@@ -0,0 +1,699 @@
...
+                         np.ones(T)))
+        # H = vstack((np.identity(K),
+        #             - y_min1_mean))
+        # y_min1 = H.T.dot(y_min1)
+
+    # the following gives the same result as the R-package tsDyn when the
+    # linear term is outside the cointegration relation:
+    # if "lt" in deterministic:  # p. 299
+    #     y_min1 = vstack((y_min1,
+    #                      np.arange(T)))
```
josef-pkt: we need this case also where linear trend is in the cointegrating equations.
AFAICS, you only have it in the diff equation

josef-pkt:
as another reference: 
Gretl also has 5 cases for deterministic
http://gretl.sourceforge.net/gretl-help/gretl-guide.pdf page number 242 at end of section 27.3

bashtage:
Eventually you will need to do a doc build which will help you see if there are small typos in the docstring  These are hard to spot.  ALternatively you can paste your docstring into http://rst.ninjs.org/

For example, oen of your docstrings should be

```
Estimates the parameters of a VECM and returns a VECMResults object.

Parameters
----------
diff_lags : int
    Number of lags in the VEC representation
method : {"ls", "egls", "ml"}
    Estimation method to use.
deterministic : str {"", "co", "cc", "lt", "s"}
    * "" - no deterministic terms
    * "co" - constant outside the cointegration relation
    * "cc" - constant within the cointegration relation
    * "lt" - linear trend
    * "s" - seasonal terms

    Combinations of these are possible (e.g. "cclt" or "colt" for
    linear trend with intercept)
coint_rank : int
    Cointegration rank, equals the rank of the matrix \Pi and the number
    of columns of \alpha and \beta

Notes
-----
Lutkepohl pp. 269-304

Returns
-------
est : VECMResults
```

bashtage:
See http://rst.ninjs.org/?n=121481bdda54fdbba9dbacd98bae7f00&theme=nature

bashtage:
You might want to also start PEP8 checking your code.  

You can run `flake8 vecm.py` to see the issues.

```
./statsmodels/tsa/vecm/tests/results/parse_jmulti_output.py:42:1: W293 blank line contains whitespace
./statsmodels/tsa/vecm/tests/results/parse_jmulti_output.py:56:29: W291 trailing whitespace
./statsmodels/tsa/vecm/tests/results/parse_jmulti_output.py:102:80: E501 line too long (80 > 79 characters)
./statsmodels/tsa/vecm/tests/results/parse_jmulti_output.py:107:80: E501 line too long (81 > 79 characters)
./statsmodels/tsa/vecm/tests/results/parse_jmulti_output.py:193:80: E501 line too long (83 > 79 characters)
./statsmodels/tsa/vecm/vecm.py:9:1: F401 'unvec' imported but unused
./statsmodels/tsa/vecm/vecm.py:9:1: F401 'vec' imported but unused
./statsmodels/tsa/vecm/vecm.py:9:1: F401 'vech' imported but unused
./statsmodels/tsa/vecm/vecm.py:13:1: E302 expected 2 blank lines, found 1
./statsmodels/tsa/vecm/vecm.py:30:1: E302 expected 2 blank lines, found 1
./statsmodels/tsa/vecm/vecm.py:31:80: E501 line too long (80 > 79 characters)
./statsmodels/tsa/vecm/vecm.py:77:80: E501 line too long (80 > 79 characters)
./statsmodels/tsa/vecm/vecm.py:141:80: E501 line too long (80 > 79 characters)
./statsmodels/tsa/vecm/vecm.py:173:5: E303 too many blank lines (2)
./statsmodels/tsa/vecm/vecm.py:213:1: E302 expected 2 blank lines, found 1
./statsmodels/tsa/vecm/vecm.py:216:80: E501 line too long (111 > 79 characters)
./statsmodels/tsa/vecm/vecm.py:230:80: E501 line too long (83 > 79 characters)
./statsmodels/tsa/vecm/vecm.py:232:1: W293 blank line contains whitespace
./statsmodels/tsa/vecm/vecm.py:235:80: E501 line too long (81 > 79 characters)
./statsmodels/tsa/vecm/vecm.py:260:80: E501 line too long (80 > 79 characters)
./statsmodels/tsa/vecm/vecm.py:271:1: W293 blank line contains whitespace
./statsmodels/tsa/vecm/vecm.py:273:9: E303 too many blank lines (2)
./statsmodels/tsa/vecm/vecm.py:317:17: E126 continuation line over-indented for hanging indent
./statsmodels/tsa/vecm/vecm.py:324:1: W293 blank line contains whitespace
./statsmodels/tsa/vecm/vecm.py:327:17: E126 continuation line over-indented for hanging indent
./statsmodels/tsa/vecm/vecm.py:329:1: W293 blank line contains whitespace
./statsmodels/tsa/vecm/vecm.py:342:17: E126 continuation line over-indented for hanging indent
./statsmodels/tsa/vecm/vecm.py:347:80: E501 line too long (113 > 79 characters)
./statsmodels/tsa/vecm/vecm.py:348:1: W293 blank line contains whitespace
./statsmodels/tsa/vecm/vecm.py:349:23: W291 trailing whitespace
./statsmodels/tsa/vecm/vecm.py:351:1: W293 blank line contains whitespace
./statsmodels/tsa/vecm/vecm.py:352:55: W291 trailing whitespace
./statsmodels/tsa/vecm/vecm.py:354:1: W293 blank line contains whitespace
./statsmodels/tsa/vecm/vecm.py:357:17: E126 continuation line over-indented for hanging indent
./statsmodels/tsa/vecm/vecm.py:372:17: E126 continuation line over-indented for hanging indent
./statsmodels/tsa/vecm/vecm.py:377:23: E124 closing bracket does not match visual indentation
./statsmodels/tsa/vecm/vecm.py:384:80: E501 line too long (80 > 79 characters)
./statsmodels/tsa/vecm/vecm.py:399:80: E501 line too long (80 > 79 characters)
./statsmodels/tsa/vecm/vecm.py:402:1: E303 too many blank lines (3)
./statsmodels/tsa/vecm/vecm.py:514:14: E111 indentation is not a multiple of four
./statsmodels/tsa/vecm/vecm.py:527:47: E127 continuation line over-indented for visual indent
./statsmodels/tsa/vecm/vecm.py:544:80: E501 line too long (84 > 79 characters)
./statsmodels/tsa/vecm/vecm.py:549:5: E303 too many blank lines (3)
./statsmodels/tsa/vecm/vecm.py:556:80: E501 line too long (80 > 79 characters)
./statsmodels/tsa/vecm/vecm.py:579:30: E127 continuation line over-indented for visual indent
./statsmodels/tsa/vecm/vecm.py:585:32: E126 continuation line over-indented for hanging indent
./statsmodels/tsa/vecm/vecm.py:588:80: E501 line too long (80 > 79 characters)
./statsmodels/tsa/vecm/vecm.py:596:5: E301 expected 1 blank line, found 0
./statsmodels/tsa/vecm/vecm.py:614:1: W293 blank line contains whitespace
./statsmodels/tsa/vecm/vecm.py:624:5: E303 too many blank lines (3)
./statsmodels/tsa/vecm/vecm.py:651:1: W293 blank line contains whitespace
./statsmodels/tsa/vecm/vecm.py:652:5: E303 too many blank lines (2)
./statsmodels/tsa/vecm/vecm.py:654:80: E501 line too long (81 > 79 characters)
./statsmodels/tsa/vecm/vecm.py:668:80: E501 line too long (90 > 79 characters)
./statsmodels/tsa/vecm/vecm.py:672:80: E501 line too long (81 > 79 characters)
./statsmodels/tsa/vecm/vecm.py:678:80: E501 line too long (84 > 79 characters)
./statsmodels/tsa/vecm/vecm.py:682:5: E303 too many blank lines (3)
./statsmodels/tsa/vecm/vecm.py:697:13: E261 at least two spaces before inline comment
./statsmodels/tsa/vecm/vecm.py:699:75: W292 no newline at end of file
./statsmodels/tsa/vector_ar/var_model.py:757:80: E501 line too long (80 > 79 characters)
```

statsmodels/tsa/vecm/vecm.py
josef-pkt: AFAIU, we don't have any problems in the estimation of the parameters with different combination of deterministic terms. But we have only critical values for a limited set of cases, and some of the conversion to the implied VAR might become available only for some cases.
So either we use options that limit the choices of users to some fixed cases, or we allow "arbitrary" combination, but then limit or raise in the part that is not supported for those cases.

In either case, I would add this add after development and debugging is finished because any exceptions limits what we can try out during development.

(Another case: the econometrics articles and textbooks mention that an implied quadratic trend for the variables in levels doesn't make much sense in macro economics, but that doesn't rule out that they might be useful in some scientific fields or applications.)

josef-pkt:
This is starting to look good, two test failures/error one missing file, one with wrong sign.

What's next on your schedule/plan?

I will try to get an overview of the current status today (but then I will have to switch to maintenance/release work)

statsmodels/tsa/vecm/vecm.py
```diff
@@ -0,0 +1,703 @@
...
+    terms. A tuple of consisting of these matrices is returned.
+
+    Parameters
+    ----------
+    endog_tot : ndarray (neqs x total_nobs)
+        The whole sample including the presample.
+    diff_lags : int
+        Number of lags in the VEC representation.
+    deterministic : str {"", "co", "ci", "lo", "li"}
+            "" - no deterministic terms
```
josef-pkt: use `"nc"` instead of empty string for no constant which is our convention in ARIMA and other models

yogabonito:
Yesterday I have taken a day off, today I have started with forecasting. I hope that I can finish this topic by tomorrow. (todo: handle deterministic terms and calculate confidence intervals)

statsmodels/tsa/vecm/vecm.py
```diff
@@ -0,0 +1,704 @@
...
+        self.y = self.endog.T  # TODO delete this line if y not necessary
+        self.neqs = self.endog.shape[1]
+        self.p = diff_lags + 1
+        self.method = method
+        self.diff_lags = diff_lags
+        self.deterministic = deterministic
+        self.seasons = seasons
+        self.coint_rank = coint_rank
+
+    def fit(self):
```
josef-pkt: Things that don't change directly the underlying data/regressor arrays can stay in `fit`
`method` is for the fit and is a `fit` argument in all models
`coint_rank`  only affects the parameters and not the underlying regressors. so, I think it should be safe to keep it in fit.

statsmodels/tsa/vecm/vecm.py
```diff
@@ -0,0 +1,704 @@
...
+        K = self.K
+        A = np.zeros((self.p, K, K))
+        A[0] = pi + np.identity(K) + gamma[:, :K]
+        A[self.p-1] = - gamma[:, K*(self.p-2):]
+        for i in range(1, self.p-1):
+            A[i] = gamma[:, K*i:K*(i+1)] - gamma[:, K*(i-1):K*i]
+        return A
+
+    def predict(self, steps=5):
+        return forecast(self.y_all.T[-self.p:], self.var_repr, 0, steps)
```
josef-pkt: great that this can be ireused and works.

Does the current vector_ar VAR handle trends in forecast correctly, i.e. can we get the forecast with added trend or only for the pure VAR part?

josef-pkt:
If the transformed VEC to VAR is compatible with vector_ar.VAR as in your predict/forecast, then impulse response function (which ignore constant/trend) might also work almost directly.
I don't know if the cov_params of the transformed/implied VAR are necessary for any of the IRF, FEVD post-estimation results.

Other issue: I looked briefly at the `summary` of statespace VARMAX and statespace mlemodel.  I think this could be largely copied and adjusted for the parameter decomposition differences.
(That is the more recent pattern for summary, while vector_ar VAR has an old summary tables created from scratch, AFAIR.)

statsmodels/tsa/vecm/vecm.py
yogabonito: Ok, I will change that.

statsmodels/tsa/vecm/vecm.py
yogabonito: The forecast function in var_model.py handles constant terms correctly but lacks support for linear trends and seasonal terms. I am about to add this functionality to the method.

josef-pkt:
about VAR forecast
issue for missing trend is #142
fix is in https://github.com/statsmodels/statsmodels/pull/3028/files#diff-e2d6024b9d0ed6c0410bf2fffc1f1254L168

@yogabonito  I think you can use those changes, and we resolve merge conflicts between PR 3028 and your PR 3070 later

yogabonito:
> about VAR forecast
issue for missing trend is #142
fix is in https://github.com/statsmodels/statsmodels/pull/3028/files#diff-e2d6024b9d0ed6c0410bf2fffc1f1254L168

> @yogabonito I think you can use those changes, and we resolve merge conflicts between PR 3028 and your PR 3070 later

Thanks Josef for pointing this out!
I have copied the body of your VAR-forecast function from PR #3028. Now forecasting works also with deterministic terms.

josef-pkt:
@yogabonito I spend time last week on reading up on weakly exogenous variables, and I think it would be good to wire them into the model after the basic model and results are finished.

In the first case, I would just support them for estimating the model, and add them essentially the same way as the current `deterministic`, in both the cointegrating equation and in the short run diff part. I think for most parts of the implementation deterministic and exog can just be treated together, without much extra work. The user interface will have to handle the additional variables.
Other `deterministic` like dummy variables or structural break could then be included in the same way.

(Some things won't be supported for those extras. E.g. EViews mentions that other deterministic are ignored for the critical values in cointegration rank testing.)

josef-pkt:
corner case: no lags of diff in VECM representation, i.e. order of VAR is 1.
Does this case work? unit tests for this case ?
If there is no constant in the difference equation, then the conditioning short run matrix would be empty.
?

josef-pkt:
@yogabonito  put the explanation and examples in the PR comments instead of the commit message. The last commit messages look pretty large. Most of the time we only see the first line of the commit message, while it's relatively easy to browse the comments for explanation and discussion.
(both get a bit messed up in a rebase with squashing and force pushing)

josef-pkt:
looks good, from the outside, but it will take some time for me to catch up with the details.

yogabonito:
Here is a [Jupyter notebook](https://gist.github.com/yogabonito/5461b26bed335cad6907aa4e613acb99#file-gsoc-vecm-ipynb) showing what the VECM-framework offers so far. I will keep working on it, adding the missing VECM-related functionality and refactoring the existing VAR-code to make it work with arbitrary deterministic terms.
Now that the GSoC is officially ending, I want to say thank you @josef-pkt and @bashtage for being my mentors, your advice has been invaluable!

josef-pkt:
@yogabonito  thanks for the successful GSOC. 

If you have time, one more high priority enhancement would be the simplest case of constrained estimation.
http://dare.uva.nl/search?field1=dai;value1=087095351;docsPerPage=1;startDoc=9
p.9 equation (10) to (12) with recursion in corollary 1 could be added to a method that returns the results with constrained parameter estimates.
The constraints are imposed on the alpha and beta separately by reparameterization, equ. (8) and (9) on page 8 above.

The calculation don't look very difficult given that you already have all the pieces. 

(I went through this during summer, not right now. I'm trying to get organized again.)

yogabonito:
Thanks @josef-pkt, it's great to hear that you are satisfied! :)
I will definitely have a look at it after the other topics are implemented.

statsmodels/tsa/vector_ar/var_model.py
```diff
@@ -1287,17 +1461,20 @@ def reorder(self, order):
 #-------------------------------------------------------------------------------
 # VAR Diagnostics: Granger-causality, whiteness of residuals, normality, etc.
 
-    def test_causality(self, equation, variables, kind='f', signif=0.05,
-                       verbose=True):
-        """Compute test statistic for null hypothesis of Granger-noncausality,
-        general function to test joint Granger-causality of multiple variables
+    def test_causality(self, causing, kind='f', signif=0.05, verbose=True):
```
josef-pkt: We cannot change the API, call signature and returns, in a backwards incompatible way without deprecation, in old/established code.

yogabonito: Thanks, Josef, for pointing this out!
How does the deprecation process work? Because I really feel like the API should be changed in future.
Here is a [Jupyter-notebook](https://gist.github.com/yogabonito/eedd44be9b8a76de7773a1e43f961d45) showing that specifying `causing` _and_ `caused` may give unexpected results - more precisely: if the union of these arguments is not the whole system of variables (this is the case in section 1 of the notebook and the results differ from those in section 2 although the causality test is done using the same variables).
[Note: the notebook was produced using the latest version of var_model.py before I changed the file for the first time - so the method's signature includes both `causing` and `caused` (called `variables` and `equation` in the old code)]

josef-pkt: I haven't looked at the details yet. (the behavior of this methods confused me several years ago.)

specifically: I think we need both causing and caused variables in the method. The remaining variables are conditioned upon.

in general: 
If we can maintain the same signature but return different results or have different keywords, then we raise a deprecation warning in the old use. 
If we cannot support and distinguish old and new cases, then it would be better to deprecate the entire method and come up with a new name for the replacement.

josef-pkt: Which is causing and which is cause in the old signature, equation or variables? The old docstring is not very informative.

I think one possibility is to allow caused to be None, and default to what you have right now, i.e. causing on all other  variables.. AFAICS, allowing None as default wouldn't break backwards compatibility.

josef-pkt: to answer my question: `equation` is the equation of the VAR, i.e. the dependent variable, or `caused` in this context. `variable` is then the `causing` variable. (i.e. wald/F test that `variable` in `equation` is insignificant.)

I don't think there is a way to switch the positional arguments in a backwards compatible way.

yogabonito: > i.e. wald/F test that variable in equation is insignificant.

That's right.
The problem I have with the current situation is that the results in section 1 of the [notebook](https://gist.github.com/yogabonito/eedd44be9b8a76de7773a1e43f961d45) I provided are not the same as those in section 2, because in the former there is one variable which is neither contained in `causing` nor in `caused`.

Also, Lutkepohl defines Granger-causality only in the context of a partition of `y` into `causing` (`x` in Lutkepohl) and `caused` (`z` in Lutkepohl) with no remaining variables. Probably this is why also JMulTi is offering to specify only one set of variables and not both `causing` and `caused`.

I think that it would make sense to make a new the method called e.g. `test_granger_causality` such that a new signature is possible and to avoid confusion between `test_causality` and `test_inst_causality` (which I plan to move to var_model.py as added VAR-feature).

josef-pkt: > The problem I have with the current situation is that the results in section 1 of the notebook I provided are not the same as those in section 2, because in the former there is one variable which is neither contained in causing nor in caused.

we need that, those test two different null hypothesis: With a left out variable, we test whether the causing variable has and additional granger causal effect.  It's a partial effect conditioning on other regressors, similar to a t-test or t-value for a single parameter.

Currently, we don't have a model for that yet, because the regressors are assumed to be the same in all `equations`, but we will need an extension where we can drop selectively irrelevant parameters or variables from just some equations. i.e. we need to be able to put zero constraints or other linear constraints on the full short-run parameter array.

I don't remember where I have seen it. Stata's vargranger only makes tests for all equations separately, i.e. all pairwise granger causality hypothesis conditional on other regressors and for `all` other variables in one equation, if I read the documentation correctly

josef-pkt:
discussion for granger causality test is inline, which might be more difficult to find again later.

We also need an option to choose the degrees of freedom calculation. I was looking for the mailing list discussion we had several years ago, and there are different choices for the degrees of freedom used in the tests. Some (?) package mentioned that they default to single equation degrees of freedom when there are no cross-equation parameter restrictions and to system degrees of freedom if there are parameter restrictions.

It might be good to add a keyword for future compatibility, even if there is no second option right now. I haven't checked what other packages are doing for granger causality.

josef-pkt:
> Some (?) package mentioned that they default to single equation degrees of freedom when there are no cross-equation parameter restrictions and to system degrees of freedom if there are parameter restrictions.

IIRC, this was in the system of equations context, maybe reg3 in Stata, but should apply in a similar way for VAR. (sysreg #2500 has several `dfk` options)

yogabonito:
> we need that, those test two different null hypothesis: With a left out variable, we test whether the causing variable has and additional granger causal effect. It's a partial effect conditioning on other regressors, similar to a t-test or t-value for a single parameter.

OK, I did not know about that since I haven't found such a definition in Lutkepohl. I will bring back the `caused` parameter and work on Granger-causality after I have refactored VAR (adding `exog`), which I plan for this week.

statsmodels/tsa/vector_ar/irf.py
```diff
@@ -26,7 +26,8 @@ class BaseIRAnalysis(object):
     able to handle known and estimated processes
     """
 
-    def __init__(self, model, P=None, periods=10, order=None, svar=False):
+    def __init__(self, model, P=None, periods=10, order=None, svar=False,
```
bashtage: Are you stuck with `P`?  `p` is a better choice.

statsmodels/tsa/vector_ar/irf.py
```diff
@@ -26,7 +26,8 @@ class BaseIRAnalysis(object):
     able to handle known and estimated processes
     """
```
bashtage: Missing a blank line

statsmodels/tsa/vector_ar/irf.py
```diff
@@ -233,11 +239,15 @@ class IRAnalysis(BaseIRAnalysis):
     -----
     Using Lutkepohl (2005) notation
     """
-    def __init__(self, model, P=None, periods=10, order=None, svar=False):
+    def __init__(self, model, P=None, periods=10, order=None, svar=False,
+                 vecm=False):
         BaseIRAnalysis.__init__(self, model, P=P, periods=periods,
```
bashtage: Same for `P` here

statsmodels/tsa/vector_ar/irf.py
```diff
@@ -381,7 +391,7 @@ def err_band_sz2(self, orth=False, repl=1000, signif=0.05,
         else:
             irfs = self.irfs
         neqs = self.neqs
-        irf_resim = model.irf_resim(orth=orth, repl=repl, T=periods, seed=seed,
+        irf_resim = model.irf_resim(orth=orth, repl=repl, T=periods, seed=seed,  # TODO: irf_resim relies on self.intercept --> solve with exog-refactor
```
bashtage: Comment should be above -- line too long

statsmodels/tsa/vector_ar/irf.py
```diff
@@ -381,7 +391,7 @@ def err_band_sz2(self, orth=False, repl=1000, signif=0.05,
...
                                    burn=100)
```
bashtage: `burn` underindented  by 1

statsmodels/tsa/vector_ar/irf.py
```diff
@@ -449,7 +459,7 @@ def err_band_sz3(self, orth=False, repl=1000, signif=0.05,
         else:
             irfs = self.irfs
         neqs = self.neqs
-        irf_resim = model.irf_resim(orth=orth, repl=repl, T=periods, seed=seed,
+        irf_resim = model.irf_resim(orth=orth, repl=repl, T=periods, seed=seed,  # TODO: irf_resim relies on self.intercept --> solve with exog-refactor
```
bashtage: Line too long

statsmodels/tsa/vector_ar/irf.py
```diff
@@ -631,7 +641,7 @@ def cum_errband_mc(self, orth=False, repl=1000,
         """
         model = self.model
         periods = self.periods
-        return model.irf_errband_mc(orth=orth, repl=repl,
+        return model.irf_errband_mc(orth=orth, repl=repl,  # TODO: irf_errband_mc relies on self.intercept --> solve with exog-refactor
```
bashtage: Line too long

bashtage:
`ipf.py` could really use a PEP clean.  There are lots of places that spaces are missing, there are `== None` comparisons that should be `is None` and there are line like ` == False` that should be simplified.

josef-pkt:
However, don't but PEP-8 cleaning in a large PR like this. It will be almost impossible to figure out what changed. And there would be even more merge conflicts with #3028

bashtage:
> However, don't but PEP-8 cleaning in a large PR like this. It will be almost impossible to figure out what changed. And there would be even more merge conflicts with #3028

It desperately needs it IMO, and it should be done before this project is completed, maybe as a last step before merge.

yogabonito:
I also think that irf.py could use some PEP-8 cleaning but I didn't write the file (I only fixed two bugs and adapted it to work with VECMs too). I thought that I shouldn't otherwise touch the files as Kevin mentioned:
> I often use the reformat code as a good starting point for producing clean code.  One warning - don't do this to existing files as many statsmodels files far from being PEP8 compliant, and so you will make a lot of diffs on code that you aren't really changing.

But of course I can fix the coding style issues if this is desired.

josef-pkt:
@bashtage If we need code cleanup, then it would be better in extra PRs when there are no other PRs changing the same code. If pep-8 cleanup is included, then it makes reviewing the final changes **very** difficult, because I have to check every line of whether it is a cleanup or whether it might introduce a backwards incompatible change.

(PRs that combine style cleanup, partially incompatible refactoring, bug fixes and enhancements are prime candidates for stalling for years because of the lack of reviewers. VAR cleanup and my robust cleanups are examples.)

We still need a lot of code cleanup, but it has to be "digestible" and worth going through several hundred or thousands of lines of changes.
(partially related: https://github.com/scipy/scipy/pull/6294#issuecomment-232168269 )

yogabonito:
I have now started to introduce `exog` into VAR (e.g. to make seasonal terms possible) and I can already estimate parameters for `endog` and `exog` correctly when seasonal terms are assumed as `exog`.

Now I wanted to do the same for the corresponding standard errors, etc. and see that things will become messy. The reason is that in current VAR estimates/std_err/... often look like this: `est_trend|est_endog` (column- or row-stacked). If I add `exog` it will look like `est_trend|est_endog|est_exog` - so trend and exog are disconnected although both may hold deterministic terms.
However rewriting the code such that `est_endog|est_exog` (with `trend` as part of `exog`) will introduce many conflicts with #3028. (Also, `exog` alone (and dealing with the disconnected `trend`- and `exog`-estimates) would lead to a lot of conflicts.)

That's why I am thinking about making a new, clean class called `VARX` instead where we might also incorporate restrictions as in Lütkepohl's chapter 10.3.1.
@josef-pkt, what is your opinion?

josef-pkt:
I think to follow the current pattern, exog should be in front of lagged endog
`est_trend|est_exog|est_endog` (or maybe better `est_exog|est_trend|est_endog`
or trend subsumed under exog
`est_exog|est_endog`

(That makes slicing easier, and most likely dropping lags and searching for number of lags computationally more efficient because we can just cut away regressors from the end of the array.)

I think it should be possible to just replace every occurrence of `trend` in the methods (except in `__init__`) by the more general `exog`, and we add trend internally to exog. (AFAIR, that's also how ARMA works)

I'm very skeptical about starting a new VAR class now because it requires too many design decisions and for general linear restrictions on parameters we need a version with stacked endog and exog/endog_lagged and nonlinear optimization. We should be able to merge the VECM class relatively soon and adding a new model class, VARX, would add a lot more work to this PR.

yogabonito:
Thanks for your feedback, Josef!
In the meantime I have finished `test_normality` in `VECMResults`. Tomorrow I will continue my work on VAR with your suggestions in mind.

statsmodels/tsa/vector_ar/var_model.py
```diff
@@ -601,13 +620,23 @@ def fit(self, maxlags=None, method='ols', ic=None, trend='c',
...
+        self.nobs = self.n_totobs - lags
+
+        # add exog to data.xnames (necessary because the length of xnames also
+        # determines the allowed size of VARResults.params)
+        if self.exog is not None:
+            x_names_to_add = [("exog%d" % i)
+                              for i in range(self.exog.shape[1])]
+            self.data.xnames = self.data.xnames[:k_trend] + \
+                               x_names_to_add + \
+                               self.data.xnames[k_trend:]
```
josef-pkt: We had problems/bugs with this pattern before, in ARMA(X) IIRC.
If `fit` is called repeatedly, then this will insert the `x_names_to_add` each time.
Can this be done in `__init__` instead?

statsmodels/tsa/vector_ar/var_model.py
```diff
@@ -620,11 +649,31 @@ def _estimate_var(self, lags, offset=0, trend='c'):
...
+            z = np.empty((x.shape[0], x.shape[1]+z.shape[1]))
+            z[:, :self.k_trend] = temp_z[:, :self.k_trend]
+            z[:, self.k_trend:self.k_trend+x.shape[1]] = x
+            z[:, self.k_trend+x.shape[1]:] = temp_z[:, self.k_trend:]
+            del temp_z, x  # free memory
+        # the following modification of z is necessary to get the same results
+        # as JMulTi for the constant-term-parameter...
+        for i in range(self.k_trend):
+            if (np.diff(z[:, i]) == 1).all():  # modify the trend-column
+                z[:, i] += lags
```
josef-pkt: this sounds like we need an offset in the trend creation, If the trend is quadratic, then this would affect only the linear term.

statsmodels/tsa/vector_ar/var_model.py
```diff
@@ -840,7 +891,31 @@ def forecast(self, y, steps):
...
+        if self.exog is not None and exog_future is None:
+            raise ValueError("Please provide an exog_future argument to "
+                             "the forecast method.")
+        trend_coefs = None if self.coefs_exog.size == 0 else self.coefs_exog.T
+
+        exogs = []
+        if self.trend.startswith("c"):  # constant term
+            exogs.append(np.ones(steps))
+        exog_lin_trend = np.arange(self.n_totobs + 1,
+                                   self.n_totobs + 1 + steps)
```
josef-pkt: I'm not sure: above the trend was adjusted/shifted by lags, Is self.n_totobs the last linear trend value in the estimation regressor?

josef-pkt:
just a partial look/review, it will take me more time to get through the details.

(github changed the review process, and I don't see whether we can just add comments.)

yogabonito:
This long list of commits suggests that something went wrong when trying to resolve the conflict. I have also a rebased branch (vecm-rebased) which is cleaner (only 55 instead of 112 commits but both branches contain the same files).
Is there a way to base this PR on the vecm-rebased branch?

statsmodels/tsa/vector_ar/var_model.py
yogabonito: The quadratic trend "ctt" is still on my todo-list. Here I tried to make test cases pass and there are no test cases for "ctt" because JMulTi does not provide this kind of trend.

statsmodels/tsa/vector_ar/var_model.py
yogabonito: Thanks, that's a good point. I will change that!

statsmodels/tsa/vector_ar/var_model.py
yogabonito: I would say yes.
Without the adjustment in line 674 the trend-part would be `1, 2, ..., nobs`.
Now, with this adjustment it is `k_ar+1, k_ar+2, ..., n_totobs`.

josef-pkt:
@yogabonito  Given that your rebased branch looks clean, you could continue with that one and open a new pull request.

According to the network only the last commit merges the original branch into the rebased version, but  the comment/commit sequence in here still shows the original sequence besides the last merge commit. According to git only the last merge commit would need to be removed, but AFAIK that would change the github comment/merge sequence. (I'm not quite sure how this matches up.)

josef-pkt:
continuation of this PR in #3246 has now been merged

tongkewu:
Hi, I cannot find vecm in the current vector_ar class. Is it removed? Thanks!

josef-pkt:
@tongkewu  It has been merged, and is in master but not in any release yet


examples
Here https://gist.github.com/josef-pkt/fda96edd7b15892a058f58ae0948ed71
is an updated version of yogabonito's notebook (updated at time of merge, but it should be still up-to-date with master)

(there is also another gist that I used for merge review https://gist.github.com/josef-pkt/b36cd8458ab2ed20ff784dc8ed3d3a1c looks messy and I don't remember.)

tongkewu:
Thank you! So if I want to implement the vecm alg, I need to pull th whole
git source?

josef-pkt:
Yes, If you want to use VECM, then you need to build from source or wait for the next release.
VECM itself is just one module, but it includes necessary changes and enhancements to VAR

tongkewu:
Got it. I really apreciate your work and I believe it will help a lot with
my data project. Thank you!

pskharade:
any timeline for the next release ( with VECM)

PR #3071, opened 2016-06-23
TST: Change appveyor clone depth

bashtage:
Set depth to 50

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,5 @@
 skip_tags: true
-clone_depth: 1
+clone_depth: 50
 
 os: Visual Studio 2015
 


bashtage:
@josef-pkt This will change the clone depth to 50 for appveyor

josef-pkt:
looks fine to me
just a question: AFAICS, we have the clone depth in the settings on appveyor (which was 10 and I changed to 50). But the clone depth in the `appveyor.yml` overrides this.
Is there an advantage of sticking to either version? 
Can we just delete the line in appveyor.yml and rely on the appveyor website setting.

bashtage:
The .yml way means you never have to mess with the web interface and
changes to appveyor are just like changes to travis, which seems cleaner to
me.

josef-pkt:
ok, I don't have a preference

I'll merge tomorrow. 
I want to see if I can first merge the maintenance commits to master to bring master in line with maintenance.

josef-pkt:
merged

@bashtage Thank you

josef-pkt:
backport if appveyor doesn't work in maintenance

PR #3073, opened 2016-06-23
TST: Add tests for datasets

bashtage:
Add tests for data sets
Fix dataset type error in fertility data


bashtage:
Probably one new issue in one of the notebooks.  I think it uses .load but
expects pandas. This is the fixed cigarette data set

PR #3112, opened 2016-07-15
PERF: Use einsum instead of broadcasting in PHReg hessian calculations

kshedden:
The broadcasting approach creates all the outer products first then reduces them by summation.  For large datasets we encountered memory errors when there wasn't enough room to store the broadcasted matrix.

The einsum approach resolves the memory problem, but I expected it to be slower for small data sets.  But actually it seems to be about 2x faster, so win-win, at least in the checking I have done.

There is possible further improvement using the `out` parameter of einsum, but I will not get to that now.

(edited for clarity)

### DIFF
--- a/statsmodels/duration/hazard_regression.py
+++ b/statsmodels/duration/hazard_regression.py
@@ -799,9 +799,10 @@ class PHReg(model.LikelihoodModel):
                     xp0 += e_linpred[ix].sum()
                     v = exog_s[ix,:]
                     xp1 += (e_linpred[ix][:,None] * v).sum(0)
-                    mat = v[None,:,:]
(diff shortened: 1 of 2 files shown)


josef-pkt:
@kshedden  Is this safe to quickly merge for 0.8?
I don't see any problem, but I used einsum so far only in a few toy examples.

kshedden:
Yes, I think it can be merged.  I can rebase if needed.

Kerby

josef-pkt:
rebase would be useful, this branch is a bit far behind.

kshedden:
Done, travis running now...

josef-pkt:
merged

@kshedden Thanks

PR #3113, opened 2016-07-16
BUG: TSA models if date index without known frequency

### LINKED ISSUE #3068: BUG: TSA models if date index without known frequency
If given a Pandas `endog`, TSA models require the index to contain dates. If the date index does not have a frequency, an attempt is made to infer it, and it appears that the intent is to raise an error if this inference fails (`statsmodels/tsa/base/tsa_model.py:56`):

```python
if not freq:
    try:
        freq = datetools._infer_freq(dates)
    except:
        raise ValueError("Frequency inference failed. Use `freq` "
                "keyword.")
```

However, `_infer_freq` doesn't usually raise an exception (from a brief look, it only ever raises an exception if the length of the series is less than 3), but it will often return `None` if the frequency can't be inferred.

I'm guessing we want to raise an exception here if the inferred frequency is None. I'll make a PR to add that behavior, unless I've misunderstood.

ChadFulton:
Closes #3068.

Puts a shim in to allow correctly creating the `DatetimeIndex` in the nanosecond case.

### DIFF
--- a/statsmodels/tsa/base/tsa_model.py
+++ b/statsmodels/tsa/base/tsa_model.py
@@ -73,8 +73,9 @@ class TimeSeriesModel(base.LikelihoodModel):
 
         # Test for nanoseconds in early pandas versions
         if freq is not None and _freq_to_pandas[freq].freqstr == 'N':
+            from distutils.version import LooseVersion
             from pandas import __version__ as pd_version
-            if pd_version < '0.14':
+            if LooseVersion(pd_version) < '0.14':
                 raise NotImplementedError('Nanosecond index not available in'
                                           ' Pandas < 0.14')
 
--- a/statsmodels/tsa/base/tsa_model.py
+++ b/statsmodels/tsa/base/tsa_model.py
@@ -71,6 +71,14 @@ class TimeSeriesModel(base.LikelihoodModel):
         self.data.dates = dates
         self.data.freq = freq
 
+        # Test for nanoseconds in early pandas versions
+        if freq is not None and _freq_to_pandas[freq].freqstr == 'N':
+            from pandas import version
+            if version.version < '0.14':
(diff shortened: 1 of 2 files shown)


ChadFulton:
The underlying issue is now fixed in Pandas (https://github.com/pydata/pandas/issues/13672), but the shim is probably still necessary.

statsmodels/tsa/base/tsa_model.py
```diff
@@ -71,6 +71,14 @@ def _init_dates(self, dates, freq):
         self.data.dates = dates
         self.data.freq = freq
 
+        # Test for nanoseconds in early pandas versions
+        if _freq_to_pandas[freq].freqstr == 'N':
+            from pandas import version
+            if version.version < '0.14':
+                raise NotImplementedError('Nanosecond index not available in'
+                                          ' Pandas < 0.14')
```
ChadFulton: This looks kind of like a hack, but I have not been able to find a simple way to test for the presence of the nanosecond bug itself. if anyone has a better way, that would be great.

ChadFulton:
I think this is going to come back green (and backport version too hopefully), in which case I'm ready to merge.

Note: coverage has to decrease here, because some of the new lines are not accessible when pandas < 0.14, and the coverage machine is pandas == 0.13.

It would be good for someone else to take a quick look (it's only a few lines of code) to make sure that I'm not relying on something (like `pd.__version__`) that I shouldn't be relying on. @josef-pkt @kshedden ?

ChadFulton:
I changed the code to use `distutils.version.LooseVersion`, so now it mirrors pandas' usage of version checking. Since there are no concerns, I will merge this and the backport. Nanoseconds should either work now where supported, or else give a meaningful error message.

PR #3120, opened 2016-07-19
Key defined twice

N-Wouda:
Nothing big, this exact key is defined about ten lines down in the same dict.

Noticed this while browsing for a good place to insert the BoxCox mixin.


ChadFulton:
Thanks for catching this and for the PR!

PR #3126, opened 2016-07-25
ENH: entry times for Kaplan-Meier

kshedden:
This PR adds entry times for `SurvfuncRight` (Kaplan-Meier estimation of the survival function).

Tests against R.


kshedden:
Also adds support to allow entry times for logrank testing (`survdiff`), added some tests against Stata.

PR #3175, opened 2016-08-20
BUG: Backport: TSA models if date index without known frequency

ChadFulton:
Backport of #3068.

### DIFF
--- a/statsmodels/tsa/base/tests/test_base.py
+++ b/statsmodels/tsa/base/tests/test_base.py
@@ -1,7 +1,7 @@
 import numpy as np
 import numpy.testing as npt
 import pandas as pd
-
+from statsmodels.tsa.base.datetools import _freq_to_pandas
 from statsmodels.tsa.base.tsa_model import TimeSeriesModel
 from statsmodels.tools.testing import assert_equal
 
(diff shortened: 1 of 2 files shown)


ChadFulton:
Failures on Pandas == 0.13.0. I think we should not support nanoseconds if Pandas < 0.14; there are too many bugs still in that version. For example:

```
>>> print(pd.to_datetime(100))
1970-01-01 00:00:00
>>> print(pd.to_datetime(1000))
1970-01-01 00:00:00.000001
>>> print(pd.to_datetime(100).to_period('N'))
1970-01-01 00:00:00.000000000
>>> print(pd.to_datetime(999).to_period('N'))
1970-01-01 00:00:00.000000000
>>> print(pd.to_datetime(1000).to_period('N'))
1970-01-01 00:00:00.000001000
```

and more importantly for us, the end result is that we can't create the extended index for nanosecond frequency because of this weird behavior:

```
>>> x = pd.to_datetime(np.arange(10))
>>> pd.DatetimeIndex(start=x[0], end=x[9], freq=pd.offsets.Nano())
<class 'pandas.tseries.index.DatetimeIndex'>
[1970-01-01 00:00:00]
Length: 1, Freq: N, Timezone: None
```

ChadFulton:
Somehow I messed up this backport PR, but I have re-created it and if it comes back green I will merge.

PR #3184, opened 2016-09-01
Fix small typo in error message

fisadev:
Just that, nothing complicated.


ChadFulton:
Thanks for the PR! I was thinking that it was meant to say "Please report if you think this **is in error**", rather than "**is an error**".

fisadev:
Oh, you are probably right (not a native english speaker, I didn't think of that option). Do you want me to change it to that?

ChadFulton:
That would be great if you could. Thanks again!

fisadev:
Done!

ChadFulton:
Thanks!

PR #3195, opened 2016-09-11
MAINT: delay numpy import during build

matthew-brett:
Delay numpy import until we build the extensions.  This allows commands
like `egg_info` to collect dependencies without crashing out with a
missing numpy.

Therefore, pip should be able to install numpy during the build process.

See: https://github.com/scipy/scipy.org/pull/179 for discussion.


bashtage:
@josef-pkt Aside from needing a rebase, this seems ready.

josef-pkt:
I thought it failed in several travis version where statespace cython code didn't find anymore some numpy math functions. But I don't see the travis links here anymore.
If it works and is green after rebase, then we can merge

bashtage:
https://travis-ci.org/statsmodels/statsmodels/builds/159058428

josef-pkt:
python 3.6 failure is unrelated
but this errors with missing cython symbol 
$ export PYTHON=3.3
$ export NUMPY=1.9
$ export SCIPY=0.14
$ export MATPLOTLIB=1.4
$ export PANDAS=0.14

matthew-brett:
Yes, I don't understand the logic for this code in `setup.py`:

```
try:
    from scipy.linalg import cython_blas
    ext_data.update(statespace_ext_data)
except ImportError:
    for name, data in statespace_ext_data.items():
        path = '.'.join([data["name"].split('.')[0], 'pyx.in'])
        append_cython_exclusion(path, CYTHON_EXCLUSION_FILE)
```

josef-pkt:
@ChadFulton  knows the details and I never looked at them.

The general problem is that we are doing currently compilation conditional on the scipy version. Some parts of the code take advantage of more recent access to the underlying linear algebra libraries  in scipy. But we still support the older scipy versions that don't have these enhancements.

ChadFulton:
> @ChadFulton knows the details and I never looked at them.

> The general problem is that we are doing currently compilation conditional on the scipy version. Some parts of the code take advantage of more recent access to the underlying linear algebra libraries in scipy. But we still support the older scipy versions that don't have these enhancements.

Exactly. But I'm sure that the setup.py could be designed better, I just don't know how :)

bashtage:
An alternative would be to move min SciPy to 0.16.  Then there is no more conditional compilation.

matthew-brett:
Updating the dependency to 0.16 would remove quite a lot of complexity here.  For example, I noticed that, if you compile statsmodels with scipy < 0.16, and later install scipy >= 0.16, statsmodels crashes, because it thinks that the extra modules are compiled and tries to import them:

```
ERROR: Failure: ImportError (cannot import name '_representation')
  File "/home/mb312/.local/lib/python3.5/site-packages/nose/failure.py", line 39, in runTest
    raise self.exc_val.with_traceback(self.tb)
  File "/home/mb312/.local/lib/python3.5/site-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/home/mb312/.local/lib/python3.5/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/home/mb312/.local/lib/python3.5/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/usr/lib/python3.5/imp.py", line 234, in load_module
    return load_source(name, filename, file)
  File "/usr/lib/python3.5/imp.py", line 172, in load_source
    module = _load(spec)
  File "<frozen importlib._bootstrap>", line 693, in _load
  File "<frozen importlib._bootstrap>", line 673, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 673, in exec_module
  File "<frozen importlib._bootstrap>", line 222, in _call_with_frames_removed
  File "/home/mb312/dev_trees/statsmodels/statsmodels/tsa/statespace/tests/test_collapsed.py", line 17, in <module>
    from statsmodels.tsa.statespace.mlemodel import MLEModel
  File "/home/mb312/dev_trees/statsmodels/statsmodels/tsa/statespace/mlemodel.py", line 14, in <module>
    from .simulation_smoother import SimulationSmoother
  File "/home/mb312/dev_trees/statsmodels/statsmodels/tsa/statespace/simulation_smoother.py", line 10, in <module>
    from .kalman_smoother import KalmanSmoother
  File "/home/mb312/dev_trees/statsmodels/statsmodels/tsa/statespace/kalman_smoother.py", line 13, in <module>
    from statsmodels.tsa.statespace.representation import OptionWrapper
  File "/home/mb312/dev_trees/statsmodels/statsmodels/tsa/statespace/representation.py", line 10, in <module>
    from .tools import (
  File "/home/mb312/dev_trees/statsmodels/statsmodels/tsa/statespace/tools.py", line 207, in <module>
    set_mode(compatibility=None)
  File "/home/mb312/dev_trees/statsmodels/statsmodels/tsa/statespace/tools.py", line 61, in set_mode
    from . import (_representation, _kalman_filter, _kalman_smoother,
ImportError: cannot import name '_representation'
```

josef-pkt:
An ImportError is not a crash, it's just an exception.

When Chad started to work on this refactoring, scipy 0.16 was still very recent. I don't know if we can drop scipy < 0.16 for a statsmodels 0.9 release.
So far this conditional compilation seems to have worked pretty well, better than I originally expected. But there are still open issues like making binaries/wheels conditional on scipy >= 0.16, but provide source distribution also for older scipy.

(Personally, I still don't feel completely friendly towards pip, even though it is now my default installer. It wasn't friendly to us for many years.)

matthew-brett:
Sure, I was using "crash" in the loose sense of dumping out with an exception, and no helpful error message.

My impression is that this stuff will work OK when compiled on a later scipy, then used with an earlier scipy, is that right?

ChadFulton:
> Updating the dependency to 0.16 would remove quite a lot of complexity here. For example, I noticed that, if you compile statsmodels with scipy < 0.16, and later install scipy >= 0.16, statsmodels crashes, because it thinks that the extra modules are compiled and tries to import them:

That's a good point. I'll make a PR to check that the modules are actually built before using trying to use them, even if scipy >= 0.16 is installed.

josef-pkt:
> My impression is that this stuff will work OK when compiled on a later scipy, then used with an earlier scipy, is that right?

AFAIU, it should not work either way. Once it's compiled, it links against the scipy/LAPACK wrapper version that it is in the compiling python, and it would not link to the another version. 
It's like upgrading numpy but using a different LAPACK library, e.g. replacing an openblas numpy with an mkl numpy.  Then scipy also wouldn't work anymore, if it was consistently compiled with openblas before.
(I think I, unintentionally, tried something like this a while ago, but don't remember what exception I got.)

(not worth the effort and extra complication: I assume the new scipy still have the old wrapper, so it could link also to the a second wrapper and dispatch during runtime. I guess there there are many things that go wrong and would require a lot of conditioning in the code.)

josef-pkt:
I'm not really trying to understand this thing, but I'm always happy seeing green.

If this or a version like this is fine, then I have no problem merging it (especially given that it looks less verbose or complicated than before.).

matthew-brett:
Ready to merge from my point of view.  I think I understand all the parts now, so I'm pretty confident this is the right way to do it.

benmosher:
I am seeing the issue precipitating this PR on an Alpine Docker build. Is there some other strategy that has superseded this?

bashtage:
Should add a pyproject.toml file with numpy.  This would prevent this issue for suitable versions of pip.

This is the correct way to implement this, although support is not fully present (e.g., tox).

josef-pkt:
@matthew-brett @bashtage 
I haven't been keeping up with build and installation topics for python packaging.
Can I just merge this after a rebase?

josef-pkt:
@matthew-brett @bashtage 
I just saw the pyproject.toml file of scipy for the first time.  (I never heard of it before.)
https://github.com/scipy/scipy/pull/7792
https://www.python.org/dev/peps/pep-0518/

As Kevin mentioned this should solve the main problems, and I like to follow scipy's lead in this kind of things. 

So can we just add a pyproject.toml file, and assume that it is enough for the relevant use cases?

josef-pkt:
(and reading through scipy #7309 reminds me why I'd rather focus on statistics and econometrics)

matthew-brett:
Josef - I think that PEP isn't even agreed yet - and when it is, I suppose it will take some time for it to filter through into the build tools, so no, I don't think that's an immediate solution.

Yes, I think this one is mergeable after a rebase.

matthew-brett:
Oops - sorry - yes the PEP's agreed - but I stand by the comment about build tools.

TomAugspurger:
The related [PEP 517](https://www.python.org/dev/peps/pep-0517/) is still being hashed out.

> but I stand by the comment about build tools.

Agreed. This PR would still be useful.

rgommers:
Had a look at this, summary:
- good idea in principle, and is fairly orthogonal to adding a ``pyproject.toml``.
- needs a rebase
- this currently breaks including generated C sources from Cython code in the sdist, that needs fixing

Also +1 for scipy >= 0.16.1, which is 2 years old by now.

rgommers:
>  @josef-pkt closed this in 6b0302f 2 hours ago

That commit was unrelated, this is not fixed.

bashtage:
I think I put the wrong issues in the big CI PR. This should be closed by
the pending setup PR.

bashtage:
#4871

rgommers:
Thanks @bashtage. gh-4871 looks good to me.

bashtage:
Can be closed now -- this was merged after some rebase and modifications this weekend.

josef-pkt:
superseded by merged #4871

PR #3205, opened 2016-09-19
ENH: Statespace: Improve/test remove_data, save, load

ChadFulton:
- Improves `remove_data` by actually removing all of the data arrays.
- Adds unit tests for `save` and `load` for each of the 4 major models.


ChadFulton:
This is a pretty simple PR, so I will merge tonight unless there are concerns.

PR #3231, opened 2016-10-08
FIX: avoid zero division and log0 in loglike calculation (resolve conflicts)

bfcondon:
This PR resolves the conflicts in #2301

### DIFF
--- a/statsmodels/genmod/families/family.py
+++ b/statsmodels/genmod/families/family.py
@@ -893,9 +893,9 @@ class Binomial(Family):
 
         else:
             return 2 * np.sum(self.n * freq_weights *
-                              (endog * np.log(endog/mu + 1e-200) +
+                              (endog * np.log(endog/(mu + 1e-200) + 1e-200) +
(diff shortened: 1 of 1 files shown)


bfcondon:
@josef-pkt I rebased the div zero PR and resubmitted it.  However, there are a large number of errors that are unrelated to these changes.

josef-pkt:
@bfcondon  thank you, I'll go through it and merge tomorrow or in the next few days.
I didn't investigate those errors yet, they have been showing up for a few days now.

josef-pkt:
temporarily prio-high, need to check various problems and proposals for genmod
 there is a suggestion to just clip instead as in other classes

jbrockmendel:
Did the clip suggestion ever get a decision?

josef-pkt:
family binomial has now a clip in the deviance computation.

I guess we still don't have a test case to check if there are problems left (e.g. loglike does not clip, and I'm not sure it should)

Issue #3232, opened 2016-10-09
Cannot perform fitting on SARIMAX model


marta-em:
The following code doesn't work on dummy data. It worked on real-life data. 

ststsmodels ver. 0.8.0rc1 

```
import pandas as pd
import statsmodels.api as sm
import random

index = pd.date_range('2015-01-01', '2015-01-31',freq='H')

df = pd.DataFrame({
     'speed': random.sample(range(100000),len(index)),
      'power': random.sample(range(100000),len(index))},
       index=index)

endog = df.ix[:, 'power']
exog = sm.add_constant(df.ix[:, 'speed'])

arima = sm.tsa.statespace.SARIMAX(endog.ix[:'2015-01-30'], exog=exog.ix[:'2015-01-30'], order=(1,0,1))
model_results = arima.fit()

arima = sm.tsa.statespace.SARIMAX(endog, exog=exog, order=(1,0,1))
res = arima.filter(model_results.params)

predict_steps = res.forecast(steps=24, exog=exog.ix['2015-01-31':])
```

Traceback (most recent call last):
  File "C:/Users/PycharmProjects/models/ARIMA_test.py", line 24, in <module>
    model_results = arima.fit()
  File "C:\Anaconda3\lib\site-packages\statsmodels\tsa\statespace\mlemodel.py", line 405, in fit
    start_params = self.start_params
  File "C:\Anaconda3\lib\site-packages\statsmodels\tsa\statespace\sarimax.py", line 967, in start_params
    endog -= np.dot(exog, params_exog)
TypeError: Cannot cast ufunc subtract output from dtype('float64') to dtype('int64') with casting rule 'same_kind'

Why the error occurs and how to fix it?

ChadFulton:
Thanks for the report! Looks like the issue occurs if `endog` is an integer array.

PR #3252, opened 2016-11-04
BUG: SARIMAX start_params when endog is integer

ChadFulton:
Fails due to attempted in-place subtraction of float array from integer array.


ChadFulton:
Closes #3232

ChadFulton:
Just the usual failures. Will merge this in a few minutes.

PR #3263, opened 2016-11-10
BUG/ENH/TST: State space: Simulate / Impulse Responses

ChadFulton:
Primarily a PR to improve test coverage on `simulate` and `impulse_responses` methods. Also:

- Produces a backwards incompatible change in `simulate` and `impulse_responses`, where the returned dimensions did not match the documentation in the case of multivariate models
- Fixes a bug when simulation was run twice and the second simulation had fewer `nobs` than the first.
- Introduces a flag (`_time_invariant`) for state space models to explicitly flag they are time invariant (rather than having to detect based on system matrix shapes)


ChadFulton:
Just the unrelated test failures

ChadFulton:
I will merge this and create a backport version of it to the 0.8 branch tonight, unless there are concerns.

PR #3294, opened 2016-11-26
[MRG] Add factor analysis

yl565:
This PR is for adding factor analysis. Already implemented Principal Axes Factor Analysis and the results are identical to R psych library `fa` when `rotation="none"` and `fm="pa"`.

To-do:
- [x] `FactorResults` class
- [x] Varimax, Quartimax and other rotations
- [x] Scree and variance explained plots
- [x] factor loading plots
- [x] Input validation
- [x] More examples


yl565:
Here is the R code to produce results of ``test_example_compare_to_R_output()`` :

```R
library(psych)
Basal = c(2.068,	2.068,	2.09,	2.097,	2.117,	2.14,	2.045,	2.076,	2.09,	2.111,	2.093,	2.1,	2.104)
Occ = c(2.07,	2.074,	2.09,	2.093,	2.125,	2.146,	2.054,	2.088,	2.093,	2.114,	2.098,	2.106,	2.101)
Max = c(1.58,	1.602,	1.613,	1.613,	1.663,	1.681,	1.58,	1.602,	1.643,	1.643,	1.653,	1.623,	1.653)
id = c(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13)
Y <- cbind(Basal, Occ, Max, id)
a <- fa(Y, nfactors=2, fm="pa", rotate="none", SMC=TRUE, min.err=1e-10)
a$loadings
a$communality
```

josef-pkt:
(I haven't looked yet)

I remember some discussions in mailing lists (not on statsmodels mailing list) about factor rotation.
There might be some reusable code out there that is license compatible, if needed.
eg. a brief search for "python varimax" shows
https://github.com/mvds314/factor_rotation
which I've never seen before. It's license compatible, looks like BSD-3, same as ours.

josef-pkt:
If it's license compatible (especially BSD-2, BSD-3 or MIT), then we can just copy the code.

For good manners and traceability of the origin:
If it's just part of a module like one or a few functions, then I just copy them. If it's almost another package, e.g. several modules, then I try to have a clean commit of the unchanged original code where we can specify the `author` in the commit message. And make adjustments to the code in follow-up commits.
For one large commit, where I essentially copied another package into statsmodels, I also contacted the author to tell him that I'm doing it.

yl565:
Factor rotation added using the [factor_rotation](https://github.com/mvds314/factor_rotation) package by @mvds314

The following rotation method produces the same results as R fa:
`None` (as 'none' in R), `"quartimax"`, `"oblimin"`

The following method produces somewhat different results:
`"varimax"`, `"equamax"`, `"promax"`, `"biquartimin"`

`"varimax"` produces the same results as the code [here](https://en.wikipedia.org/wiki/Talk%3aVarimax_rotation)

yl565:
@josef-pkt Can I translate some R code into python?

josef-pkt:
Almost all R code is GPL licensed which is NOT compatible with our license, BSD-3, and we cannot translate that code.
We can only translate R code, either if it has the BSD or MIT type license, there are a very few packages with it, or we have the explicit permission by the author.

josef-pkt:
In the past (after my transition from matlab to python), I was looking sometimes on the matlab fileexchange for some details, most of it is BSD licensed but not all.
e.g. 
https://www.mathworks.com/matlabcentral/fileexchange/?term=tag%3A%22factor+analysis%22
Antonio Trujillo-Ortiz  has a nice collection of statistical tests that I looked at (more for inspiration or checking some detail than actual translation)
Julia has a multivariate package but it doesn't have factor models yet. Most Julia packages are MIT licensed and license compatible for us.

yl565:
@josef-pkt I think this is ready for you to review. After it merged, I'm planning to implement other factor analysis method like maximum likelihood

mvds314:
Good to see that this the code on github is of benefit!

> The following method produces somewhat different results:
"varimax", "equamax", "promax", "biquartimin"

Both the R package and the Python package use the algorithms described here:
http://www.stat.ucla.edu/research/gpa/
The Python package contains unittests that test whether the code produces the same output as the Matlab/Octave code on the site. In that sense, the code already has been double checked for errors. If there are inconsistencies with the R code you are referring to that need to be resolved, I might be able to help.

The Python package is (with explicit permission of the original authors) published under a  BSD license on github.

Just to mention, the code package on github might be not be working under Python 3 due to the way I imported stuff in the __init__.py file. (This was the first time I put code in an actual package :)

statsmodels/multivariate/factor_rotation/_analytic_rotation.py
```diff
@@ -0,0 +1,275 @@
...
+    #solve procrustes problem
+    S=procrustes(A,H) #np.linalg.inv(A.T.dot(A)).dot(A.T).dot(H);
+    #normalize
+    d=np.sqrt(np.diag(np.linalg.inv(S.T.dot(S))));
+    D=np.diag(d)
+    T=np.linalg.inv(S.dot(D)).T
+    #return
+    return A.dot(T), T
+
+class unittests(unittest.TestCase):
```
josef-pkt: unit tests should go into a `statsmodels/multivariate/factor_rotation/tests` directory so they are picked up by nose for automatic testing

statsmodels/multivariate/factor_rotation/_gpa_rotation.py
```diff
@@ -0,0 +1,976 @@
...
+    if L is None:
+        assert(A is not None and T is not None)
+        L=rotateA(A,T,rotation_method='orthogonal')
+    q=np.linalg.norm(W*(L-H),'fro')**2
+    return q
+
+
+
+
+class unittests(unittest.TestCase):
```
josef-pkt: same here, separate unit tests from main modules

yl565:
@josef-pkt and @mvds314, I wonder how we can collaborate on this. Maybe @mvds314 can submit a separate pull request of the rotation package since it can be used by PCA as well.

josef-pkt:
I only looked roughly at the structure in this PR.

From my perspective the inclusion of the rotation looks ok. There could be minor git surgery, but nothing serious. We are not squashing all commits before merging (like some other packages), so the inclusion of the rotation will be visible as separate commits.
Usage of rotation for PCA and other models when they are available can be added in follow-up PRs.

One other part, scree_plot and similar look like overlapping/code duplication from PCA, and we might move the common parts into a reusable form instead of keeping duplicated code.

yl565:
@mvds314 [Here](https://github.com/statsmodels/statsmodels/pull/3294#issuecomment-263078836) is the R code I used to produce the results.
## Varimax
#### R loadings output

      factor 1    factor 2
Basal 0.9695076 0.2293770
Occ   0.9526215 0.2554104
Max   0.7864679 0.5889237
id    0.1462892 0.5742083
#### python loadings output
Basal         0.9883        -0.1259
Occ           0.9742        -0.1535
Max           0.8442        -0.5027
id            0.2060        -0.5556

## equamax
#### R loadings
Basal 0.9893560 -0.11719005
Occ   0.9824271 -0.08694325
Max   0.9407897  0.28333295
id    0.3344228  0.48915954
#### Python loadings
BBasal         0.9918         0.0946
Occ           0.9842         0.0646
Max           0.9341        -0.3047
id            0.3232        -0.4966

## promax
#### R loadings
Basal  0.9725400 0.04112070
Occ    0.9414561 0.07602842
Max    0.5991637 0.51127594
id    -0.1074315 0.64634944

#### Python loadings
Basal         0.9883        -0.1259
Occ           0.9742        -0.1535
Max           0.8442        -0.5027
id            0.2060        -0.5556

## biquartimin
#### R loadings
Basal  0.9725400 0.04112070
Occ    0.9414561 0.07602842
Max    0.5991637 0.51127594
id    -0.1074315 0.64634944
#### Python loadings
Basal           1.0862           0.1144
Occ             1.0371           0.0639
Max             0.4758          -0.5578
id             -0.3790          -0.8541

yl565:
I moved the `factor_rotation` unit test to `tests` folder and fixed pep8 problems.

mvds314:
@yl565 I am not sure if completely understand what you comparison you made between R and Python. I would leave the factor computation itself out of it and just focus on the rotation of the factors. Factor construction itself depends on a lot of details (whether or not to normalize and stuff like that, see for example the R function fa) which is not the aim of this package.

What we should compare is this R package:
https://cran.r-project.org/web/packages/GPArotation/GPArotation.pdf
with the Python package I put on git. For example, choose for favorite matrix Y and rotate it with varimax. The R code then becomes:
`GPArotation::Varimax(Y,eps=1e-5,maxit=100000)`
And the Python code becomes
`L1,T= fr.rotate_factors(df.values,'varimax', max_tries=100000,tol=1e-5)`
I only did a very brief check, but I don't see any differences so far up to several decimals in the examples I looked at.

How does this fit which what you found so far?

yl565:
After looking at fa.R source code, it looks like using rotate='varimax' will use stats::varimax while using rotate='Varimax' will call GPArotation::Varimax. I can confirm if I use the following code will produce the same results as python:
```R
Basal = c(2.068,	2.068,	2.09,	2.097,	2.117,	2.14,	2.045,	2.076,	2.09,	2.111,	2.093,	2.1,	2.104)
Occ = c(2.07,	2.074,	2.09,	2.093,	2.125,	2.146,	2.054,	2.088,	2.093,	2.114,	2.098,	2.106,	2.101)
Max = c(1.58,	1.602,	1.613,	1.613,	1.663,	1.681,	1.58,	1.602,	1.643,	1.643,	1.653,	1.623,	1.653)
id = c(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13)
Y <- cbind(Basal, Occ, Max, id)
a <- fa(Y, nfactors=2, fm="pa", rotate="Varimax", SMC=TRUE, min.err=1e-10)
```
Looks like a lot of rotation in R `fa` does not use GPArotation library. I think we are fine here since this is exploratory analysis and different way to rotate the loadings may produce different results.

yl565:
@mvds314 Looks like the factor_rotation library is failing the unittest under [python 2.7](https://travis-ci.org/statsmodels/statsmodels/jobs/179250945) Can you please help take a look?

mvds314:
@yl565 I took a look at the failed tests. I must say I don't know much about version control, automatic testing and stuff.

The unittests of the github code are running fine on my computer (both on Python 2.7 and 3.5).

I don't know much about git, project management and automatic testing on a build server. I cannot find any tests failing here due to the factor rotation code:
https://travis-ci.org/statsmodels/statsmodels/builds/179298396

statsmodels/multivariate/factor.py
```diff
@@ -0,0 +1,245 @@
...
+        elif rotation == 'promax':
+            A, T = promax(A)
+
+        if rotation is not None:  # Rotated
+            c = np.power(A, 2).sum(axis=1)
+        self.loadings = A
+        self.rotation = rotation
+        return FactorResults(self)
+
+    def plot_scree(self, ncomp=None):
```
josef-pkt: this should be a results method

statsmodels/multivariate/factor.py
```diff
@@ -0,0 +1,245 @@
...
+            Whether or not to apply squared multiple correlations
+
+        -------
+
+        """
+        if rotation not in [None, 'varimax', 'quartimax', 'biquartimax',
+                            'equamax', 'oblimin', 'parsimax', 'parsimony',
+                            'biquartimin', 'promax']:
+            raise ValueError('Unknown rotation method %s' % (rotation))
+        R = pd.DataFrame(self.endog).corr().values
```
josef-pkt: use numpy.corrcoef(self.endog, rowvar=0)

statsmodels/multivariate/factor.py
```diff
@@ -0,0 +1,245 @@
...
+            c = np.ones([1, len(R)])
+
+        # Iterative communality estimation
+        eigenvals = None
+        for i in range(n_max_iter):
+            # Get eigenvalues/eigenvectors of R with diag replaced by
+            # communality
+            for j in range(len(R)):
+                R[j, j] = c[j]
+            L, V = eig(R)
```
josef-pkt: use `eigh` ?

statsmodels/multivariate/factor.py
```diff
@@ -0,0 +1,245 @@
...
+            # Calculate new loadings and communality
+            A = V.dot(sL)
+            c = np.power(A, 2).sum(axis=1)
+            if norm(c_last - c) < tolerance:
+                break
+        self.eigenvals = eigenvals
+        self.communality = c
+        # Perform rotation of the loadings
+        self.loadings_no_rot = np.array(A)
+        if rotation in ['varimax', 'quartimax', 'biquartimax', 'equamax',
```
josef-pkt: possible structural change:  given that we can use different rotation with same raw FA
use pattern as for cov_type
- make rotation a fit keyword
- set the rotation in the results class `__init__`
- try to avoid any rotation specific attributes in `Factor` and attach those to results instance

to get a new results instance we could make the optimization loop conditional on not having `loadings_not_rot` as Factor attribute. 

maybe rename `loadings_not_rot` -> `loadings_raw` (or just `loadings` if it's unambiguous in the `Factor` class. (while FactorResults has the rotated loading).

statsmodels/multivariate/factor.py
```diff
@@ -0,0 +1,245 @@
...
+            included the same as the number of maximum possible loadings
+
+        Returns
+        -------
+        fig : figure
+            Handle to the figure
+        """
+        return plot_scree(self.eigenvals, self.n_comp, ncomp)
+
+    def plot_loadings(self, loading_pairs=None, plot_prerotated=False):
```
josef-pkt: this should also be a results method because it requires that `fit` has been called

statsmodels/multivariate/factor.py
josef-pkt: as part of data, this could be an attribute.
It doesn't change in repeated calls to fit

statsmodels/multivariate/plots.py
```diff
@@ -0,0 +1,141 @@
+import matplotlib.pyplot as plt
```
josef-pkt: matplotlib import needs to be `try ... except` protected because it is not a required dependency

statsmodels/multivariate/plots.py
```diff
@@ -0,0 +1,141 @@
...
+import numpy as np
+
+
+def plot_scree(eigenvals, total_var, ncomp=None, x_label='factor'):
```
josef-pkt: needs `fig` keyword to follow our usual pattern

statsmodels/multivariate/factor_rotation/tests/test_rotation.py
```diff
@@ -0,0 +1,581 @@
...
+        0.80985  -0.37650
+        0.77040   0.52371
+        0.65774   0.47826
+        0.58020   0.46189
+        0.63656   0.35255
+        """)
+        self.assertTrue(np.allclose(L, L_required, atol=1e-05))
+        T = target_rotation(A, H, full_rank=True)
+        L = A.dot(T)
+        self.assertTrue(np.allclose(L, L_required, atol=1e-05))
```
josef-pkt: use assert_allclose imported from numpy.testing, also in other places

statsmodels/multivariate/factor_rotation/tests/test_rotation.py
```diff
@@ -0,0 +1,581 @@
+import unittest
+import numpy as np
+
+from statsmodels.multivariate.factor_rotation._wrappers import rotate_factors
+from statsmodels.multivariate.factor_rotation._gpa_rotation import *
+from statsmodels.multivariate.factor_rotation._analytic_rotation import *
+
+
+class TestAnalyticRotation(unittest.TestCase):
```
josef-pkt: those should subclass object, we use now pytest and don't use unittest.TestCase

statsmodels/multivariate/factor_rotation/tests/test_rotation.py
```diff
@@ -0,0 +1,581 @@
...
+         .672  .458
+         .594  .444
+         .647  .333
+        """)
+
+    @classmethod
+    def get_quartimin_example(cls):
+        A = cls.get_A()
+        table_required = cls.str2matrix("""
+          0.00000    0.42806   -0.46393    1.00000
```
josef-pkt: extra blank, I don't know if that affects the parsing

josef-pkt:
needs rebase, also in the meantime we have switched from nosetests to pytests.

TravisCI doesn't retrieve the log files for the test runs anymore.
appveyor has an unrelated compat failure that should be fixed after rebase on master.
other failure: I have currently no guess about shape mismatch

```
ERROR: test_oblimin (statsmodels.multivariate.factor_rotation.tests.test_rotation.TestGPARotation)
  File "C:\projects\statsmodels\statsmodels\multivariate\factor_rotation\tests\test_rotation.py", line 459, in test_oblimin
    self.assertTrue(np.allclose(table, table_required, atol=1e-05))
  File "C:\Py\lib\site-packages\numpy\core\numeric.py", line 2372, in allclose
    res = all(isclose(a, b, rtol=rtol, atol=atol, equal_nan=equal_nan))
  File "C:\Py\lib\site-packages\numpy\core\numeric.py", line 2453, in isclose
    return within_tol(x, y, atol, rtol)
  File "C:\Py\lib\site-packages\numpy\core\numeric.py", line 2436, in within_tol
    result = less_equal(abs(x-y), atol + rtol * abs(y))
ValueError: operands could not be broadcast together with shapes (19,4) (17,4)
```

yl565:
Thanks for the review @josef-pkt! I just passed my PhD defense and now have more time. I'll work on this sometime next week.

josef-pkt:
@yl565 Thanks, I target this for 0.10, And this time I will help with debugging.

Congratulations to the PhD defense

yl565:
@josef-pkt It has been a long time since I last worked on this. Could you please educate me how I rebase the upstream master? Can I just use the following commands:
git fetch upstream-rw
git rebase upstream-rw/master

Also, there is some uncommitted changes in the code. Maybe I should commit the changes first?

josef-pkt:
yes, that's it
commit all changes, then fetch and rebase as in your two commands
then force push

I would recommend making a backup copy of the branch, i.e. just create a new branch from the current one, just in case something goes wrong and the rebase needs to be repeated. 

The first thing that might need fixing after the rebase are the unit tests so that pytest runs without problems.

statsmodels/multivariate/factor_rotation/tests/test_rotation.py
```diff
@@ -0,0 +1,581 @@
...
+        vgQ = lambda L=None, A=None, T=None: oblimin_objective(
+            L=L, A=A, T=T, gamma=1/2, rotation_method='oblique')
+        L, phi, T, table = GPA(A, vgQ=vgQ, rotation_method='oblique')
+        self.assertTrue(np.allclose(table, table_required, atol=1e-05))
+        self.assertTrue(np.allclose(L, L_required, atol=1e-05))
+        # quartimin derivative free
+        out = self.get_biquartimin_example_derivative_free()
+        A, table_required, L_required = out
+        ff = lambda L=None, A=None, T=None: oblimin_objective(
+            L=L, A=A, T=T, gamma=1/2, rotation_method='oblique',
```
josef-pkt: here is an integer division that causes failure python 2 
add `from __future__ import division` to top of file

statsmodels/multivariate/plots.py
josef-pkt: still todo, causes import error on Travis with python 2 without matplotlib

josef-pkt:
causes for Travis test failure/error, AFAICS (tests using python 3 with matplotlib available are green)

- integer division in a test file
- unprotected matplotlib import causes import error

josef-pkt:
I just realized that the rotation unit tests were extracted from the factor_rotation package. I think we can leave those unchanged for now, so we can keep it separate from statsmodels code for now. pytests seems to work fine with those unit tests..

josef-pkt:
travis is all green, no more py 2 problems

yl565:
Any thing else needs to be done?

josef-pkt:
Based on my previous review comments, and if we want to add more estimators, I think we should move all rotation specific attributes and code to the Results class. The Factor class itself can then be reused with different estimators and different rotations.

yl565:
@josef-pkt I applied the changes you suggested. All checks have passed. Do I still need to change to pytest syntax?.

josef-pkt:
We don't change the rotation tests for now. It's better to keep them close to the original package, as long as that is maintained and the tests don't cause problems.

(I will have to go over it again, and run some examples to see how well it works.)

josef-pkt:
I was doing a bit of general background reading in this neighborhood but not specific to factor models, eg. #4143
I think following the model-results pattern is the right way to go, and we can add additional features later, I guess we can get some inferential methods and some more formal functions to choose the number of components (besides scree plot).

I think inplace changes of attributes in the rotate method is too fragile, i.e. might result in surprising side effects for users after they change the rotation while holding on to the "old" instance with a previous rotation. I'm not sure yet how we could best handle that.

yl565:
About determining the number of factors, criteria such as Kaiser’s eigenvalue greater than one rule and total variance explained is easy and straight forward. Horn’s parallel analysis use the mean eigenvalue of randomly generated data matrices as the cutoff. It is also easy to implement but could be more time consuming to run. I think implementing them should be good enough for a first PR on FA. We can add more approaches later in new PR. What do you think?

josef-pkt:
The main one I found is Velicer's which from what I have seen is the main competition to parallel analysis but much simpler. Parallel analysis looks good in the references that I have seen, but might depend on the way that the simulation data is generated. (I didn't read what assumptions underlie the simulations.)
(Initially I was looking at usage of Bartlett's test for sphericity or similar, but, based on some remarks, it doesn't work well for choosing the number of components.)

These are from my notes with pdf file names, but I didn't read much:
"Determining the number of components from the matrix of partial correlations_A.3.Velicer-MAP-1976-Psychometrika.pdf"
    !!! implement this, printed, not read, parallel Monte Carlo based is main alternative
"Type I and Type II Error Rates and Overall Accuracy of the Revised Parallel Analysis Method for Determining the Number of Factors_0013164414546566.pdf"
"Performance of Velicer’s Minimum Average Partial Factor Retention Method With Categorical Variables_0013164410389489.pdf"

aside: kmo test is in PR #3842

yl565:
@josef-pkt May be we could merge this and leave the enhancements (such as determine # of factors) and #4156 in separate PRs? This PR already incorporated Principal Axes FA and the rotation. It seems to me a little cumbersome to incorporate more features.

josef-pkt:
@yl565  Yes I would like to merge this without extra features, Those can go in later. The main thing is to think about whether the current version is designed well enough for enhancements.

(Something to keep in mind, but every new feature is "experimental" for at least a release and we can still change it before changes become subject to serious backwards compatibility restrictions.)

josef-pkt:
`SMC` in `fit` does not have unit tests and raises an exception for me when I set `fit(SMC=False)` using the example from the test case.

SMC in fit will make an old model instance switch attributes, i.e. SMC changes state in model and will cause inplace modification of existing model instances. Those options either need to go into the `model.__init__` or not change model attributes. My guess is that moving it to `__init__` is better.
Also the name `SMC` is not informative and self-explanatory, and should be a lower case name (naming convention).

inplace modification by `rotate` in the results does currently not cause problems, except that users cannot get two instances with different rotation (without calling `fit` again to create a new unrotated instance.)  The main reason to still change this now is precautionary, to avoid that an existing instance changes behavior which will become fragile if we add cached attributes.

yl565:
I was wondering how to store SMC value in the result class.
Perhaps n_max_iter=50, tolerance=1e-6 should go to __init__ too?

statsmodels/multivariate/factor.py
```diff
@@ -0,0 +1,272 @@
...
+                             (n_max_iter))
+        if tolerance <= 0 or tolerance > 0.01:
+            raise ValueError('tolerance must be larger than 0 and smaller than'
+                             ' 0.01! Got %f instead' % (tolerance))
+        #  Initial communality estimation
+        if SMC:
+            c = 1 - 1 / np.diag(inv(R))
+            self.SMC = np.array(c)
+        else:
+            c = np.ones([1, len(R)])
```
josef-pkt: this should be 1 dimensional, AFAICS
`c = np.ones(len(R)) `
then there is no exception

unit tests ?

josef-pkt:
> Perhaps n_max_iter=50, tolerance=1e-6 should go to __init__ too?

Those are optimizer options that we always have in fit. In this case they could also cause "stale state" or better rewritten state, but I don't see a reason why users should hold on to a instance that didn't fit correctly, or that they would mind getting the latest estimate. It's not completely clean, but having optimizer options in fit is more important.

yl565:
Apart from working on smc and unit test, I'm thinking maybe we should store `n_max_iter=50`, `tolerance=1e-6`, 'smc=True`, etc. in the `FactorResults` so people can know the model and fitting specification from the results instance. What's the usual practice?

josef-pkt:
possibilities:
- the factor instance is not attached to the class
- all that is needed to create the Factor results instance are loadings, eigenvalues, communality and information n_comp and endog_names. I gues n_comp could be inferred from loadings.shape

If we change this to explicit argumens in `results.__init__`, then the factor class wouldn't need to have the decomposition as attributes. However those might become convenient if we want to reuse them.

In general: I feel like converting it to the full model results patter, but in this case it is not necessary (yet). Currently, it's still a bit closer to a stripped down decomposition method like PCA and CCA.
But I'm not sure where we are going with this. e.g. semi-random thought in #4159 

I think we leave it with small changes as is and merge, and leave the `FactorModel` design for the future.

josef-pkt:
> we should store n_max_iter=50, tolerance=1e-6, 'smc=True, etc. in theFactorResults` so people can know the model and fitting specification from the results instance.

The standard is to use `mle_settings` dictionary with the optimizer settings, and `mle_retvals` for the optimizer outcome/return. That is the pattern when we use scipy optimizers but is not followed by all of the other optimization methods, I ran into this recently in #2401.

Also a `converged` flag is useful, in case maxiter was too small.

statsmodels/multivariate/factor.py
```diff
@@ -0,0 +1,272 @@
...
+                             ' of columns of endog! %d > %d' %
+                             (n_factor, endog.shape[1]))
+        self.n_factor = n_factor
+        self.loadings = None
+        self.communality = None
+        self.eigenvals = None
+        self.method = method
+        super(Factor, self).__init__(endog, exog)
+
+    def fit(self, **kwargs):
```
josef-pkt: use explicit keywords because they are part of the documentation

naming convention maxiter=100, tol=1e-8  (copied from GLM)

statsmodels/multivariate/factor.py
```diff
@@ -0,0 +1,272 @@
...
+        """
+        Extract factors
+        """
+        if self.method == 'pa':
+            return self.fit_pa(**kwargs)
+        else:
+            raise ValueError("Unknown factor extraction approach '%s'" % self.method)
+
+    def fit_pa(self, n_max_iter=50, tolerance=1e-6, SMC=True):
```
josef-pkt: I think we should make this private `_fit_pa`
There is no reason for users to use it directly, 
e.g. GLM has two private `_fit_xxx` methods called via `fit`
and two public methods `fit_regularized` and `fit_constrained` that do something different than what is estimated by fit.

josef-pkt:
I'm giving up on the rotate question. Leave as is for now, and we change it when there is a need for a more careful non-inplace version.

josef-pkt:
@yl565 @kshedden cov, corr or multicorr?
Right now the unit tested version is FA for SMC (squared mutliple correlation)
This can be turned off to plain correlation.
Do we need FA for a covariance matrix instead of correlation matrix? 
It was my impression from skimming some of the related literature that we need a cov version.
(PCA has a `standardize` option)

@yl565  What Kerby suggested in https://github.com/statsmodels/statsmodels/pull/4156#issuecomment-350516954 to use cov or corr? directly as argument instead of endog, can still be added.

kshedden:
The ML method works for a covariance matrix.  It would be equivalent to run
it on a correlation matrix, then do ML factor analysis, then convert back
to the original scale.  So there's nothing really lost by requiring a
correlation matrix.  Perhaps the convergence will be faster on a
correlation matrix, not sure.  But it is more general to allow either a
covariance or correlation matrix.

statsmodels/multivariate/factor.py
```diff
@@ -0,0 +1,272 @@
...
+
+    Supported rotations:
+        'varimax', 'quartimax', 'biquartimax', 'equamax', 'oblimin',
+        'parsimax', 'parsimony', 'biquartimin', 'promax'
+
+    Parameters
+    ----------
+    endog : array-like
+        Variables in columns, observations in rows
+
```
josef-pkt: no blank lines between arguments in Parameters and Returns

statsmodels/multivariate/factor.py
```diff
@@ -0,0 +1,272 @@
...
+    n_factor : int
+        The number of factors to extract
+
+    method : str
+        Specify the method to extract factors.
+        'pa' - Principal axis factor analysis
+
+    """
+    def __init__(self, endog, n_factor, exog=None, method='pa', **kwargs):
```
josef-pkt: exog is not used, so we should remove it here and use exog=None in super call

josef-pkt: `**kwargs` are not forwarded to super
remove it so that it doesn't silently swallow wrong arguments.
It should be possible to add `missing` as specific keyword

josef-pkt:
I checked briefly with the Stata documentation, and they also allow correlation as alternative to data. Also it looks like only correlation matter according to their description also

To summarize, there are only two changes so it can be merged
- SMC: better name an in `__init__` and at least some unit tests
- `corr` in `__init__` as alternative to endog, 
  - if corr is not a DataFrame, then we would also need a way for users to set `endog_names`
  - if corr is not None, then endog could be None, and if endog is None, then the super `__init__` call needs to be skipped.

e.g. one possibility is to use:

`__init__(self, endog, n_factor, corr=None, method='pa', multicorr=True, missing='drop')`

usage: call with corr needs explicit None for endog if not provided, e.g.
`Factor(None, 3, corr=my_correlation_matrix)`

When we want to compute explicit arrays of factors in the future, then we need to raise if endog was not provided, or need to add endog as and argument to a get_factors results method.

other: docstring will need another review and making sure that those are up-to-date and complete.

kshedden:
I think we will also want a keyword argument n_obs=None that provides the
sample size in the case where the user passes a correlation matrix instead
of endog (needed for standard errors and other inference statistics)

josef-pkt:
semi-random thought:

one reason to add endog is to get outlier diagnostics and residual plots from the results.
If there are outliers, then they could distort the results because, AFAIK, those methods are not outlier robust (unless a robust correlation matrix has been used). A residual plot for each endog could help in identifying possible data problems that make appropriateness of the standard FA doubtful.

josef-pkt:
> I think we will also want a keyword argument n_obs=None

`nobs`

Yes I though about that but wasn't sure we should add it already, given that we don't have inference yet.
(I had to add `nobs` as required argument for all covariance and correlation tests that didn't use the original data.)

yl565:
The reason I use `SMC` as the variable name is that it is used in both [SAS](https://support.sas.com/documentation/cdl/en/statug/63347/HTML/default/viewer.htm#statug_factor_sect006.htm) and [R](https://cran.r-project.org/web/packages/psych/psych.pdf). Maybe we just use the lower case `smc`?  Another idea i have is to use `sm_cor` stand for squared multiple correlation

josef-pkt:
The SAS link mentions SMC under priors, R sounds like it's just starting parameters.

question: does SMC have an effect on the final estimate or is it just to get the iterations started?

I had thought it applies FA to a differently scaled correlation matrix, but if it is only for starting values, then it might still belong into `fit` and not into `__init__`.

yl565:
It specifies the initial guess for communality and it has an effect on the final results.

statsmodels/multivariate/factor.py
```diff
@@ -0,0 +1,330 @@
...
+        'pa' - Principal axis factor analysis
+    smc : True or False
+        Whether or not to apply squared multiple correlations
+    endog_names: str
+        Names of endogeous variables.
+        If specified, it will be used instead of the column names in endog
+
+    """
+    def __init__(self, endog, n_factor, corr=None, method='pa', smc=True,
+                 missing='drop', endog_names=None):
```
josef-pkt: add also `nobs` as Kerby pointed out. 
It's not currently used but will be necessary in future when we get some inferential methods.

statsmodels/multivariate/factor.py
```diff
@@ -0,0 +1,330 @@
...
+
+        # Check validity of corr
+        if corr is not None:
+            if corr.shape[0] != corr.shape[1]:
+                raise ValueError('Correlation matrix corr must be a square')
+            if endog is not None and endog.shape[1] != corr.shape[0]:
+                    raise ValueError('The number of columns in endog must be '
+                                     'equal to the number of columns and rows corr')
+        self.corr = corr
+        self.endog_names = endog_names
```
josef-pkt: corr could be a pandas DataFrame
We need `self.corr = np.asarray(corr)` to make sure that we have an array for further processing.
endog_names could be taken from the index or column of the DataFrame,
e.g. `if hasattr(corr, index): endog_names=...`

statsmodels/multivariate/factor.py
```diff
@@ -0,0 +1,330 @@
...
+    @property
+    def endog_names(self):
+        """Names of endogenous variables"""
+        if self._endog_names is not None:
+            return self._endog_names
+        else:
+            if self.endog is not None:
+                return self.data.ynames
+            else:
+                return None
```
josef-pkt: as default if not available you could make up some names
(that's what we usually do by default in the models)

statsmodels/multivariate/factor.py
```diff
@@ -0,0 +1,330 @@
...
+
+    @endog_names.setter
+    def endog_names(self, value):
+        # Check validity of endog_names:
+        if value is not None:
+            if self.corr is not None and len(value) != self.corr.shape[0]:
+                raise ValueError('The number of elements in endog_names must '
+                                 'be equal to the number of columns and rows in corr')
+            if self.endog is not None and len(value) != self.endog.shape[1]:
+                raise ValueError('The number of elements in endog_names must '
```
josef-pkt: You could define `k_endog` in `__init__` to store corr.shape[0] or endog.shape[1]
The the check would be simpler here

statsmodels/multivariate/factor.py
```diff
@@ -0,0 +1,330 @@
...
+                                 'be equal to the number of columns in endog')
+            self._endog_names = np.asarray(value)
+        else:
+            self._endog_names = None
+
+    def fit(self, n_max_iter=50, tolerance=1e-6):
+        """
+        Extract factors
```
josef-pkt: fit is the public function and needs the full docstring

maxiter and tol need renaming

yl565:
Changes applied

statsmodels/multivariate/factor.py
```diff
@@ -48,18 +48,26 @@ class Factor(Model):
...
         If specified, it will be used instead of the column names in endog
+    n_obs: int
+        The number of observations. To be used together with `corr`
+        Should be equals to the number of rows in `endog`.
 
     """
     def __init__(self, endog, n_factor, corr=None, method='pa', smc=True,
-                 missing='drop', endog_names=None):
-        # CHeck validity of n_factor
+                 missing='drop', endog_names=None, n_obs=None):
```
josef-pkt: naming is `nobs`, which we use in all model

josef-pkt:
comment to a commit (which might not be very visible): 
naming is `nobs`, which we use in all model
i.e. n_obs -> nobs

josef-pkt:
@yl565 Can you rebase this on master and push to a NEW PR?
Using a new PR avoids that we mess up the history of this PR, I don't think we have to squash commits.
(If you have problems, then I can do the rebase. )

I think we need to add "status: experimental" to the docstrings because I'm pretty sure we have to change things as we add more features.
(I'm still not familiar enough we FA to tell what we will need. The latest I tried, is to get the factors and prediction on endog based on factors, i.e. observation specific information. However, it's not as easy as in PCA. Additionally, we need the reconstituted cov_endog from the factor model. )

yl565:
OK, I will create a new branch, rebase on the new branch and push it to a new PR. Is it right?

josef-pkt:
more precise: rebase the new branch on master
Otherwise, yes that's it. 

I think I will merge after another spot checking. 
But there is still work to do after the merge, e.g. AFAICS, smc=False still doesn't have unit tests.

statsmodels/multivariate/tests/test_factor.py
```diff
@@ -0,0 +1,129 @@
...
+def test_unknown_fa_method_error():
+    # Test raise error if an unkonwn FA method is specified in fa.method
+    mod = Factor(X.iloc[:, 1:-1], 2, method='ab')
+    assert_raises(ValueError, mod.fit)
+
+
+def test_example_compare_to_R_output():
+    # No rotation without squared multiple correlations prior
+    # produce same results as in R `fa`
+    mod = Factor(X.iloc[:, 1:-1], 2, smc=False)
```
yl565: Actually, the unit test is here

josef-pkt: I guess I need better reading glasses.
Thanks

yl565: No problem, just let me know if there is anything else need to be done

josef-pkt:
If you have time to work on this some more, then there are a few features to get closer to what e.g. Stata has, e.g. methods for factor scoring.

I will look at the new PR and most likely will merge it tomorrow.

What we still need is a paragraph, or so, with a brief description for each of the new `multivariate` features that you added for inclusion in the release notes.

I will check inclusion in api.py and in rst documentation

josef-pkt:
rebased version #4161 merged

josef-pkt:
I opened #4164 to keep track of enhancement and refactoring ideas.

Issue #749, opened 2013-04-03
Repeated Measures ANOVA


jseabold:
[Request](http://stackoverflow.com/questions/15790708/how-can-i-perform-a-2-way-repeated-measures-anova-using-statsmodels/15793366) to support repeated measures ANOVA

see #786 for PR

tomwallis:
Just wanted to add another bump here. I'm teaching a stats course using Python / statsmodels, and it would be great to have a repeated-measures ANOVA implemented. It's an extremely common design in biomed, psychology and neuroscience. Would help out if I could but I am only an intermediate Python user.

It may help to look into R's implementation of the problem. One way to think about repeated-measures ANOVA is that the error term that a within-subjects factor is tested against is *stratified* or *nested* within subjects. For an example using R, see [here](http://www.psych.upenn.edu/~baron/rpsych/rpsych.html#htoc60). Perhaps Patsy's nesting can provide some help here?

josef-pkt:
I was looking at this recently again, trying to replicate some of the test cases of the PR as a linear model, however without success.

The reference looks helpful. For example I never figured out what the corresponding contrast for the tests are, something like  the following paired t-test equivalent
`t.test(hays.mat %*% c(1, 1, -1, -1))`
which uses the difference between means instead of a coding of the difference from the total within subject mean (which is one of the ways of removing subject specific unobserved heterogeneity used in econometrics)

My impression is that we could do most of it with patsy but we still need to create the encodings or contrasts for the different within or between effects.
Also we have anova_lm or now wald_test_terms #2132 as pattern to create F or chisquare tests for a sequence contrasts.

josef-pkt:
The other thing I haven't figured out yet in terms of equivalence to a linear model is how much code needs to depend on the specific experimental design, split-plot, cross-over and so on, and how much could be independent of the structure of the experimental design by just using general patsy formulas and contrasts.

josef-pkt:
And one more. Are the different test statistics in repeated measures anova separate tests or based on one model? anova_lm has the different types, like type 2 and type 3 anova, while in wald_test_terms I only looked at type 3 (or similar) which tests contrasts on coefficients estimated in the full model.
(AFAIR, in a completely balanced or orthogonal design it doesn't make a difference under standard assumptions.)

subhacom:
Waiting eagerly for this.

TheChymera:
:+1: we need ANOVA and prospectively even more flexible linear models for python!

[...]

PR #3303, opened 2016-12-02
[MRG] Repeated measures ANOVA

yl565:
This PR is for adding repeated measures ANOVA
Tested up to three factors and it reproduces R ezANOVA
#### R Code for producing the results in  `test_two_factors_repeated_measures_anova`
```R
library(ez)
DV = c(7, 3, 6, 6, 5, 8, 6, 7, 7, 11, 9, 11, 10, 10, 11, 11, 8, 14, 10, 11, 12, 10, 11, 12, 16, 7, 11, 9, 10, 11, 8, 8, 16, 10, 13, 10, 10, 14, 11, 12, 24, 29, 10, 22, 25, 28, 22, 24)
id = c(1, 2, 3, 4, 5, 6, 7, 8, 1,  2, 3,  4,  5,  6,  7,  8, 1,  2,  3,  4,  5,  6,  7,  8,  1, 2,  3, 4,  5,  6, 7, 8,  1,  2,  3,  4,  5,  6,  7,  8,  1,  2,  3,  4,  5,  6,  7,  8)
A = c('a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','b','b','b','b','b','b','b','b','b','b','b','b','b','b','b','b','b','b','b','b','b','b','b','b')
B = c('a','a','a','a','a','a','a','a','b','b','b','b','b','b','b','b','c','c','c','c','c','c','c','c','a','a','a','a','a','a','a','a','b','b','b','b','b','b','b','b','c','c','c','c','c','c','c','c')
data =data.frame(A, B, id, DV)
result = ezANOVA(data=data, wid=id, within=.(A, B), dv=DV)
print(result)
```

Planning to add support for mixed factor design and Greenhouse-Geisser correction in the future


yl565:
PR to #749

jona-sassenhagen:
Seems the CI bug is a problem with Py26. Any chance of finishing this one before Xmas? :D

josef-pkt:
I think this should go into statsmodels.stats. We don't want a new subpackage just for ANOVA.

If it doesn't get much larger than this, e.g. a few hundred lines, then it could be added directly to the current statsmodels/stats/anova.py otherwise it could also be in it's own module (but maybe private and imported into stats.anova.

As a `stats` hypothesis testing function it doesn't have to follow the relatively strict design and pattern of our models.

Does this only or mainly apply to balanced, orthogonal designs? 
I gave a brief look at the Rutherford book and he focuses on balanced designs, but I didn't look enough whether that is a requirement for the valid interpretation of the results or just makes the results more robust.

In general for this: I have no idea or preferences for the requirements or required functionality (and still have not much idea about the underlying theory). So, this is mostly up to whoever is interested in this and would be using it, and it needs lots of examples and unit tests.

hoechenberger:
Since I don't know much about the implementation details, I'll simply ask directly: does this implementation also support Type 3 sums of squares? In R, I (and all of my colleagues in psychology/neuroscience) use [afex](https://github.com/singmann/afex) or [JASP](https://github.com/jasp-stats/jasp-desktop), which both default to Type 3 SS, which allows users to compare the results to those produced by commercial statistical packages like SPSS. It's more or less required to allow users to switch to a new software with confidence.

yl565:
@hoechenberger Currently, the implementation only support specifying within-subject factors.

hoechenberger:
@yl565 You're right, sorry for not checking that first!

jona-sassenhagen:
@yl565 you still on it? :)

Only one of the Travis runs gives an error, but I don't understand it ...

yl565:
@jona-sassenhagen I'm waiting for review. @josef-pkt mentioned that it is fine if the oldest version of python reports error

josef-pkt:
I didn't know it's ready for review. You can ping me if a forget about a PR (which does not guarantee that I will look closely at it if I'm stuck in a completely different topic.)

I will review the code soon, but I will not be able to say much about the provided functionality.

statsmodels/stats/anova.py
```diff
@@ -362,6 +367,198 @@ def anova_lm(*args, **kwargs):
...
+        self.subject = subject
+        self._check_data_balanced()
+
+    def _check_data_balanced(self):
+        factor_levels = 1
+        for wi in self.within:
+            factor_levels *= len(self.data[wi].unique())
+
+        cell_count = {}
+        for index in range(self.data.shape[0]):
```
josef-pkt: this loops looks like a candidate for performance optimization
(I'm not sure yet how)

statsmodels/stats/anova.py
```diff
@@ -362,6 +367,198 @@ def anova_lm(*args, **kwargs):
...
+        y = self.data[self.dv].values
+
+        def _not_slice(slices, slices_to_exclude, n):
+            ind = np.array([True]*n)
+            for term in slices_to_exclude:
+                s = slices[term]
+                ind[s] = False
+            return ind
+
+        def _ssr_reduced_model(y, x, term_slices, params, keys):
```
josef-pkt: I'm not a fan of nested functions except for trivial or necessary ones. This looks like it could be useful more generally and could go into the module (as private function).

statsmodels/stats/anova.py
```diff
@@ -362,6 +367,198 @@ def anova_lm(*args, **kwargs):
...
+            """
+            Residual sum of squared of OLS model excluding factors in `keys`
```
josef-pkt: again squared -> squares

statsmodels/stats/anova.py
```diff
@@ -362,6 +367,198 @@ def anova_lm(*args, **kwargs):
...
+            x : independent variables
+            term_slices : a dict of slices
+                term_slices[key] is a boolean array specifies the parameters
+                associated with the factor `key`
+            params : OLS solution of y = x * params
+            keys : factors to be excluded
+
+            Returns
+            -------
+            residual sum of squared and degree of freedom
```
josef-pkt: sum of squares

statsmodels/stats/anova.py
```diff
@@ -362,6 +367,198 @@ def anova_lm(*args, **kwargs):
...
+                    df2 = df_resid1 - df_resid
+                    mse = (ssr1 - ssr) / df2
+                F = msm / mse
+                p = stats.f.sf(F, df1, df2)
+                term = key.replace('C(', '').replace(', Sum)', '')
+                anova_table.loc[term, 'F Value'] = F
+                anova_table.loc[term, 'Num DF'] = df1
+                anova_table.loc[term, 'Den DF'] = df2
+                anova_table.loc[term, 'Pr > F'] = p
+        self.anova_table = anova_table.iloc[:, [1, 2, 0, 3]]
```
josef-pkt: is there a reason to attach `anova_table` to `self`, i.e. the AnovaRM instance.
It would be cleaner to provide it as argument to `ANOVARestults.__init__`.

(Currently irrelevant, but better in case there will be fit options)

josef-pkt:
possible enhancement to API:
instead of specifying data as dataframe and strings for the columns, those could be arrays or individual pandas Series or DataFrames.
AFAICS, that would be consistent with current API and would not require a backwards incompatibility

statsmodels/stats/anova.py
```diff
@@ -362,6 +367,198 @@ def anova_lm(*args, **kwargs):
...
+class AnovaRM(object):
+    """
+    Repeated measures ANOVA using least square regression
+
+    The full model regression residual sum of squared is
+    used to compared with reduced model for calculating the
+    within subject effect sum of squared [1]
+
+    Between subject effect and Greenhouse-Geisser correction is not yet
+    supported.
```
yl565: Both should be able to added without backwards compatible break.  I currently have no idea how to implement the support for between subject effect using the current regression based approach though.

josef-pkt: As long as there is some extendable or (roughly) backwards compatible way forward, we can always add it later, e.g. add and delegate to a new internal algorithm.

statsmodels/stats/anova.py
jona-sassenhagen: and degree*s* of freedom?

jona-sassenhagen:
Very cool.

yl565:
@josef-pkt I applied the suggested changes.

jona-sassenhagen:
Travis error (py26 build) still looks unrelated, other Travis runs are fine.

statsmodels/stats/anova.py
```diff
@@ -362,6 +367,197 @@ def anova_lm(*args, **kwargs):
...
+                term = key.replace('C(', '').replace(', Sum)', '')
+                anova_table.loc[term, 'F Value'] = F
+                anova_table.loc[term, 'Num DF'] = df1
+                anova_table.loc[term, 'Den DF'] = df2
+                anova_table.loc[term, 'Pr > F'] = p
+
+        return ANOVAResults(anova_table.iloc[:, [1, 2, 0, 3]])
+
+
+class ANOVAResults(object):
```
hoechenberger: I think this should better be renamed to `AnovaResults` to be consistent with `AnovaRM`. Also looks prettier, imho. ;)

yl565:
@josef-pkt Is there any other changes you want me to make?

josef-pkt:
I will be back to reviewing and merging tomorrow.
I didn't see any problems in a quick skimming.

statsmodels/stats/anova.py
```diff
@@ -362,6 +367,197 @@ def anova_lm(*args, **kwargs):
...
+    used to compared with reduced model for calculating the
+    within subject effect sum of squares [1]
+
+    Between subject effect and Greenhouse-Geisser correction is not yet
+    supported.
+
+    Parameters
+    ----------
+    data : DataFrame
+    dv : string
```
josef-pkt: our usual term is `endog`

The signature is non-traditional for us, I'm not sure if we need to change it.

josef-pkt: use `depvar` instead of `dv` 
and we leave it otherwise as is.

It's not the same kind of pattern as the models, but this is a hypothesis test "function"

josef-pkt:
just one change dv -> depvar
Can you rebase this on master, given that it's a bit far behind?

then it's ready for merging

yl565:
@josef-pkt Is the rebase successful or I did something wrong?

josef-pkt:
Based on the commit list something  went wrong. I can look at it in an hour, and it would be better not to do anything with this branch until this is cleaned up.

josef-pkt:
Based on the commit list something  went wrong. I can look at it in an hour, and it would be better not to do anything with this branch until this is cleaned up.

josef-pkt:
Based on the network graph, afaics, you did a fast forward merge of master into your branch instead of a rebase.

I think the "easiest" is to do an interactive rebase on master and drop all the commits that come from master.
see two comments here https://github.com/statsmodels/statsmodels/pull/3395#issuecomment-275716950
I don't know if we have specific developer docs for this

Best is to try on a temporary branch and when everything is fine, then force push back here, so this doesn't show the master commits anymore. You could also open a PR with the temporary or rebased branch to see the results on github.

interactive rebase is a useful tool that should eventually be learned, but given that this PR is essentially ready for merge, I could do the rebase/cleanup and then have it ready for merge.

yl565:
@josef-pkt Please help me rebase/cleanup!

josef-pkt:
I can help tonight or tomorrow. (Right now I'm trying to do a monthly computer restart which takes a few hours)

jona-sassenhagen:
What's the status here? Can I help in some way?

josef-pkt:
the rebased version is now in #3663.
There is not supposed to be a change in code, just removing commits that don't belong in the PR

bertrandhermann:
Hi, 

Does anybody know if between-subject factor will be implemented in a near future ?

I use rpy2 to do mixed design repeated measures anovas, but since I have to do permutations on this (cluster permutation test for EEG data analysis), it takes very long. I think a fully python solution with statsmodel will be great. 
I'd love to help but I fear I'm not capable of.

Thanks

jona-sassenhagen:
[mne-python](https://github.com/mne-tools/mne-python) has repeated-measures ANOVAs for permutation tests of EEG data, thanks to @dengemann

bertrandhermann:
Yes, 
But this doesn't allow for between subject design; 
I discussed this issue with him directly and it cannot be done in mne-python
Thanks for your prompt reply

dengemann:
Yep. Someone needs to sit down and add the code for mixed-effects models :)

PR #3305, opened 2016-12-04
BUG: State space: Simulation smoothing with observation intercept.

### LINKED ISSUE #3304: BUG: State space: Simulation smoothing with observation intercept.
Incorrect results when `obs_intercept` is not zero and using the simulation smoother.

```python
nobs = 10
endog = np.ones(nobs) * 100
mod = sm.tsa.UnobservedComponents(endog, 'rwalk', exog=np.ones(nobs))
# the second element here, 100, is the size of the intercept
mod.update([1, 100])
sim = mod.simulation_smoother()
sim.simulate(disturbance_variates=np.zeros(mod.nobs * 2), initial_state_variates=np.zeros(1))
print(sim.simulated_state[0])
```

In the `UnobservedComponents` models, `exog` enters into the observation equation through the observation intercept. Here the exog is just an intercept (all ones), and the coefficient on it is 100.

Since I've set the disturbance variates to zero and the model is a random walk, the state should just stay at its initial level forever (0) and because there is an `obs_intercept` of 100, the predicted value for the observed series would then be `100` as it ought to be.

Instead, the output of the print statement is:

```
[-100. -100. -100. -100. -100. -100. -100. -100. -100. -100.]
```

I have already identified the error and will submit a PR shortly with a unit test and a fix.

ChadFulton:
This problem was caused due to using a short-cut for simulation smoothing where we filter the difference y_t - y_t^+ rather than separately filtering y_t and y_t^+ (see Durbin and Koopman, 2002). When filtering each separately, the `obs_intercept` should be kept as in the usual model, but when filtering the difference, the `obs_intercept` must be set to zero.

Durbin, J., and S. J. Koopman. 2002. “A Simple and Efficient Simulation Smoother for State Space Time Series Analysis.” Biometrika

Closes #3304

### DIFF
--- a/statsmodels/tsa/statespace/_simulation_smoother.pyx.in
+++ b/statsmodels/tsa/statespace/_simulation_smoother.pyx.in
@@ -205,6 +205,12 @@ cdef class {{prefix}}SimulationSmoother(object):
             self.secondary_simulated_smoother = {{prefix}}KalmanSmoother(
                 self.secondary_simulated_model, self.secondary_simulated_kfilter, smoother_output
             )
+        # In the case of non-missing data, the Kalman filter will actually
+        # be run over y_t^* = y_t - y_t^+, which means the observation equation
+        # intercept should be zero; make sure that it is
+        else:
+            dim2[0] = self.model.k_endog; dim2[1] = self.model.obs_intercept.shape[1];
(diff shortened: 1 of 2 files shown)


ChadFulton:
Will just merge this if everything is green since it's pretty straightforward.

PR #3351, opened 2017-01-02
BUG: Correct handling of missing in PCA

### LINKED ISSUE #3347: PCA with missing values in Pandas DataFrame
This falls over at data[mask] = projection_masked for a pandas df. 

When passed df.values seems to work ok.

bashtage:
Fix PCA to correctly handle missing values when input is pandas
Add tests for alternative drop methods

closes #3347


josef-pkt:
This looks like it should be backported to 0.8. Should it?

bashtage:
Probably.  It is a bug.

PR #3382, opened 2017-01-21
REF: Replace pd.TimeSeries to Series

sinhrks:
Remove ``pd.TimeSeries`` which has been alias of ``Series`` for a while, and dropped in 0.20.


ChadFulton:
Looks good to me. I'm surprised we had so few left around. Thanks!

PR #3394, opened 2017-01-27
TST: Fix issue with test order dependence

bashtage:
Fix test order dependence in test_sarimax.Test_ar_no_enforce
where changing order of tests could result in unexpected failure

### DIFF
--- a/statsmodels/tsa/statespace/tests/test_sarimax.py
+++ b/statsmodels/tsa/statespace/tests/test_sarimax.py
@@ -16,7 +16,6 @@ from statsmodels.tsa import arima_model as arima
 from .results import results_sarimax
 from statsmodels.tools import add_constant
 from numpy.testing import assert_equal, assert_almost_equal, assert_raises, assert_allclose
-from nose.exc import SkipTest
 
 current_path = os.path.dirname(os.path.abspath(__file__))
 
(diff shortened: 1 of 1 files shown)


bashtage:
@ChadFulton test code modified the model and so the order that the individual tests run matters.  Do my changes still make the test meaningful?

ChadFulton:
Thank you for looking into and fixing this. I think the test is still valid. I just had one question which I'll put inline.

statsmodels/tsa/statespace/tests/test_sarimax.py
```diff
@@ -1170,16 +1167,23 @@ def setup_class(cls, *args, **kwargs):
         # of states if enforce_stationarity = False
         cls.model.ssm.loglikelihood_burn = 0
 
-    def test_loglike(self):
```
ChadFulton: Was this test causing problems? I can't tell why we needed to eliminate it.

bashtage: This test seems to pass fine using the version in the parent class (pasted below).  It seem that the reduced tolerance wasn't necessary, and so this is just a little clean up.  The `init_keys_replication` needed some fixed to pass.  Not sure if these are strictly kosher or not -- "fixing the test" might be proving that it doesn't work as expected.

```
    def test_loglike(self):
        self.result = self.model.filter(self.true_params)

        assert_allclose(
            self.result.llf,
            self.true_loglike,
            atol=0.7 * 10**(-self.decimal)
        )
```

josef-pkt:
Is the ready for merging?

bashtage:
I'm happy with it.  It doesn't change any tests that are run, just makes
them order invariant.

josef-pkt:
Good, merging

Thanks

PR #3397, opened 2017-01-28
Migrate to pytest

bashtage:
pytest is an alternative testing framework that supports unittest-like tests but also new, simpler methods to construct test.  For example, it is easy to parameterize tests using a simple decorator to achieve the same thing that yield tests accomplish with nose. 

May not be popular

Advantages
* Good multithreaded test support.  Can run entire test suit in < 3 minutes on my desktop.
* Fixtures in addition to class based test design
* Actively developed and no nose/nose2 uncertainty
* Requires cleaning up some test code which improves code base
* Other complex project (pandas) moving to it

Disadvantages:
* Large one-time commit needed to make all tests run
* Still have dep on nose through NumPy


.travis.yml
```diff
@@ -101,24 +102,27 @@ before_install:
...
   - SRCDIR=$PWD
-  - python setup.py install
-  - mkdir -p "${SRCDIR}/travis-test"; cd "${SRCDIR}/travis-test"
   # Show versions
   - python -c 'import statsmodels.api as sm; sm.show_versions();'
   # Compose a script to run testing with coverage
-  - echo 'import statsmodels as sm; a=sm.test(label="full", verbose=3); import sys; sys.exit((len(a.failures)+len(a.errors))>0)' > test.py
-  - if [ ${COVERAGE} = true ]; then coverage run --rcfile=${SRCDIR}/.travis_coveragerc test.py; else python test.py; fi
+  - echo pytest --skip-slow -n auto ${COVERAGE_OPTS} statsmodels
+  - pytest --skip-slow -n auto ${COVERAGE_OPTS} statsmodels
```
josef-pkt: we cannot skip-slow on travis. We need to have a full test run to verify changes.

josef-pkt:
It's a bit too early for my taste to switch to pytest. I've seen it mentioned in scipy related mailing lists, but I prefer to be at the end of the packages that switch so the other ones have to figure out the problems. I just read on matplotlib gitter that they are discussing some pytest problems with python 2.

However, many changes in this PR are test cleanups that are useful whether nose or pytest is actually used.  The only question for merging those is whether they create merge conflicts with PRs waiting to be merged, or almost ready to be merged. That should affect at most a small number of modules.
So I guess it would be better to split this PR in two, one for unit test cleanup and one for the actual switch from nose to pytest.

question: Does the commandline `nosetests <path_to_module>` still work after these changes?

Once we switch, we might have to adjust our own workflow. I have been using nosetests for 10 years or so.

statsmodels/emplike/tests/test_regression.py
```diff
@@ -13,11 +13,12 @@ class GenRes(object):
     Loads data and creates class instance ot be tested
 
     """
-    def __init__(self):
+    @classmethod
+    def setup_class(cls):
```
josef-pkt: are those changes, replacing `__init__` by `setup_class` required or just to save on repeated `__init__` computation>

josef-pkt:
@bashtage  Do you know how overall memory usage changes from the switch from nose to pytest and replacing `__init__` by `setup_class`?
Since we never really ran into a problem, I never looked closely at changes over the years in nose that affected when tests are collected and classes are created.

BTW: nosetests can also run on multiple processors, but I haven't used it for some time.

josef-pkt:
both travis and appveyor are happy with this, coverage decrease by 9%  which might (largely) be because of `skip-slow`

BTW: I just realized that appveyor is fast now and finishes almost in 10 minutes. So, we could run now also the slow tests on appveyor without hitting the time limit.

.travis.yml
bashtage: Eventually it would be best to split the slow and the non-slow into different test runs.

statsmodels/emplike/tests/test_regression.py
bashtage: `__init__` is mostly bad practice.  But in this case pytest doesn't run and tests in classes with `__init__`.  Memory use was very mild on my machine ~175MB/process max.

bashtage:
> BTW: nosetests can also run on multiple processors, but I haven't used it for some time.

Multiprocess nose was always clunky and never worked correctly for me.  Pytest has a plugin that does proper load balancing at the test unit (e.g. a class or individual test) which has worked flawlessly for me.

josef-pkt:
(github is getting annoying: It's hiding the comments and I don't find them anymore from the email link)

> `__init__` is mostly bad practice. But in this case pytest doesn't run and tests in classes with `__init__`. Memory use was very mild on my machine ~175MB/process max.

We (numpy, ...) will need updated testing guidelines if some things don't work anymore. We want to avoid writing unit tests that never run.

Using `__init__` was perfectly fine and maybe needed in some cases when we want test/instance specific setup and not class method setup. E.g. we might have some cases where a test changes the state of the attributes created in `__init__` and we need to recreate them before each `test_xxx` method is run.
I don't remember when or where I ran into this.

josef-pkt:
`statsmodels/base/tests/test_shrink_pickle.py `  might be a case and we need to make sure the unit tests still run as intended.
remove_data wipes arrays from the result instance, and I guess it was intentional not to use the class method `setup_class`

josef-pkt:
here is a comment relating to state across test_methods:
https://github.com/statsmodels/statsmodels/pull/3397/files#diff-0463d424b8471427c194fc4b385ff98fR208

AFAIR, nose uses `setup` before a test method is run. and `setup_class` is used only once for all methods

bashtage:
> Using __init__ was perfectly fine and maybe needed in some cases when we want test/instance specific setup and not class method setup. E.g. we might have some cases where a test changes the state of the attributes created in __init__ and we need to recreate them before each test_xxx method is run.
> I don't remember when or where I ran into this.

If something needs to be reset, it is better to reset it explicitly rather than relying on creating an entirely new instance to do the same.  I would say using an __init__ is not needed and so ends up as a bad idea.  From what I saw in the code, most of these were written a long time ago.

josef-pkt:
Compared to the previous run in master, number of unit tests dropped from
`Ran 5651 tests in 674.248s OK (SKIP=54)` to `5605 passed, 58 skipped in 807.77 seconds` if the counting works the same
https://s3.amazonaws.com/archive.travis-ci.org/jobs/194274349/log.txt

bashtage:
It isn't the same.  The total in nose is equal to the number of run +
number of skips in purest.  So 5651 vs 5605+58.

bashtage:
> both travis and appveyor are happy with this, coverage decrease by 9% which might (largely) be because of skip-slow

Unfortunately the current coverage using nose isn't correct.  In particular files that have 0 lines exexuted are not counted.  You can see the difference by looking at the number of files covered here (537)

https://coveralls.io/builds/9880591

and (655)

https://coveralls.io/builds/9889137

For example sandbox/tsa/garch.py is missing from the current coverage by is included the pytest coverage. This is the reason for the large drop.  It seems correct to count coverage as lower, at least if sandbox should be included (which has terrible coverage ~40% and lots of LOC).

bashtage:
Thanks - I did it mostly manually yesterday.  The only part that was meaningfully different was how skipped tests are counted when a module or class raises `SkipTest`. pytest counts the individual tests while nose only counts a single skip.

josef-pkt:
@bashtage  you have 4 test runs in the queue. Do you need all those?
https://travis-ci.org/statsmodels/statsmodels/pull_requests

bashtage:
They should fast fail since I am amending so that the commit won't exist.  But no.  I'll close and reopen later.

josef-pkt:
they were still running. I canceled all except for the latest

PR #3402, opened 2017-01-31
FIX: Allow statespace models to be pickled in Python 3.6

bashtage:
Minimal patch to allow pickling.
Likely is not correct in the sense that the pickled objects would
not behave the same after pickling.

### DIFF
--- a/statsmodels/tsa/statespace/_kalman_filter.pxd
+++ b/statsmodels/tsa/statespace/_kalman_filter.pxd
@@ -175,6 +175,7 @@ cdef class sKalmanFilter(object):
     cdef void numerical_stability(self)
     cdef void check_convergence(self)
     cdef void migrate_storage(self)
+    cdef void _reinitialize_pointers(self) except *
 
     cdef void _forecasting(self)
     cdef np.float32_t _inversion(self)
@@ -300,6 +301,7 @@ cdef class dKalmanFilter(object):
     cdef void numerical_stability(self)
     cdef void check_convergence(self)
     cdef void migrate_storage(self)
+    cdef void _reinitialize_pointers(self) except *
 
     cdef void _forecasting(self)
     cdef np.float64_t _inversion(self)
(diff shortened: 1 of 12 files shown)


bashtage:
Fixes on Python 3.6. Probably doesn't contain enough to actually correctly run the three objects that can now be pickled since there is a lot of state.  Correct method will require `__reduce___` and `__setstate__` to be defined so that the class can be initialized and then any state that can change can be restored.

bashtage:
Pinging @ChadFulton

ChadFulton:
If this fixes the tests, then that sounds great to me.

bashtage:
Unfortunately pickling is much harder than this. Probably not worth it.

bashtage:
I think it words correctly now.  I probably copied more state that was strictly necessary since I wasn't sure what should persist across iterations and what was strictly temporary.

bashtage:
@ChadFulton I think this is ready.  Would be helpful if you could check the variables that I save in `state` in the method `__reduce__`.   If any of these don't need to be saved, I could reduce the size of the pickle.  For example, I am even saving temporary arrays which I suspect are always over written when used.   But in case they are not, it seemed safer to save them. 

I might make one final commit where I add the pickle tests to their own file `test_pickle.py`

bashtage:
closes #3401

bashtage:
@josef-pkt as long as @ChadFulton is happy with this, then it is ready.

josef-pkt:
Thanks, merging,
test failures are unrelated to this

bashtage:
You need to use master.

nruchika21:
Master of what ?

bashtage:
Master version of stats models from github.com.

nruchika21:
I am using what comes with python 3.6, How do I use the master version ?

Thanks,
Ruchika

bashtage:
Somethin like

pip install git+https://github.com/statsmodels/statsmodels

This requires you to a compiler for your version of Python.

PR #3432, opened 2017-02-05
CLN: Remove FutureWarnings

bashtage:
Remove FutureWarnings from NumPy
Clean up test to remove docstrings


bashtage:
@josef-pkt Should solve FutureWarnings once and for all
@TomAugspurger

TomAugspurger:
lgtm, though I don't have much experience with recarrays.

PR #3453, opened 2017-02-07
DOC: Small doc fixes

bashtage:
Fix layout problem whihc excluded css
Fix broken URL
Remove missing todo document

[skip ci]

### DIFF
--- a/docs/make.bat
+++ b/docs/make.bat
@@ -7,7 +7,7 @@ if "%SPHINXBUILD%" == "" (
 )
 set BUILDDIR=build
 set TOOLSPATH=../tools
-set DATASETBUILD  = dataset_rst.py
+set DATASETBUILD=dataset_rst.py
 set NOTEBOOKBUILD=nbgenerate.py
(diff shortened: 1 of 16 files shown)


bashtage:
@josef-pkt A few small doc problems that led to 404 errors

bashtage:
@josef-pkt @TomAugspurger 

There are no doc build errors after this patch on Windows except:

  * x12/13 needs to be installed
  * graphics are output in the wrong place, easy to manually fix
  
All of the WARNINGs are to do with missing toctree which can be ignored.

bashtage:
Doesn't fix notebooks in #3345 so that patch is needed too

TomAugspurger:
Looks good to me, @josef-pkt ?

> graphics are output in the wrong place, easy to manually fix

Are these the graphics produced by code blocks, or just some static assets? What would fixing that entail?

bashtage:
They come from code blocks using @savefig.  But it dumps them in docs/source and then the log shows a complaining about them not being somewhere else.

Issue #2129, opened 2014-12-15
Fail more gracefully on bad exog for ARMAX.


jseabold:
See example [here](https://stackoverflow.com/questions/27057445/how-to-fit-an-armax-model-using-statsmodels/).

Juanlu001:
This issue is no longer present in 0.10.0:

```
$ cat test.py
import numpy as np
import statsmodels.api as sm
# some stupid data

y = np.random.randn(100)
u = np.ones((100,2))
armax = sm.tsa.ARMA(y, order=(1, 1), exog=u).fit()
$ python test.py
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/juanlu/.pyenv/versions/st37/lib/python3.7/site-packages/statsmodels/tsa/arima_model.py", line 904, in fit
    k_trend, exog = _make_arma_exog(endog, self.exog, trend)
  File "/home/juanlu/.pyenv/versions/st37/lib/python3.7/site-packages/statsmodels/tsa/arima_model.py", line 397, in _make_arma_exog
    exog = add_trend(exog, trend='c', prepend=True, has_constant='raise')
  File "/home/juanlu/.pyenv/versions/st37/lib/python3.7/site-packages/statsmodels/tsa/tsatools.py", line 108, in add_trend
    raise ValueError(msg)
ValueError: x contains a constant. Adding a constant with trend='c' is not allowed.
```

Juanlu001:
Specifically, it was fixed in https://github.com/statsmodels/statsmodels/commit/119fcf90b218b95e48897a6fcaff6ef64dfd314c

bashtage:
Closing as fixed.

PR #3460, opened 2017-02-08
Fail more gracefully on bad exog for ARMAX

rishabmarkand:
fixes https://github.com/statsmodels/statsmodels/issues/2129

Let me know if any changes are to be made, especially with the error message.

### DIFF
--- a/statsmodels/tsa/arima_model.py
+++ b/statsmodels/tsa/arima_model.py
@@ -493,6 +493,11 @@ class ARMA(tsbase.TimeSeriesModel):
         exog = self.exog
         if k != 0:
             ols_params = GLS(endog, exog).fit().params
+            if k != len(ols_params):
+                raise ValueError("Number of exogenous variables "
+                                 "(including constants) is not equal "
+                                 "to the paramaters of generalised "
+                                 "least square fit")
             start_params[:k] = ols_params
             endog -= np.dot(exog, ols_params).squeeze()
         if q != 0:


bashtage:
When does this occur?  Can you write a more direct test rather than producing all do the values that fit makes?

bashtage:
You should also add a test that will trigger this error that checks that this error was raised.

rishabmarkand:
It appears to occur when the exog matrix has one or more constant columns and trend='c'

bashtage:
I think you can just check for a constant in exog and if trend == 'c'.  You can check for a constant using something like

```
nobs = x.shape[0]
rank = matrix_rank(np.hstack((ones((nobs, 1)), x)))
has_constant = rank == x.shape[1]
if has_constant and trend == 'c':
    raise ValueError()
```

rather than a full call to GLS.

bashtage:
At a minimum this needs a test that will raise the error to ensure it is correct.  

The PR would be better it if directly tested the rank rather than relying on a full fit of GLS which could be slow if the dataset is large.

rishabmarkand:
```python
def safe_is_const(s):
                try:
                    return 0.0 in np.ptp(s,axis=0) and np.any(s != 0.0)
                except:
                    return False
            if safe_is_const(exog) and trend == 'c':
                raise ValueError("Shape mismatch between exog and endog"
                                " matrices. Check that the exog matrix"
                                " is not comprised of constant columns.")
```
After checking for constant column in the exog in the above manner, numerous other (previously written) test cases produce errors. However, after the running the GLS fit, if I check the equality of k (the no. of exogenous vars including constant) and the number of ols parameters, the test cases are passing.

```python
if k != len(ols_params):
    raise ValueError("Shape mismatch between exog and endog"
                                " matrices. Check that the exog matrix"
                                " is not comprised of constant columns.")
```
This snippet seems to fix the issue. Let me know what you think!

bashtage:
This specific issue appears to be fixed in master.

PR #3477, opened 2017-02-12
BoxCox implementation and tests

N-Wouda:
See #2925 for details. Older, and much more concise: #1309.

### DIFF
--- a/statsmodels/base/transform.py
+++ b/statsmodels/base/transform.py
@@ -15,6 +15,7 @@ class BoxCox(object):
 
         Parameters
         ----------
+        x : array_like
         lmbda : float
             The lambda parameter for the Box-Cox transform. If None, a value
             will be estimated by means of the specified method.
@@ -22,11 +23,12 @@ class BoxCox(object):
             The method to estimate the lambda parameter. Will only be used if
             lmbda is None, and defaults to 'guerrero', detailed in Guerrero
             (1993). 'loglik' maximizes the profile likelihood.
-        **kwargs: dict
+        **kwargs
             Options for the specified method.
-            * For 'guerrero', this entails _R_, the grouping parameter,
-              and _scale_, the dispersion measure.
-            * For 'loglik', this dict is not used.
+            * For 'guerrero', this entails window_length, the grouping
+              parameter, scale, the dispersion measure, and options, to be
+              passed to the optimizer.
+            * For 'loglik': options, to be passed to the optimizer.
 
         Returns
         -------
@@ -117,11 +119,11 @@ class BoxCox(object):
         method : {'guerrero', 'loglik'}
             The method by which to estimate lambda. Defaults to 'guerrero', but
             the profile likelihood ('loglik') is also available.
-        **kwargs : dict
-            Options dict for the specified method.
-            * For 'guerrero': R (int), the seasonality/grouping parameter. Scale
-              ({'mad', 'sd'}), the dispersion measure. Options (dict), to be
-              passed to the optimizer.
+        **kwargs
+            Options for the specified method.
+            * For 'guerrero': window_length (int), the seasonality/grouping
+              parameter. Scale ({'mad', 'sd'}), the dispersion measure. Options
+              (dict), to be passed to the optimizer.
             * For 'loglik': Options (dict), to be passed to the optimizer.
 
         Returns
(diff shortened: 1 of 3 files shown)


N-Wouda:
@josef-pkt this is the finished PR (same end result as in #2925).

Once the checks are finished, I think it should be ready for merging.

Do you happen to know why the CI builds are failing? Is that related to this PR? (I've not kept up with everything that has changed in `statsmodels` for a couple of months)

josef-pkt:
There is a problem with python 3.6  with some time indexing in tsa, and a possible cython problem on appveyor that shows up frequently but not always.
I haven't checked yet the run for this PR, but I guess you don't get any additional failures or errors.

N-Wouda:
Ah, good to know (but of course, also unfortunate!). I am sure that will be remedied in the future.

Since this is the same as the previous PR, I am confident the results will return the same also.

statsmodels/base/transform.py
```diff
@@ -0,0 +1,222 @@
...
+        if method == 'guerrero':
+            lmbda = self._guerrero_cv(x, bounds=bounds, **kwargs)
+        elif method == 'loglik':
+            lmbda = self._loglik_boxcox(x, bounds=bounds)
+        else:
+            raise ValueError("Method '{0}' not understood.".format(method))
+
+        return lmbda
+
+    def _guerrero_cv(self, x, bounds, R=4, scale='sd', options={'maxiter': 25}):
```
josef-pkt: `R` is not really a nice parameter, maybe change it to `window_length` and mention that `guerrero` uses non-overlapping windows for the coefficient of variation.

N-Wouda: This is what Guerrero uses in both (1993) and (2004). What does `statsmodels` use internally in similar reshaping operations?

Since Guerrero also describes this (AFAIR) as the grouping parameter, so perhaps `group_length` if we want to follow the literature?

statsmodels/base/transform.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+class BoxCox(object):
+    """
+    Mixin class to allow for a Box-Cox transformation.
+    """
+
+    def transform_boxcox(self, x, lmbda=None, method='guerrero', **kwargs):
+        """
+        Performs a Box-Cox transformation on the data array x. If lmbda is None,
```
bashtage: Is `lam` better than `lmbda` for this parameters?

statsmodels/base/transform.py
```diff
@@ -0,0 +1,222 @@
...
+        Parameters
+        ----------
+        lmbda : float
+            The lambda parameter for the Box-Cox transform. If None, a value
+            will be estimated by means of the specified method.
+        method : {'guerrero', 'loglik'}
+            The method to estimate the lambda parameter. Will only be used if
+            lmbda is None, and defaults to 'guerrero', detailed in Guerrero
+            (1993). 'loglik' maximizes the profile likelihood.
+        **kwargs: dict
```
bashtage: **kwargs is not a dict, but are additional keyword arguments that the function accepts. These should be usual python variable names, so `r` and `scale`.

statsmodels/base/transform.py
josef-pkt: We want to have informative names that can be interpreted without looking at the original articles. In general we don't use one letter names because they are non-informative.
I prefer window in analogy to time series approaches, and because this guerrero approach assumes a data series with ordered observations. grouping sounds more general than the sequential group/window that this uses.

statsmodels/base/transform.py
N-Wouda: I am reasonably sure we should not want to diverge too much from the literature. In any case, SciPy also uses `lmbda` for their implementation of the BoxCox transform (see https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.boxcox.html).

statsmodels/base/transform.py
N-Wouda: You are obviously correct, and I have no good reason why I put `dict` there. I should've had another cup of coffee! :)

statsmodels/base/transform.py
N-Wouda: I'll change this to `window_length`, and explain in the docstring.

N-Wouda:
@josef-pkt (and @bashtage) I changed the argument names (and docstrings) according to your suggestions. I have the tests passing locally (nothing has changed here), and I believe that once the checks are completed this should be ready to merge.

bashtage:
Would be a good idea to add to the docs.

N-Wouda:
>Would be a good idea to add to the docs.

Seems reasonable, as it does entail a new feature. What do you think @josef-pkt?

N-Wouda:
Also, @bashtage, where in the docs do you think this should be mentioned? AFAICT, we do not have a `docs/base` or related.

bashtage:
Probably in regression since I think this is the main application.

josef-pkt:
at the current stage it might fit better in the tools, similar to add_constant or detrend for transforming endog an exog.
It depends a bit on how we tie it into the model classes. One main application especially for guerrero will be tsa.

N-Wouda:
Perhaps we can leave incorporation into the docs to a future date? It depends largely where we will use this BoxCox class (since I don't think it's necessarily appropriate for a direct standalone application).

bashtage:
I think it is better to put it somewhere - it can always be moved later
since docs don't really have stability concerns.  Wha ti smore annoying is
scrambling around to remember new features to document. :-)

josef-pkt:
The problems is that import paths can only be changed with deprecation (once it's released). So those should be relatively stable once we add it as public function/classes.

I look at it tomorrow (offline tonight), maybe we start with documenting regression tools somewhere even if they are advertised as mainly for internal purposes or for extending classes.

josef-pkt:
I think it can be useful as it is now. I used something similar for time series forecasting in a talk last year.

guerrero is a bit time series specific with predefined sequential structure, but this came up and is useful in more general settings.
However, the details are still experimental and internal at this stage, until we have our own applications, wrappers or examples for it.

N-Wouda:
This was initially meant as a specific mixin for other (`tsa`) models, so that would be the main application. I am not sure if I want to pursue that myself anytime soon, but I could open an issue to keep track of it (I'd definitely not get back to it before, say, July).

In the meantime, we could merge this and see where it is used in practice?

josef-pkt:
merged as is, instead of ...

@N-Wouda Thank you

PR #3481, opened 2017-02-13
CLN/TST: Remove nan issues with NumPy

bashtage:
Clean up nans that are being passsed to NumPy and SciPy
Remove calls to print in tests

### DIFF
--- a/statsmodels/nonparametric/tests/test_kernels.py
+++ b/statsmodels/nonparametric/tests/test_kernels.py
@@ -19,9 +19,9 @@ fname = 'results/results_kernel_regression.csv'
 results = np.recfromcsv(os.path.join(curdir, fname))
 
 y = results['accident']
-x = np.log(results['service'])
-use_mask = ~np.isnan(x)
-x = x[use_mask]
+x = results['service']
+use_mask = x >= 0
+x = np.log(x[use_mask].copy())
 y = y[use_mask]
 xg = np.linspace(x.min(), x.max(), 40) # grid points default in Stata
 
--- a/statsmodels/nonparametric/tests/test_kernels.py
+++ b/statsmodels/nonparametric/tests/test_kernels.py
@@ -20,9 +20,9 @@ results = np.recfromcsv(os.path.join(curdir, fname))
 
 y = results['accident']
 x = results['service']
-use_mask = x >= 0
-x = np.log(x[use_mask].copy())
-y = y[use_mask]
+positive = x >= 0
+x = np.log(x[positive])
+y = y[positive]
 xg = np.linspace(x.min(), x.max(), 40) # grid points default in Stata
 
 #kern_name = 'gau'
--- a/statsmodels/stats/contrast.py
+++ b/statsmodels/stats/contrast.py
@@ -55,7 +55,12 @@ class ContrastResults(object):
                 self.pvalue = self.dist.sf(self.statistic, df_denom)
             else:
                 "normal"
-                self.pvalue = self.dist.sf(np.abs(value)) * 2
+                # TODO: Need a simpler method to handle NaNs
+                pvalue = np.empty_like(value)
+                pvalue.fill(np.nan)
+                not_nan = np.logical_not(np.isnan(value))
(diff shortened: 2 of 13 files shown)


josef-pkt:
this contains several refactorings and nan enhancements that need review, it's not just TST

statsmodels/tools/tools.py
```diff
@@ -354,24 +354,37 @@ def pinv_extended(X, rcond=1e-15):
...
     equal to 0 to 0. Therefore, it presumes that X should be positive in
     general.
     """
-    x = np.maximum(np.asarray(X).astype(np.float64), 0)
-    return np.greater(x, 0.) / (x + np.less_equal(x, 0.))
 
+    x = np.asarray(x)
+    out = np.zeros_like(x, dtype=np.float64)
+    nans = np.isnan(x.flat)
+    pos = np.logical_not(nans)
```
jbrockmendel: Is there an agreed-upon convention in favor of `logical_not(foo)` over `~foo`?

bashtage:
I don't think so - there are 5 uses of logical_not .

statsmodels/nonparametric/tests/test_kernels.py
```diff
@@ -19,10 +19,10 @@
 results = np.recfromcsv(os.path.join(curdir, fname))
 
 y = results['accident']
-x = np.log(results['service'])
-use_mask = ~np.isnan(x)
-x = x[use_mask]
-y = y[use_mask]
+x = results['service']
+positive = x >= 0
+x = np.log(x[positive].copy())
```
jbrockmendel: why is the `copy()` needed?

statsmodels/stats/contrast.py
```diff
@@ -55,7 +55,10 @@ def __init__(self, t=None, F=None, sd=None, effect=None, df_denom=None,
                 self.pvalue = self.dist.sf(self.statistic, df_denom)
             else:
                 "normal"
-                self.pvalue = self.dist.sf(np.abs(value)) * 2
+                pvalue = np.full_like(value, np.nan)
+                not_nan = ~np.isnan(value)
+                pvalue[not_nan] = self.dist.sf(np.abs(value[not_nan])) * 2
+                self.pvalue = pvalue
```
jbrockmendel: No objection to be explicit and careful, but are there cases where the 1-line version messes up?  My n=1 test looked the same.

jbrockmendel:
Looks like appveyor error is the old recarray problem.  I think that has been resolved (or silenced), so that should go away with a CI re-run.

statsmodels/nonparametric/tests/test_kernels.py
bashtage: It is not.

statsmodels/stats/contrast.py
bashtage: What is the 1 line version?

jbrockmendel: GH can't find the diff anymore, but I think I was referring to the status-quo version.

bashtage:
When a nan comes up it triggers a warning.  Can't run nans through sf

jbrockmendel:
Makes sense, thanks for clearing that up.

+1

jbrockmendel:
Travis error is unrelated test_arima111_predict_exog_2127.  This looks good to go.

jbrockmendel:
@josef-pkt 5 dollar bill right here on the sidewalk, nearing its first birthday.

bashtage:
@josef-pkt Still ready to be merged. Silences a lot of the noise in test runs

josef-pkt:
merged
I'm not sure we are supposed to have nans at a few places like _calc_survfunc_right or whether it indicates a different problem.

I changed the merge message because it's not just CLN. I don't have a good overview where we want to allow for nans that are internally produced. I think we never really consistently reviewed this.
An alternative to avoiding the scipy distribution noise that has been showing up for some time now, would be to use catch warnings. I guess that in the distribution case the nan check is duplicate, once by statsmodels and once by scipy.stats. (But I don't remember the details or whether that has changed in the scipy.stats.distributions.)

@bashtage @jbrockmendel  thank you

PR #3483, opened 2017-02-14
ENH: Add nan- and inf-protected matrix_rank

bashtage:
Add a matrix rank that protects against inf anf nan to
prevent LinearAlgebraErrors in NumPy

### DIFF
--- a/examples/notebooks/distributed_estimation.ipynb
+++ b/examples/notebooks/distributed_estimation.ipynb
@@ -9,7 +9,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "metadata": {
     "collapsed": false
    },
@@ -52,7 +52,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": null,
    "metadata": {
     "collapsed": true
    },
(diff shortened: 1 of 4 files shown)


bashtage:
This was needed in the distributed notebook to fix the bug.  Presumably this is really a bug in the distributed estimation code that is in the notebook since it shouldn't try and estimate a model using matrices with `nan`s

josef-pkt:
I looked at this.  I think we check for nans when we compute ptp in this case which is cheaper, and doesn't require some duplicate computation.

Also, I think we need to put nan policy as prio-high for 0.9, i.e. change the default behavior.

In the distributed notebook: Was this in an example, where we could use `missing='drop'` or is this a consequence of internal computation?

bashtage:
It came from something inside the distributor    original data is nan
free.  So probably a bug.  May be coming from the iteration of GLM though
for one of the group's fitted.

josef-pkt:
Thanks, that sounds like a bug in distributed and we need an issue for it.

bashtage:
I think it happens because of a nan entering the calculations in an interative step -- since this checks rank when creating a new model to estimate, the bug is tripped. Probably need to stop the nan somewhere much earlier -- this fix doesn't actually stop the fit from working.

bashtage:
> Also, I think we need to put nan policy as prio-high for 0.9, i.e. change the default behavior.

IMO models that doesn't want to explicitly handle NaNs (e.g., not PCA) should simply raise and return.

PR #3497, opened 2017-02-22
ENH: Add basic time series outlier functions

Kevin-McIsaac:
I've created a simple time series outlier detection model as I had proposed in #3285.

The processes is:
1. Remove any trend component using rlm then create 'residues'
2. Use a robust measure of dispersion (ie., MAD or IQR) on the residues to identify potential outliers.

This implements 
* isoutliers:  find outliers in a time 
* replace: replace outliers with NaN or an interpolated value from the neighbours.
* plot: chart the time series overlaying the trend line and the outliers as red stars  

While simplistic it work well enough and is a starting point. It might be interesting to extend this to use autoregression to create the residues instead of rlm.

As I'm new to creating code for statsmodels, I'm happy to talk any feedback on programming style and advice on what else is needed, e.g., additional documentation etc.

### DIFF
--- a/statsmodels/tsa/outlier.py
+++ b/statsmodels/tsa/outlier.py
@@ -1,0 +1,189 @@
+# A simple module for visualising, identifing and replace outliers
+import pandas as pd
+from matplotlib.pyplot import gca
+
+from scipy.stats import norm as Gaussian
+import statsmodels.formula.api as smf
+from statsmodels.sandbox.regression.predstd import wls_prediction_std
+import numpy as np 
+
+
+def isoutlier(ts, find="IQR", N=2, detrend=True, dates=None, level=0):
+    '''
+    Find outliers in a time series.
+    
+    Parameters
+    ----------
+    ts : 1D array-like 
+         A panda Series with a DateTime index or a 1D array
+       
+    find : {'MAD', 'IQR, 'ZScore'}
+         Method for finding outliers
+              
+    N : float
+        if |MAD or ZScore| > N observation is an outlier
+       
+    detrend : boolean
+        remove the trend before finding ourliers?
+    
+    dates : array-like DateTime, optional
+        The datetime values for the time series    
+       
+    level : int
+            index level if the Series is multi-index
+    
+    Returns
+    -------
+    array-like boolean
+        True when the observation is an outlier
+    '''
+    
+    ts_len = len(ts)
+    numpy = False
+    if isinstance(ts, np.ndarray):
+        ts = pd.Series(ts)
+        numpy = True
+        
+    if detrend:
+        ts = residue(ts, dates=dates, level=level) 
+            
+    if find == "MAD":
+        outliers = abs(ts - ts.median()) > ts.mad() * (N / Gaussian.ppf(3/4.)) # reordered to avoid vector division
+        
+    elif find == "ZScore":
+        outliers = (ts - ts.mean()) > ts.std() * N # reordered to avoid vector division
+        
+    elif find == "IQR": # Note uses a fixed value rather than N
+        q = ts.quantile([0.25,.75])
+        IQR = q[0.75] - q[0.25]
+        outliers = (ts < (q[0.25] - 1.5*IQR)) | (ts > (q[0.75] + 1.5*IQR))
+        
+    else:
+        raise ValueError('find must be one of "MAD", "ZScore" or "IQR"') 
+
+    assert ts_len == len(outliers), "Returned result is incorrect length"
+    return outliers.values if numpy else outliers
+
+def trend(ts, dates=None, level=0):
+    '''
+    Return the line of best fit
+    '''
+    
+    if (~ts.isnull()).sum() < 2 : #need at least two points
+        return ts
+    else:
+        return rlm_ts(ts, dates=dates, level=level).fit().fittedvalues
+
+def residue(ts, dates=None, level=0):
+    '''
+    Return the residue from the line of best fit
+    '''
+    if (~ts.isnull()).sum() < 2 : #need at least two points
+        return ts
+    else:
+        return ts.where(ts.isnull(), rlm_ts(ts, dates=dates, level=level).fit().resid) #Necessary to deal with NaN values in ts
+
+def rlm_ts(ts, dates=None, level=0):
+    '''
+    Fit a linear model using robust estimates
+    '''
+    df = ts.to_frame() 
+    df['__X__'] = dates if dates else df.index.get_level_values(level)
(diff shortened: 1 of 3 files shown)


statsmodels/tsa/outlier.py
```diff
@@ -0,0 +1,189 @@
...
+
+def isoutlier(ts, find="IQR", N=2, detrend=True, dates=None, level=0):
+    '''
+    Find outliers in a time series.
+    
+    Parameters
+    ----------
+    ts : 1D array-like 
+         A panda Series with a DateTime index or a 1D array
+
```
bashtage: Too many spaces between parameters.  See examples

josef-pkt:
The main structural problem is that each function computes the trend regression or outlier detection. This repeats the same computations, but also makes it difficult to swap optionally the underlying algorithm.

Our main pattern is to use a class that holds all the results and add things like plot as a method. If plot is also useful as function to be shared across classes, then it could take all the relevant information as argument, e.g. from a results instance, to avoid recomputing parts.

I haven't looked at all the details yet, but for example trend will need an option for the polynomial order or for a non or semi parametric trend estimate.

Kevin-McIsaac:
@josef-pkt that is an interesting thought. I can see a class this makes sense for a model, i.e., regression, but doesn't for some simple methods like isnan(), MAD() or zscore().

Originally I thought of:
*  isoutliers() as a pd.Series predicate like isnull, 
*  replace as a pd.Series method like where, and
*  plot as a pd.DataFrame method like plot().

Now that I'm making this part of statsmodels, does it make sense to have isoutliers, replace and plot be new methods in model like RLM and AR?

That is, instead of writing 
```
tsa.isoutlier(ts)
```
to get a bool array or series I'd write
``` 
robust.robust_linear_model.RLM(endog, exog).fit.isoutlier() 
```
to get back a bool array, or I might write
``` 
robust.robust_linear_model.rlm(formula, ts).isoutlier() 
```

I could  write
```
tsa.ar_model.AR(endog).plot() 
```
to plot the time series showing the trend and the outliers

Do you think this fits statsmodels better? If so I'll try this first with RLM.

josef-pkt:
@Kevin-McIsaac 

OLS has an outlier test and outlier influence statistics that are similar to this, second part that you are suggesting. However, the idea there is to evaluate a given model with endog and exog.
Note OLS missing handling assumes independent observations, because the usual is to drop observations which would destroy an equal spaced time spacing for time series.
This should be extended to other models, and we also have some open issues discussion outlier detection for tsa models, but no code yet.

However, your initial proposal which I think is a good idea, goes into the direction of creating a task specific functionality instead of a model specific functionality.
One difference would be that we only use trend and similar for the "explained" part, instead of a full user specified model as in the former case. Adding nonparametric options for the trend or splines or more general polynomials would be pretty flexible but restricted to the univariate, single endog case.

eg. something like the following  
(`OutlierDetection` could be a class, but a function `detect_outliers` might be enough)

```
def detect_outliers(..., method='xxx', ...):
    ...
    res = OutlierResults(...)
    return res

class OutlierResults(object):

    def __init__(...):

        results_auxiliary =  results_instance_of_RLM
        outlier_mask = ...
        summary_frame = ...

    def plot():

...
```

so the call for just getting an outlier mask would be

`detect_outliers(x, ...).outlier_mask`

but if we want to look at it more closely, we keep the results instance

```
res = detect_outliers(x, ...)
res.outlier_mask
res.plot()
...

ChadFulton:
As an source of residuals from models with various types of trends, etc., state space models generally and the `UnobservedComponents` model for univariate series in particular can generate a number of residuals for testing relevant hypotheses.

The state space model in general is along the lines of:

y_t = Z \alpha_t + \epsilon_t
\alpha_t+1 = T \alpha_t + \eta_t

and as part of the Kalman smoother we can compute "auxiliary residuals" for the measurement equation and any of the state equations that can be used to test for deviations from "the model", at any time period, where "the model" depends on the equation in question.

For example, a local level model (basically a model that allows a time-varying intercept) is:

y_t = \mu_t + \epsilon_t
\mu_t+1 = \mu_t + \eta_t

If the auxiliary residual at time t associated with the measurement equation is large, that suggests that y_t might be an outlier.

If the auxiliary residual at time t associated with the state equation is large, that suggests that there may be a shift in the level of the series at time t.

This can be generalized in a variety of ways with more complex models (e.g. test for changes in trends, seasonals, cycles).

I worked up a more detailed example using the Nile dataset if you're interested, that can be found here: http://nbviewer.jupyter.org/gist/ChadFulton/c79ab90930c433b85ec751680cf5c185

Kevin-McIsaac:
@josef-pkt  I don't see why a task specific approach is better? 

I think it is just as clear to a user if we  extend models (RLM, AR, statemodel) with three standard methods, i.e.:
* isoutlier(method={"IQR", "MAD"}, N=2): return a bool indicating which points in the time series are outliers.
* plot(trend=True, interval=True, outliers=True ): return a plot of the time series showing the trend (prediction), the 95% confidence interval and the outliers.
* replace(fill=NaN): return the original time series with the outliers replaced by NaN or an interpolated value.

I see this has the following benefits:
* For novice users there is a simple way to find an outlier, isoutlier(), and a simple way to visualise, i.e., a plot that highlights the outliers and optionally shows a trend (prediction) and a 95% confidence interval.
* Advanced users can use a more complex of a model to generate the residues then apply isoutlier etc.
* The model implements outlier detection appropriate to the model. I suspect this is the same in each case, i..e, find the residues, assume they are near normal, us an appropriate measure of dispersion and threshold, e.g, IQR or MAD > N standard deviations/
* This approach works for all types of models
* There is no need to introduce a new class/object

Kevin-McIsaac:
@ChadFulton interesting notebook. Based on my limited understanding, the general approach to finding outliers is 
1. Calculate the residue error for the model. This is model dependant.
2. Assume the residues are near normal,  define outliers as observations that are along way from the other and then apply a test of dispersion. This part is not dependant on the mode and is common done using MAD or IQR. 

The Harvey, Koopman, and Penzer (1999) approach appears to using this approach (i.e., standardised errors) however not clear what the de Jong and Penzer approach is actually doing.

Kevin-McIsaac:
@ChadFulton just out of interest I applied [bayesian changepoint detection](https://github.com/Kevin-McIsaac/Nightlife/blob/master/Nile%20Level%20Changes.ipynb) to the Nile data to compare results.

It predicted the level change was most likely at 31-Dec-1899. I then fitted a linear regression to each segment to get the following. 

image

Very similar result

PR #3507, opened 2017-02-25
BUG: SARIMAX initialization: ergodic mean with trend

ChadFulton:
Pull request #3405 was the general case fixing this bug when stationary initialization is handled through `self.ssm.initialize_stationary()`, but since SARIMAX does a manual stationary initialization (so that it can also accomodate non-stationary states, e.g. for differencing and for regression coefficients in the state vector), it was not fixed there.

(Practically the effects of this are pretty minor, but still obviously not good to be wrong :)


ChadFulton:
I will also merge this PR today if it comes back green, since it too is pretty simple.

PR #3512, opened 2017-02-26
Apparent typo in sumofsq; axis kwarg should either be passed through …

jbrockmendel:
…or not exist.


ChadFulton:
Thanks for the PR! Test failures are unrelated (timeout of `webuse`). Seems like a good fix to me (although the value of the `sumofsq` function itself is somewhat dubious...).

PR #3525, opened 2017-03-01
FIX: Replace .ix usage

bashtage:
.ix is deprecated.  Replace with correct loc/iloc as needed
Change indices to since iloc does not include end point


TomAugspurger:
This looks good on a quick skim. Didn't see any other `.ix`es on a grep either. All ready @bashtage?

bashtage:
I think it is ready.

TomAugspurger:
Thanks!

PR #3597, opened 2017-04-06
COMPAT/CI: remove Widepanel and pandas dev

TomAugspurger:
There are a couple stray WidePanel uses in sandbox, but I wasn't able to run those anyway since my matplotlib is too new

```bash
statsmodels/compat/pandas.py:34:    data_klasses = (pandas.Series, pandas.DataFrame, pandas.WidePanel,
statsmodels/sandbox/examples/thirdparty/findow_0.py:11:uses DataFrame and WidePanel to hold data downloaded from yahoo using matplotlib.
statsmodels/sandbox/examples/thirdparty/findow_0.py:57:# combine into WidePanel
statsmodels/sandbox/examples/thirdparty/findow_0.py:58:pawp = pa.WidePanel.fromDict(dmall)
statsmodels/sandbox/examples/thirdparty/findow_1.py:11:uses DataFrame and WidePanel to hold data downloaded from yahoo using matplotlib.
statsmodels/sandbox/examples/thirdparty/findow_1.py:61:# combine into WidePanel
statsmodels/sandbox/examples/thirdparty/findow_1.py:62:pawp = pa.WidePanel.fromDict(dmall)
```

I've also changed the CI run using pandas 0.14 to instead use pandas' [daily wheels](https://mail.google.com/mail/u/1/#search/daily+wheels/15af30f4fb754e5e) so that we can catch these ahead of time.

Closes statsmodels#3580

### DIFF
--- a/statsmodels/compat/pandas.py
+++ b/statsmodels/compat/pandas.py
@@ -13,7 +13,8 @@ if version >= LooseVersion('0.17.0'):
         return df.sort_values(*args, **kwargs)
 else:
     def sort_values(df, *args, **kwargs):
-        kwargs.setdefault('inplace', False)  # always set inplace with 'False' as default
+        # always set inplace with 'False' as default
+        kwargs.setdefault('inplace', False)
         return df.sort(*args, **kwargs)
 
 try:
@@ -45,8 +46,12 @@ except ImportError:
 
 if version >= '0.20':
     from pandas.tseries import frequencies
+    data_klasses = (pandas.Series, pandas.DataFrame, pandas.Panel)
 else:
     try:
         import pandas.tseries.frequencies as frequencies
     except ImportError:
-        from pandas.core import datetools as frequencies
+        from pandas.core import datetools as frequencies  # noqa
+
+    data_klasses = (pandas.Series, pandas.DataFrame, pandas.Panel,
+                    pandas.WidePanel)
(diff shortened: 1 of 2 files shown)


TomAugspurger:
Waiting on https://github.com/pandas-dev/pandas/issues/15156 to actually test against pandas-master (after pandas release 0.20 in a few weeks probably).

bashtage:
Is that URL truly stable?  It looks like it si generated by something. Would be bad to change the test environment frequently to put out this fire. Is there not conda channel that hosts recent dev builds?

bashtage:
This change will also up the minimum to 0.15 effectively.  I'm happy with that -- my preference would be to use something like 0.18 for master/next release, but others are more conservative.

TomAugspurger:
> Is that URL truly stable? 

Yeah, pandas uses it for testing against NumPy and scipy dailies. I suppose we could test against those as well? I don't think conda packages are regularly uploaded anywhere. This is building off the multibuild stuff all these projects use for wheels.

> This change will also up the minimum to 0.15 effectively. 

Yeah.. I wasn't sure about which one to replace, so I just went with the oldest. 0.15 is pretty far back. I'm not super concerned about where in pandas 0.14 - 0.20 to draw the line, but I think statsmodels would be well-served to just require pandas 1.0 when that is out.

bashtage:
Need a rebase -- I would skip the change to the target and just implement the compat fix.

bashtage:
Conda and/or Travis seem to be broken ATM.  I think pandas is also seeing unexpected build failures.

PR #3599, opened 2017-04-06
Attach dates to VAR Forecasts

jbrockmendel:
Addresses #3531.  Quoting from there:

> Your _get_forecast_index should have an equivalent in the current ARMA and statespace models, which are supposed to handle date and time index correctly if available.

In this PR, `_get_forecast_index` method is in a new `TimeSeriesProcess` class that `VARProcess` inherits from.  `ARMA` etc should be able to inherit in the same way, but I'd like to get an OK on this before making changes all over the place.

> I think if you override forecast in VARResults so that it doesn't require the user to pass in data and then add 'forecast': 'dates', to the _methods dictionary in VARResultsWrapper it should work (but probably there are some fiddly things that won't work right away even so).

@ChadFulton I like the idea of handling this further up the hierarchy, but I think the "fiddly things" are exactly what is handled in `datetools.get_forecast_index`.  

> I think this question is better directed at VARResults, since I think VARProcess is meant to be data-agnostic, and also it doesn't have access to a model object from which to get e.g. model.data.row_labels or model.endog_names.

At the moment the forecast method checks the input data for an index, then falls back to look for a `self.model.data` and an associated index.   I like the idea of getting rid of the `self.model` check and overriding forecast in `VARResults`.


ChadFulton:
Thanks for following up on the VAR forecast dates (and all the pep errors :), and I'm sorry I wasn't more helpful before other than that cryptic message.

My opinion is that we should not add more special-case date handling in general, and especially not for a "weird" case like `VARProcess` (which doesn't really follow any conventions). One of the biggest problems is that it's hard to make sure that it all stays consistent (and I just stripped out much as I could from `datetools.py` :)

From a quick test, what do you think about simply replacing the `VARProcess.forecast` function with the following:

```python
    def forecast(self, y, steps):
        """Produce linear minimum MSE forecasts for desired number of steps
        ahead, using prior values y

        Parameters
        ----------
        y : ndarray (p x k)
        steps : int

        Returns
        -------
        forecasts : ndarray (steps x neqs)

        Notes
        -----
        Lutkepohl pp 37-38
        """
        model = getattr(self, 'model', tsbase.TimeSeriesModel(y))
        start, end, out_of_sample, prediction_index = (
            model._get_prediction_index(model.nobs + 1, model.nobs + steps))
        fcast = forecast(y, self.coefs, self.intercept, steps)
        return model.data.attach_dates(fcast)
```

It seems to work for me - maybe you can check that it does what you expect?

jbrockmendel:
> It seems to work for me - maybe you can check that it does what you expect?

When I run the unit test exactly as in the PR, the index is attached to the `y` input, not to `model.data`.  In that case it breaks.  If I pass the data to the VAR constructor as a DataFrame from the get-go, then it works as expected.

`model = getattr(self, 'model', tsbase.TimeSeriesModel(y))`

This is a clever workaround, but I want to advocate against calls that make copies.

ChadFulton:
> This is a clever workaround, but I want to advocate against calls that make copies.

`TimeSeriesModel` doesn't make any copies as far as I know.

jbrockmendel:
Dropped this for a while to focus on statespace, time to wrap this up.

> TimeSeriesModel doesn't make any copies as far as I know.

You're right.  I was confusing this was an unrelated case where I was passing in several columns from a DataFrame.  I eventually want to get around to things like `add_constant` which may make avoidable copies.

As I mentioned above, delegating to `attach_dates` fails when (as in the test) `y` is a DataFrame but the original model's data is not.  Potential mismatch between `y` and `self.model.data` can cause other problems or ambiguity, much of which I think can be resolved by re-attaching dates as early as possible.

> My opinion is that we should not add more special-case date handling in general, and especially not for a "weird" case like VARProcess

This seems like an argument for patching this weird case in the short run, then a) getting rid of the weird case and b) handling the general case up in tsabase.  I'd much rather spend this time  bike-shedding my next PR over whether recession bars should be "peak-to-trough" or [...]

Side-note: in var_model.py we have `import scipy.linalg as L`.  I have a deep dislike for single-letter names.  Any objection to changing this?  My top choice would be the verbose `scipy.linalg`

jbrockmendel:
After further thought, I want to make another pitch for the implementation in the PR.

The `forecast` method is in `VARProcess`, which represents the abstract process without any data needed.  If we're going to attach dates to the output, we should get those dates from the index (if any exists) on the input `y`.

jbrockmendel:
If there's an Issue for this, this can be closed.

ChadFulton:
To hopefully also help moving this along: personally I still think like my solution above is the best for this case, since it uses the existing date handling infrastructure and doesn't create special cases.

However, I don't have strong views, except that I think it's hard to see what is actually related to the VAR forecasts part here because of all the cleaning up in this PR.

Was this cleanup ever merged in a different PR? If not, would it be possible to separate this into two PRs?

jbrockmendel:
This was from when I was just getting to know the code.  Even if this approach is right (which it probably isn't), it will be easier to do fresh than to rebase at this point.  Closing.

Issue #3559, opened 2017-03-17
statsmodels 0.8.0 is missing smf.WLS.fit_regularized 


Sadrpour:
I have been using statsmodels 0.6.0 and was able to train models using
```
import statsmodels.api as smf
smf.WLS.fit_regularized()
```
This allowed me to fit a weighted least squares estimation with L1/L2 regularization.
I recently upgraded to statsmodels 0.8.0 to access improved time series models, but do not see the WLS regularized module anymore. 

In statsmodels 0.8.0 we have:
```
import statsmodels.api as smf
smf.OLS.fit_regularized()
```
which is not the same thing and missing the weight parameters. How can i apply smf.WLS.fit_regularized() in statsmodels 0.8.0

kshedden:
This may be something that happened when we made fit_regularized more generic, I will try to look into it later today.

Sadrpour:
Dr. Shedden thank you much, we are trying to use statsmodels 0.8.0 in production and your response will help us move forward. By the way, i was in your Stats650 class, an amazing class, you are one of the smartest people i have met.

PR #3625, opened 2017-04-25
Ridge regression fix  (closes #3624 first part)

### LINKED ISSUE #3624: Ridge regression gives wrong and inconsistent results.
statsmodel.api.OLS.fit_regularized goes through a shortcut when L1_wt is 0. However, that shortcut gives a wrong result when alpha is an array instead of a constant. The following code illustrates this issue with statsmodels version 0.8.0.
```python
import statsmodels.api as sm
import numpy as np
X = np.random.randn(1000, 100)
b = np.random.randn(100)
y = np.dot(X, b) + np.random.randn(1000) + 1 # y.mean() == 1.2861581555386801
res = sm.OLS(y, sm.add_constant(X)).fit_regularized(alpha=np.array([0] + [1e10] * 100), L1_wt=1e-8)
print('correct:', res.params[0]) # res.params[0] == 1.28615815554, which is correct.
sm.add_constant(X)).fit_regularized(alpha=np.array([0] + [1e10] * 100), L1_wt=0)
print('wrong:', res.params[0]) # res.params[0] == -0.00191408618425, which is wrong.
```

There is another peculiarity when L1_wt is zero. statsmodel.api.OLS.fit_regularized().params is a numpy array when L1_wt is zero. However, it becomes a pandas Series when L1_wt is not zero. This is inconsistent and confusing.

kshedden:
The ridge regression shortcut we used before was not correct when alpha is a non-constant vector.  This fixes that problem.

#3624

### DIFF
--- a/statsmodels/regression/linear_model.py
+++ b/statsmodels/regression/linear_model.py
@@ -1018,9 +1018,18 @@ class OLS(WLS):
 
         u, s, vt = np.linalg.svd(self.exog, 0)
         v = vt.T
-        s2 = s*s + alpha * self.nobs
-        params = np.dot(u.T, self.endog) * s / s2
-        params = np.dot(v, params)
+        q = np.dot(u.T, self.endog) * s
+        s2 = s * s
+        if np.isscalar(alpha):
+            sd = s2 + alpha * self.nobs
+            params = q / sd
+            params = np.dot(v, params)
+        else:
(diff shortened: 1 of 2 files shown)


kshedden:
This will conflict a bit with #3559, I will merge that first if no objections (after making it green).

PR #3646, opened 2017-05-07
ENH: Allow expected hessian in scipy optimizer fits

thequackdaddy:
Hello,

I think this is a rather simple approach to exposing the expected Hessian for scipy optimizer fits. 

I thought about using functools `partial` but ran into some obstacles. The real issue is that context matters and not all the `hessian` methods have an `observed` parameter. So this was as simple as I could make it without much effort. 

xref #3257

### DIFF
--- a/statsmodels/base/model.py
+++ b/statsmodels/base/model.py
@@ -253,7 +253,7 @@ class LikelihoodModel(Model):
 
     def fit(self, start_params=None, method='newton', maxiter=100,
             full_output=True, disp=True, fargs=(), callback=None, retall=False,
-            skip_hessian=False, **kwargs):
+            skip_hessian=False, observed_hessian=True, **kwargs):
         """
         Fit method for likelihood based models
 
@@ -436,14 +436,17 @@ class LikelihoodModel(Model):
         nobs = self.endog.shape[0]
         f = lambda params, *args: -self.loglike(params, *args) / nobs
         score = lambda params, *args: -self.score(params, *args) / nobs
-        try:
+        if observed_hessian:
             hess = lambda params, *args: -self.hessian(params, *args) / nobs
-        except:
-            hess = None
+        else:
+            hess = lambda params, *args: -self.hessian(params, *args, observed=False) / nobs
 
         if method == 'newton':
             score = lambda params, *args: self.score(params, *args) / nobs
-            hess = lambda params, *args: self.hessian(params, *args) / nobs
+            if observed_hessian:
+                hess = lambda params, *args: self.hessian(params, *args) / nobs
+            else:
+                hess = lambda params, *args: self.hessian(params, *args, observed=False) / nobs
             #TODO: why are score and hess positive?
 
         warn_convergence = kwargs.pop('warn_convergence', True)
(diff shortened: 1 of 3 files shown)


statsmodels/base/model.py
```diff
@@ -436,14 +436,17 @@ def fit(self, start_params=None, method='newton', maxiter=100,
         nobs = self.endog.shape[0]
         f = lambda params, *args: -self.loglike(params, *args) / nobs
         score = lambda params, *args: -self.score(params, *args) / nobs
-        try:
+        if observed_hessian:
             hess = lambda params, *args: -self.hessian(params, *args) / nobs
-        except:
-            hess = None
+        else:
+            hess = lambda params, *args: -self.hessian(params, *args, observed=False) / nobs
```
josef-pkt: nice trick
I was wondering why this doesn't break things in the standard case where observed is not an available keyword.
But you don't change the code on the default path.

I don't know why we had the try with catch-all except.

josef-pkt:
statespace models which are the only ones that currently have a similar option use `optim_hessian` with 3 choices http://www.statsmodels.org/devel/generated/statsmodels.tsa.statespace.sarimax.SARIMAX.fit.html

(IIRC or if I guess correctly, then OPG and EIM are the same in GLM, at least for fixed scale/dispersion)

thequackdaddy:
As far as I could tell at runtime, lamdas only have their parameters evaluated. Not until the lambda is called will the function fully evaluate. Additionally, the variables that are used must in the scope of the calling module. The nobs divisor needs to be declared in the calling module and you have to physically pass "True"

Honestly, this really feels like something that you should use functools partial... but the divide by nobs part and using negative for all optimizers but newton made that do-able.

statsmodels/base/model.py
```diff
@@ -253,7 +253,7 @@ def hessian(self, params):
 
     def fit(self, start_params=None, method='newton', maxiter=100,
             full_output=True, disp=True, fargs=(), callback=None, retall=False,
-            skip_hessian=False, **kwargs):
+            skip_hessian=False, observed_hessian=True, **kwargs):
```
josef-pkt: the new keyword here is not used anymore after the last commit

josef-pkt:
We might have to change observed=None, because hessian would ignore the observed keyword when called directly and _optim_hessian is set, which would be a difficult source of bugs to find.

statsmodels/genmod/generalized_linear_model.py
```diff
@@ -605,6 +604,8 @@ def hessian(self, params, scale=None, observed=True):
         hessian : ndarray
             Hessian, i.e. observed information, or expected information matrix.
         """
+        if hasattr(self, '_optim_hessian') and self._optim_hessian == 'eim':
```
josef-pkt: switch to `observed=None` in signature.
and protect this with `if observed is None:   ....  else observed=True`

josef-pkt:
What happened to the observed=None in`hessian(self, params, scale=None, observed=True)` which I thought I saw in an earlier version of the last commit?  Was there a problem with it?

thequackdaddy:
It still [there](https://github.com/thequackdaddy/statsmodels/blob/9b02c1237f85c19355f9081e68e2939eab7eec8a/statsmodels/genmod/generalized_linear_model.py#L587).

I amended the commit because I had logic that essentially said if `observed` is not `None`, the `observed = True`... so when you tried to compare to the EIM during non-fits, you got the OIM instead.

josef-pkt:
Good, github changes view showed the old signature.

merging, Thanks @thequackdaddy

Issue #3701, opened 2017-05-25
remove duplicated \begin{tabular} and \end{tabular} in .as_latex() output


natashawatkins:
When using `.as_latex()` output on `summary_col()`, the method creates two tables and joins them together.

This isn't great if the columns don't align...
![image](https://cloud.githubusercontent.com/assets/21038129/26444162/8ce1ccbc-417e-11e7-8b1b-1105d3624bb5.png)

Here is the output that creates this table:
```
\begin{table}
\caption{OLS Regressions}
\begin{center}
\begin{tabular}{lcccccccc}
\hline
                       &  Whole  &  Base    &  Whole  &  Whole  &  Base    &  Base    &  Whole  &  Base     \\
                       &  world  &  sample  &  world  &  world  &  sample  &  sample  &  world  &  sample   \\
                       &    (1)  &    (2)   &    (3)  &    (4)  &    (5)   &    (6)   &    (7)  &    (8)    \\
\hline
\hline
\end{tabular}
\begin{tabular}{lllllllll}
avexpr                 & 0.53    & 0.52     & 0.46    & 0.39    & 0.47     & 0.40     & 0.45    & 0.46      \\
                       & (0.04)  & (0.06)   & (0.06)  & (0.05)  & (0.06)   & (0.06)   & (0.04)  & (0.06)    \\
lat\_abst              &         &          & 0.87    & 0.33    & 1.58     & 0.88     &         &           \\
                       &         &          & (0.49)  & (0.45)  & (0.71)   & (0.63)   &         &           \\
asia                   &         &          &         & -0.15   &          & -0.58    &         &           \\
                       &         &          &         & (0.15)  &          & (0.23)   &         &           \\
africa                 &         &          &         & -0.92   &          & -0.88    &         &           \\
                       &         &          &         & (0.17)  &          & (0.17)   &         &           \\
other                  &         &          &         & 0.30    &          & 0.11     &         &           \\
                       &         &          &         & (0.37)  &          & (0.38)   &         &           \\
R squared              & 0.61    & 0.54     & 0.62    & 0.72    & 0.57     & 0.71     & 0.55    & 0.49      \\
Number of observations & 111     & 64       & 111     & 111     & 64       & 64       & 102     & 61        \\
\hline
\end{tabular}
\end{center}
\end{table}
```
This can be fixed by deleting
```
\end{tabular}
\begin{tabular}{lllllllll}
```
[...]

PR #3707, opened 2017-05-26
remove duplicate tabular objects created when latexing summary_col

natashawatkins:
Best solution I've thought of to #3701 

- uses a regex to find where duplicate tabular objects are created from summary_col, and replaces with two \midrule
- doesn't appear to affect regular regression output as aligns are different


natashawatkins:
My output now from `summary_col.as_latex()`

image

josef-pkt:
Hi, I had only a brief look
style/pep-8: needs line break to stay below 80 characters per line

> doesn't appear to affect regular regression output as aligns are different

That sounds a bit fragile given that different summarys go through the same code. 
Is it possible to add a flag as attribute to `self` and condition on it, `if self.merge_latex_table: ...`?

Also can you make a simple unit test with just a few columns. Given that there doesn't seem to be anything that changes in the table, like dates or floating point noise, it would be easy just to compare the generated string. The regular expression might easily break if there are any formatting changes, and without unit tests it would be slow to be found.   
(Most of the time we only use smoke tests for summary, but in this case a full test to verify the output wouldn't be difficult.)

josef-pkt:
(insert comment before first comment :)

Thanks, the output looks much better after the change.

natashawatkins:
Thanks for your comments @josef-pkt 

I've gone ahead and added the changes you've suggested

The unit tests simply test whether the \end{tabular}\begin{tabular} still exist or not for OLS vs. summary_col output.

I've also deleted brackets causing the caption error (#3700) because it keeps giving me compilation issues and it's a very simple fix.

statsmodels/iolib/tests/test_summary2.py
```diff
@@ -0,0 +1,52 @@
+import unittest
+from statsmodels.regression.linear_model import OLS, add_constant
+from summary2 import summary_col
```
josef-pkt: needs full import path, fails on all travis runs

josef-pkt:
looks good to me, one import error that needs to be fixed (travis test errors)

can you squash this (so we don't have the reverted file copy and redundant changes)

statsmodels/iolib/tests/test_summary2.py
natashawatkins: oops! done thanks

josef-pkt:
@natashawatkins  I merged this after rebase and some changes to the unit tests in #3963

Thank you for the contribution

PR #3788, opened 2017-06-27
ENH/BUG Contingency table improvements

kshedden:
* Add `__str__` methods so that the table objects and results returned as bunches are informative when printed.

* Fix problems with `make_df_square` (it wasn't tested and was not working with current Pandas)

* Allow nested lists, e.g. [[3, 5], [2, 1]] to be passed to the Table2x2 initializer.

* Fix issues with the two `from_data` methods

* flake8 edits


kshedden:
See also #3579

PR #3796, opened 2017-06-28
Refactor SaveLoadMixin used by both Wrapper and Results

jbrockmendel:
`save` and `load` methods are essentially identical in ResultsWrapper and LikelihoodModelResults.  

This PR merges them, along with `remove_data` into a `SaveLoadMixin` class in wrapper.py.  Besides de-duplicating code this is part of an effort to reduce the number of points-of-contact for the model.data object.

This is the first of several small PRs to address topics discussed in #3570.

### DIFF
--- a/statsmodels/base/model.py
+++ b/statsmodels/base/model.py
@@ -844,7 +844,9 @@ class Results(object):
 
 
 #TODO: public method?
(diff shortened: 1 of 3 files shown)


jbrockmendel:
Looks like error is a false-positive from coveralls.

PR #3817, opened 2017-07-13
Adding Holt Winters Functionality to tsa

tvanzyl:
Would like to add the holt winters functionality to tsa.
This includes all the additive, multiplicative, trend, damped and seasonal variations.
Also included are two helper functions for Simple Exponential Smoothing and Holt's Trend Method.

Thanks for all your efforts.


bashtage:
See #1489 for reference

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,472 @@
...
+import numpy as np
+import pandas as pd
+import warnings
+
+from collections import namedtuple
+from scipy.stats import boxcox
+try:
+    from scipy.special import inv_boxcox
+except ImportError:
+    inv_boxcox = lambda x, lam6da: (x**lam6da-1)/lam6da if lam6da != 0 else np.log(x)
```
bashtage: Is that a 6 in `lam6da`?

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,472 @@
...
+from scipy.optimize import minimize, basinhopping, brute
+
+Fit = namedtuple("Fit", ["fcast","fitted","alpha", "beta", 
+                         "gamma", "m", "phi", "l0", "b0", 
+                         "s0", "SSE", "level", "slope", 
+                         "season", "opt", "AIC", "BIC", "AICc",
+                         "resid", "k"])
+
+def holt_winters(data, alpha=None, beta=None, gamma=None, m=None, h=1, trend=None, 
+                 damped=False, seasonal=None, phi=None, optimised=True,
```
bashtage: I think American English is used in statsmodels.

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,472 @@
...
+                 boxcoxed=False, remove_bias=False, try_harder=False):
```
bashtage: Probably just `boxcox`

bashtage: `try_harder` is not a good name.

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,472 @@
...
+
+    Returns
+    -------
+    results : namedtuple
+        A namedtuple with fcast, fitted and other attributes
+
+    Notes
+    -----
+    This is a full implementation of the holt winters exponential smoothing as 
+    per [1]. This includes all the unsatble methods as well as the stable methods.
```
bashtage: Typo undstble

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,472 @@
...
+    The implmentaion of the library follows as per the R library as much as possible
```
bashtage: Did you use the R code?  It is not license compatible as far as I know.  Or just look at documentation, which is OK.

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,472 @@
...
+        set then this value will be used as the value.
+    m : int
+        The number of seasons to consider for the holt winters.
+    h : int
+        The number of time steps to forecast ahead.
+    trend : str {"add", "mul", None}
+        Type of trend component.
+    damped : bool
+        Should the trend component be damped.
+    seasonal : str {"add", "mul", None}
```
bashtage: Should be {"add", "mul", None}, optional.  No str since None isnt a str.

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,472 @@
...
+        The beta value of the holts trend method, if the value is
+        set then this value will be used as the value.
+    gamma : float, optional
+        The gamma value of the holt winters seasonal method, if the value is
+        set then this value will be used as the value.
+    m : int
+        The number of seasons to consider for the holt winters.
+    h : int
+        The number of time steps to forecast ahead.
+    trend : str {"add", "mul", None}
```
bashtage: Same

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,472 @@
...
+    alpha : float, optional
+        The alpha value of the simple exponnetial smoothing, if the value is
+        set then this value will be used as the value.
+    beta :  float, optional
+        The beta value of the holts trend method, if the value is
+        set then this value will be used as the value.
+    gamma : float, optional
+        The gamma value of the holt winters seasonal method, if the value is
+        set then this value will be used as the value.
+    m : int
```
bashtage: , optional
Multiple places

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,472 @@
...
+    [1] Hyndman, Rob J., and George Athanasopoulos. Forecasting: principles and practice. OTexts, 2014.
+    """
+    opt=None
+    trending = trend in ['mul','add']
+    seasoning = seasonal in ['mul','add']    
+    if damped and not trending:
+        raise NotImplementedError('Can only dampen the trend component')
+    phi = phi if damped else 1.0
+    N = len(data)
+    #if not isinstance(data, pd.Series):
```
bashtage: Delete?

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,472 @@
...
+        #raise NotImplementedError('Only Pandas Series Supported')
+    if boxcoxed == True:
+        y, lam6da = boxcox(data)
+    elif boxcoxed == 'log':
+        lam6da = 0.0
+        y = boxcox(data, 0.0)
+    else:
+        if isinstance(data, pd.Series):
+            y = data.values.flatten()
```
bashtage: Feels wrong to blindly flatten a (M,N) array.  Probably need a guard here -- `squeeze` and then check ndim==1.

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,472 @@
...
+    ----------
+    [1] Hyndman, Rob J., and George Athanasopoulos. Forecasting: principles and practice. OTexts, 2014.
+    """
+    opt=None
+    trending = trend in ['mul','add']
+    seasoning = seasonal in ['mul','add']    
+    if damped and not trending:
+        raise NotImplementedError('Can only dampen the trend component')
+    phi = phi if damped else 1.0
+    N = len(data)
```
bashtage: Should be lower case

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,472 @@
...
+    l = np.zeros((N,))
+    b = np.zeros((N,))
+    if seasoning:
+        if (m is None or m == 0):
+            raise NotImplementedError('Unable to detect season automatically')
+    else:
+        m = 0
+    s = np.zeros((N+m-1,))
+    p = np.zeros(6+m)
+    max_seen = np.finfo(float).max
```
bashtage: For clarity maybe use np.double since there is np.float/float differences?

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,472 @@
...
+    def holt_init(x, xi):
+        p[xi] = x
+        alpha,beta,_,l0,b0,phi = p[:6];
+        alphac = 1 - alpha
+        betac = 1 - beta
+        y_alpha[:] = alpha*y
+        l[:] = 0; b[:] = 0;
+        l[0] = l0; b[0] = b0
+        warnings.filterwarnings("error")
```
bashtage: Should not make a global change here.

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,472 @@
...
+    s = np.zeros((N+m-1,))
+    p = np.zeros(6+m)
+    max_seen = np.finfo(float).max
+    def holt_init(x, xi):
+        p[xi] = x
+        alpha,beta,_,l0,b0,phi = p[:6];
+        alphac = 1 - alpha
+        betac = 1 - beta
+        y_alpha[:] = alpha*y
+        l[:] = 0; b[:] = 0;
```
bashtage: should pass `l` and use local variables.

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,472 @@
...
+        s0 = []
+    if optimised:        
+        init_alpha = alpha if alpha is not None else 0.5/max(m,1)
+        init_beta = beta if beta is not None else 0.1*init_alpha
+        init_gamma = None
+        init_phi = phi if phi is not None else 0.99
+        if seasoning:            
+            init_gamma = gamma if gamma is not None else 0.05*(1-init_alpha) 
+            xi = np.array([alpha is None, beta is None, gamma is None, True, trending, phi is None and damped]+[True]*m)
+            if seasonal=='mul':
```
bashtage: More pythonic would be to stick these in a dict[(seasonal,trend)].  This is like C 👍

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,472 @@
...
+        The phi value of the damped method, if the value is
+        set then this value will be used as the value.
+    optimised : bool
+        Should the values that have not beem set above be optimised automatically?
+    boxcoxed : bool
+        Should the boxcox tranform be applied to the data first?
+    remove_bias : bool
+        Should the bias be removed from the fcast and fitted values before being
+        returned?
+    try_harder : bool
```
bashtage: use_basinhopping?

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,472 @@
...
+         phi=None, optimised=True):
+    """
+    Holt's Exponential Smoothing helper wrapper to holt_winters(...)
+
+    Parameters
+    ----------
+    data : array-like
+        Time series
+    alpha : float, optional
+        The alpha value of the simple exponnetial smoothing, if the value is
```
bashtage: Typo

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,472 @@
...
+        fitted = pd.Series(fitted, name=data.name)
+    resid = data-fitted[:-h]
+    if remove_bias:        
+        fitted += resid.mean()
+    if not damped: phi = None
+    return Fit(fitted[-h:], fitted[:-h], alpha, beta, 
+               gamma, m, phi, l0, b0, s0, SSE, level, 
+               slope, season, opt, AIC, BIC, AICc, resid, k)
+
+def ses(data, alpha=None, h=1, optimised=True):
```
bashtage: ses isn't helpful. Maybe simple_exp_smoothing?

statsmodels/tsa/tests/test_holtwinters.py
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python3
```
bashtage: Should be version agnostic

statsmodels/tsa/tests/test_holtwinters.py
```diff
@@ -0,0 +1,96 @@
...
+
+import numpy as np
+import pandas as pd
+from numpy.testing import assert_almost_equal, assert_equal, assert_raises
+from statsmodels.tsa.holtwinters import holt_winters, ses, holt
+from pandas import DataFrame, DatetimeIndex
+
+class TestHoltWinters(object):
+    @classmethod
+    def setupClass(cls):
```
bashtage: Please use `setup_class`

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,472 @@
...
+        alpha,beta,phi,alphac,betac = holt_init(x, xi)        
+        try:            
+            for i in range(1, N):
+                l[i] = (y_alpha[i-1]) + (alphac * (l[i-1]))
+            return sqeuclidean(l, y)
+        except RuntimeWarning:
+            return max_seen
+        finally:
+            warnings.filterwarnings("once")
+    def holt_mul_dam(x, xi):
```
bashtage: This patter of nested function and globals make unit testing really hard since it is difficult to directly test the smallest unit (e.g. a test hat holt_mul_dam works correctly).

statsmodels/tsa/holtwinters.py
tvanzyl: Yes it is the name of the parameter, I just use a "6" instead of a "b", maybe give it another name?

statsmodels/tsa/holtwinters.py
tvanzyl: Will change

statsmodels/tsa/holtwinters.py
tvanzyl: No, I never used any R code.

statsmodels/tsa/holtwinters.py
tvanzyl: changed

statsmodels/tsa/holtwinters.py
tvanzyl: Deleted

statsmodels/tsa/holtwinters.py
tvanzyl: Done

statsmodels/tsa/holtwinters.py
tvanzyl: done

josef-pkt:
I only gave it a very quick look so far.
The main issue I saw is the use of namedtuple. We don't use those because they cannot be changed in future without breaking backwards compatibility. Instead this should use a simple results class, that could also get a plot (plot_forecast) method, for example.

statsmodels/tsa/holtwinters.py
tvanzyl: done

statsmodels/tsa/holtwinters.py
tvanzyl: Sorry, I don't fully understand. Can you give me an example then I can fix it.

statsmodels/tsa/tests/test_holtwinters.py
tvanzyl: Yeah I realised once the tests had been run, I was going to have to change a few things.
Trying to figure out what the lowest common denominator will be to get this all working.

statsmodels/tsa/holtwinters.py
bashtage: I think there are some exampels using either `lam` or `lamda`

bashtage:
BTW, I think this is a good idea and there is a lot of demand for this type of function, and so, once ready, it should definitely go in.

statsmodels/tsa/holtwinters.py
tvanzyl: fixed all over

statsmodels/tsa/holtwinters.py
bashtage:
```
fns = {('mul','mul'): holt_win_mul_mul_dam,
           ('mul','add'): holt_win_add_mul_dam}

func = fns[(seasonal, trend)]
```

statsmodels/tsa/tests/test_holtwinters.py
tvanzyl: Changed

statsmodels/tsa/holtwinters.py
tvanzyl: Ok wait, got it I think, will change.

statsmodels/tsa/tests/test_holtwinters.py
tvanzyl: done

statsmodels/tsa/holtwinters.py
tvanzyl: changed

statsmodels/tsa/holtwinters.py
tvanzyl: changed

statsmodels/tsa/holtwinters.py
tvanzyl: done

statsmodels/tsa/holtwinters.py
tvanzyl: done

statsmodels/tsa/holtwinters.py
tvanzyl: done

statsmodels/tsa/holtwinters.py
tvanzyl: done

statsmodels/tsa/holtwinters.py
tvanzyl: done

statsmodels/tsa/holtwinters.py
tvanzyl: removed this, will allow the warnings through

statsmodels/tsa/holtwinters.py
tvanzyl: done

statsmodels/tsa/holtwinters.py
tvanzyl: done

tvanzyl:
Made the changes as per josef-pkt 
Added a Result class like the one in seasonal.py

statsmodels/tsa/holtwinters.py
tvanzyl: done

statsmodels/tsa/tests/test_holtwinters.py
tvanzyl: Done

statsmodels/tsa/api.py
```diff
@@ -24,3 +24,4 @@
 from .statespace.dynamic_factor import DynamicFactor
 from .regime_switching.markov_regression import MarkovRegression
 from .regime_switching.markov_autoregression import MarkovAutoregression
+from .holtwinters import holt_winters, holt, simple_exp_smoothing
```
bashtage: Should this module be called exponential_smoothing to enhance discoverability?

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,444 @@
+"""
+Created on Wed Jul 12 09:35:35 2017
+
```
bashtage: You should add your full name.  Also any references would be appreciated.

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,444 @@
...
+import pandas as pd
+import warnings
+
+from statsmodels.compat.python import iteritems
+
+from scipy.stats import boxcox
+try:
+    from scipy.special import inv_boxcox
+except ImportError:
+    inv_boxcox = lambda x, lmbda: np.exp(np.log1p(lmbda * x) / lmbda) if lmbda != 0 else np.exp(x)
```
bashtage: Should be a def function.    No assigning lambdas.

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,444 @@
...
+    
+from scipy.spatial.distance import sqeuclidean
```
bashtage: Could you alphabetize the scipy imports.

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,444 @@
...
+    This is a full implementation of the holt winters exponential smoothing as 
+    per [1]. This includes all the unstable methods as well as the stable methods.
+    The implementaion of the library follows as per the R library as much as possible
+    whilest still being pythonic.
+    
+    References
+    ----------
+    [1] Hyndman, Rob J., and George Athanasopoulos. Forecasting: principles and practice. OTexts, 2014.
+    """
+    opt=None
```
bashtage: Missing a space.   Might as well autopep8 this file since new.

bashtage:
My big thought is whether this should be a class with most methods private.  This would fit better with most other methods.   Ideally they he computations wouldn’t happen until fit() was called. 

This really should make it in.   If you don’t have time to convert let me know and I’ll take a short from your branch.

statsmodels/tsa/api.py
tvanzyl: Called the class ExponentialSmoothing

statsmodels/tsa/holtwinters.py
tvanzyl: Done

statsmodels/tsa/holtwinters.py
tvanzyl: done

statsmodels/tsa/holtwinters.py
tvanzyl: done, I order by full module names?

statsmodels/tsa/holtwinters.py
tvanzyl: done

tvanzyl:
Moved it all into a class inherited from TimeSeriesModel, 
Tried to follow the ARIMA model of Results and ResultsWrapper and inherited from TimeSeriesModel

Little issues might need someone who is more familiar with the code base to iron out since TimeSeriesModel extends LikelihoodModel, Would be nice if we could refactor this out a bit more, since all TimeSeriesModels are not Likelihood models.

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,821 @@
...
+        Type of trend component.
+    damped : bool, optional
+        Should the trend component be damped.
+    seasonal : {"add", "mul", None}, optional
+        Type of seasonal component.
+    m : int, optional
+        The number of seasons to consider for the holt winters.
+    """
+
+    def __init__(
```
bashtage: Normal style here is something like 

```
def __init__(self, endog, trend=None, damped=False, seasonal=None, m=None, 
             dates=None, freq=None, missing='none', **kwargs):
```

This is true in a few places below.

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,821 @@
...
+        References
+        ----------
+        [1] Hyndman, Rob J., and George Athanasopoulos. Forecasting: principles and practice. OTexts, 2014.
+        """
+        es = ExponentialSmoothing(endog)
+        res = es.fit(alpha=alpha, optimized=optimized)
+        return es._predict(h=h, **res.params)
+
+    @staticmethod
+    def holt(
```
bashtage: Same here.

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,821 @@
...
+                                   season=season,
+                                   AIC=AIC,
+                                   BIC=BIC,
+                                   AICc=AICc,
+                                   resid=resid,
+                                   k=k)
+        return HoltWintersResultsWrapper(hwfit)
+
+    @staticmethod
+    def holt_winters(
```
bashtage: Same here.

bashtage: Should there be a "simple" subclass

```
class HoltWinters(ExponentialSmoothing):
```

Or which is the base and which is the restricted?

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,821 @@
...
+                              l0=l0,
+                              b0=b0,
+                              s0=s0,
+                              use_boxcox=use_boxcox,
+                              lamda=lamda,
+                              remove_bias=remove_bias)
+        hwfit.opt = opt
+        return hwfit
+
+    def _predict(self,
```
bashtage: Same here.

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,821 @@
...
+                        b,
+                        s,
+                        m,
+                        n,
+                        max_seen),
+                    bounds=bounds[xi])
+            p[xi] = res.x
+            [alpha, beta, gamma, l0, b0, phi] = p[:6]; s0 = p[6:]
+            opt = res
+        hwfit = self._predict(h=0,
```
bashtage: Same here.

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,821 @@
...
+            [alpha, beta, gamma, l0, b0, phi] = p[:6]; s0 = p[6:]
+            #bounds = np.array([(0.0,1.0),(0.0,1.0),(0.0,1.0),(0.0,None),(0.0,None),(0.8,1.0)] + [(None,None),]*m)
+            if use_basinhopping:
+                # Take a deeper look in the local minimum we are in to find the best
+                # solution to parameters, maybe hop around to try escape the local
+                # minimum we may be in.
+                res = basinhopping(
+                    func,
+                    p[xi],
+                    minimizer_kwargs={
```
bashtage: Same here.

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,821 @@
...
+                        'args': (
```
bashtage: Same here.

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,821 @@
...
+
+            # txi [alpha, beta, gamma, l0, b0, phi, s0,..,s_(m-1)]
+            # Have a quick look in the region for a good starting place for alpha etc.
+            # using guestimates for the levels
+            txi = xi & np.array(
+                [True, True, True, False, False, True] + [False] * m)
+            bounds = np.array([(0.0, 1.0), (0.0, 1.0), (0.0, 1.0),
+                               (0.0, None), (0.0, None), (0.0, 1.0)] + [(None, None), ] * m)
+            res = brute(
+                func,
```
bashtage: Same here.

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,821 @@
...
+            endog,
+            trend=None,
+            damped=False,
+            seasonal=None,
+            m=None,
+            dates=None,
+            freq=None,
+            missing='none',
+            **kwargs):
+        super(
```
bashtage: Same here.

josef-pkt:
This pull request introduces 2 alerts - [view on lgtm.com](https://lgtm.com/projects/g/statsmodels/statsmodels/rev/pr-00054c1aa1b9af10412cafde25cc32656490f4a9)

**new alerts:**

---

statsmodels/tsa/holtwinters.py
tvanzyl: That was what that autopep8 command did.
--aggressive maybe tone it down a bit leave that off?

statsmodels/tsa/holtwinters.py
tvanzyl: Sort of same style as the AR ARMA ARIMA

bashtage:
Looking really good.  Two more-or-less final asks:

1. Add to documentation -- somewhere in tsa, perhaps right before VAR (grouping with univariate)
2. (Very helpful, but some work) Illustrate the use of these in an example notebook -- feel free to take existing examples of exp smoothing and just show how write them using your code.  If you need to you could even add data, although you can also let the notebook grab data from the web (e.g. using pandas-datareader and FRED or reading from R).  If you add data it needs to be license compatible.

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,690 @@
...
+            raise NotImplementedError('Can only dampen the trend component')
+        if self.seasoning:
+            if (m is None or m == 0):
+                raise NotImplementedError(
+                    'Unable to detect season automatically')
+            self.m = m
+        else:
+            self.m = 0
+
+    def predict(self, params, start=None, end=None):
```
bashtage: A doc string here would be helpful.

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,690 @@
...
+
+    Returns
+    -------
+    results : SimpleExpSmoothing class
+
+    Notes
+    -----
+    This is a full implementation of the simple exponential smoothing as
+    per [1].
+
```
bashtage: Perhaps a 

```
See Also
---------
Exponential Smoothing
Holt
```

(not sure if you need full path or not, there are examples in SM)

josef-pkt:
This pull request introduces 2 alerts - [view on lgtm.com](https://lgtm.com/projects/g/statsmodels/statsmodels/rev/pr-90056dd69ad9c578e80e2db9238159ec3c15cb65)

**new alerts:**

---

statsmodels/tsa/holtwinters.py
tvanzyl: done

josef-pkt:
This pull request introduces 2 alerts - [view on lgtm.com](https://lgtm.com/projects/g/statsmodels/statsmodels/rev/pr-abd98e8512b92dc81fef752e105fb3d3dc371e8d)

**new alerts:**

---

tvanzyl:
Do you know if the data contained within the R package is licence compatible?

bashtage:
Most R data files will be GPL I reckon.  The only reliable option is to use r2py, although this is a bit clunky.  I woudl suggest using some Macro-type data from FRED. Alt if the data set is here you can just load it from the web:

https://github.com/vincentarelbundock/Rdatasets

josef-pkt:
for notebooks rdataset is good.
unit tests need to work without internet connection, which is not allowed, for example in Debian

some datasets are public domain or fair use. AFAIU, the R packages cannot relicense the data itself under GPL, unless they have an agreement from an original copyright holder. This is a large grey area, e.g. I saw recently this discussion:
https://www.rd-alliance.org/group/rdacodata-legal-interoperability-ig/post/front-lines-data-licensing-research-institution

We also have quartely macrodata in the datasets. All US data published by government agencies are public domain (unless explicitly restricted access), AFAIK.

josef-pkt:
This pull request introduces 2 alerts - [view on lgtm.com](https://lgtm.com/projects/g/statsmodels/statsmodels/rev/pr-6f62d65d3694f859571d922564a105694ebf5bff)

**new alerts:**

---

josef-pkt:
This pull request introduces 2 alerts - [view on lgtm.com](https://lgtm.com/projects/g/statsmodels/statsmodels/rev/pr-57d18c3127cf0eee0ee8d1efd05e60546e8b7299)

**new alerts:**

---

tvanzyl:
Added Notebook.

ChadFulton:
Jumping in at the end here, but the code and notebook look nice to me.

- You might construct the date indexes in the notebook directly as `DatetimeIndex` objects, e.g. the first one would be `pd.DatetimeIndex(start='1996', end='2007', freq='A')`
- I notice that the IC aren't in the unit tests.

Nice job!

bashtage:
@josef-pkt This should get serious consideration for inclusion in the near term.

josef-pkt:
The notebook should be empty, i.e. have no generated output. We want to avoid increasing the repo size with autogenerated results and plots/pictures. The output would also create large changesets if some details change when the notebook is updated. 
Given that it is the last commit, amending it would still work.

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,726 @@
...
+    y_alpha = alpha * y
+    y_gamma = gamma * y
+    l[:] = 0; b[:] = 0; s[:] = 0
+    l[0] = l0; b[0] = b0; s[:m] = s0
+    return alpha, beta, gamma, phi, alphac, betac, gammac, y_alpha, y_gamma
+
+#(,M)
+
+
+def _holt_win__mul(x, xi, p, y, l, b, s, m, n, max_seen):
```
josef-pkt: add at least minimal docstrings also to private function
eg. """Recursion for multiplicative Holt-Winters"""

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,726 @@
...
+        return max_seen
+    for i in range(1, n):
+        l[i] = (y_alpha[i - 1] / s[i - 1]) + \
+            (alphac * (l[i - 1] + phi * b[i - 1]))
+        b[i] = (beta * (l[i] - l[i - 1])) + (betac * phi * b[i - 1])
+        s[i + m - 1] = (y_gamma[i - 1] / (l[i - 1] + phi *
+                                          b[i - 1])) + (gammac * s[i - 1])
+    return sqeuclidean((l + phi * b) * s[:-(m - 1)], y)
+
+#(M,M) (Md,M)
```
josef-pkt: pep-8 space after comment `#`

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,726 @@
...
+    for i in range(1, n):
+        l[i] = (y_alpha[i - 1]) - (alpha * s[i - 1]) + \
+            (alphac * (l[i - 1] * b[i - 1]**phi))
+        b[i] = (beta * (l[i] / l[i - 1])) + (betac * b[i - 1]**phi)
+        s[i + m - 1] = y_gamma[i - 1] - \
+            (gamma * (l[i - 1] * b[i - 1]**phi)) + (gammac * s[i - 1])
+    return sqeuclidean((l * phi * b) + s[:-(m - 1)], y)
+
+
+class HoltWintersResults(Results):
```
josef-pkt: docstring missing

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,726 @@
...
+
+    def predict(self, start=None, end=None):
+        return self.model.predict(self.params, start, end)
+
+    def forecast(self, steps=1):
+        try:
+            start = self.model._index[-1] + 1
+            end = self.model._index[-1] + steps
+            return self.model.predict(self.params, start=start, end=end)
+        except:
```
josef-pkt: What's the specific exception that this captures?
We never use a unspecified except, since it hides everything.

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,726 @@
...
+        seasonal = self.seasonal
+        m = self.m
+        opt = None
+        phi = phi if damped else 1.0
+        n = len(data)
+        if use_boxcox == 'log':
+            lamda = 0.0
+            y = boxcox(data, 0.0)
+        elif use_boxcox:
+            y, lamda = boxcox(data)
```
josef-pkt: What's the default method for choosing lambda?
This might need a `use_boxcox = a_float` options, so users can choose the lambda

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,726 @@
...
+            b[:i] = dampen(b[:i], phi)
+            b[i:] = dampen(b[i], phi_h)
+            fitted = trended(l, b)
+        level = l[:i].copy()
+        if use_boxcox or use_boxcox == 'log':
+            fitted = inv_boxcox(fitted, lamda)
+            # TODO: Does it make sense to have a inv_boxcox transform of the level and trend?
+            #level = inv_boxcox(level, lamda)
+            #slope = inv_boxcox(slope, lamda)
+            # TODO: How do we deal with the inv_boxcox transform on negative seasonal components?
```
josef-pkt: my guess here is that seasonal and other components need to be computed as the difference between inv_boxcox transformed sums.
e.g. `seasonal_effect = fitted - inv_boxcox(fitted_without_seasonal)`
(just a guess because I haven't gone through much of the PR/algorithm yet.)

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,726 @@
...
+    ----------
+    endog : array-like
+        Time series
+    trend : {"add", "mul", None}, optional
+        Type of trend component.
+    damped : bool, optional
+        Should the trend component be damped.
+    seasonal : {"add", "mul", None}, optional
+        Type of seasonal component.
+    m : int, optional
```
josef-pkt: needs more informative name
`season_length` or whatever is used in statespace (not `freq`)

statsmodels/tsa/tests/test_holtwinters.py
```diff
@@ -0,0 +1,116 @@
...
+        
+    def test_forecast(self):
+        fit1 = ExponentialSmoothing(self.aust, m=4, trend='add', seasonal='add').fit()
+        assert_almost_equal(fit1.forecast(steps=4), [60.9542,36.8505,46.1628,50.1272], 3)
+        
+    def test_simple_exp_smoothing(self):
+        fit1 = SimpleExpSmoothing(self.oildata_oil).fit(0.2,optimized=False)
+        fit2 = SimpleExpSmoothing(self.oildata_oil).fit(0.6,optimized=False)
+        fit3 = SimpleExpSmoothing(self.oildata_oil).fit()
+        assert_almost_equal(fit1.forecast(1), [484.802468], 4)
```
josef-pkt: in general we prefer no `assert_allclose` because it is more precise and flexible to set atol and rto instead of decimals.
However, assert_almost_equal looks appropriate here with fixed print precision in reference numbers.

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,726 @@
...
+    alphac = 1 - alpha
+    betac = 1 - beta
+    y_alpha = alpha * y
+    l[:] = 0; b[:] = 0;
+    l[0] = l0
+    b[0] = b0
+    return alpha, beta, phi, alphac, betac, y_alpha
+
+
+def _holt__(x, xi, p, y, l, b, s, m, n, max_seen):
```
josef-pkt: why two trailing underscores?

tvanzyl: It was to help organise it when I was coding.
The spaces between underscores helped with knowing which components of the exponential smoothing model are left out.

statsmodels/tsa/holtwinters.py
tvanzyl: done

statsmodels/tsa/holtwinters.py
tvanzyl: done

statsmodels/tsa/holtwinters.py
tvanzyl: done

statsmodels/tsa/holtwinters.py
tvanzyl: done

statsmodels/tsa/holtwinters.py
tvanzyl: done

josef-pkt:
This pull request introduces 2 alerts - [view on lgtm.com](https://lgtm.com/projects/g/statsmodels/statsmodels/rev/pr-3fa6b7ffb6f97789c984bd1dc725527730f6a913)

**new alerts:**

---

statsmodels/tsa/holtwinters.py
tvanzyl: done

tvanzyl:
@ChadFulton

statsmodels/tsa/holtwinters.py
ChadFulton: I think I see what you mean, but (as are as I can tell) the number of underscores isn't consistently used across the functions, so maybe it's best just to eliminate underscores at the end and eliminate double underscores?

ChadFulton:
Looks pretty good to me. Happy to help out with more review if necessary, anything in particular that needs to be done?

josef-pkt:
If possible, I would like to merge this as experimental for 0.9 (given the long standing demand for it).

@ChadFulton The main review is to check whether there are any "obvious" problems or needed refactorings. As "experimental" we can still make changes and enhance later. For example, I didn't check whether datetime support is consistent with other tsa models. Another one is performance evaluation and to check whether it is urgent to cythonize the filtering loops. Both can be postponed if necessary IMO.

ChadFulton:
Great, I'll check the date time and the performance and see if anything else occurs to me.

ChadFulton:
> The implementaion of the library follows as per the R library as much as possible whilest still being pythonic.

Just to check, this means that you followed the R API as much as possible, not that you translated the (GPL licensed) R code, correct?

ChadFulton:
Again, very nice work! As to the two specific things I was investigating:

1. Datetime support seems to be fine (uses the inherited methods as intended, so that seems like it should work, and nothing was problematic in my investigation)
2. Performance: I think performance is okay for now. It takes a noticeable amount of time for longer datasets, especially with seasonal effects, but right now this seems to mostly be measured in seconds, not minutes.

Here are some other thoughts. Some of these are important now (e.g. the bugs and test misses), others could be postponed or ignored.

## Comments

### Framework

It might be nice to structure in the same way as state space:
- `filter` method performs the iterations for a given set of parameters and returns a `FilterResults` object with `level`, `slope`, `seasonal`, `forecasts`, `forecasts_error` (and maybe more) attributes.
- `loglike`, `sse`, `mse`, `amse`, `mae` functions are objective functions for a given set of parameters
- `start_params` is a property that chooses basic starting parameters and initialization values (e.g. put the `brute` call in here)
- `fit` performs the actual fitting, returns a `ExponentialSmoothingResults` object (that extends from `FilterResults`)

This will be nice if we ever want to move to Hyndman et al.'s more general structure in the "Forecasting with Exponential Smoothing" book.

### General structure

- Equations should be provided in some docstring: level, growth, seasonal, forecast.
- It looks to me like `ExponentialSmoothing._predict` contains a separate implementation of the procedures implemented in the `_holt_*` functions?
- Should have a `summary` output table (two tables maybe, one for the paramters `alpha, beta, gamma, phi`, and one for the initial values `l0, b0, s0, ..., s_(m-1)`.
- `params` should be an array or pd.Series that only consists of the actual parameters used in the model; for example in simple exponential smoothing, it should only include `alpha` and `l0`.
- Box-cox transformations should maybe be separate? Is there any reason it needs to be part of the model per se?
- Results object - should not set results values using keyword arguments passed to base `Results` constructor?

### Naming

- Arguments: maybe use "additive" and "multiplicative" instead of "add" and "mul"? Or maybe allow both?
- Although I know we're trying to mirror Hyndman's package, we should consider using more user-friendly names; e.g. instead of `alpha, beta, gamma`, maybe `smoothing_level, smoothing_slope, smoothing_seasonal, damping_slope`, and instead of `l0, b0, s_i` maybe `initial_level, initial_slope, initial_seasons`.
- `season_length` argument - probably `seasonal_periods` for consistency with `UnobservedComponents`
- Sometimes you use `n`, it would be more like other models if you set and used `self.nobs`

### Bugs

- Need to double-check the indexes in the `ExponentialSmoothing._predict` function. I think the `level` and `slope` variables be e.g. `level = l[1:i+1].copy()` rather than `level = l[:i].copy()`, since `l[0]` is the initialization value while `l[i]` should be the level associated with the last observed value. It seems like the other indexes are okay, but I haven't worked it out for sure.
- Docstring for `Holt` is not up to date.
- We should raise exceptions for the invalid models - e.g. multiplicative trend when data has negative or zero values; `ets` says "Inappropriate model for data with negative or zero values" but the current implementation runs the recusions resulting in e.g. the level going to infinity, or else infs and nans show up.

### Tests

- Some of the tests are pretty low precision - is it possible to improve this? Also would be nice if there was an R script that generated the results.
- Should have tests for all values (e.g. `level`, `slope`, `seasonal`, information criteria, etc. - these would have caught the bug above).

### Miscellaneous

- The `remove_bias` argument should maybe be explained better in docstring? I guess it enforces average residual equal to zero, but that should be documented.

tvanzyl:
@ChadFulton No code was taken from the R implementation. It used to follow a similar API to the R version, but we have since completely restructured to be more object oriented. So I think we should probably take that that statement out the description.

ChadFulton:
> No code was taken from the R implementation.

Great!

statsmodels/tsa/holtwinters.py
```diff
@@ -359,8 +359,8 @@ class ExponentialSmoothing(TimeSeriesModel):
     -----
     This is a full implementation of the holt winters exponential smoothing as
     per [1]. This includes all the unstable methods as well as the stable methods.
-    The implementaion of the library follows as per the R library as much as possible
-    whilest still being pythonic.
+    The implementation of the library covers the functionality of the R 
+    library as much as possible whilest still being pythonic.
```
bashtage: while or whilst (no e)

tvanzyl: fixed

tvanzyl:
**Framework**

_It might be nice to structure in the same way as state space:_
(Following the ARIMA framework, does a fit and then predict and passes the params)
(Maybe need to consolidate to single structure later)

**General structure**

_Equations should be provided in some docstring: level, growth, seasonal, forecast._
Its a lot of equations all of these https://www.otexts.org/sites/default/files/fpp/images/Table7-8.png

_It looks to me like ExponentialSmoothing._predict contains a separate implementation of the procedures implemented in the _holt_* functions?_
(The _holt_ are optimised for SSE not for getting the actual values)

_Should have a summary output table_
(next round)

_params should be an array or pd.Series_
(next round)

_Box-cox transformations should maybe be separate? Is there any reason it needs to be part of the model per se?_
(happy to take it out, was part of Hyndman's  package, wanted to get similar results to his, but yes would be simpler, pity its not available, elsewhere)

_Results object - should not set results values using keyword arguments passed to base Results constructor?_
(was copying the framework of ARIMA)

**Naming**

_Arguments: maybe use "additive" and "multiplicative" instead of "add" and "mul"? Or maybe allow both?_
(Done)

Although I know we're trying to mirror Hyndman's package, we should consider using more user-friendly names; e.g. instead of alpha, beta, gamma, maybe smoothing_level, smoothing_slope, smoothing_seasonal, damping_slope, and instead of l0, b0, s_i maybe initial_level, initial_slope, initial_seasons.

season_length argument - probably seasonal_periods for consistency with UnobservedComponents
Sometimes you use n, it would be more like other models if you set and used self.nobs
(Done)

**Bugs**

_Need to double-check the indexes_
(Done)

_Docstring for Holt is not up to date._
(Done)

_We should raise exceptions for the invalid models_ 
(Done)

**Tests**

_Some of the tests are pretty low precision - is it possible to improve this? Also would be nice if there was an R script that generated the results._
(Used the values from the text book as unit-test. In most cases the scipy optimizations are getting better SSE than the R, so then being in the same region of values is as good a comparison as we can check.)

_Should have tests for all values (e.g. level, slope, seasonal, information criteria, etc. - these would have caught the bug above)._
(Done)

**Miscellaneous**

_The remove_bias argument should maybe be explained better in docstring?_ 
(Done)

josef-pkt:
just one item:

I would leave box-cox transformation in as part of the options.
#2925 We started to implement box-cox more generally, but it's not reused in models or forecasting functions yet.
I haven't looked closely enough here to tell whether it would be easy to connect or reuse.

josef-pkt:
This pull request introduces 2 alerts - [view on lgtm.com](https://lgtm.com/projects/g/statsmodels/statsmodels/rev/pr-730dd4849f843afbbcd6e634e6e7698e4619a570)

**new alerts:**

---

Juanlu001:
Would love to see this included in statsmodels! :) cc @lpcampoy

ChadFulton:
Thanks for addressing all those things @tvanzyl, and I'm sorry about my delay in response!

My general feeling is as follows: the underlying functionality is clearly very good and desirable to have, but I personally don't see the way it is structured as being very natural in the context of usual Statsmodels models.

@josef-pkt What are your thoughts on the general structure? My comments are essentially those from "Framework" and "General structure" in my post above. I personally still think those are desirable and would improve this model quite a bit.

I see nothing wrong with merging essentially as-is (although maybe I'll take another run through as a double-check that the model results continue to be right) to ensure the functionality gets in, but in my mind it would be best to immediately introduce a new PR to refactor things (I would be happy to be the one to do this refactoring PR, but I don't want to step on anyone's toes).

@tvanzyl what do you think - are you opposed in principle to my suggestions (and I don't mind if you are opposed)? or do you just feel like we should get this in now and worry about the structure later?

(Trying to keep things moving here :)

josef-pkt:
I started initially a comment, but deleted it before submitting because I don't have a strong opinion.

I see the advantages of bringing it closer to our usual general structure of models. On the other hand the main purpose, AFAIU, is to have a simple reasonably flexible way to forecast. So, any structure is fine with me if it makes it easy to get/estimate/calibrate the model and forecast.

It might be a bit similar to seasonal_decompose versus for example unobserved components. seasonal_decompose doesn't require much thinking about the specification and works in many cases.
e.g. questions and comments here
https://stackoverflow.com/questions/47076771/statsmodels-seasonal-decompose-what-is-naive-about-it

Even if it is initially just some forecasting function, we might eventually (!?) go to full models in this area, e.g. also #2892, that are either fancier (more complete, more options) versions of the forecasting functions or something that simpler forecasting functions could delegate to.
I'm not sure if we want to have, for example, inference available for those kind of function if they are just build up in pieces and e.g. we don't have an equivalent MLE defined for it. I don't know enough about those models, but one possibility is that they are just combinations of "ad-hoc" types of optimized parts which work fine for forecasting but don't necessarily specify a correct loglike/llf.
(e.g. seasonal decompose is just a combination of convolutions.)

josef-pkt:
I would like to merge this this week, especially because of the high demand for this.
(I'm not planning on a full review but try to play with the notebook.)

tvanzyl:
@ChadFulton Sorry, not enough time around exams to tackle all of this. Happy to get the functionality in as I just wanted to have something in a statsmodels release to use in my course next year.
Happy to work with you on the PR. You know the framework better than I. 
Also want to collaborate on a state space based implementation in then new year. I believe it will fit with the statsmodels framework and will make it easier to get confidence intervals etc.

josef-pkt:
This pull request introduces 2 alerts - [view on lgtm.com](https://lgtm.com/projects/g/statsmodels/statsmodels/rev/pr-1b9a6016ef1d093c13d862e57594175b1059b8a0)

**new alerts:**

---

josef-pkt:
The notebook looks very good.

(I will add some comments here as a try out a few things in the notebook)

aic and similar should be lower case, see eg. RegressionResult

josef-pkt:
from Chad's comment

>Although I know we're trying to mirror Hyndman's package, we should consider using more user-friendly names; e.g. instead of alpha, beta, gamma, maybe smoothing_level, smoothing_slope, smoothing_seasonal, damping_slope, and instead of l0, b0, s_i maybe initial_level, initial_slope, initial_seasons.

I have no idea what any of the Greeks are. Even if we keep those greek letters for now, then they need to be explained somewhere, like a language dictionary. I only checked one docstring and it doesn't mention what each greek parameter stands for.

examples/notebooks/exponential_smoothing.ipynb
```diff
@@ -0,0 +1,483 @@
...
+   "execution_count": null,
+   "metadata": {
+    "ExecuteTime": {
+     "end_time": "2017-12-06T09:09:03.149848Z",
+     "start_time": "2017-12-06T09:08:54.296071Z"
+    }
+   },
+   "outputs": [],
+   "source": [
+    "fit1 = ExponentialSmoothing(aust, seasonal_periods=4, trend='add', seasonal='add', use_boxcox=True).fit()\n",
```
josef-pkt: incorrect usage
`use_boxcox` has no effect here.
It took me a while to realize that it has to be a fit keyword.

josef-pkt:
params is a dict, which is inconsistent with our methods in general.
I think it's ok in this case given that we don't inherit anything besides datahandling and predict.

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,858 @@
...
+    The implementation of the library covers the functionality of the R 
+    library as much as possible whilst still being pythonic.
+
+    References
+    ----------
+    [1] Hyndman, Rob J., and George Athanasopoulos. Forecasting: principles and practice. OTexts, 2014.
+    """
+
+    def __init__(self, endog, trend=None, damped=False, seasonal=None,
+                 seasonal_periods=None, dates=None, freq=None, missing='none', **kwargs):
```
josef-pkt: I think we should remove the `**kwargs` here. AFAICS, they are not used in this case, and just swallow without feedback any incorrect keywords

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,858 @@
...
+        seasoning = self.seasoning
+        trending = self.trending
+        trend = self.trend
+        seasonal = self.seasonal
+        m = self.seasonal_periods
+        phi = phi if damped else 1.0
+        if use_boxcox == 'log':
+            lamda = 0.0
+            y = boxcox(data, 0.0)
+        elif use_boxcox:
```
josef-pkt: this is missing the case when lamda is already a float.

josef-pkt:
general pattern: only the results should know about fit results.
Currently the model has params attached, which I guess is unused.

having use_boxcox as fit argument and not as model argument might be a bit fragile. I haven't managed yet to think may way through boxcox forecasting in the code.
There is currently duplicated computation of the boxcox transformation in fit and _predict, AFAICS. Also each call to _predict does the full boxcox transformation


should `slope` be renamed to `trend`, to follow eg seasonal_decompose and some statespace models, or is this something different?

josef-pkt:
general, theoretical question:

I was surprised in the plot that seasonal component is not centered at zero.
How is the constant included or split up between componenets? Is it a fixed rule or does it depend on how the recursive structure or time series iteration is initialized?

ChadFulton:
> @ChadFulton Sorry, not enough time around exams to tackle all of this. Happy to get the functionality in as I just wanted to have something in a statsmodels release to use in my course next year.
Happy to work with you on the PR. You know the framework better than I.
Also want to collaborate on a state space based implementation in then new year. I believe it will fit with the statsmodels framework and will make it easier to get confidence intervals etc.

Thanks for getting all of this setup. We can move slowly, but we're glad to get it in!

> I would like to merge this this week, especially because of the high demand for this.

I agree, great!

examples/notebooks/exponential_smoothing.ipynb
tvanzyl: removed the **kwargs should ensure we don't let this through in future

statsmodels/tsa/holtwinters.py
tvanzyl: done

statsmodels/tsa/holtwinters.py
tvanzyl: done

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,858 @@
...
+                season = (fitted - inv_boxcox(trend, lamda))[:i]
+            elif seasonal == 'mul':
+                season = (fitted / inv_boxcox(trend, lamda))[:i]
+            else:
+                pass
+        SSE = sqeuclidean(fitted[:-h - 1], data)
+        # (s0 + gamma) + (b0 + beta) + (l0 + alpha) + phi
+        k = m * seasoning + 2 * trending + 2 + 1 * damped
+        AIC = self.nobs * np.log(SSE / self.nobs) + (k) * 2
+        AICc = AIC + (2 * (k + 2) * (k + 3)) / (self.nobs - k - 3)
```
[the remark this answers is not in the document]
tvanzyl: done

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,858 @@
...
+                       'gamma': gamma,
+                       'phi': phi,
+                       'l0': l0,
+                       'b0': b0,
+                       's0': s0,
+                       'use_boxcox': use_boxcox,
+                       'lamda': lamda,
+                       'remove_bias': remove_bias}
+        hwfit = HoltWintersResults(self, self.params, fittedfcast=fitted, fittedvalues=fitted[:-h - 1],
+                                   fcast=fitted[-h - 1:], SSE=SSE, level=level,
```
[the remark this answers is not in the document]
tvanzyl: done

statsmodels/tsa/holtwinters.py
```diff
@@ -0,0 +1,858 @@
...
+                # solution to parameters
+                res = minimize(func, p[xi], args=(
+                    xi, p, y, l, b, s, m, self.nobs, max_seen), bounds=bounds[xi])
+            p[xi] = res.x
+            [alpha, beta, gamma, l0, b0, phi] = p[:6]
+            s0 = p[6:]
+            opt = res
+        hwfit = self._predict(h=0, alpha=alpha, beta=beta, gamma=gamma, phi=phi, l0=l0,
+                              b0=b0, s0=s0, use_boxcox=use_boxcox, lamda=lamda, remove_bias=remove_bias)
+        hwfit.opt = opt
```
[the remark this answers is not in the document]
tvanzyl: done

josef-pkt:
This pull request introduces 2 alerts - [view on lgtm.com](https://lgtm.com/projects/g/statsmodels/statsmodels/rev/pr-3c55368340cbc1c3f77e1d6c8b342c5aa7ac5c92)

**new alerts:**

---

josef-pkt:
So far, I only looked at the latest changes, and they look good to me. I will try again later today.

I think the greek translation should also go in the docstring, either in notes or in explanation for each parameter. It is useful to have the link between statsmodels names and shortcut names used in the reference(s) if there is a dominant reference or consistent Letters/Greeks across references.
        alpha = smoothing_level
        beta = smoothing_slope
        gamma = smoothing_seasonal
        phi = damping_slope

One thing I think I forgot to mention: Result statistics like aic can go into the results class now. They can be computed lazily on demand, see for example linear_model.RegressionResults. fittedvalues and similar are computed in the fit method and cannot be moved to the results class as in RegressionResults.

One more that I did't finish looking at yesterday: Should we already have optimizer options in `fit` in case the default doesn't work, e.g. just for the `minimize` case, e.g. `minimize_kwds` as extra dict in fit arguments?

josef-pkt:
To check that the notebook examples work with a different dataset, and it does. The forecast looks good. The only time I needed to spend is on getting the pandas index.

with classical airpassenger dataset, which has increasing seasonal amplitude

```
df_ap = pd.read_csv("AirPassengers.csv")
index = pd.DatetimeIndex(start='1949', end='1960-M12', freq='MS')
ap = pd.Series(df_ap['AirPassengers'].values, index)

res_ap = ExponentialSmoothing(ap, seasonal_periods=12, trend='add', seasonal='add').fit(use_boxcox=True)

ax = ap.plot(figsize=(10,6), marker='o', color='black', title="Forecasts from Holt-Winters' multiplicative method" )
ax.set_ylabel("International visitor night in Australia (millions)")
ax.set_xlabel("Year")
res_ap.fittedvalues.plot(ax=ax, style='--', color='red')
res_ap.forecast(36).plot(ax=ax, style='--', marker='o', color='red', legend=True)
print("Figure 7.6: Forecasting airpassenger using Holt-Winters method with both additive and multiplicative seasonality.")
plt.show()
```

josef-pkt:
A `plot_forecast` method would be nice for quick inspection. AFAIR, we have it in some other models.

I think I'm stopping with reviewing. It seems to work well out of the box and we can add or change things during the next release cycle.
We need to rebase this and push the rebased version to a new PR so the history in this PR stays unchanged. There are a few commits that need squashing, especially those with the full notebook output. I can do the interactive rebase, unless you are familiar enough with git to do it.

josef-pkt:
for minimal squashing:
"Added a jupyter notebook  Oct 5 adds" full notebook
second commit after "fixed the unit tests and cleared the jupyter notebook" clears notebook

"fixed the notebook with the latest changes to the api as requested" on Dec 6 changes to full notebook
next commit clears notebool

josef-pkt:
I merged the rebased version in PR #4176

@tvanzyl  Thanks for this extension, many users will be glad this is in
@bashtage @ChadFulton Thanks for reviewing

It's open for extensions and some refactoring for the next release.

Juanlu001:
Awesome! When is the next stable release? :)

josef-pkt:
> When is the next stable release?

My guess is now mid January.

PR #3859, opened 2017-08-07
Caching in omega_forc_cov

jbrockmendel:
`VARResults._omega_forc_cov` cached `bpow` as an optimization, but that turns out not to be a major hotspot in this function.  This PR implements caching for the expensive `trace` and `chain_dot` operations.

According to past-me (#3630), in existing test cases this gives speedups of 15% for `steps=3` up to 59% by `steps=15`.


jbrockmendel:
The appveyor log doesn't seem to have any error message, just a "exited with status 1".  Any idea what the issue is?

PR #3862, opened 2017-08-07
Bugfix for pacf_ols (issue #3839)

nikhase:
see #3839 

* For pacf with "method=ols", results differ quite a lot from
"method={yw, ld}" (biased and unbiased setting)

* Demeaning and minor changes in the OLS parameters improves the
result for smaller time series (emprirically evaluated)

* Added "unbiased" parameter for ols

* Added reference in docstring

### DIFF
--- a/statsmodels/tsa/stattools.py
+++ b/statsmodels/tsa/stattools.py
@@ -520,9 +520,8 @@ def pacf_yw(x, nlags=40, method='unbiased'):
     return np.array(pacf)
 
 
-#NOTE: this is incorrect.
-def pacf_ols(x, nlags=40):
-    '''Calculate partial autocorrelations
+def pacf_ols(x, nlags=40, unbiased=True):
+    '''Calculate partial autocorrelations via OLS
 
     Parameters
     ----------
@@ -539,20 +538,29 @@ def pacf_ols(x, nlags=40):
     Notes
     -----
     This solves a separate OLS estimation for each desired lag.
-    '''
-    #TODO: add warnings for Yule-Walker
-    #NOTE: demeaning and not using a constant gave incorrect answers?
-    #JP: demeaning should have a better estimate of the constant
-    #maybe we can compare small sample properties with a MonteCarlo
-    xlags, x0 = lagmat(x, nlags, original='sep')
-    #xlags = sm.add_constant(lagmat(x, nlags), prepend=True)
-    xlags = add_constant(xlags)
-    pacf = [1.]
-    for k in range(1, nlags+1):
-        res = OLS(x0[k:], xlags[k:, :k+1]).fit()
-         #np.take(xlags[k:], range(1,k+1)+[-1],
 
-        pacf.append(res.params[-1])
+    References
+    ----------
+    .. [1] Box, G. E., Jenkins, G. M., Reinsel, G. C., & Ljung, G. M. (2015).
+    Time series analysis: forecasting and control.
+    John Wiley & Sons, p. 66
+
+    '''
+    n = len(x)
+    # Demean the series
+    x = x - np.mean(x)
+    # Create lags for multivariate OLS
+    xlags, x0 = lagmat(x, nlags, original='sep')
+    pacf = [1.]
+    for k in range(1, nlags + 1):
+        res = OLS(x0, xlags[:, :k]).fit()
+        # Last coefficient corresponds to PACF value (see [1])
+        coeff = res.params[-1]
+        if unbiased:
+            coeff = coeff * n / (n - k)
+
+        pacf.append(coeff)
+
     return np.array(pacf)
 
 


nikhaas:
Well, seeing travis' results, I haven't seen the tests in the first place. Let's discuss the suggested fix at first.

josef-pkt:
I didn't think about the details yet.

Where does the bias correction come from, is this in Box Jenkins. (I don't remember much about bias corrections for (auto-)correlation coefficients)

Given the comments by Skipper and past changes, we might have to make this optional. AFAIR, Skipper had changed away from demeaning to match the Stata results.
In the plot for nobs=100 and 250, "prior_ols_implementation" is closer to the true zero pacf at medium lags, than the others. I don't know if this will be true on average. 
(If the initial observation have some startup behavior, then dropping those observations might provide a better estimate of the long run pacf function.)

josef-pkt:
@nikhaas general question:
What is your motivation of using pacf_ols if/when it produces the same results as the other methods?

We had a discussion many years ago, but I never saw much of a difference, with the exception that the existing ols version uses a different assumption on the mean.

jbrockmendel:
+1 for fixing this.  That comment `#NOTE: this is incorrect` is really bothersome.

josef-pkt:
> That comment #NOTE: this is incorrect is really bothersome.

I think that comment refers to the initial version which, AFAIU, is the same as the version proposed here.
or it refers to the append=False versus append=True change in default of add_constant.

jbrockmendel:
> I think that comment refers to the initial version which, AFAIU, is the same as the version proposed here.

I continue to offer positive reinforcement.

nikhaas:
About my motivation: I need `pacf` for an implementation in [tsfresh](https://github.com/blue-yonder/tsfresh/issues/223). I stumbled over the very different results for OLS and Yule-Walker in the first place, which was unsatisfying to me. 

I am particularly interested because

* I definetely think this is a bug
* I aggree with @jbrockmendel that this comment is really bothersome
* imho, OLS is more intuitive than Yule-Walker or Levinson Durbin, it is easier to get what is going on
* Matlab says it uses this OLS approach internally, and many open source libraries try to mimic Matlab
* method = Yule-Walker is the slowest option (but default anyway). Levinson recursion is much faster, and a short internal investigation has shown, that the (old) OLS implementation is competitive to Levinson-Recursion (OLS missing in this figure, but same as LD):

comparison pacf methods

For the bias correction: Partial Autocorrelation is not so straightforward (and mainstream) as the ACF, and literature is more sparse. But for ACF, the denominator in the following formula is performing the bias correction:

ACF autocorrelation

So all i did was "mimicking" this for PACF, and the results correspond to the unbiased and biased options (`method={"ld", "ldb"} )

All I did was following straightforward the proposal from the Box-Jenkins book:

> The partial autocorrelations may be estimated by fitting successively autoregressive models of orders 1, 2, 3, ... by least squares and picking out the estimates $\phi_{11}$, $\phi_{22}$, $\phi_{33}$ ... of the last coefficient fitted at each stage.

I already spend a few hours on theoretical research, and I cannot spend more time on this. I neither have (full) access to the literature, so from a theoretical POV, this is the best I can propose.

josef-pkt:
to make sure about the comparison with Stata, I checked a quick example
(data from a coint test csv file)

We agree at floating point precision for OLS with Stata, This means `pacf_ols` was correctly rewritten to match Stata. 

`tsat.pacf_ols(df_coint.values[:,1])[:15]`

difference pacf_ols versus Stata (with added 1 at zero lag:
`[0.0, 0.0, -1.4988010832439613e-15, -1.5456386170953351e-15, -4.067059189427624e-15, -1.5890067039947553e-15, 9.055256544598933e-16, 8.378714388967978e-16, -1.2212453270876722e-15, 1.3461454173580023e-15, -1.0408340855860843e-15]
`

```
>>> np.max(np.abs(pf[:11] - r0))
4.0670591894276242e-15
```

This means it's not a BUG, adding more options would be an enhancement, which is no problem.
Separate decision is whether to change the default in `pacf_ols` which would be a backwards incompatible redefinition. I'm not sure we should do that, but even though we verify a lot against Stata, we don't always use the same defaults, (if we don't like them, i.e. we have a good reason to use a different default). Without backwards compatibility consideration, I would like the new and original demean version better.

nikhaas:
Thanks for validating with a quick example. Is the `df_coint` dataframe public? No association comes into my mind when I read
> coint test file

I managed to get access to Matlab and - surprisingly to me - their results coincide with the already implemented OLS results. It surprises me, because from my experience, the methods `LD` and `YW` agree, whereas `OLS` shows a consistent difference, which, imho, should not be the case. Just for info: in Matlabs `parcorr`, they use the [qr decomposition](https://de.mathworks.com/help/matlab/ref/qr.html) to solve the least squares regression.

See the following plot (with my well-known AR process):
pacf_yw_ld_ols_vs_matlab_len_25
pacf_yw_ld_ols_vs_matlab_len_100

### Conclusion
* I really do not know the reason for the consistent difference between LD/YW and OLS
* In this context, I do not know if the OLS solution is biased / unbiased and how to correct for it
* I propose the switch to Levinson-Durbin Recursion for default, because it yields the same results as Yule-Walker, but seems to be as fast as OLS. Whether unbiased (`ld`) or biased (`ldb`) might be discussed

josef-pkt:
Thanks for checking against matlab

Now that the "correctness" issue is out of the way:

I still like the proposed implementation, and "When in doubt, add options." is my favorite policy for this kind of decisions.
My guess is that neither Matlab nor Stata is based on a strong reason, it might be just run the default OLS regression and take the last coefficient. Stata adds a constant by default.

There is a small sample bias in OLS for estimating coefficients for the lagged variables, I just don't remember a reference nor the details for the bias correction (and I thought I remember that the coefficient is downward and not upward biased). The nice feature of your bias correction, at least based on your plots, is that it prevents the higher lag pacf to increase. I have seen that problem before and seen it mentioned in some reference or blog, which can make pacf at higher lags misleading.

It's also nice that OLS can replicate the other versions based on autocorrelation, acf.

So, the two questions are
- how to we add the options?
- do we change defaults in a backwards incompatible way?

nikhaas:
> It's also nice that OLS can replicate the other versions based on autocorrelation, acf.

Can you explain this in more detail? I do not get what you want to say.

In the end, I do not want to *disimprove* anything. What are the major differences to the prior OLS implementation:
1. demean the series at first
2. do not use a constant
3. for a lag k and lagmat `xlags`, I use `xlags[:,:k]` instead of `xlags[k:, :k]`, i.e. the preceeding `0`s are taken into account from OLS (if it does not reject them internally)
4. I do the self-defined, heuristic bias correction `coeff = coeff * n / (n - k)`

---
>  The nice feature of your bias correction, at least based on your plots, is that it prevents the higher lag pacf to increase.

How is this possible? ` n / (n - k) > 1`, so `coeff` should be increased with the correction. This might be an effect caused in cooperation with 3.

---
Imho, the only robust suggestions that are left from the discussion are:

* disambiguate the comments in `pacf_ols`
* Make default to `ld` because it is faster than `yw` 
---
One last investigation: AR(1) and `np.random.randn` with 20 and 100 points respectively

pacf_new_ols_ar_and_randn

josef-pkt:
Maybe I misinterpreted some plots, and I need to go over it again when I'm less distracted.

Aside: my coint csv file was the first time series file in my stata work directory that I found. Based on the file name, I think it is the write_csv of https://github.com/statsmodels/statsmodels/pull/3146/files#diff-8d4eb599f6121222f9a183e3a7205f9bR269

jbrockmendel:
it Looks like this reached a concensus that the comment about the function being wrong is itself incorrect.  Is this acccurate?

josef-pkt:
There is nothing "incorrect". The current code corresponds to the intended definitions (which follow other packages).
However, we might want to add another ols version based on a different definition/algorithm.

jbrockmendel:
Woah there cowboy.  Reflexive disagreement is a great heuristic, I get it, but I'm literally referring to a comment that says 

```
#NOTE: this is incorrect
```

So yes.  Something is incorrect.  Either the code or the comment itself.  Unless the comment is referring to itself and we've

josef-pkt:
I was referring to code and definitions, not docs or code comments. It's and ENH issue and not a BUG issue.

jbrockmendel:
Right, and I was referring to the comment (the way you can tell is by reading what I wrote "that the comment about the function being wrong is itself incorrect") because someone might accidentally read it and get the wrong idea.  If the function is correct, then this is anti-documentation.

jbrockmendel:
So... the conclusion here is that "X" and "not X" are not mutually exclusive, so all of ZFC needs to be scrapped?

statsmodels/tsa/stattools.py
```diff
@@ -539,20 +538,29 @@ def pacf_ols(x, nlags=40):
...
+    # Create lags for multivariate OLS
     xlags, x0 = lagmat(x, nlags, original='sep')
-    #xlags = sm.add_constant(lagmat(x, nlags), prepend=True)
-    xlags = add_constant(xlags)
     pacf = [1.]
-    for k in range(1, nlags+1):
-        res = OLS(x0[k:], xlags[k:, :k+1]).fit()
-         #np.take(xlags[k:], range(1,k+1)+[-1],
+    for k in range(1, nlags + 1):
+        res = OLS(x0, xlags[:, :k]).fit()
```
bashtage: I like this PR but would make much more sense to avoid OLS and just call either `pinv` or `lstsq` directly .  OLS computes a ton of stuff that isn't needed.

jbrockmendel:
@nikhase needs rebase.

josef-pkt:
see https://github.com/statsmodels/statsmodels/pull/3862#issuecomment-381403526

This PR cannot be merged because it changes the algorithm/definition and looses Stata (and matlab) and backwards compatibility, and it's not a BUG fix.

bashtage:
I think it is good to go. Thanks.

PR #3886, opened 2017-08-26
refactor descriptivestats.__main__ into test_descriptivestats

jbrockmendel:
Rebases and replaces #3732


jbrockmendel:
It looks like the test errors here are errors in the tests that are currently in `descriptivestats.__main__` and do not currently get run.  So instead of being just a refactor/cleanup, this is now a bughunt.

jbrockmendel:
`xfail`ed the error in `test_descriptivestats` in py3.  The remaining travis error is `test_arima111_predict_exog_2127`

jbrockmendel:
False-positive test errors a higher priority than cleaning up cruft.  Closing.

PR #3893, opened 2017-08-29
Mark smoke tests

jbrockmendel:
I'm going through and marking smoke tests with `@attr('smoke')` so that they can be skipped in specific runs.  The idea is that it would be helpful to know where we have "real" test coverage.

I used a pretty simple heuristic to choose what tests to mark: if it contains a comment about being a smoke test, it gets marked.  So some tests without comments likely got missed, and some tests where a smoky comment only applies to a few lines definitely got over-marked.  These can be refined in a follow-up.

### DIFF
--- a/statsmodels/base/tests/test_optimize.py
+++ b/statsmodels/base/tests/test_optimize.py
@@ -1,3 +1,5 @@
+from nose.plugins.attrib import attr
+
 from numpy.testing import assert_
 from statsmodels.base.optimizer import (_fit_newton, _fit_nm,
                                         _fit_bfgs, _fit_cg,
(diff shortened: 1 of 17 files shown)


bashtage:
Would it be better to just make a decorator `smoke` or `smoke_test` in `compat.testing`, rather than importing attr everywhere?

bashtage:
Should also add `--skip-smoke-tests` and/or `--smoke-tests-only` to conftest.py.

jbrockmendel:
@bashtage I like both of your suggestions.  Will update, though probably not today.

jbrockmendel:
Test robustness not a priority.  Closing.

PR #3974, opened 2017-10-04
De-duplicate covtype func

jbrockmendel:
Last one for the day.


jbrockmendel:
Incrementally addresses a code smell without requiring expert attention.  If it's not broken, don't fix it.  Closing.

PR #3997, opened 2017-10-08
BUG: State space: memory error if k_posdef > k_states (#2)

### LINKED ISSUE #3833: BUG: State space: memory error if k_posdef > k_states

ChadFulton:
Closes #3833.

The same PR as #3834, but I'm curious if a fresh PR will trigger some different coverage results.


ChadFulton:
Never mind, the coverage noticed I had the same commit hash and didn't do anything new, I will try again.

PR #4015, opened 2017-10-10
Implement deprecated_alias, with tests

jbrockmendel:
Inconsistent names are An Issue.  But we're not willing to go through and excise names that don't conform to current naming conventions.

This PR implements a `deprecated_alias` property-like attribute that warns on access but otherwise behaves normally.  i.e. we can bring some consistency to the world, warn users of coming deprecations and maybe even eventually follow up on those deprecations.

One example is implemented in the basically-unmaintained `tsa.varma_process`.  `VarmaPoly.nvars` corresponds to what we would now call `neqs`.

### DIFF
--- a/statsmodels/tools/decorators.py
+++ b/statsmodels/tools/decorators.py
@@ -9,6 +9,28 @@ __all__ = ['resettable_cache', 'cache_readonly', 'cache_writable',
 
 
 def deprecated_alias(old_name, new_name, remove_version=None):
+    """
+    Older or less-used classes may not conform to statsmodels naming
+    conventions.  `deprecated_alias` lets us bring them into conformance
+    without breaking backward-compatibility.
+
+    Example
+    -------
+    Instances of the `Foo` class have a `nvars` attribute, but it _should_
+    be called `neqs`:
+
+    class Foo(object):
+        nvars = deprecated_alias('nvars', 'neqs')
+
+        def __init__(self, neqs):
+            self.neqs = neqs
+
+    >>> foo = Foo(3)
+    >>> foo.nvars
+    __main__:1: FutureWarning: nvars is a deprecated alias for neqs
+    3
+
+    """
     msg = '%s is a deprecated alias for %s' % (old_name, new_name)
     if remove_version is not None:
         msg += ', will be removed in version %s' % remove_version
--- a/statsmodels/tools/decorators.py
+++ b/statsmodels/tools/decorators.py
@@ -14,11 +14,11 @@ def deprecated_alias(old_name, new_name, remove_version=None):
         msg += ', will be removed in version %s' % remove_version
 
     def fget(self):
-        warnings.warn(msg)
+        warnings.warn(msg, DeprecationWarning, stacklevel=2)
         return getattr(self, new_name)
 
     def fset(self, value):
-        warnings.warn(msg)
+        warnings.warn(msg, DeprecationWarning, stacklevel=2)
         setattr(self, new_name, value)
 
     res = property(fget=fget, fset=fset)
--- a/statsmodels/tools/decorators.py
+++ b/statsmodels/tools/decorators.py
@@ -14,11 +14,11 @@ def deprecated_alias(old_name, new_name, remove_version=None):
         msg += ', will be removed in version %s' % remove_version
 
     def fget(self):
-        warnings.warn(msg, DeprecationWarning, stacklevel=2)
+        warnings.warn(msg, FutureWarning, stacklevel=2)
         return getattr(self, new_name)
 
     def fset(self, value):
-        warnings.warn(msg, DeprecationWarning, stacklevel=2)
+        warnings.warn(msg, FutureWarning, stacklevel=2)
         setattr(self, new_name, value)
 
     res = property(fget=fget, fset=fset)
--- a/statsmodels/tools/decorators.py
+++ b/statsmodels/tools/decorators.py
@@ -4,7 +4,26 @@ from numpy.testing import assert_equal
 from statsmodels.compat.python import get_function_name
 import warnings
 
-__all__ = ['resettable_cache', 'cache_readonly', 'cache_writable']
+__all__ = ['resettable_cache', 'cache_readonly', 'cache_writable',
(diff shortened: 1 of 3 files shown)


josef-pkt:
an you check whether that attribute deprecation is already handled by the similar numpy function.
We use that one to deprecate functions that have a new version or name.

quick search: np.deprecate
usage, e.g. here
https://github.com/statsmodels/statsmodels/pull/2478/files

jbrockmendel:
I took a look at that and AFAICT it is best suited to functions.  I didn't see anything I could crib from pandas either.

statsmodels/tools/tests/test_decorators.py
```diff
@@ -0,0 +1,39 @@
...
+
+from statsmodels.tools.decorators import deprecated_alias
+
+
+class TestDeprecatedAlias(object):
+
+	@classmethod
+	def setup_class(cls):
+		
+		class Dummy(object):
```
josef-pkt: there is either a tab instead of spaces or 8 instead of 4 spaces

statsmodels/tsa/varma_process.py
```diff
@@ -464,11 +466,14 @@ class VarmaPoly(object):
 
 
     '''
+
+    nvars = deprecated_alias('nvars', 'neqs', '0.11.0')
```
josef-pkt: this could have a very short test (hopefully it can be added to an existing test)
I will need to be deleted again later

statsmodels/tsa/varma_process.py
```diff
@@ -464,11 +466,14 @@ class VarmaPoly(object):
...
+
     def __init__(self, ar, ma=None):
         self.ar = ar
         self.ma = ma
         nlags, nvarall, nvars = ar.shape
-        self.nlags, self.nvarall, self.nvars = nlags, nvarall, nvars
+        self.nlags, self.nvarall, self.neqs = nlags, nvarall, nvars
```
josef-pkt: Do we really make this change and deprecate this?

statsmodels/tools/decorators.py
```diff
@@ -4,7 +4,26 @@
...
+    msg = '%s is a deprecated alias for %s' % (old_name, new_name)
+    if remove_version is not None:
+        msg += ', will be removed in version %s' % remove_version
+
+    def fget(self):
+        warnings.warn(msg)
+        return getattr(self, new_name)
+
+    def fset(self, value):
+        warnings.warn(msg)
```
josef-pkt: does this need a warnings class, e.g. DeprecationWarning.
what kind of warn does numpy deprecate issue?

josef-pkt:
Ok, then this looks fine to me. a few inline comments. (I didn't think hard about what the decorator function is doing) 

To check: this warns only once, or does it warn repeatedly?

statsmodels/tools/decorators.py
jbrockmendel: Looks like np.deprecate does `warnings.warn(depdoc, DeprecationWarning, stacklevel=2)`.  This is easy to implement, a bit harder to test.  I'll take a look for examples elsewhere.

statsmodels/tsa/varma_process.py
jbrockmendel: This is mainly an example case chosen to be small, clear, and out of the way.  `VARResults` is a much more relevant case.

I'm not necessarily the person to ask; I think `varma_process` should be removed entirely and most of its functionality covered by `wold`.

statsmodels/tsa/varma_process.py
```diff
@@ -464,11 +466,14 @@ class VarmaPoly(object):
 
 
     '''
+
+    nvars = deprecated_alias('nvars', 'neqs', '0.11.0')
+
     def __init__(self, ar, ma=None):
         self.ar = ar
         self.ma = ma
         nlags, nvarall, nvars = ar.shape
```
bashtage: `nvars` -> `negs` here

jbrockmendel: To keep the diff small, I only changed it in the one line that sets `self.nvars = nvars` to set `self.neqs = nvars` instead.

jbrockmendel:
OK, changed the warning to specifically issue a DeprecationWarning.  testing for that is a PITA, but is now implemented.

statsmodels/tools/decorators.py
```diff
@@ -4,7 +4,26 @@
 from statsmodels.compat.python import get_function_name
 import warnings
 
-__all__ = ['resettable_cache', 'cache_readonly', 'cache_writable']
+__all__ = ['resettable_cache', 'cache_readonly', 'cache_writable',
+           'deprecated_alias']
+
+
+def deprecated_alias(old_name, new_name, remove_version=None):
+    msg = '%s is a deprecated alias for %s' % (old_name, new_name)
```
bashtage: Sound you use modern syntax `msg = '{0} is a ...'.format(old_name...`

bashtage: Would a short docstring with an example help others find/use this?

bashtage: Should modern string formatting be used `'{0}'.format(var)`?

jbrockmendel: I think the pytest precedent is that its OK for me to take my time getting comfortable with that.

jbrockmendel: Good idea, will implement.

bashtage: I would autopep all of SM, PRs be damned.

jbrockmendel:
I was wrong about the tests for DeprecationWarning working (but right about it being a PITA).  Just pushed a commit that emits a FutureWarning instead, doesn't require any magic to test.

statsmodels/tools/decorators.py
jbrockmendel: > I would autopep all of SM, PRs be damned.

My foolish hope is that jpkt will delegate design-decision-free fixups to you and I so we can a) clear out the PR backlog and b) clean up the flaketastrophe.

bashtage:
Merry Christmas 😀

PR #4020, opened 2017-10-11
Cycov

jbrockmendel:
This is a WIP attempt to get coverage results for cython modules.  So far it is successfully getting the TRACE variables written to the generated C files, but the results are still not showing up in the coverage reports.

I've gotten this working in a different package, copied the setup there where possible.  Best guess is that remaining differences in setup.py/cythonize.py are the show-stopper.

Three ideas for how to go forward from here:
1) See if someone else poking at it can move the ball further down field.
2) Try to modernize setup.py/cythonize.py and hope the issue resolves itself.
3) Ping the cython mailing list for suggestions.

I'll wait a bit to let folks weigh in here, then try the plan 3.

CC @bashtage, @ChadFulton 
noci

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,8 @@
+cython >= 0.27.1
+numpy >= 1.13.1
+scipy >= 0.19.1
+pandas >= 0.20.3
+
+patsy >= 0.4.1
+
+matplotlib >= 2.1.0
--- a/tools/cythonize.py
+++ b/tools/cythonize.py
@@ -48,6 +48,19 @@ try:
 except NameError:
     WindowsError = None
 
+
+# Determine whether or not to enable coverage in cython files.
+linetrace = os.environ.get('linetrace', False)
+CYTHON_TRACE = str(int(bool(linetrace)))
+
+directives = {'linetrace': False}
+macros = []
+if linetrace:
+    # https://pypkg.com/pypi/pytest-cython/f/tests/example-project/setup.py
+    directives['linetrace'] = True
+    macros = [('CYTHON_TRACE', '1'), ('CYTHON_TRACE_NOGIL', '1')]
+
+
 #
 # Rules
 #
@@ -64,6 +77,11 @@ def process_pyx(fromfile, tofile):
     flags = ['--fast-fail']
     if tofile.endswith('.cxx'):
         flags += ['--cplus']
+
+    from Cython.Build import cythonize
+    cythonize(fromfile, compiler_directives=directives)
+
+    return
 
     try:
         try:
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,9 +1,14 @@
 # .coveragerc to control coverage.py
 [run]
 branch = False
+plugins = Cython.Coverage
 
 [report]
 show_missing = True
+
+omit=
+    */tests/*
+    */datasets/*
 
 # Regexes for lines to exclude from consideration
 exclude_lines =
--- a/setup.py
+++ b/setup.py
@@ -65,8 +65,9 @@ DOWNLOAD_URL = ''
 # These imports need to be here; setuptools needs to be imported first.
 from distutils.extension import Extension
 from distutils.command.build import build
-from distutils.command.build_ext import build_ext as _build_ext
 
+# from distutils.command.build_ext import build_ext as _build_ext
+from Cython.Distutils import build_ext as _build_ext
 
 class build_ext(_build_ext):
     def build_extensions(self):
@@ -81,13 +82,16 @@ class build_ext(_build_ext):
 
 def generate_cython():
     cwd = os.path.abspath(os.path.dirname(__file__))
+    tools_dir = os.path.join(cwd, 'tools')
+
     print("Cythonizing sources")
-    p = subprocess.call([sys.executable,
-                          os.path.join(cwd, 'tools', 'cythonize.py'),
-                          'statsmodels'],
-                         cwd=cwd)
-    if p != 0:
-        raise RuntimeError("Running cythonize failed!")
+    try:
+        sys.path.insert(0, tools_dir)
+        import cythonize
+        cythonize.find_process_files('statsmodels')
+    finally:
+        sys.path.remove(tools_dir)
+    return
 
 
 def init_cython_exclusion(filename):
@@ -453,6 +457,18 @@ except ImportError:
         append_cython_exclusion(path.replace('/', os.path.sep),
                                 CYTHON_EXCLUSION_FILE)
 
+# Determine whether to build the cython extensions with coverage
+# measurement enabled.
+linetrace = os.environ.get('linetrace', False)
+CYTHON_TRACE = str(int(bool(linetrace)))
+
+directives = {'linetrace': False}
+macros = []
+if linetrace:
+    # https://pypkg.com/pypi/pytest-cython/f/tests/example-project/setup.py
+    directives['linetrace'] = True
+    macros = [('CYTHON_TRACE', '1'), ('CYTHON_TRACE_NOGIL', '1')]
+
 extensions = []
 for name, data in ext_data.items():
     data['sources'] = data.get('sources', []) + [data['name']]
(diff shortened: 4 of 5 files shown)


requirements.txt
```diff
@@ -0,0 +1,8 @@
+cython >= 0.27.1
+numpy >= 1.13.1
+scipy >= 0.19.1
+pandas >= 0.20.3
+
+patsy >= 0.4.1
+
+matplotlib >= 2.1.0
```
bashtage: Do these need to be bleeding edge?

jbrockmendel: Not at all, I just needed to put _something_ there.  This isn't intended to be merged in anything near its current state.

josef-pkt:
This pull request introduces 1 alert - [view on lgtm.com](https://lgtm.com/projects/g/statsmodels/statsmodels/rev/pr-6bddc3987703caf76e31d931fc916ae1d35d4572)

**new alerts:**

* 1 for Unreachable code

---

ChadFulton:
Is the end result (the coverage output) something in codecov, or coveralls, or something local with pytest?

jbrockmendel:
> Is the end result (the coverage output) something in codecov, or coveralls, or something local with pytest?

Documentation FTW.  Locally running "tox -e coverage" should build the extensions, run the tests, and print the coverage results.  Figuring out the CI is a separate step.

tools/cythonize.py
```diff
@@ -48,6 +48,19 @@
 except NameError:
     WindowsError = None
 
+
+# Determine whether or not to enable coverage in cython files.
+linetrace = os.environ.get('linetrace', False)
+CYTHON_TRACE = str(int(bool(linetrace)))
+
+directives = {'linetrace': False}
```
bashtage: Need `'binding': True` it appears.  I got one run in `arch` for a single cython function using your setup + binding. 

https://stackoverflow.com/questions/24144931/python-line-profiler-and-cython-function

bashtage: `binding` isn't needed, sorry.

bashtage:
I got it to work in a very simple case:

https://codecov.io/gh/bashtage/arch/src/cython-coverage/arch/bootstrap/_samplers.pyx

The recipe was similar to what you did but it was _essential_ that `python setup.py develop` was used instead of a straight install. One could probably do 

```
python setup.py build_ext --inplace -f
python setup.py install
```

as well.

bashtage:
I got it fully working on arch.  One thing I noticed is that it can make execution very slow.  Runtime increased from ~2 minutes to ~8, so it may be difficult to implement in a larger package on travis.

jbrockmendel:
> essential that python setup.py develop was used instead of a straight install

I think I probably used the `--inplace -f` version, but really I tried so many different variants its hard to tell which actually matter.  Regardless, I'm psyched some good has come of this in the form of getting this working for arch.

Any ideas on how to get work this into the CI here?  I have zero experience with Travis config.

bashtage:
One breakthrough for this code -- you have to use the usual route of calling cythonize on the extension containing .pyx files to get coverage to work.  This will require a large rewrite of setup to redo how cython is used. I was easily able to get coverage for the nonparametric code using this strategy.

bashtage:
> Is the end result (the coverage output) something in codecov, or coveralls, or something local with pytest?

Any file that has a .pyx in the repo will show up in online coverage.  Files with only pyx.in will not since the source file isn't available.  These will show up in local coverage though, say in the html report.

bashtage:
Here is the proof:

screenshot-2017-10-14 coverage for statsmodels nonparametric _smoothers_lowess pyx 92

bashtage:
Overall I think statsmodels setup to too complicated to allow this to be effectively implemented.  The cythonization requires a full rewrite so that `Cython.Build.cythonize(extensions,...)` can be used when coverage is needed, while still supporting distibution of C source files so that Cython isn't a build requirement. My preference would be to make Cython a straight build requirement which would allow setup to be dramatically simplified -- the only real step required would be to process `pyx.in` to `pyx` files and to set their timestamps to match the `pyx.in`, and then Cython would handle the rest. 

Give that I don't think this would make it in, I can't find the motivation to do the work.

josef-pkt:
@bashtage  I was thinking about the setup.py problem, especially because I assume we don't want this in production code if it makes it much slower.

Can we just create a new setup_coverage.py and run that when we want to get the cython coverage?
For travis we could create a PR branch or a second branch in the statsmodels that we can rebase on master.

This would allow us to get coverage when cython or cython related code changes, and give us some time to gain experience with it, without changing our setup for the development and release master.

bashtage:
When it isn't enabled it has no effect on performance.  The code to profile
isn't even present when linetrace is false.

What about switching to versioneer?  This would simplify versioning a lot
since it just uses the last tag.

josef-pkt:
@ChadFulton  was doing the last rewrite of cythonizing/extension building on master, and he has most of the cython code. So that part of setup.py is up to him.

about versioneer: I never heard about it or looked at it.
After reading a bit it sounds like it's replacing human fragility and errors by machine fragility and errors

they mention git describe, and I get
M:\...\statsmodels_py34_pr>git describe
v0.6.0rc1-2578-g326ae0c

and paying attention to it for the first time, I see in git gui
Follows: v0.6.0rc1

which is maybe `https://github.com/warner/python-versioneer/issues/159`  (commented out to prevent github linking)

My guess is that this comes from never merging the tag/release branch back into master, which I thought is a useless step for us.

bashtage:
It would mean a different process -- release from master, which imo is a
much simpler process that removes most of the pain of releasing.

josef-pkt:
> It would mean a different process -- release from master,

That sounds difficult, without release branch, master would be blocked between pre-release and final release. And if we ever do a bug-fix release, we still would need a release branch.

ChadFulton:
I'm out traveling this weekend but will try to get up to speed tomorrow night.

bashtage:
Not that difficult.   Two strategies.

1. Always release from master. This requires staging big changes for short
merge windows and trickling in bug patches.  Then bug fixes can be merged
in.  Requires getting PRs in good shape.  I think pandas uses a model like
this.  Help to have shorter release cycles or calendar targets.

2. Release major from master.  Bug fix releases are then branches from tags
where bug fixes are cherry picked from master.

It feels to me like releases should always correspond to a commit in
master.

jbrockmendel:
No idea about versioneer.  I was just kind of assuming the CI would have an option analogous to "tox -e coverage".

setup.py+tools/cythonize.py could be simplified with some modernization.  In particular, I think cython now handles the caching.  pandas also uses pyx.in files; they might have something worth cribbing.

Speaking of pandas, they don't have coverage for cython code.  Getting it working here in a reproducible way would serve as a public good for a lot of the ecosystem (mixed metaphors for the win).

jbrockmendel:
Taking a look at setup.py, a few things jump out at me (largely tangential to the coverage goal)

A `if __name__ ==` block is not something I'm used to seeing in a setup.py file.  If keeping that is important, then at least the constants in it (e.g. `min_versions`, `extras`) should be moved outside that block so as to be visible externally.

`min_versions['numpy']` is 1.6.2 which was released mid-2012.  It's probably long past time for a bump.  Best guess it has in fact been bumped but didn't get updated here.  If correct, then there should be a Single Source of Truth about min versions.

`write_version_py` and `git_version` look like a workaround to accomplish something similar to versioneer.  If that's accurate, then at some point a transition should be considered.  Poeple with comparative advantage appear to have decided versioneer is the preferred solution to this problem.  Transition costs are non-trivial so I'm not advocating that anytime soon.

Last, specific to cython, pandas effectively does `generate_cython` [inside of ](https://github.com/pandas-dev/pandas/blob/master/setup.py#L147) their `build_ext`.  I'm inclined to crib their solutions (and to use subprocess less).

ChadFulton:
> setup.py+tools/cythonize.py could be simplified with some modernization. In particular, I think cython now handles the caching. pandas also uses pyx.in files; they might have something worth cribbing.

Currently we follow scipy's setup.py file and our `cythonize.py` is from them.

It looks like Pandas does things more like you're thinking; they used to do some things I didn't understand that required a "fake" pyrex file, but taking a look at their setup.py, I guess that's not necessary anymore.

I think that I recall that 0.9 will be the last version of statsmodels to support scipy < 0.14, and so after 0.9 we can drop the conditional compilation stuff currently in our setup.py.

jbrockmendel:
Is there a natural point at which setup.py will be worked on anyway?  If so, this can probably wait until then.  (I can run coverage locally and post results; the cython code is slow-changing).

jbrockmendel:
Discussion about modernizing setup.py notwithstanding, I think this is now subsumed by #4035

bashtage:
Probably can be closed.  The good bits live on in #4035

PR #4029, opened 2017-10-12
REF: Align descriptions between mlemodel and markov_switching

jbrockmendel:
@ChadFulton 

This implements the description strings discussed in #3675.

One point that merits extra attention is the description for `self.cov_type == 'robust'`.  The description in markov_switching for that cases matches the string in mlemodel for 'robust_approx' (which is the case that is broken over there, addressed in the next PR).  It looks like markov_switching's cov_params_robust does the same thing as mlemodel's cov_params_robust_approx, so it may be that the descriptions are correct as is.


ChadFulton:
> One point that merits extra attention is the description for self.cov_type == 'robust'. The description in markov_switching for that cases matches the string in mlemodel for 'robust_approx' (which is the case that is broken over there, addressed in the next PR). It looks like markov_switching's cov_params_robust does the same thing as mlemodel's cov_params_robust_approx, so it may be that the descriptions are correct as is.

Yes, `robust` means sandwich covariance matrix and then in state space, the suffix `approx` or `oim` indicates whether the covariance matrix was computed using numerical differentiation (`approx`) or the method of Harvey (`oim`). So you're right, in Markov switching models, we only have `robust_approx` (and `robust_oim` isn't an option to implement as far as I know).

So we shouldn't have `robust_approx` raise an exception, but it doesn't matter to me whether `robust_approx` is allowed as an alias or not.

jbrockmendel:
> So we shouldn't have robust_approx raise an exception

I put that there as a reminder; note it doesn't actually change in current behavior, just raises in its own block instead of the following catch-all block.

Recall a major part of the issue in #3675 is that the `self.cov_type == `robust_approx'` branch in mlemodel dispatches to `res.cov_params_robust` and not `res.cov_params_robust_approx`.  I'm saving that particular fix for a separate PR in to try to keep tight focus.  So the "extra attention" note is just to check that the descriptions in regime_switching will still be accurate after that fix to mlemodel is made in the follow-on PR.

jbrockmendel:
@ChadFulton would you rather we fix the incorrect call in MLEResults before doing this?  There are 2-3 follow-ups I have in mind here, want to strike while the iron is hot.

ChadFulton:
I think we can merge this first before the fix to MLEModel, but I think we should remove the added if statement (even though you're right that the exception doesn't change) because the TODO isn't quite accurate - `robust_approx` is already "implemented" (in the sense that it's just the same thing as `robust`), we just aren't calling it that.

Happy to merge this asap if we fix that.

ChadFulton:
I guess I'm not sure that we need the custom covariance matrix here, although it doesn't do any harm. We did it for `MLEModel` so that we could support the Recursive Least Squares case there. Maybe we shouldn't add it until we need it?

jbrockmendel:
> I think we can merge this first before the fix to MLEModel, but I think we should remove the added if statement

Just changed and pushed.

> I guess I'm not sure that we need the custom covariance matrix here, although it doesn't do any harm

Part of the MO is to go Highlander on the ~5 implementations of `get_robustcov_results`.

ChadFulton:
Will merge when green (tomorrow).

ChadFulton:
This needs to be rebased, I guess, since #4056 was merged.

jbrockmendel:
@ChadFulton The appveyor fail is segfault on test_simulate; coveralls fail is ignorable.  I think this is good to go.

ChadFulton:
Thanks!

PR #4055, opened 2017-10-21
DOC: Clean up small errors and warnings

bashtage:
Clean up small errors and warnings in doc build

### DIFF
--- a/statsmodels/tsa/statespace/dynamic_factor.py
+++ b/statsmodels/tsa/statespace/dynamic_factor.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 """
 Dynamic factor model
 
@@ -136,7 +137,7 @@ class DynamicFactor(MLEModel):
 
     References
     ----------
-    .. [*] Lutkepohl, Helmut. 2007.
+    .. [*] Lütkepohl, Helmut. 2007.
        New Introduction to Multiple Time Series Analysis.
        Berlin: Springer.
 
--- a/statsmodels/tsa/statespace/dynamic_factor.py
+++ b/statsmodels/tsa/statespace/dynamic_factor.py
@@ -136,7 +136,7 @@ class DynamicFactor(MLEModel):
 
     References
     ----------
-    .. [1] Lutkepohl, Helmut. 2007.
+    .. [*] Lutkepohl, Helmut. 2007.
        New Introduction to Multiple Time Series Analysis.
        Berlin: Springer.
 
--- a/statsmodels/tsa/vector_ar/var_model.py
+++ b/statsmodels/tsa/vector_ar/var_model.py
@@ -1,9 +1,10 @@
+# -*- coding: utf-8 -*-
 """
 Vector Autoregression (VAR) processes
 
 References
 ----------
-Lutkepohl (2005) New Introduction to Multiple Time Series Analysis
+Lütkepohl (2005) New Introduction to Multiple Time Series Analysis
 """
 
 from __future__ import division, print_function
@@ -116,7 +117,7 @@ def var_acf(coefs, sig_u, nlags=None):
 
     Notes
     -----
-    Ref: Lutkepohl p.28-29
+    Ref: Lütkepohl p.28-29
 
     Returns
     -------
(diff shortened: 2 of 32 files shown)


statsmodels/tsa/vector_ar/vecm.py
```diff
@@ -805,7 +805,9 @@ class VECM(tsbase.TimeSeriesModel):
     References
     ----------
     .. [1] Lutkepohl, H. 2005. *New Introduction to Multiple Time Series Analysis*. Springer.
-    .. [2] Johansen, S. 1995. *Likelihood-Based Inference in Cointegrated Vector Autoregressive Models*. Oxford University Press.
+
+    .. [2] Johansen, S. 1995. *Likelihood-Based Inference in Cointegrated *
+           *Vector Autoregressive Models*. Oxford University Press.
```
jbrockmendel: It looks like everywhere else you changed `[2]` to `[*]`.  Why is this different?

bashtage:
Uncited reference create warnings.  [*] Is the correct way in Sphinx to add
a reference that isn't cited in the doc string.

statsmodels/tsa/statespace/dynamic_factor.py
```diff
@@ -136,7 +136,7 @@ class DynamicFactor(MLEModel):
 
     References
     ----------
-    .. [1] Lutkepohl, Helmut. 2007.
+    .. [*] Lutkepohl, Helmut. 2007.
```
jbrockmendel: Should we standardized on the umlaut `Lütkepohl`?

bashtage:
Do you mean ue or ü?  Does the latter potentially cause issues on Python
2.7?

jbrockmendel:
In e.g. vecm we have the non-ascii version, which I assume is more-correct.  My understanding is that putting `# -*- coding: utf-8 -*-` takes care of the interpreter.

josef-pkt:
about non-ASCII names

As long as the encoding is correctly specified `# -*- coding: utf-8 -*-` there should not be any problems, also not in python 2.
I thought we have some cases, but I don't find any in the modules that I checked as likely candidates.
`Lutkepohl` seems to be consistently spelled this way in vector_ar VAR and VECM. That's the closes mapping to ASCII. I'm not sure anybody will understand the German umlaut "ue" conversion, although For Hardle, I might be also using that instead of "Haerdle"
In references in docstrings we should switch to utf-8 and native spelling (but getting the utf-8 for non-ascii is not trivial when we don't have the corresponding keyboard).

e.g. random reference copied from my Zotero

Lütkepohl, Helmut, and Aleksei Netšunajev. 2017. “Structural Vector Autoregressions with Heteroskedasticity: A Review of Different Volatility Models.” Econometrics and Statistics 1 (January):2–18. https://doi.org/10.1016/j.ecosta.2016.05.001.

josef-pkt:
jstor converts to simple ASCII

Hardle, W., J. Hart, J. S. Marron, and A. B. Tsybakov. 1992. “Bandwidth Choice for Average Derivative Estimation.” Journal of the American Statistical Association 87 (417):218–26. https://doi.org/10.2307/2290472.

statespace/mlemodel.py line 1905 uses the correct utf-8 "ü"

statsmodels/tsa/vector_ar/var_model.py
```diff
@@ -1716,17 +1720,24 @@ def test_inst_causality(self, causing, signif=0.05):
...
+              The p-value corresponding to the test statistic.
+
+            "df" : float
+              The degrees of freedom of the Chi^2-distribution.
+
+            "conclusion" : str {"reject", "fail to reject"}
+              Whether H0 can be rejected or not.
+
+            "signif" : float
+              Signifigance level
```
josef-pkt: was this necessary
I thought the list with would work as rst

jbrockmendel: Signifigance --> Significance

josef-pkt:
LGTM
I didn't know how to cleanup the references that are not cited.
properly spelled Lutkepohl looks good

(I will be back to merging in a few hours)

jbrockmendel:
I don't know rst/sphinx well enough to give this a "LGTM" stamp (hoping jspkt recognizes that there is a non-trivial signal in places where I _do_ apply this stamp), but it does look internally consistent.

bashtage:
The dictionary output was broken.  This is standard RST.

bashtage:
Fixed.

bashtage:
image

bashtage:
LGTM

josef-pkt:
@bashtage Thanks
@jbrockmendel  Thanks for reviewing

merged

PR #4056, opened 2017-10-22
BUG: State space: Fix label cov_params_robust --> cov_params_robust_approx

jbrockmendel:
@ChadFulton The discussed fix for `MLEResults._get_robustcov_results`

### DIFF
--- a/statsmodels/tsa/statespace/mlemodel.py
+++ b/statsmodels/tsa/statespace/mlemodel.py
@@ -1004,9 +1004,10 @@ class MLEModel(tsbase.TimeSeriesModel):
         partials = np.zeros((self.nobs, n))
         k_endog = self.k_endog
         for t in range(self.nobs):
-            for i in range(n):
-                inv_forecasts_error_cov = np.linalg.inv(
+            inv_forecasts_error_cov = np.linalg.inv(
                     res.forecasts_error_cov[:, :, t])
+
+            for i in range(n):
                 partials[t, i] += np.trace(np.dot(
                     np.dot(inv_forecasts_error_cov,
                            partials_forecasts_error_cov[:, :, t, i]),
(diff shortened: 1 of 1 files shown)


ChadFulton:
This looks fine, can merge when green.

Not sure what our policy is on including misc. cleanups in unrelated PRs? Should the cleaning code be in a separate PR? I'm guilty of this also.

statsmodels/tsa/statespace/mlemodel.py
```diff
@@ -1004,9 +1004,10 @@ def _score_obs_harvey(self, params, approx_complex_step=True,
         partials = np.zeros((self.nobs, n))
         k_endog = self.k_endog
         for t in range(self.nobs):
-            for i in range(n):
-                inv_forecasts_error_cov = np.linalg.inv(
+            inv_forecasts_error_cov = np.linalg.inv(
                     res.forecasts_error_cov[:, :, t])
+
```
jbrockmendel: Noticed this while skimming: `inv_forecasts_error_cov` is re-calculated in each step of the loop over `i`.  The diff is a bit hard to read; it just moves that calculation outside the loop.  There are some other repeated calculations inside this loop, but they don't get stored as intermediate variables so it is _conceivable that there is a memory tradeoff.  Leaving this with a big "moderately-low hanging fruit" sign attached.

jbrockmendel:
> Not sure what our policy is on including misc. cleanups in unrelated PRs? Should the cleaning code be in a separate PR? I'm guilty of this also.

My experience has been that cleanup and other technical-debt-paydown PRs tend not to get accepted.  This has been more than a little bit frustrating.  If you want to do things differently in your fiefdom, let me know.

ChadFulton:
> My experience has been that cleanup and other technical-debt-paydown PRs tend not to get accepted. This has been more than a little bit frustrating. If you want to do things differently in your fiefdom, let me know.

Well I don't have many strong feelings on these points, and am happy to merge this PR. I bring it up because in the past I have been slower than I'd like to be at reviewing PRs, and the fewer unrelated things going on in a PR the more likely it is that I could review it in a few minutes (in which case I would do it sooner) whereas if there are more things that I feel like I need to dig into (even though by themselves they're trivial), I may not get to it for a while.

statsmodels/tsa/statespace/mlemodel.py
ChadFulton: Yes, I think this looks good.

ChadFulton:
Thanks!

PR #4064, opened 2017-10-24
Natashawatkins add drop omitted summary2

bashtage:
Rebase #3702 and add a test

### DIFF
--- a/statsmodels/iolib/summary2.py
+++ b/statsmodels/iolib/summary2.py
@@ -1,14 +1,17 @@
 from statsmodels.compat.python import (lrange, iterkeys, iteritems, lzip,
                                        reduce, itervalues, zip, string_types,
                                        range)
-from statsmodels.compat.collections import OrderedDict
+
+from collections import OrderedDict
+import datetime
+import re
+import textwrap
+
 import numpy as np
 import pandas as pd
-import datetime
-import textwrap
+
 from .table import SimpleTable
 from .tableformatting import fmt_latex, fmt_txt
-import re
 
 
 class Summary(object):
@@ -426,23 +429,24 @@ def _make_unique(list_of_names):
     return header
 
 
-def summary_col(results, float_format='%.4f', model_names=[], stars=False,
-                info_dict=None, regressor_order=[], drop_omitted=False):
+def summary_col(results, float_format='%.4f', model_names=(), stars=False,
+                info_dict=None, regressor_order=(), drop_omitted=False):
     """
     Summarize multiple results instances side-by-side (coefs and SEs)
 
     Parameters
     ----------
     results : statsmodels results instance or list of result instances
-    float_format : string
+    float_format : string, optional
         float format for coefficients and standard errors
         Default : '%.4f'
-    model_names : list of strings of length len(results) if the names are not
+    model_names : list of strings, optional
+        Must have same length as the number of results. If the names are not
         unique, a roman number will be appended to all model names
     stars : bool
         print significance stars
     info_dict : dict
-        dict of lambda functions to be applied to results instances to retrieve
+        dict of functions to be applied to results instances to retrieve
         model info. To use specific information for different models, add a
         (nested) info_dict with model name as the key.
         Example: `info_dict = {"N":..., "R2": ..., "OLS":{"R2":...}}` would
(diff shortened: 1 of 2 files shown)


statsmodels/iolib/tests/test_summary2.py
```diff
@@ -36,6 +39,23 @@ def test_summarycol(self):
...
+        y1 = [6, 4, 2, 7, 4]
+        y2 = [8, 5, 0, 12, 4]
+        reg1 = OLS(y1, x).fit()
+        reg2 = OLS(y2, x2).fit()
+        actual = summary_col([reg1, reg2], regressor_order=['const', 'x1'],
+                             drop_omitted=True)
+        assert 'x2' not in str(actual)
+        actual = summary_col([reg1, reg2], regressor_order=['x1'],
+                             drop_omitted=False)
+        assert 'const' in str(actual)
```
josef-pkt: no plain assert in unit test, use numpy.testing  `assert_`

bashtage:
Plain assert is preferred with pytest since you get clear output in case of
a failure.

josef-pkt:
I don't care much about it in this case.
But in other code I don't want to have any automatic code rewriting (i.e. pytest vodoo. That's one reason I always preferred nosetests. The other reason was that I found it "silly" to use a `py.` namespace.)

`--pdb` (--pdb-failures) is good to investigate the details of a test failure, but I think having the original code is safer than having to work through come rewritten, autogenerated code.

bashtage:
pytest also supports running with --pdb.  

The big advantage of raw asserts when appropriate is that you don't get the stack trace messed up with some other function -- the assert is asserted in teh place it should be. 

Plus, nose is dead.  Even conservative numpy is moving to pytest in release after next. 

> I think having the original code is safer than having to work through come rewritten, autogenerated code.

I don't follow this -- when I see a test failure I can directly step through the test inspecting as I go (I use PyCharm, but I'm sure it is no different in other debuggers).

josef-pkt:
It's partially less bad and even worse
in pytest --pdb it lists the original code, there is no rewritten code.
Reading the small print that I didn't pay attention to in my previous reading http://pybites.blogspot.ca/2011/07/behind-scenes-of-pytests-new-assertion.html
pytest is rewriting the AST, so we don't even see what is actually executed.

Aside: I don't trust pytest at all. 
And I haven't found out what the import hook is doing
https://github.com/statsmodels/statsmodels/issues/3082#issuecomment-337320993
(I gave up because we cannot get rid of pytest during runtime/import when not testing.)

We are switching to pytest because we have to, not necessarily because we are big fans of it.

jbrockmendel:
I'll offer jpkt some moral support: I'm still not comfortable with pytest's "magic".

That said, my understanding is that bashtage is right that nosetests is no longer maintained and that pytest is the grown-up way to do things.  If we're going to use the tool, we should (eventually) use it in the supported manner.  Trust that smart people have worked hard on this and come to a carefully-considered conclusion.

statsmodels/iolib/summary2.py
```diff
@@ -1,14 +1,16 @@
+import datetime
+import re
+import textwrap
+
+import numpy as np
+import pandas as pd
+
+from statsmodels.compat.collections import OrderedDict
```
jbrockmendel: compat.collections is a py26 compat shim; just import from collections

statsmodels/iolib/summary2.py
```diff
@@ -1,14 +1,16 @@
+import datetime
+import re
+import textwrap
```
jbrockmendel: +1 for the stdlib --> 3rd party --> 1st party import ordering

jbrockmendel:
The "assert" issue aside, this PR is excellent and should be merged post-haste.

josef-pkt:
Asides: statsmodels was the last one of the related packages in our neighborhood to switch to git. But after a few years I got used to it.

I became very averse to pip after it wiped a few times my carefully constructed virtual environments on Windows. Still after 5 years or more it seems to work as one of the two standards.

I guess pytest will also work out eventually, but I don't like to do the bug hunting and bug reporting, especially of invisible, behind-the-scenes "magic".

PR also LGTM

statsmodels/iolib/summary2.py
bashtage: It is correct except it should be compat, std lib, 3rd, 1st.

bashtage:
Added a test for an other reported problem -- the problem had already been corrected but test improves coverage a bit.

josef-pkt:
@bashtage Thanks

merged

PR #4072, opened 2017-10-25
do validation checks before expensive computations

jbrockmendel:
This should be a no-brainer

### DIFF
--- a/statsmodels/discrete/count_model.py
+++ b/statsmodels/discrete/count_model.py
@@ -200,6 +200,11 @@ class GenericZeroInflated(CountModel):
             alpha=0, trim_mode='auto', auto_trim_tol=0.01, size_trim_tol=1e-4,
             qc_tol=0.03, **kwargs):
 
+        if method not in ['l1', 'l1_cvxopt_cp']:
+            # FIXME: This should be a ValueError right?
+            raise TypeError("argument method == %s, which is not handled" %
+                            method)
+
         if np.size(alpha) == 1 and alpha != 0:
             k_params = self.k_exog + self.k_inflate
             alpha = alpha * np.ones(k_params)
@@ -223,12 +228,7 @@ class GenericZeroInflated(CountModel):
                 alpha=alpha, trim_mode=trim_mode, auto_trim_tol=auto_trim_tol,
                 size_trim_tol=size_trim_tol, qc_tol=qc_tol, **kwargs)
 
-        if method in ['l1', 'l1_cvxopt_cp']:
-            discretefit = self.result_class_reg(self, cntfit)
-        else:
-            raise TypeError(
-                    "argument method == %s, which is not handled" % method)
-
+        discretefit = self.result_class_reg(self, cntfit)
         return self.result_class_reg_wrapper(discretefit)
 
     fit_regularized.__doc__ = DiscreteModel.fit_regularized.__doc__
--- a/statsmodels/discrete/discrete_model.py
+++ b/statsmodels/discrete/discrete_model.py
@@ -465,7 +465,7 @@ class BinaryModel(DiscreteModel):
                 full_output=full_output, disp=disp, callback=callback,
                 alpha=alpha, trim_mode=trim_mode, auto_trim_tol=auto_trim_tol,
                 size_trim_tol=size_trim_tol, qc_tol=qc_tol, **kwargs)
-        
+
         discretefit = L1BinaryResults(self, bnryfit)
         return L1BinaryResultsWrapper(discretefit)
     fit_regularized.__doc__ = DiscreteModel.fit_regularized.__doc__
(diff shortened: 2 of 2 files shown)


josef-pkt:
Yes, it should raise ValueError in all cases, wrong value of argument.

statsmodels/discrete/count_model.py
```diff
@@ -200,6 +200,11 @@ def fit_regularized(self, start_params=None, method='l1',
             alpha=0, trim_mode='auto', auto_trim_tol=0.01, size_trim_tol=1e-4,
             qc_tol=0.03, **kwargs):
 
+        if method not in ['l1', 'l1_cvxopt_cp']:
+            # FIXME: This should be a ValueError right?
+            raise TypeError("argument method == %s, which is not handled" %
```
bashtage: Please no old style format strings.  Also, needs test coverage since isn't hit.  Same for other.

In case you don't have the codecov plugin for Chrome -- you see ticks like this on github.

image

statsmodels/discrete/discrete_model.py
```diff
@@ -454,16 +454,19 @@ def fit_regularized(self, start_params=None, method='l1',
             maxiter='defined_by_method', full_output=1, disp=1, callback=None,
             alpha=0, trim_mode='auto', auto_trim_tol=0.01, size_trim_tol=1e-4,
             qc_tol=0.03, **kwargs):
+
+        if method not in ['l1', 'l1_cvxopt_cp']:
+            # FIXME: This should be a ValueError, right?
+            raise Exception(
```
bashtage: No bare Exception.  ValueError?

statsmodels/discrete/discrete_model.py
```diff
@@ -859,16 +862,18 @@ def fit_regularized(self, start_params=None, method='l1',
...
```
bashtage: Same, probably ValueError.

statsmodels/discrete/discrete_model.py
```diff
@@ -1048,16 +1053,19 @@ def fit_regularized(self, start_params=None, method='l1',
...
```
bashtage: Same.

statsmodels/discrete/discrete_model.py
```diff
@@ -1411,6 +1419,11 @@ def fit_regularized(self, start_params=None, method='l1',
             alpha=0, trim_mode='auto', auto_trim_tol=0.01, size_trim_tol=1e-4,
             qc_tol=0.03, **kwargs):
 
+        if method not in ['l1', 'l1_cvxopt_cp']:
+            # FIXME: This should be a ValueError, right?
+            raise Exception(
```
bashtage: Same

statsmodels/discrete/discrete_model.py
```diff
@@ -2780,6 +2788,11 @@ def fit_regularized(self, start_params=None, method='l1',
...
```
bashtage: Same.

statsmodels/discrete/discrete_model.py
```diff
@@ -3137,6 +3146,7 @@ def fit_regularized(self, start_params=None, method='l1',
             qc_tol=0.03, **kwargs):
 
         if method not in ['l1', 'l1_cvxopt_cp']:
+            # FIXME: This should be a ValueError, right?
```
bashtage: Yes, ValueError

statsmodels/discrete/count_model.py
```diff
@@ -200,6 +200,11 @@ def fit_regularized(self, start_params=None, method='l1',
             alpha=0, trim_mode='auto', auto_trim_tol=0.01, size_trim_tol=1e-4,
             qc_tol=0.03, **kwargs):
 
+        if method not in ['l1', 'l1_cvxopt_cp']:
+            # FIXME: This should be a ValueError right?
```
bashtage: Yes, ValueError

jbrockmendel:
@josef-pkt Sounds like we have a concensus.  Let's merge this and I'll follow up with a PR to raise ValueErrors

@bashtage That's a neat plugin; I'll try it out.  Also, why the strong opinion on string formatting?

bashtage:
I like new and shiny.  I also find '{variable} does
something'.format(variable=variable) clearer.  FWIW I think pandas is
moving to these.

josef-pkt:
the old string interpolation is still valid python.
I never got used to the new `format` style and only use the old `%`.
Except in long strings I find it easier to read, e.g. if there are only one or two, three places to replace.

jbrockmendel:
Can't argue with shiny.  For now let's help the boss keep eyes on the prize.

jbrockmendel:
So... we have concensus on this right?

bashtage:
Maybe at least change the Exception to a specific exception. Either ValueError or NotImplementedError.

jbrockmendel:
> Maybe at least change the Exception to a specific exception. Either ValueError or NotImplementedError.

The choice is basically merge this reviewed+CIed PR and then follow up to make those changes vs. edit the PR.  It's an additional 3-5 minutes of labor for me either way.  I strongly prefer the former, since I want to avoid having trivial PRs sit here for months on end.

bashtage:
Merge away then. IDD wrt small PRs that are strictly improvements.

jbrockmendel:
@josef-pkt codecov error is a dummy; this should be good to go (has already been reviewed by you among others)

jbrockmendel:
Let’s rock.

PR #4079, opened 2017-10-27
xfail tests that had been disabled via name-mangling

jbrockmendel:
I looked through base, datasets, and discrete.  Still a bunch more directories to hunt these down in.


bashtage:
@josef-pkt LGTM.  This is the correct method to handle tests that cannot pass.  Easier to find `xfail` than to find munged test names.

jbrockmendel:
@josef-pkt did ya hear? That politician you dislike says name-mangling is _great_!

PR #4083, opened 2017-10-27
Replace specific-case pickling methods with general-case versions

jbrockmendel:
Low priority.

Ideally we'd be able to separate econometrics from "plumbing".  The latter includes any pickle-related hijinx.  This PR replaces hard-coded special cases with more general logic.


jbrockmendel:
I may circle back to this eventually, but for now am closing to de-clutter the queue.

PR #4089, opened 2017-10-29
Prepare CheckTweedieSpecial for parameterization; cleanup

jbrockmendel:
This is cleanup, should change zero logic.

That said, the motivation here is following #3988 to try to move towards parameterized tests that let us be more systematic.  The relevant part of the diff to look at is `CheckTweedieSpecial`.

I haven't figured out how to make pytest parameterized tests play nicely with `setup_class`-style test classes.  It seems likely that there is a best-practice way of doing this.


jbrockmendel:
Error is in test-simulate, unrelated

PR #4149, opened 2017-12-02
DOC: Fix a typo in an error message

fuglede:
This turns "timeseries index witha freq" into "timeseries index with a freq".


ChadFulton:
Looks good, thanks for the PR!

Issue #3109, opened 2016-07-14
FAQ: Time series support in Statsmodels


pcdinh:
Hi,

I learned from /r/datascience 

> I didn't know they finally added SARIMA support to statsmodels. That being said, having time series support with just ARIMA and basic exponential smoothing is like having a standard statistics package with nothing but standard linear and logistic regression. There's basically no support for ARCH and GARCH models, no support for incorporating transfer functions, and no support for multiple seasonal trends. Although somehow, R doesn't have support for the last one either but SAS does.

https://www.reddit.com/r/datascience/comments/4siy0n/learning_time_series_analysis/d5cbp3c

I am quite new to statsmodels. Is his statement true to current features of statsmodels? If it is the case, is the any plan to support the missing parts?

Thanks

ChadFulton:
There are some true things about that statement and some false things, as
well as a dose of that commenter's opinions. Here's an update on the
Statsmodels time series library.

- The emphasis in econometric time series generally is on ARIMA models, and
they're used a huge amount.
- Statsmodels v0.6.x also has support for VAR models, which are again huge
workhorses in econometric time series.
- I don't know the whole story here, but ARCH and GARCH are implemented in
the "arch" package by one of the major statsmodels contributors. I'm not
sure why it's not in Statsmodels itself, but does it need to be? Not all of
R time series is in a single package.
- The commenter seems to think that transfer functions and multiple
seasonal trends are very important time series models, and maybe they are
in some particular discipline, but they're not universally considered to be
basic models everyone must support. Stata, for example, doesn't support
them (except through their state space framework, but see below).

Starting with v0.8 (which has a release candidate out, and hopefully will
be released in August), statsmodels will get:

- SARIMAX models, as the commenter noted
- A general framework for state space models; this can incorporate any
linear / Gaussian time series model you want, including transfer functions
and multiple seasonal trends.
- Unobserved components, dynamic factors, and VARMAX models.
- Also, introduced are Markov switching regressions and autoregressions,
which are non-linear Gaussian models. This is a good example of something
that is quite hard to find implemented in R.

In the future (v0.9?), we can expect:

- Markov switching state space models, which opens up a huge range of
non-linear Gaussian models

So ultimately, I think Statsmodels will very soon incorporate a huge range
of important time series models. On the other hand, I'm sure that for
certain disciplines there are specific models which are not implemented. Of
course we would be delighted to get contributions from individuals in those
fields!

Chad

[...]

PR #4250, opened 2018-02-05
ENH: Add frequency domain seasonal components to UnobservedComponents

Noobivsho:
Previously, we had implemented the main time domain seasonal component, which
is restrictive when the periodicity is very high or the seasonality can be
more parsimoniously represented by trigonometric terms.  Also, we were only
 able to use a single seasonal component.  This commit allows for multiple
 seasonal components in the frequency domain and for mixing of the two types.

Finally, we added an ipython notebook as an example showcasing the new
functionality.

Also
   - Change offset for regression coefficients when cycle is included.
     Cycle is two states regardless of if it is stochastic.  Therefore,
     the offset should always be 2 instead of possibly 1 for a deterministic
     cycle.

Signed-off-by: Jordan Yoder <jordan.yoder@gmail.com>

---- Additional commentary outside of commit message -------
This is as in Durbin and Koopman (2006) equations (3.7) and (3.8).  De Livera et al. (2010) mention this approach as well with the modifications to a configurable number of harmonics employed here.

I attempted to add to the previous signature, so I did not fold the current seasonal argument
into this, although I considered it.  If ya'll want, I am happy to discuss consolidations.

I searched through the issues for related items.  I found a couple...
Issues:
#4221
#3109

Pull requests:
#2892 -- I don't think we have conflicts in the code, although the purpose of his PR looks very similar.   He has some cool features around box-cox transforms and detecting the "correct" number of seasons that I don't.  Thus, I think my PR is a more spartan implementation, but it is in one of the existing frameworks, so there's that.

### DIFF
--- a/statsmodels/tsa/statespace/structural.py
+++ b/statsmodels/tsa/statespace/structural.py
@@ -60,6 +60,13 @@ class UnobservedComponents(MLEModel):
         `level` must also be True.
     seasonal : int or None, optional
         The period of the seasonal component, if any. Default is None.
+    freq_seasonal: list of dicts or None, optional.
+        Whether (and how) to model seasonal component(s) with trig. functions.
+        If specified, there is one dictionary for each frequency-domain
+        seasonal component.  Each dictionary must have the key, value pair for
+        'period' -- integer and may have a key, value pair for
+        'harmonics' -- integer. If 'harmonics' is not specified in any of the
+        dictionaries, it defaults to the floor of period/2.
     cycle : bool, optional
         Whether or not to include a cycle component. Default is False.
     ar : int or None, optional
@@ -73,7 +80,11 @@ class UnobservedComponents(MLEModel):
     stochastic_trend : bool, optional
         Whether or not any trend component is stochastic. Default is False.
     stochastic_seasonal : bool, optional
-        Whether or not any seasonal component is stochastic. Default is False.
+        Whether or not any seasonal component is stochastic. Default is True.
+    stochastic_freq_seasonal: list of bools, optional
+        Whether or not each seasonal component(s) is (are) stochastic.  Default
+        is True for each component.  The list should be of the same length as
+        freq_seasonal.
     stochastic_cycle : bool, optional
         Whether or not any cycle component is stochastic. Default is False.
     damped_cycle : bool, optional
@@ -203,6 +214,39 @@ class UnobservedComponents(MLEModel):
     time series is available in the results class in the `seasonal`
     attribute.
 
+    ** Freq. Seasonal**
+
+    Each frequency-domain seasonal component is modeled as:
+
+    .. math::
+        \gamma_t &=& \sum_{j=1}^h \gamma_{j, t}
+        \gamma_{j, t+1} &=& \gamma_{j, t}\cos(\lambda_j) \\
+                        && + \gamma^{*}_{j, t}\sin(\lambda_j) + \omega_{j,t} \\
+        \gamma^{*}_{j, t+1} &=^ -\gamma^{(1)}_{j, t}\sin(\lambda_j) \\
+                            &&+ \gamma^{*}_{j, t}\cos(\lambda_j) \\
+                            + \omega^{*}_{j, t}, \\
+        \omega^{*}_{j, t}, \omega_{j, t} &\sim& N(0, \sigma_{\omega^2}
+        \lambda_j = \frac{2 \pi j}{s}
+    where j ranges from 1 to h.
+
+    The periodicity (number of "seasons" in a "year") is s and the number of
+    harmonics is h.  Note that h is configurable to be less than s/2, but
+    s/2 harmonics is sufficient to fully model all seasonal variations of
+    periodicity s.  Like the time domain seasonal term (cf. Seasonal section,
+    above), the inclusion of the error terms allows for the seasonal effects to
+    vary over time.  The argument stochastic_freq_seasonal can be used to set
+    one or more of the seasonal components of this type to be non-random,
+    meaning they will not vary over time.
+
+    This component results in one parameter to be fitted using maximum
+    likelihood: :math: `\sigma_{\omega^2}`, and up to two parameters to be
+    chosen, the number of seasons s and optionally the number of harmonics
+    h, with :math: 1 \leq h \leq \floor(s/2).
+
+    After fitting the model, each unobserved seasonal component modeled in the
+    frequency domain is available in the results class in the `freq_seasonal`
+    attribute.
+    
     **Cycle**
 
     The cyclical component is intended to capture cyclical effects at time
@@ -299,9 +343,13 @@ class UnobservedComponents(MLEModel):
     """  # noqa:E501
 
     def __init__(self, endog, level=False, trend=False, seasonal=None,
-                 cycle=False, autoregressive=None, exog=None, irregular=False,
-                 stochastic_level=False, stochastic_trend=False,
-                 stochastic_seasonal=True, stochastic_cycle=False,
+                 freq_seasonal=None, cycle=False, autoregressive=None,
+                 exog=None, irregular=False,
+                 stochastic_level=False,
+                 stochastic_trend=False,
+                 stochastic_seasonal=True,
+                 stochastic_freq_seasonal=None,
+                 stochastic_cycle=False,
                  damped_cycle=False, cycle_period_bounds=None,
                  mle_regression=True,
                  **kwargs):
@@ -311,6 +359,15 @@ class UnobservedComponents(MLEModel):
         self.trend = trend
         self.seasonal_periods = seasonal if seasonal is not None else 0
         self.seasonal = self.seasonal_periods > 0
+        if freq_seasonal:
+            self.freq_seasonal_periods = [d['period'] for d in freq_seasonal]
+            self.freq_seasonal_harmonics = [d.get(
+                'harmonics', int(np.floor(d['period'] / 2))) for
+                d in freq_seasonal]
+        else:
+            self.freq_seasonal_periods = []
+            self.freq_seasonal_harmonics = []
+        self.freq_seasonal = any(x > 0 for x in self.freq_seasonal_periods)
         self.cycle = cycle
         self.ar_order = autoregressive if autoregressive is not None else 0
         self.autoregressive = self.ar_order > 0
@@ -319,6 +376,16 @@ class UnobservedComponents(MLEModel):
         self.stochastic_level = stochastic_level
         self.stochastic_trend = stochastic_trend
         self.stochastic_seasonal = stochastic_seasonal
+        if stochastic_freq_seasonal is None:
+            self.stochastic_freq_seasonal = [True] * len(
+                self.freq_seasonal_periods)
+        else:
+            if len(stochastic_freq_seasonal) != len(freq_seasonal):
+                raise ValueError(
+                    "Length of stochastic_freq_seasonal must equal length"
+                    " of freq_seasonal: {!r} vs {!r}".format(
+                        len(stochastic_freq_seasonal), len(freq_seasonal)))
+            self.stochastic_freq_seasonal = stochastic_freq_seasonal
         self.stochastic_cycle = stochastic_cycle
 
         self.damped_cycle = damped_cycle
@@ -414,6 +481,8 @@ class UnobservedComponents(MLEModel):
                 (self.level and self.stochastic_level) or
                 (self.trend and self.stochastic_trend) or
                 (self.seasonal and self.stochastic_seasonal) or
+                (self.freq_seasonal and any(
+                    self.stochastic_freq_seasonal)) or
                 (self.cycle and self.stochastic_cycle) or
                 self.autoregressive):
             warn("Specified model does not contain a stochastic element;"
@@ -423,6 +492,13 @@ class UnobservedComponents(MLEModel):
         if self.seasonal and self.seasonal_periods < 2:
             raise ValueError('Seasonal component must have a seasonal period'
                              ' of at least 2.')
+
+        if self.freq_seasonal:
+            for p in self.freq_seasonal_periods:
+                if p < 2:
+                    raise ValueError(
+                        'Frequency Domain seasonal component must have a '
+                        'seasonal period of at least 2.')
 
         # Create a bitmask holding the level/trend specification
         self.trend_mask = (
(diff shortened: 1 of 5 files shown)


ChadFulton:
This looks great, thanks very much for the contribution! I'll try to take a look it over closely as soon as I can.

One quick question - is the unit test a regression test or is it testing against a published result / or other statistical package?

Noobivsho:
It's merely a regression test.  The ipython notebook can be modified into a unit test for accuracy of the method by simulating a lot of data and comparing the estimated variances to the truth.  I was doing this when developing it.  I didn't add it in as a test of that nature because it didn't seem to fit with the rest of the tests on the module, although now I am guessing those were against a published result.

Do you prefer published result over other package?  I can definitely add one.

Noobivsho:
After doing more reading on that test file, I see you used KFAS in R to make your test cases.  KFAS has an option for configurable harmonics, but I believe it is currently bugged.  I made an issue on their project:

https://github.com/helske/KFAS/issues/30

I could just do full number of functions to avoid the bug.  Or use a different package?  Or "manually" create the component as in https://github.com/helske/KFAS/issues/21 but with the adjusted indexing logic?

ChadFulton:
Since the state space framework is pretty well tested at this point, I guess all we really need are tests that the state space system matrices are being set up and updated properly.

Why don't we keep the current regression test, and add a new test that just sets up a "somewhat complicated" UC model (e.g. maybe local linear trend with damped cycle and two of your trigonometric seasonal components) and then call `mod.update()` with some parameter vector and test that the state space matrices (most importantly design, transition, selection, and state_cov) are what you expect them to be.

Then we can make a separate issue to test against KFAS if / when that bug gets fixed?

ChadFulton:
> #2892 -- I don't think we have conflicts in the code, although the purpose of his PR looks very similar. He has some cool features around box-cox transforms and detecting the "correct" number of seasons that I don't. Thus, I think my PR is a more spartan implementation, but it is in one of the existing frameworks, so there's that.

I think your PR will be very valuable here. #2892 does look really interesting, but unfortunately it's been stalled for quite some time since the original author ran out of time / interest. It also uses / adds support for the single source of error state space models that Hyndman uses. These would be great to finish, since they would allow us to use our state space framework to encompass the exponential smoothing methods.

Noobivsho:
1) Sounds good about your test idea.  I do think I should mark it somehow to distinguish it from your other tests, whose results you generated in the .R code.

2) Re: 

> Then we can make a separate issue to test against KFAS if / when that bug gets fixed?

The author invited me to make a pull request to fix the bug (see the linked issue).  The fix is very simple--a one liner.  While I am slightly uncomfortable using my own code to "independently" verify my other code, I suppose his review of the change could make it independent enough to not seem too cozy?

Noobivsho:
Still working on this.  I think I got a nice test with KFAS (I avoided the bugged portion of the seasonal code by allowing full harmonics), but am still setting up the various matrices test of the moderately complicated model as per your suggestion.

ChadFulton:
That sounds great, thanks for your work on this. Let me know if there's anything I can do to help.

Noobivsho:
I think you could give the initialization a good look once I update it.  I realized that since I had a state that was a linear combination of other states to represent the actual seasonal term (for convenience in the summary object mostly), the initialization of the state cov matrix needs to be handled carefully for that part (only some of the states are diffuse).  I'll point out location once I update the PR with an additional commit.

Noobivsho:
Belay that about initialization.  I convinced myself to go ahead and remove the extra book-keeping state to keep things simple w.r.t. initialization.  It does make the summary object code slightly more complicated, since the component people would want to look at is actually a summation of multiple other components.

statsmodels/tsa/statespace/tests/test_structural.py
```diff
@@ -141,7 +141,7 @@ def test_fixed_slope():
     run_ucm('fixed_slope')
 
 
-def test_fixed_slope():
```
Noobivsho: Renamed this so it wouldn't conflict with the other test of the same name

statsmodels/tsa/statespace/tests/results/test_ucm.R
```diff
@@ -94,12 +94,23 @@ print(exp(res_cycle$optim.out$par)) # [37.57197224]
 print(res_cycle$optim.out$value)    # 672.3102588
 
 # Seasonal (with fixed period=4, seasonal variance=0.1)
-mod_seasonal <- SSModel(dta$unemp ~ -1 + SSMseasonal(4, Q=matrix(0.1), sea.type='dummy', P1=diag(1)*1e6), H=matrix(NA))
+# Note: matching this requires setting loglikelihood_burn=0
+mod_seasonal <- SSModel(dta$unemp ~ -1 + SSMseasonal(4, Q=matrix(0.1), sea.type='dummy', P1=diag(rep(1e6, 3))), H=matrix(NA))
```
Noobivsho: I noticed that the first P1 initialization is 1e6 oo', where o is the vector of ones of length 3.  I think we wanted the diagonal version.

statsmodels/tsa/statespace/structural.py
```diff
@@ -10,7 +10,6 @@
 from statsmodels.compat.collections import OrderedDict
 
 import numpy as np
-import pandas as pd
```
Noobivsho: This was an unused import.  I went ahead and removed it.

Noobivsho:
I don't really understand the travis-ci build error.  I saw it popped up on another recent PR as well.  Any help?

josef-pkt:
about travis 3.6 problem, I saw it this morning in a test run, it's unrelated to any statsmodels changes.
There is a problem installing R for the docbuild on that machine. R/rpy2 is currently only used in the example notebooks, AFAIK.  (I didn't even know or remember that we install R on travis.)

statsmodels/tsa/statespace/tests/test_structural.py
```diff
@@ -355,3 +359,89 @@ def test_predict_custom_index():
     res = mod.smooth(mod.start_params)
     out = res.predict(start=1, end=1, index=['a'])
     assert_equal(out.index.equals(pd.Index(['a'])), True)
+
+
+def test_matrices_somewhat_complicated_model():
```
Noobivsho: Here is the requested test on matrices

statsmodels/tsa/statespace/tests/results/results_structural.py
```diff
@@ -164,11 +164,21 @@
...
     'models': [{'irregular': True, 'seasonal': 4}],
     'params': [38.1704278, 0.1],
-    'llf': -655.3337155,
-    'kwargs': {},
+    'llf': -678.8138005,
+    'kwargs': {'loglikelihood_burn': 0},
     'rtol': 1e-6
 }
 
+freq_seasonal = {
```
Noobivsho: This test case is now against the KFAS results in test_ucm.R instead of just a regression test.

ChadFulton:
This looks great, thank you!

I probably won't get time to review until the weekend.

Noobivsho:
You bet!  While I've got you, can you confirm that the purpose of calling initialize_state at the end of every update is that the actual filter only holds the current state information and we need to restart it at the first timestep?  I had to timebox figuring this out.  If this is the case, I think stashing it to for reset later would be a cheap speed-up in the case of AR models, since we are just solving the same system

Noobivsho:
@ChadFulton 
Bump on this.  I don't want to be forgotten : )

If wanted I can rebase to a different PR, as I see some movement on your end for some cool stuff with initialization.

ChadFulton:
> Bump on this. I don't want to be forgotten : )

Thanks for the reminder! I would like to get this in for v0.9 if possible (depends on timing, but seems likely to me), and it slipped through the cracks (my memory).

> If wanted I can rebase to a different PR, as I see some movement on your end for some cool stuff with initialization.

When you get time, please rebase on current master - unless you want to wait for merging until after 0.9 is released (timeline is indeterminate right now), it's probably best to not rebase on the diffuse initialization branch yet, since I won't be merging that until after 0.9 comes out.

ChadFulton:
> You bet! While I've got you, can you confirm that the purpose of calling initialize_state at the end of every update is that the actual filter only holds the current state information and we need to restart it at the first timestep? I had to timebox figuring this out. If this is the case, I think memoizing it for reset later would be a cheap speed-up in the case of AR models, since we are just solving the same system

(I'm interpreting this as asking why `self._initialize_state()` is called in the function `KalmanFilter._filter` - let me know if I misunderstood).

In general, it needs to be reset because it may have changed based on the parameters. In the case of stationary initialization (which is used by e.g. AR models), it does change at every iteration because the unconditional variance changes if the AR parameter or error variance changes (e.g. in an AR(1) model, it is sigma2 / (1 - phi**2)).

In the case of diffuse initialization, it doesn't need to be recomputed, but I think in that case the call is not doing very much and the speed up would be negligible.

josef-pkt:
I'm really really trying to finish up a 0.9rc1 by the end of this week. (There are a few more days of merges, and then docs and release preparation.)

Noobivsho:
> (I'm interpreting this as asking why self._initialize_state() is called in the function KalmanFilter._filter - let me know if I misunderstood).

Sorry,  I had meant why self.initialize_state() is called in the function UnobservedComponents.update().  I think your answer still makes sense though, particularly

> In the case of diffuse initialization, it doesn't need to be recomputed, but I think in that case the call is not doing very much and the speed up would be negligible.

Noobivsho:
I believe I have successfully pushed my rebase.  It was suspiciously painless.

ChadFulton:
This PR is very impressive - you did a great job integrating the new components into `structural.py` and the notebook is really helpful.

I'm going to take another look through the details a little more carefully than my first pass, but I'm of the opinion that we can merge this tomorrow or this weekend.

The only argument against it would be that it's pretty new code, so we could alternatively merge it in right after 0.9 is released leave it in master for a while (i.e. to be released in 0.10) but it doesn't seem like that's necessary, and I think having the multiple seasonals is pretty useful (and the ability to model long-period seasonal effects with a smaller number of harmonics is great too).

ChadFulton:
If there are no other comments / concerns, I will plan to merge tomorrow.

statsmodels/tsa/statespace/structural.py
```diff
@@ -203,6 +214,39 @@ class UnobservedComponents(MLEModel):
     time series is available in the results class in the `seasonal`
     attribute.
 
+    ** Freq. Seasonal**
+
+    Each frequency-domain seasonal component is modeled as:
+
+    .. math::
```
ChadFulton: I don't think this is all formatted correctly (.. math:: will wrap in in \begin{split} \end{split} - is that what you were looking for?)

statsmodels/tsa/statespace/structural.py
```diff
@@ -203,6 +214,39 @@ class UnobservedComponents(MLEModel):
...
+        \gamma_t &=& \sum_{j=1}^h \gamma_{j, t}
+        \gamma_{j, t+1} &=& \gamma_{j, t}\cos(\lambda_j) \\
+                        && + \gamma^{*}_{j, t}\sin(\lambda_j) + \omega_{j,t} \\
+        \gamma^{*}_{j, t+1} &=^ -\gamma^{(1)}_{j, t}\sin(\lambda_j) \\
```
ChadFulton: &=^ should probably be &=&, right?

statsmodels/tsa/statespace/structural.py
```diff
@@ -203,6 +214,39 @@ class UnobservedComponents(MLEModel):
...
+                            &&+ \gamma^{*}_{j, t}\cos(\lambda_j) \\
+                            + \omega^{*}_{j, t}, \\
+        \omega^{*}_{j, t}, \omega_{j, t} &\sim& N(0, \sigma_{\omega^2}
```
ChadFulton: Closing parentheses on the Gaussian specification

statsmodels/tsa/statespace/structural.py
```diff
@@ -203,6 +214,39 @@ class UnobservedComponents(MLEModel):
...
+        \lambda_j = \frac{2 \pi j}{s}
+    where j ranges from 1 to h.
+
+    The periodicity (number of "seasons" in a "year") is s and the number of
+    harmonics is h.  Note that h is configurable to be less than s/2, but
+    s/2 harmonics is sufficient to fully model all seasonal variations of
```
ChadFulton: Is it better to say "s/2 harmonics is sufficient" or "s/2 harmonics are required"? IMO the first makes it seem like you might sometimes use more, while I think your point is that you might sometimes use less for parsimony but if you want to model all seasonal variations then you need s/2?

Am I understanding correctly?

statsmodels/tsa/statespace/structural.py
```diff
@@ -203,6 +214,39 @@ class UnobservedComponents(MLEModel):
...
+    periodicity s.  Like the time domain seasonal term (cf. Seasonal section,
+    above), the inclusion of the error terms allows for the seasonal effects to
+    vary over time.  The argument stochastic_freq_seasonal can be used to set
+    one or more of the seasonal components of this type to be non-random,
+    meaning they will not vary over time.
+
+    This component results in one parameter to be fitted using maximum
+    likelihood: :math: `\sigma_{\omega^2}`, and up to two parameters to be
+    chosen, the number of seasons s and optionally the number of harmonics
+    h, with :math: 1 \leq h \leq \floor(s/2).
```
ChadFulton: I think you need :math:`...` here (i.e. the ` are missing).

ChadFulton:
I had a couple of comments on the documentation formatting, but nothing that would prevent me from merging this as is, I / we can always fix that in a followup.

If you happen to have a chance to take a look at the comments in the next few hours, great. Otherwise, I will still merge this today.

ChadFulton:
Thanks @Noobivsho, this is a great added model!

statsmodels/tsa/statespace/structural.py
Noobivsho: Yes you have my main point.  I was also thinking that for artificial seasonal components (i.e. nature is a single sin wave), "required" isn't correct.  I like your phrasing better though, since virtually any "real" periodic function will not be perfectly modeled by less than s/2 trig. pairs.

Noobivsho:
Thanks for taking such a close look and cleaning up after my sloppy docs!  I was out of town last weekend, so this is a nice surprise to come back to : )

Issue #3604, opened 2017-04-10
Inconsistent results for mixed linear models under Python 2.7 and 3.4 


sosata:
_statsmodels.mixedlm_ generates different results under Python 2.7 and Python 3.4. Moreover, in Python 3.4, the model does not converge when _re_formula_ is used with variable slope.

Here is a simple example:

```python
import statsmodels.formula.api as smf
import pandas as pd
import numpy as np
from copy import deepcopy

np.random.seed(seed=0)

data1 = pd.DataFrame(columns=['x','y','id'])
data2 = pd.DataFrame(columns=['x','y','id'])
data1['x'] = np.random.normal(size=100)
data1['y'] = deepcopy(data1['x'])*3 + np.random.normal(size=100)
data1['id'] = np.ones(100)
data2['x'] = np.random.normal(size=100)
data2['y'] = deepcopy(data2['x'])*2 + np.random.normal(size=100)
data2['id'] = np.ones(100)*2

data = pd.concat([data1,data2])
data = data.reset_index(drop=True)

md = smf.mixedlm('y ~ x', data, groups=data['id'], re_formula='~ x')
mdf = md.fit(method='cg')
print(mdf.summary())
````

Python 2.7 converges and generates the following results:

```
Mixed Linear Model Regression Results
==============================================================
Model:               MixedLM   Dependent Variable:   y        
No. Observations:    200       Method:               REML    
No. Groups:          2         Scale:                0.9636  
Min. group size:     100       Likelihood:           -286.7528
Max. group size:     100       Converged:            Yes      
Mean group size:     100.0                                    
--------------------------------------------------------------
                   Coef.  Std.Err.   z    P>|z| [0.025 0.975]
--------------------------------------------------------------
Intercept           -0.048    0.532 -0.091 0.928 -1.091  0.994
x                    2.631    1.673  1.572 0.116 -0.649  5.911
Intercept RE         0.558                                    
Intercept RE x x RE  1.660                                    
x RE                 5.597                                    
==============================================================
```

Python 3.4 does not converge and gives these results:

```
/usr/local/lib/python3.4/dist-packages/statsmodels/base/model.py:496: ConvergenceWarning: Maximum Likelihood optimization failed to converge. Check mle_retvals
  "Check mle_retvals", ConvergenceWarning)
/usr/local/lib/python3.4/dist-packages/statsmodels/base/model.py:496: ConvergenceWarning: Maximum Likelihood optimization failed to converge. Check mle_retvals
[...]

PR #4320, opened 2018-03-06
Test: Add more MixedLM tests

kshedden:
work in progress...


kshedden:
Looking into #3604

kshedden:
Using `fit(method='powell')` eliminates many of the convergence warnings.

PR #4366, opened 2018-03-20
small steps of algebra simplification

jbrockmendel:
The NegBin and GeneralizedPoisson likelihood/score/hessians are big hotspots for profiling.  They are _also_ complicated spots of algebra that are difficult to read/inspect.

This cleans them up a little bit and does a _few_ substitutions to reduce redundant calculations.  There are several more steps worth of algebraic simplification available, but this is a place where its easy to make mistakes, so let's do this in bite-size pieces.

### DIFF
--- a/statsmodels/discrete/discrete_model.py
+++ b/statsmodels/discrete/discrete_model.py
@@ -1468,8 +1468,10 @@ class GeneralizedPoisson(CountModel):
         dmudb = mu * exog
 
         dalpha = (mu_p * (y * ((y - 1) / a2 - 2 / a1) + a2 / a1**2))
-        dparams = dmudb * (-a4 / a1 + a3 * a2 / (a1 ** 2) + (1 + a4) *
-                  ((y - 1) / a2 - 1 / a1) + 1 / mu)
+        dparams = dmudb * (-a4 / a1 +
+                           a3 * a2 / (a1 ** 2) +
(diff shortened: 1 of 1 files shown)


jbrockmendel:
Any objections to this?  Doing this as a first-step will make subsequent algebraic steps much easier.

jbrockmendel:
This is mostly just changing where line-breaks are to break on addition and subtraction and avoid breaking on multiplication and division.

jbrockmendel:
green button

jbrockmendel:
@ChadFulton little help?  Because this is an algebra-simplification thing, it needs to be broken into many individually-trivial steps.  That plan only works if there's a quick-ish turnaround.

ChadFulton:
I reviewed the changes and they look fine to me (I didn't see any substitutions that you mentioned in the OP?) although I don't have an opinion on whether they represent an improvement over the status quo.

I will plan to merge tomorrow night unless there are concerns.

jbrockmendel:
You're right; the main thing this does is change e.g.

```
(foo*bar + baz *
zap - baz * bar *
foo / baz)
```

to 

```
(foo * bar +
baz * zap -
baz * bar * foo / baz)
```

The idea being that this makes subsequent steps much more obvious.

jbrockmendel:
ping

ChadFulton:
Thanks @jbrockmendel

PR #4401, opened 2018-03-27
dont ignore use_self

jbrockmendel:


jbrockmendel:
@ChadFulton free lunch

jbrockmendel:
ping BUG

bashtage:
@josef-pkt Trivial bug fix.

jbrockmendel:
> use_self is from an earlier version that is not in use in the current version. (I think use_self is left over from having originally an identical signature between private and old public method.)

Then FFS it should be gotten rid of.

> I'm closing it so we have fewer open PRs)

The reasonable way to have fewer open PRs is to merge the many extremely easy ones that have been open for days, weeks, or in some cases years.

PR #4414, opened 2018-04-01
parametrize test, use smaller n_grp to (try to) expose numerical issue

jbrockmendel:


jbrockmendel:
CI errors reflect the PR being correct and the existing code being problematic.

kshedden:
@jbrockmendel thanks for working on this.

Here are a couple of ways to make the test problem a bit easier:

Option 1: Comment out this line at around line 108 of the test file:

```
slopes[:, -1] *= 2
```

Option 2: Change this line at around line 135 of the test file:

```
cov_re = np.random.normal(size=(k_re, k_re))
```

to something like this

```
cov_re = np.random.normal(size=(5*k_re, k_re))
```

For me, doing either of these things makes the gradients accurate to at least 10-4 or better, based on monitoring

```
np.max(np.abs((gr - ngr) / ngr))
```

I don't know what fundamentally is going on here, but I suspect that there may be some roundoff errors creeping in when calculating the log likelihood and/or score far from the MLE.  The changes above make it so that the test points are closer to the MLE.  The MLE for cov_re is `[[1, 0][0, 4]]`, but taking option 1 ` above changes it to `[[1, 0], [0, 1]]`.  The test values of `cov_re` are generated from a low-order Wishart distribution which can produce values very far from the mean (which is the 2x2 identity matrix).  Taking option 2 above samples from a higher order Wishart distribution so that the values will fall closer to the center of the distribution.

It would be nice to know whether the analytic score or numerical score are closer to the truth here, but I have no way to assess that.  Fortunately, it probably doesn't matter much because the gradient doesn't need to be highly accurate until you get close to the MLE.

I doubt that there is an actual error in the analytic score, because if there were, I think the values would be off by much more and would be off more consistently.  On the other hand, there is a lot of room for numerical error to creep into these calculations.  I haven't reviewed the code to try to identify places where algebraic rearrangement could improve numerical accuracy.

jbrockmendel:
@kshedden Not sure if this is mutually exclusive with your hypothesis, but AFAICT there is a code-consistency issue and a numerical stability issue.  For the next block of code I'm going to assume that you've run the contents of the test up until the first AssertionError (for me this happens at `kr==2`)

The test is checking that a numerical derivative of loglike matches score, but it _also_ is computing these using different representations of the parameters.  It looks like the parameters part is problem:

```
>>> model.loglike(params, profile_fe=profile_fe)
-272.4360448588463
>>> model.loglike(params_vec, profile_fe=profile_fe)
-273.815074867071
```

I'm _assuming_ these should match; is that assumption incorrect?  I _think_ that the mismatch is related to the hard-coding of `has_fe` in the `if type(params) is not MixedLMParams:` blocks in `score`/`loglike`/`hessian` (discussed elsewhere; I'd have to look it up)

Calling `model.score(params_vec, profile_fe=profile_fe)` raises a `ValueError` with mis-matched shapes.

Then there's the issue of numerical stability.  Let's approximate a numerical derivative in the relevant coordinate of `params_vec`:

```
def func(h):
    pvec2 = params_vec.copy()
    pvec2[-2] += h
    numerator = model.loglike(pvec2, profile_fe=profile_fe) - model.loglike(params_vec, profile_fe=profile_fe)
    return numerator / h

>>> func(1e-8)
0.013443468560581096
>>> -gr[-2]
0.013432134266906277
# Makes sense.

>>> func(1e-9)
0.013358203432289883
>>> -ngr[-2]
0.0133514404296875
# Checks out.

>>> func(1e-10)
0.013073986337985843
>>> func(1e-11)
0.011368683772161603
>>> func(1e-12)
0.056843418860808015
```

This seems like the sort of thing that could be a problem.

jbrockmendel:
@kshedden with the new smw_foo tests you've implemented (much appreciated) it looks like that isn't the issue here.  Any other ideas for where to look?

And uh unrelated, if you happen to get a chance to look at #4501, it's super-obvious and will soon be old enough to get a driver's license.

jbrockmendel:
Closing; #5802 can take over for discussion of this flaky test.

PR #4443, opened 2018-04-08
remove redundant method, comments

jbrockmendel:


statsmodels/tsa/vector_ar/irf.py
```diff
@@ -687,18 +688,11 @@ def H(self):
         Kkk = tsa.commutation_matrix(k, k)
         Ik = np.eye(k)
 
-        # B = chain_dot(Lk, np.eye(k**2) + commutation_matrix(k, k),
-        #               np.kron(self.P, np.eye(k)), Lk.T)
-
-        # return np.dot(Lk.T, L.inv(B))
-
```
jbrockmendel: Its straightforward to confirm that the commented-out code is equivalent to the version in place (at least algebraically; maybe there's a floating point difference?)

PR #4457, opened 2018-04-10
ENH: Add support for RangeIndexes

ChadFulton:
As of Pandas 0.18, the default index is `RangeIndex` rather than `Int64Index`. We only directly supported `Int64Index` (along with date indexes) so a warning would be issued by tsa models that the index would be ignored.

This adds direct support for `RangeIndex`es of all types.

### DIFF
--- a/statsmodels/tsa/base/tsa_model.py
+++ b/statsmodels/tsa/base/tsa_model.py
@@ -360,7 +360,7 @@ class TimeSeriesModel(base.LikelihoodModel):
             #   (e.g. a non-numeric string) will raise a ValueError if the
             #   index is RangeIndex (otherwise will raise an IndexError)
             #   (as of Pandas 0.22)
-            except IndexError, ValueError as e:
+            except (IndexError, ValueError) as e:
                 raise KeyError(str(e))
             loc = key
         else:
--- a/statsmodels/tsa/base/tsa_model.py
+++ b/statsmodels/tsa/base/tsa_model.py
@@ -352,8 +352,15 @@ class TimeSeriesModel(base.LikelihoodModel):
             try:
                 index[key]
             # We want to raise a KeyError in this case, to keep the exception
-            # consistent across index types
-            except IndexError as e:
+            # consistent across index types.
+            # - Attempting to index with an out-of-bound location (e.g.
+            #   index[10] on an index of length 9) will raise an IndexError
+            #   (as of Pandas 0.22)
+            # - Attemtping to index with a type that cannot be cast to integer
+            #   (e.g. a non-numeric string) will raise a ValueError if the
+            #   index is RangeIndex (otherwise will raise an IndexError)
+            #   (as of Pandas 0.22)
+            except IndexError, ValueError as e:
                 raise KeyError(str(e))
             loc = key
         else:
--- a/statsmodels/tsa/base/tsa_model.py
+++ b/statsmodels/tsa/base/tsa_model.py
@@ -229,7 +229,7 @@ class TimeSeriesModel(base.LikelihoodModel):
 
         if ((date_index and has_freq) or (int_index and is_increment) or
                 range_index):
-            _index = index.copy()
+            _index = index
         else:
             _index = increment
             index_generated = True
@@ -361,10 +361,6 @@ class TimeSeriesModel(base.LikelihoodModel):
 
         # Check if we now have a modified index
         index_was_expanded = index is not base_index
-
-        # (Never return the actual index object)
-        if not index_was_expanded:
-            index = index.copy()
 
         # Return the index through the end of the loc / slice
         if isinstance(loc, slice):
--- a/statsmodels/tsa/base/tsa_model.py
+++ b/statsmodels/tsa/base/tsa_model.py
@@ -7,12 +7,18 @@ import warnings
 import numpy as np
 from pandas import (to_datetime, Int64Index, DatetimeIndex, Period,
                     PeriodIndex, Timestamp, Series, Index)
+# RangeIndex only introduced in Pandas 0.18, so we include a shim until
+# Statsmodels requires that version.
+try:
+    from pandas import RangeIndex
+except ImportError:
+    class RangeIndex(object):
+        pass
 from pandas.tseries.frequencies import to_offset
 
 from statsmodels.base import data
 import statsmodels.base.model as base
 import statsmodels.base.wrapper as wrap
-from statsmodels.tsa.base import datetools
 from statsmodels.tools.sm_exceptions import ValueWarning
 
 _tsa_doc = """
@@ -204,12 +210,13 @@ class TimeSeriesModel(base.LikelihoodModel):
         has_index = index is not None
         date_index = isinstance(index, (DatetimeIndex, PeriodIndex))
         int_index = isinstance(index, Int64Index)
+        range_index = isinstance(index, RangeIndex)
         has_freq = index.freq is not None if date_index else None
         increment = Int64Index(np.arange(self.endog.shape[0]))
         is_increment = index.equals(increment) if int_index else None
 
         # Issue warnings for unsupported indexes
-        if has_index and not (date_index or is_increment):
+        if has_index and not (date_index or range_index or is_increment):
             warnings.warn('An unsupported index was provided and will be'
                           ' ignored when e.g. forecasting.', ValueWarning)
         if date_index and not has_freq:
@@ -220,7 +227,8 @@ class TimeSeriesModel(base.LikelihoodModel):
         # Construct the internal index
         index_generated = False
 
-        if (date_index and has_freq) or (int_index and is_increment):
+        if ((date_index and has_freq) or (int_index and is_increment) or
+                range_index):
             _index = index.copy()
         else:
             _index = increment
@@ -270,11 +278,25 @@ class TimeSeriesModel(base.LikelihoodModel):
 
         index = base_index
         date_index = isinstance(base_index, (PeriodIndex, DatetimeIndex))
+        int_index = isinstance(base_index, Int64Index)
+        range_index = isinstance(base_index, RangeIndex)
         index_class = type(base_index)
         nobs = len(index)
 
+        # Special handling for RangeIndex
+        if range_index and isinstance(key, (int, long, np.integer)):
+            # Negative indices (that lie in the Index)
+            if key < 0 and -key <= nobs:
+                key = nobs + key
+            # Out-of-sample (note that we include key itself in the new index)
+            elif key > nobs - 1:
+                stop = base_index._start + (key + 1) * base_index._step
+                index = RangeIndex(start=base_index._start,
+                                   stop=stop,
+                                   step=base_index._step)
+
         # Special handling for Int64Index
-        if (isinstance(index, Int64Index) and not date_index and
+        if (not range_index and int_index and not date_index and
                 isinstance(key, (int, long, np.integer))):
             # Negative indices (that lie in the Index)
             if key < 0 and -key <= nobs:
@@ -319,9 +341,23 @@ class TimeSeriesModel(base.LikelihoodModel):
                                             periods=len(index) + 1,
                                             freq=base_index.freq)
 
-        # Get the location (note that get_loc will throw a KeyError if key is
-        # invalid)
-        loc = index.get_loc(key)
+        # Get the location
+        if date_index:
+            # (note that get_loc will throw a KeyError if key is invalid)
+            loc = index.get_loc(key)
+        elif int_index or range_index:
+            # For Int64Index and RangeIndex, key is assumed to be the location
+            # and not an index value (this assumption is required to support
+            # RangeIndex)
+            try:
+                index[key]
+            # We want to raise a KeyError in this case, to keep the exception
+            # consistent across index types
+            except IndexError as e:
+                raise KeyError(str(e))
+            loc = key
+        else:
+            loc = index.get_loc(key)
 
         # Check if we now have a modified index
         index_was_expanded = index is not base_index
--- a/statsmodels/tsa/base/tsa_model.py
+++ b/statsmodels/tsa/base/tsa_model.py
@@ -212,7 +212,7 @@ class TimeSeriesModel(base.LikelihoodModel):
         int_index = isinstance(index, Int64Index)
         range_index = isinstance(index, RangeIndex)
         has_freq = index.freq is not None if date_index else None
-        increment = Int64Index(np.arange(self.endog.shape[0]))
+        increment = Index(range(self.endog.shape[0]))
         is_increment = index.equals(increment) if int_index else None
 
         # Issue warnings for unsupported indexes
--- a/statsmodels/tsa/tests/test_tsa_indexes.py
+++ b/statsmodels/tsa/tests/test_tsa_indexes.py
@@ -24,7 +24,8 @@ try:
     from pandas import RangeIndex
     has_range_index = True
 except ImportError:
-    pass
+    class RangeIndex(object):
+        pass
 
 from numpy.testing import (assert_allclose, assert_almost_equal, assert_equal,
                            assert_raises)
@@ -209,7 +210,7 @@ def test_instantiation_valid():
 
             mod = tsa_model.TimeSeriesModel(endog)
             assert_equal(isinstance(mod._index,
-                                    (pd.Int64Index, pd.RangeIndex)), True)
+                                    (pd.Int64Index, RangeIndex)), True)
             assert_equal(mod._index_none, True)
             assert_equal(mod._index_dates, False)
             assert_equal(mod._index_generated, True)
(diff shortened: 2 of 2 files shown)


ChadFulton:
Everything green and I think this is pretty straightforward, but I will wait a few days to merge in case anyone has a chance to look at it.

statsmodels/tsa/base/tsa_model.py
```diff
@@ -204,12 +210,13 @@ def _init_dates(self, dates=None, freq=None):
         has_index = index is not None
         date_index = isinstance(index, (DatetimeIndex, PeriodIndex))
         int_index = isinstance(index, Int64Index)
+        range_index = isinstance(index, RangeIndex)
         has_freq = index.freq is not None if date_index else None
         increment = Int64Index(np.arange(self.endog.shape[0]))
```
jbrockmendel: In cases where RangeIndex is available, using that for `increment` prevents a memory allocation.

statsmodels/tsa/base/tsa_model.py
```diff
@@ -319,9 +341,23 @@ def _get_index_loc(self, key, base_index=None):
...
+            loc = index.get_loc(key)
+        elif int_index or range_index:
+            # For Int64Index and RangeIndex, key is assumed to be the location
+            # and not an index value (this assumption is required to support
+            # RangeIndex)
+            try:
+                index[key]
+            # We want to raise a KeyError in this case, to keep the exception
+            # consistent across index types
+            except IndexError as e:
```
jbrockmendel: a) why do we care that it is always a KeyError?  I'm not clear on why index.get_loc doesn't work in this case.
b) Personal preference for the "# We want to raise [...]" comment to go after the `except IndexError as e:`, with an extra level of indentation.  YMMV.
c) This method is pretty big, could stand to be refactored into smaller pieces.  This block (344-360) would be a nice scope because it (almost) just depends on `index` and `key`.  self-contained tests!

statsmodels/tsa/base/tsa_model.py
```diff
@@ -319,9 +341,23 @@ def _get_index_loc(self, key, base_index=None):
...
+                raise KeyError(str(e))
+            loc = key
+        else:
+            loc = index.get_loc(key)
```
jbrockmendel: Out of scope, but commenting on 331/367 below: why is copying index necessary?  `pd.Index` objects are immutable(ish).

statsmodels/tsa/base/tsa_model.py
```diff
@@ -220,7 +227,8 @@ def _init_dates(self, dates=None, freq=None):
         # Construct the internal index
         index_generated = False
 
-        if (date_index and has_freq) or (int_index and is_increment):
+        if ((date_index and has_freq) or (int_index and is_increment) or
+                range_index):
```
jbrockmendel: Should have asked about the index.copy() here.

Also, w/r/t 236-241, are all of these attributes necessary?  ("Out Of Scope/Not My Problem is an acceptable answer here")  `_index_inferred_freq` is never used.  `_index_generated` is used, but it isn't entirely obvious why (also it is used identically in three places, needs to be made a function)

statsmodels/tsa/tests/test_tsa_indexes.py
```diff
@@ -307,6 +322,20 @@ def test_instantiation_valid():
         assert_equal(mod.data.dates, None)
         assert_equal(mod.data.freq, None)
 
+        if has_range_index:
```
jbrockmendel: Can this become a separate test?  flake8 complains about this test being too complicated already

ChadFulton:
Thanks @jbrockmendel for the review! I will reply in-line.

statsmodels/tsa/base/tsa_model.py
ChadFulton: First, to be honest, I do not fully understand the design decisions between `KeyError` and `IndexError` in Pandas. For example, the following:

```python
x = pd.Int64Index(np.arange(10))
x.get_loc(11)  # raises KeyError
x[11]          # raises IndexError
```

Now, this part of the code is relying on Pandas to tell me if the key is invalid (i.e. we don't want to have to figure it out ourselves), and we'd ideally like to check using `index.get_loc(key)` so that it works with date strings, etc. This also works fine for incrementing `Int64Index`es, but the problem is that for `RangeIndex`, the index values are integers too, so for example:

```python
x = pd.RangeIndex(start=0, stop=10, step=2)
print(x.get_loc(2))  # 1
print(x[2])          # 4
```

Now in this code I want to check that the key is a valid part of the index, so for a `RangeIndex` I have to do that using `index[key]` rather than `index.get_loc(key)` and for some reason this raises an `IndexError`. Intuitively, though, I'm searching for the existence of a key, and I want to raise a `KeyError` if it doesn't exist.

If there's a better way to do this check consistently, that's good too.

ChadFulton: (c) I suppose we could refactor it into `_get_index_loc` which then calls `_get_index_loc_int64index`, `_get_index_loc_rangeindex`, and `_get_index_loc_datetindex`. What do you think?

statsmodels/tsa/base/tsa_model.py
ChadFulton: I didn't realize that they were immutable. If so, I guess there's no reason we couldn't return the original index.

statsmodels/tsa/base/tsa_model.py
ChadFulton: This is a good idea. I'll try to play around with it to see if I can support pre-0.18 w/ `Int64Index` and post-0.18 with `RangeIndex` without too much special casing code.

statsmodels/tsa/base/tsa_model.py
ChadFulton: I think these attributes are useful.

> _index_generated is used, but it isn't entirely obvious why

It is necessary to know if the user provided an index that cannot be used for forecasting (in which case the forecasting index is a generated index). This is because in-sample prediction can still be done using keys from the provided index.

> (also it is used identically in three places, needs to be made a function)

I don't understand this comment.

statsmodels/tsa/base/tsa_model.py
jbrockmendel: Another option is just `Index(range(self.endog.shape[0])``.  That defaults to Int64Index in older versions and RangeIndex in newer versions.

statsmodels/tsa/base/tsa_model.py
jbrockmendel: > I think these attributes are useful.

Does that include _index_inferred_freq?  Because that's never used.  The what-attributes-exist guessing game is one of my pet peeves in sm, so I'm pushing to trim unused attributes where possible.


>> (also it is used identically in three places, needs to be made a function)

> I don't understand this comment.

sarimax:
            # Reset the ModelData datasets and cache
            self.data.endog, self.data.exog = (
                self.data._convert_endog_exog(endog, exog))

            # Reset indexes, if provided
            new_length = self.data.endog.shape[0]
            if self.data.row_labels is not None:
                self.data._cache['row_labels'] = (
                    self.data.row_labels[orig_length - new_length:])
            if self._index is not None:
                if self._index_generated:
                    self._index = self._index[:-(orig_length - new_length)]
                else:
                    self._index = self._index[orig_length - new_length:]
```

markov_autoregression
```
        # Reset indexes, if provided
        if self.data.row_labels is not None:
            self.data._cache['row_labels'] = (
                self.data.row_labels[self.order:])
        if self._index is not None:
            if self._index_generated:
                self._index = self._index[:-self.order]
            else:
                self._index = self._index[self.order:]
```

arima_model
```
        # Reset index
        orig_length = len(self._index)
        new_length = self.endog.shape[0]
        if self.data.row_labels is not None:
            self.data._cache['row_labels'] = (
                self.data.row_labels[orig_length - new_length:])
        if self._index is not None:
            if self._index_generated:
                self._index = self._index[:-(orig_length - new_length)]
            else:
                self._index = self._index[orig_length - new_length:]
```

File: statsmodels/tsa/base/tsa_model.py (hunk 2 above)
Comment (jbrockmendel) [reply statsmodels/tsa/base/tsa_model.py]: I'll take a look at the KeyError vs IndexError issue.

On the splitting-up-big-methods, feel free to copy/paste from [here](https://github.com/jbrockmendel/sm2/blob/master/sm2/tsa/base/tsa_model.py).

File: statsmodels/tsa/tests/test_tsa_indexes.py (hunk 5 above)
Comment (jbrockmendel) [reply statsmodels/tsa/tests/test_tsa_indexes.py]:
```
@pytest.mark.skipif(not has_range_index, reason="RangeIndex not available with older pandas")
def test_instantiation_valid_range_index():
     for base_endog in dta[2:4]:
```

... followed by lines 326-337, but with one less level of indentation.

Also if you're interested in a copy/paste option to split up this giant test (pre-this-PR), [here ya go](https://github.com/jbrockmendel/sm2/blob/master/sm2/tsa/tests/test_tsa_indexes.py.

File: statsmodels/tsa/tests/test_tsa_indexes.py [hunk 6]
```diff
@@ -710,6 +739,105 @@ def test_prediction_increment_pandas_dates_nanosecond():
     assert_equal(prediction_index.equals(desired_index), True)
 
 
+@pytest.mark.skipif(not has_range_index, reason='No RangeIndex')
```
jbrockmendel: I haven't looked at the rest of this carefully yet, will circle back in a bit.

statsmodels/tsa/base/tsa_model.py
ChadFulton: Thanks!

ChadFulton:
I'm going to keep this PR more tightly scoped than breaking up a bunch of methods, but thanks again for the review.

statsmodels/tsa/base/tsa_model.py
ChadFulton: > Does that include _index_inferred_freq? Because that's never used.

Good question, but it's the only way to tell if the index has an inferred frequency versus one explicitly passed by the user, and also has a self-explanatory name, so I think it's okay to keep.

statsmodels/tsa/tests/test_tsa_indexes.py
```diff
@@ -17,6 +17,16 @@
 import pandas as pd
 import pytest
 
+# RangeIndex only introduced in Pandas 0.18, so we include a shim until
+# Statsmodels requires that version.
```
jbrockmendel: what version _does_ statsmodels require?  setup.py says 0.13, INSTALL.txt says 0.14, and both of those are absurdly old (0.23.0 is being released before long)

statsmodels/tsa/tests/test_tsa_indexes.py
```diff
@@ -710,6 +739,105 @@ def test_prediction_increment_pandas_dates_nanosecond():
     assert_equal(prediction_index.equals(desired_index), True)
 
 
+@pytest.mark.skipif(not has_range_index, reason='No RangeIndex')
+def test_range_index():
+    tsa_model.__warningregistry__ = {}
+
+    endog = pd.Series(np.random.normal(size=5))
+    assert_equal(isinstance(endog.index, pd.RangeIndex), True)
+    # Warning should not be given
```
jbrockmendel: Does this mean that the call shouldn't emit any warnings, or that the test should suppress the warnings?  If the former, then better not to catch them.

statsmodels/tsa/tests/test_tsa_indexes.py
```diff
@@ -710,6 +739,105 @@ def test_prediction_increment_pandas_dates_nanosecond():
...
+        assert_equal(len(w), 0)
+
+
+@pytest.mark.skipif(not has_range_index, reason='No RangeIndex')
+def test_prediction_rangeindex():
+    index = supported_increment_indexes[2][0]
+    endog = pd.Series(dta[0], index=index)
+    mod = tsa_model.TimeSeriesModel(endog)
+
+    # Basic prediction: [0, end]
```
jbrockmendel: What makes the three cases (Basic prediction, Negative Index, Forecasting) special?  Does this cover the space of Cases That Should Be Tested?  Basically asking for a comment at the top of this test

statsmodels/tsa/tests/test_tsa_indexes.py
```diff
@@ -710,6 +739,105 @@ def test_prediction_increment_pandas_dates_nanosecond():
...
+
+    assert_equal(start, 1)
+    assert_equal(end, 4)
+    assert_equal(out_of_sample, 1)
+    desired_index = pd.RangeIndex(start=-4, stop=1, step=1)
+    assert_equal(prediction_index.equals(desired_index), True)
+
+
+@pytest.mark.skipif(not has_range_index, reason='No RangeIndex')
+def test_prediction_rangeindex_withstep():
```
jbrockmendel: Same as above re asking for a comment.  Also this is a good use case for parametrizing these two tests and getting rid of a bunch of duplicated code.

jbrockmendel:
A few new requests for comments in the tests notwithstanding, LGTM.

statsmodels/tsa/tests/test_tsa_indexes.py
ChadFulton: I think I remember that the policy is that we support whatever is in Travis, so that would be 0.14. I think after 0.9 the plan is to boost the minimums. For example, we will at least have a minimum of scipy = 0.16 so that we can drop the compatibility layer for state space models. I don't know if there's a particular version target for Pandas.

statsmodels/tsa/tests/test_tsa_indexes.py
ChadFulton: This means that prior to this PR (i.e. in master right now), a warning would be issued with any `RangeIndex` (including the default index for any pandas series for which an index was not explicitly given by the user). The test is that after the PR a warning is *not* issued for `RangeIndex`.

statsmodels/tsa/base/tsa_model.py
ChadFulton: > Another option is just `Index(range(self.endog.shape[0])``. That defaults to Int64Index in older versions and RangeIndex in newer versions.

This doesn't seem to be true for me?

```python
import pandas as pd
print(pd.__version__)
print(pd.Index(range(3)))
```

```
0.22.0
Int64Index([0, 1, 2], dtype='int64')
```

ChadFulton: Maybe there is a difference between Python 2 and Python 3 behavior here.

statsmodels/tsa/tests/test_tsa_indexes.py
ChadFulton: I think these are the three most common use cases, but are not exhaustive.

statsmodels/tsa/tests/test_tsa_indexes.py
ChadFulton: > Also this is a good use case for parametrizing these two tests and getting rid of a bunch of duplicated code.

I agree, but will leave it for later :)

ChadFulton:
Assuming it comes back green, I plan to merge tonight unless there are concerns.

ChadFulton:
Thanks very much for the review @jbrockmendel (even though I didn't get all the improvements into the PR :).

Merging.

statsmodels/tsa/tests/test_tsa_indexes.py
```diff
@@ -17,6 +17,16 @@
 import pandas as pd
 import pytest
 
+# RangeIndex only introduced in Pandas 0.18, so we include a shim until
```
jbrockmendel: Should this file be in tsa.base.tests?

statsmodels/tsa/base/tsa_model.py
```diff
@@ -204,12 +210,13 @@ def _init_dates(self, dates=None, freq=None):
         has_index = index is not None
         date_index = isinstance(index, (DatetimeIndex, PeriodIndex))
         int_index = isinstance(index, Int64Index)
+        range_index = isinstance(index, RangeIndex)
         has_freq = index.freq is not None if date_index else None
-        increment = Int64Index(np.arange(self.endog.shape[0]))
+        increment = Index(range(self.endog.shape[0]))
         is_increment = index.equals(increment) if int_index else None
```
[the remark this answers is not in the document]
ChadFulton: The `is_increment` is only used in combination with a passed integer index, so we can avoid the comparison via the short circuit.

(The issue is that we can only accept the index provided by the user if we can unambiguously extend it when forecasting. We can definitely do that for any date index with an associated frequency or any range index, but any random combination of integers can be an int64 index, so we only assume that we can extend the in64 index if it is identical to the basic increment index.)

statsmodels/tsa/tests/test_tsa_indexes.py
ChadFulton: It definitely could be.

PR #4469, opened 2018-04-12
 BUG: Fix simulation smoothing with state intercept

ChadFulton:
This fixes a bug with simulation smoothing when there is a state intercept. It's basically the same underlying issue as was in #3305, but now in the state intercept case.

This issue and the fix is discussed in https://www.ecb.europa.eu/pub/pdf/scpwps/ecbwp1867.en.pdf


ChadFulton:
Merging since it's green and it's a pretty straightforward bug fix.

PR #4473, opened 2018-04-12
partial cleanup of stattools files

jbrockmendel:
`stats.stattools` and `tsa.stattools` are especially important to get right, so merit some extra effort.

There are a number of places, particularly in `tsa.stattools`, where code is commented-out or needs attention from @josef-pkt.  This PR cleans up everything _other_ than those places, making it much easier to identify spots in need of attention.  Specifically, anything where there is a `#foo` instead of a `#foo` is commented-out code that needs to be either documented or deleted.

### DIFF
--- a/statsmodels/stats/stattools.py
+++ b/statsmodels/stats/stattools.py
@@ -62,8 +62,8 @@ def omni_normtest(resids, axis=0):
     -------
     Chi^2 score, two-tail probability
     """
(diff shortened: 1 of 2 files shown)


jbrockmendel:
BTW stats.stattools is in pretty good shape.  Only real todo-list item is a TODO note saying durbin_watson needs should be tested

PR #4532, opened 2018-04-22
fix distracting whitespace

jbrockmendel:
Getting this out of the way will make follow-up work in this module moderately easier.


jbrockmendel:
@josef-pkt rebased (tried to do it the way you like; LMK).  I know you don't care about whitespace but this will make it easier for me to work in this module so I'm asking nicely.

jbrockmendel:
@josef-pkt rebase like you like

PR #4594, opened 2018-05-05
TEST/DOC: SMW linalg routines documentation and test

kshedden:
#4455

### DIFF
--- a/statsmodels/regression/tests/test_lme.py
+++ b/statsmodels/regression/tests/test_lme.py
@@ -2,12 +2,14 @@
 import warnings
 import numpy as np
 import pandas as pd
-from statsmodels.regression.mixed_linear_model import MixedLM, MixedLMParams
+from statsmodels.regression.mixed_linear_model import (
+    MixedLM, MixedLMParams, _smw_solver, _smw_logdet)
 from numpy.testing import (assert_almost_equal, assert_equal, assert_allclose,
                            assert_)
 from . import lme_r_results
 from statsmodels.base import _penalties as penalties
 import pytest
+from scipy import sparse
 import statsmodels.tools.numdiff as nd
 import os
 import csv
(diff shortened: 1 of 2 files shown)


kshedden:
In addition to working on the docstrings, I also did some minor refactoring of `_smw_solver` and `_smw_logdet`, and modified the sparse matrix handling a bit so that product matrices that are not very sparse are stored in dense form.

statsmodels/regression/tests/test_lme.py
```diff
@@ -1094,6 +1096,62 @@ def test_random_effects_getters():
...
+    A = np.random.normal(size=(p, q))
+    AtA = np.dot(A.T, A)
+
+    B = np.zeros((q, q))
+    c = np.random.normal(size=(r, r))
+    B[0:r, 0:r] = np.dot(c.T, c)
+    di = np.random.uniform(size=s)
+    B[r:q, r:q] = np.diag(1/di)
+    Qi = np.linalg.inv(B[0:r, 0:r])
+    s = 0.5
```
jbrockmendel: @kshedden thanks for putting this together.  I'm interested in putting together something like hypothesis-style tests for this (recall I'm ultimately interested in debugging #4414).  To that end: is the construction above effectively the general case for constructing valid inputs for these two functions>\?

kshedden: Yes, I think this test is fairly general, except for the dimensions which are hard-coded.

jbrockmendel: Thanks.  Any restrictions on valid (or reasonable) dimensions?

kshedden: No restrictions, it should work for a wide variety of dimensions.

jbrockmendel: Substituting in `np.random.randint(0, 5)` for each of `p, q, r, s` it looks like we need the restriction `q = r + s`.  The dimension errors that pop up after that are harder to pin down; I'll list them here as I figure them out.

jbrockmendel: It looks like the setup in test_smw_logdet is slightly different from test_smw_solver:

```
    c = np.random.normal(size=(r, r))
    B[0:r, 0:r] = np.dot(c.T, c)
```

vs

```
    B[0:r, 0:r] = np.random.normal(size=(r, r))
```

Is that intentional?  My naive prior is that the setups would be identical.

jbrockmendel: Bounding `p >= 1` fixes all the failures in test_smw_logdet; but test_smw_solver fails (raises with mismatches dimensions) in ~40% of cases.

PR #4686, opened 2018-05-26
TEST: Docstring edits and variable name changes for clarity

kshedden:
@jbrockmendel if you have other suggestions for naming, documentation, and testing of the two SMW routines let me know, we can add them here.  So far, I don't see any indication that there is anything materially wrong with this code.

### DIFF
--- a/statsmodels/regression/tests/test_lme.py
+++ b/statsmodels/regression/tests/test_lme.py
@@ -391,6 +391,9 @@ class TestMixedLM(object):
         # library(geepack)
         # rm = lmer(Weight ~ Time + (1 | Pig), data=dietox)
         # rm = lmer(Weight ~ Time + (1 | Pig), REML=FALSE, data=dietox)
+        #
+        # Comments below are R code used to extract the numbers used
+        # for comparison.
 
         cur_dir = os.path.dirname(os.path.abspath(__file__))
         rdir = os.path.join(cur_dir, 'results')
@@ -450,6 +453,9 @@ class TestMixedLM(object):
         # library(geepack)
         # r = lmer(Weight ~ Time + (1 + Time | Pig), data=dietox)
         # r = lmer(Weight ~ Time + (1 + Time | Pig), REML=FALSE, data=dietox)
+        #
+        # Comments below are the R code used to extract the constants
+        # for comparison.
 
         cur_dir = os.path.dirname(os.path.abspath(__file__))
         rdir = os.path.join(cur_dir, 'results')
--- a/statsmodels/regression/tests/test_lme.py
+++ b/statsmodels/regression/tests/test_lme.py
@@ -13,7 +13,6 @@ import statsmodels.tools.numdiff as nd
 import os
 import csv
 import scipy
-import itertools
 import nose.tools
 
 # TODO: add tests with unequal group sizes
@@ -620,10 +619,7 @@ class TestMixedLM(object):
         model1 = MixedLM(endog, exog, groups, exog_re=exog_re, exog_vc=exog_vc)
         result1 = model1.fit()
 
-        df = pd.DataFrame(
-            exog[:, 1:], columns=[
-                "x1",
-            ])
+        df = pd.DataFrame(exog[:, 1:], columns=["x1"])
         df["y"] = endog
         df["re1"] = exog_re[:, 0]
         df["re2"] = exog_re[:, 1]
@@ -767,8 +763,8 @@ class TestMixedLMSummary(object):
         # Test that the summary correctly includes all variables.
         summ = self.res.summary()
         desired = ["const", "x1", "x2", "Group Var"]
-        actual = summ.tables[
-            1].index.values  # Second table is summary of params
+        # Second table is summary of params
+        actual = summ.tables[1].index.values
         assert_equal(actual, desired)
 
     def test_summary_xname_fe(self):
@@ -857,9 +853,9 @@ fnames = [x for x in fnames if x.startswith("lme") and x.endswith(".csv")]
 
 # Copied from bashtage's #3847
 @nose.tools.nottest
-@pytest.mark.parametrize('fname,reml,irf',
-                         itertools.product(fnames, [False, True],
-                                           [False, True]))
+@pytest.mark.parametrize('fname', fnames)
+@pytest.mark.parametrize('reml', [False, True])
+@pytest.mark.parametrize('irf', [False, True])
 def test_r(fname, reml, irf):
     ds_ix = int(fname[3:5])
     do1(reml, irf, ds_ix)
@@ -1187,10 +1183,10 @@ def test_smw_solver():
         y1 = f(x)
         assert_allclose(y1, y2)
 
-    for p in 5, 10:
-        for q in 4, 8:
-            for r in 2, 3:
-                for s in 0, 0.5:
+    for p in (5, 10):
+        for q in (4, 8):
+            for r in (2, 3):
+                for s in (0, 0.5):
                     tester(p, q, r, s)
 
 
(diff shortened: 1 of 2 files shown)


statsmodels/regression/tests/test_lme.py
```diff
@@ -47,8 +49,7 @@ def __init__(self, meth, irfs, ds_ix):
         self.loglike = getattr(lme_r_results, "loglike" + bname)
 
         if hasattr(lme_r_results, "ranef_mean" + bname):
-            self.ranef_postmean = getattr(lme_r_results, "ranef_mean"
-                                          + bname)
+            self.ranef_postmean = getattr(lme_r_results, "ranef_mean" + bname)
```
jbrockmendel: Is this going to cause flake8 complaints about lines longer than 80 characters?  Not a big deal, but if we're being picky...

statsmodels/regression/tests/test_lme.py
```diff
@@ -327,18 +339,19 @@ def test_vcomp_3(self):
...
-                        np.r_[-0.101549, 0.028613, -0.224621, -0.126295],
-                        rtol=1e-3)
+        assert_allclose(
+            result.resid.iloc[0:4],
+            np.r_[-1.180753, 0.279966, 0.578576, -0.667916],
+            rtol=1e-3)
+        assert_allclose(
+            result.fittedvalues.iloc[0:4],
+            np.r_[-0.101549, 0.028613, -0.224621, -0.126295],
+            rtol=1e-3)
```
jbrockmendel: The older formatting here is a pattern I kinda like.  Any particular reason for disliking it?

statsmodels/regression/tests/test_lme.py
```diff
@@ -417,8 +440,8 @@ def test_dietox(self):
         assert_allclose(result.cov_re, 39.82097, rtol=1e-5)
 
         # logLik(rm)
```
jbrockmendel: `logLik` is almost certainly out of date, doesn't serve as a very useful comment.  Not sure what its replacement would be.

statsmodels/regression/tests/test_lme.py
```diff
@@ -583,7 +620,10 @@ def test_vcomp_formula(self):
         model1 = MixedLM(endog, exog, groups, exog_re=exog_re, exog_vc=exog_vc)
         result1 = model1.fit()
 
-        df = pd.DataFrame(exog[:, 1:], columns=["x1", ])
+        df = pd.DataFrame(
+            exog[:, 1:], columns=[
+                "x1",
+            ])
```
jbrockmendel: Breaking the columns list into three lines is pretty ugly.  Standard thing to do here would be to keep the original but remove the trailing `, ` inside the `columns` list.

statsmodels/regression/tests/test_lme.py
```diff
@@ -724,23 +767,27 @@ def test_summary(self):
         # Test that the summary correctly includes all variables.
         summ = self.res.summary()
         desired = ["const", "x1", "x2", "Group Var"]
-        actual = summ.tables[1].index.values # Second table is summary of params
+        actual = summ.tables[
+            1].index.values  # Second table is summary of params
```
jbrockmendel: +1 for adding the extra space before the `#`, but does the line need to be split like this?  It's not a pattern most readers are used to.

statsmodels/regression/tests/test_lme.py
```diff
@@ -808,18 +854,17 @@ def do1(reml, irf, ds_ix):
 fnames = os.listdir(rdir)
 fnames = [x for x in fnames if x.startswith("lme") and x.endswith(".csv")]
 
-import itertools
-import nose.tools
-import pytest
```
jbrockmendel: +1 for imports at the top of the file (and I think removing a duplicated import)

statsmodels/regression/tests/test_lme.py
```diff
@@ -808,18 +854,17 @@ def do1(reml, irf, ds_ix):
...
 
 # Copied from bashtage's #3847
 @nose.tools.nottest
 @pytest.mark.parametrize('fname,reml,irf',
-                         itertools.product(fnames, [False, True], [False, True]))
+                         itertools.product(fnames, [False, True],
+                                           [False, True]))
```
jbrockmendel: As long as this is being futzed with, the idiomatic way to do this would be

```
@pytest.mark.parametrize('fname', fnames)
@pytest.mark.parametrize('reml', [False, True])
@pytest.mark.parametrize('irf', [False, True])
def test_r(fname, reml, irf):
```

statsmodels/regression/tests/test_lme.py
```diff
@@ -986,51 +1031,53 @@ def test_handle_missing():
...
     # provide start_params to speed up convergence
-    sp1 = np.array([-1.26722599,  1.1617587 ,  0.19547518])
-    mod1 = MixedLM.from_formula('Y ~ X', d, groups=d ['IDS'])
+    sp1 = np.array([-1.26722599, 1.1617587, 0.19547518])
+    mod1 = MixedLM.from_formula('Y ~ X', d, groups=d['IDS'])
     results1 = mod1.fit(start_params=sp1)
-    sp2 = np.array([ 3.48416861,  0.55287862,  1.38537901])
-    mod2 = MixedLM.from_formula('X ~ Y', d, groups=d ['IDS'])
+    sp2 = np.array([3.48416861, 0.55287862, 1.38537901])
+    mod2 = MixedLM.from_formula('X ~ Y', d, groups=d['IDS'])
```
jbrockmendel: +1 for whitespace fixups here

statsmodels/regression/tests/test_lme.py
```diff
@@ -1057,99 +1104,125 @@ def test_random_effects_getters():
...
+        assert_allclose(y1, y2)
 
-    f = _smw_solver(s, sparse.csr_matrix(A), sparse.csr_matrix(AtA), Qi, di)
-    y1 = f(x)
-    assert_allclose(y1, y2)
+    for p in 5, 10:
+        for q in 4, 8:
+            for r in 2, 3:
+                for s in 0, 0.5:
+                    tester(p, q, r, s)
```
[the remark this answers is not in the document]
jbrockmendel: Oh and brackets around the for-loop args pls

statsmodels/regression/tests/test_lme.py
```diff
@@ -1057,99 +1104,125 @@ def test_random_effects_getters():
...
+        assert_allclose(d1, d2)
+
+    for p in 5, 10:
+        for q in 4, 8:
+            for r in 2, 3:
+                for s in 0, 0.5:
+                    tester(p, q, r, s)
 
 
 if __name__ == "__main__":
```
jbrockmendel: if deleting this is an option, it's ready to go

jbrockmendel:
A few comments, but generally I like this a lot.

What do you think of putting the smw functions in tools.linalg (and moving corresponding tests)?  I advocate it based on a) if a reader were to look for such a function, that'd be the natural place, and b) collecting easy-to-test-in-isolation code in tools can help us move toward cordoning off well-tested from not-well-tested code.  (Totally OK for the idea to be out-of-scope for this PR)

statsmodels/regression/tests/test_lme.py
```diff
@@ -1057,99 +1104,125 @@ def test_random_effects_getters():
...
+
+    def tester(p, q, r, s):
+
+        d = q - r
+        A = np.random.normal(size=(p, q))
+        AtA = np.dot(A.T, A)
+
+        B = np.zeros((q, q))
+        c = np.random.normal(size=(r, r))
+        B[0:r, 0:r] = np.dot(c.T, c)
```
jbrockmendel: The other `tester` func sets `B[0:r, 0:r] = c` (effectively).  Is the difference intentional, and if not, can these share some code?

kshedden: The construction of B needs to be different.  In the log determinant calculation, B needs to be positive definite, but for the solver B only needs to be non-singular.

kshedden:
@jbrockmendel, thanks for the review. I updated some of the formatting based on your suggestions.

If no further comments I will merge.

Moving the smw functions to the tools module sounds like a good idea.  Let's do that as a separate PR.

jbrockmendel:
Sounds good, thanks for explaining.  BTW the randomized tests are fixed by enforcing q < p.

statsmodels/regression/tests/test_lme.py
kshedden: These comments are the R code required to extract the relevant values for comparison.

kshedden:
Merging (the appveyor failure is due to a timeout on their end, and it returned green before the last commit, which only changed a comment string)

PR #4707, opened 2018-06-03
import instead of copy/paste

jbrockmendel:
very nearly all of sandbox.utils_old exists in copy/paste (or strictly improved) form elsewhere.  This just imports instead of re-implementing.

### DIFF
--- a/statsmodels/sandbox/utils_old.py
+++ b/statsmodels/sandbox/utils_old.py
@@ -1,49 +1,14 @@
 import numpy as np
-import numpy.linalg as L
-import scipy.interpolate
 import scipy.linalg
+
+from statsmodels.tools.tools import (  # noqa:F401
+    fullrank, recipr, recipr0, clean0)
+from statsmodels.distributions.empirical_distribution import (  # noqa:F401
+    StepFunction, monotone_fn_inverter, ECDF)
+from statsmodels.robust.scale import mad  # noqa:F401
 
 __docformat__ = 'restructuredtext'
 
-def recipr(X):
-    """
-    Return the reciprocal of an array, setting all entries less than or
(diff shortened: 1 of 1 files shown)


jbrockmendel:
@ChadFulton little help?

ChadFulton:
I guess I'm staying out of `sandbox` issues. I know that this isn't helpful, but I just don't understand this part of statsmodels and I'm afraid I'd do more harm than good.

For example, this PR is clearly a good idea in tested production code, but I don't know what the consequences are in `sandbox`.

My only contribution is that I still think it's a good idea to issue warnings at the top of the `sandbox` files (the ones that are not imported by Statsmodels main api anyway), so that we are clear that we don't have to support them anymore.

If more clarity were to arise about how we should move forward in a general way about `sandbox` stuff, I would be happy to be more helpful.

jbrockmendel:
> If more clarity were to arise about how we should move forward in a general way about sandbox stuff, I would be happy to be more helpful.

> the ones that are not imported by Statsmodels main api anyway

AFAICT there are a small number of sandbox modules that are imported into non-sandbox parts of the code.  I think those should be moved out of sandbox so there is a clear line between production code and nostalgia code.  Without that clear line, there cannot be a clear policy as to "what sandbox means".

If such a line existed, I could ignore problems in sandbox.  Without it, I have to worry about it silently "infecting" code I rely on, which I'm not OK with.  So I try to make (easily reviewed) baby steps towards bringing it up to code.

jbrockmendel:
@josef-pkt what say you give me and [Chad? Kevin? Bizzaro-Me?] the go-ahead to fix Sufficiently Obvious things like this.  It's a one-time cost that benefits all future readers and contributors.

PR #4743, opened 2018-06-15
BUG: state space: exact diffuse filtering in complex data type case

ChadFulton:
Bug causes the exact diffuse filter to never transition to the usual Kalman filter.


ChadFulton:
Obvious name typo with one-line fix, so I will merge this if it comes back green.

PR #4755, opened 2018-06-22
ENH: RecursiveLS: Add recursive t-test

ChadFulton:
Along with some improvements / fixes, adds the ability to compute a t-test for any subset of the data after estimating a model via `RecursiveLS`. For example:

```python
dta = sm.datasets.macrodata.load_pandas().data
dta.index = pd.PeriodIndex(start='1959Q1', end='2009Q3', freq='Q')
endog = dta.infl
exog = sm.add_constant(dta[['m1', 'unemp']])

# Run RecursiveLS over the entire model
mod = sm.RecursiveLS(endog, exog)
res = mod.fit()

# Compute the t-test using the first 100 observations
t = 99
restriction = 'm1 + unemp = 1, const = 0.5'
ttest_rls = res.t_test(restriction, loc=t)

# Now, run OLS over the restricted sample and compute the t-test
mod_ols = sm.OLS(endog[:t], exog[:t])
res_ols = mod_ols.fit()
ttest_ols = res_ols.t_test(restriction)

# Compare the results
print(ttest_rls)
print(ttest_ols)
```

yields

```
                             Test for Constraints                             
                 coef    std err          t      P>|t|      [0.025      0.975]
c0            -1.4091      0.227    -10.628      0.000      -1.859      -0.959
c1             4.2860      0.985      3.845      0.000       2.332       6.240
                             Test for Constraints                             
                 coef    std err          t      P>|t|      [0.025      0.975]
c0            -1.4091      0.227    -10.628      0.000      -1.859      -0.959
c1             4.2860      0.985      3.845      0.000       2.332       6.240
```

### DIFF
--- a/statsmodels/regression/recursive_ls.py
+++ b/statsmodels/regression/recursive_ls.py
@@ -266,6 +266,171 @@ class RecursiveLSResults(MLEResults):
             'k_exog': self.model.k_exog,
             'k_constraints': self.model.k_constraints})
 
+    def t_test(self, r_matrix, loc=None, cov_p=None, scale=None, use_t=None):
+        """
+        Compute a t-test for a each linear hypothesis of the form Rb = q
+
+        Parameters
+        ----------
+        r_matrix : array-like, str, tuple
+            - array : If an array is given, a p x k 2d array or length k 1d
+              array specifying the linear restrictions. It is assumed
+              that the linear combination is equal to zero.
+            - str : The full hypotheses to test can be given as a string.
+              See the examples.
+            - tuple : A tuple of arrays in the form (R, q). If q is given,
+              can be either a scalar or a length p row vector.
+        loc : int, str, or datetime, optional
+            Zero-indexed observation number at which to compute the t-test.
+            Can also be a date string to parse or a datetime type. Default is
+            to compute the t-test using the full-sample estimates. Cannot be
+            used in combination with `cov_p`.
+        cov_p : array-like
+            An alternative estimate for the parameter covariance matrix.
+            If None is given, self.normalized_cov_params is used.
+        scale : float, optional
+            An optional `scale` to use.  Default is the scale specified
+            by the model fit.
+        use_t : bool, optional
+            If use_t is None, then the default of the model is used.
+            If use_t is True, then the p-values are based on the t
+            distribution.
+            If use_t is False, then the p-values are based on the normal
+            distribution.
+
+        Returns
+        -------
+        res : ContrastResults instance
+            The results for the test are attributes of this results instance.
+            The available results have the same elements as the parameter table
+            in `summary()`.
+
+        Examples
+        --------
+        >>> import numpy as np
+        >>> import statsmodels.api as sm
+        >>> data = sm.datasets.longley.load()
+        >>> data.exog = sm.add_constant(data.exog)
+        >>> results = sm.OLS(data.endog, data.exog).fit()
+        >>> r = np.zeros_like(results.params)
+        >>> r[5:] = [1,-1]
+        >>> print(r)
+        [ 0.  0.  0.  0.  0.  1. -1.]
+
+        r tests that the coefficients on the 5th and 6th independent
+        variable are the same.
+
+        >>> T_test = results.t_test(r)
+        >>> print(T_test)
+                                     Test for Constraints
+        ==============================================================================
+                         coef    std err          t      P>|t|      [0.025      0.975]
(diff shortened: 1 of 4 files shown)


statsmodels/regression/recursive_ls.py
```diff
@@ -267,6 +266,176 @@ def __init__(self, model, params, filter_results, cov_type='opg',
...
+        -4.0167754636411717
+        >>> T_test.pvalue
+        0.0015163772380899498
+
+        Alternatively, you can specify the hypothesis tests using a string
+
+        >>> from statsmodels.formula.api import ols
+        >>> dta = sm.datasets.longley.load_pandas().data
+        >>> formula = 'TOTEMP ~ GNPDEFL + GNP + UNEMP + ARMED + POP + YEAR'
+        >>> results = ols(formula, dta).fit()
```
jbrockmendel: I'm guessing this docstring is copy/pasted from elsewhere.  Can the existing version just be pinned here?  i.e. so if it changes down the road this one doesn't become inaccurate?

statsmodels/regression/recursive_ls.py
```diff
@@ -267,6 +266,176 @@ def __init__(self, model, params, filter_results, cov_type='opg',
...
+                raise ValueError("r_matrix and q_matrix must have the same "
+                                 "number of rows")
+
+        if use_t is None:
+            # switch to use_t false if undefined
+            use_t = (hasattr(self, 'use_t') and self.use_t)
+
+        _t = _sd = None
+
+        # If we've been given a location
```
jbrockmendel: It looks like most of the code is copied from elswhere, with the next bits looking less familiar than the others.  Can any of it be shared?

statsmodels/regression/recursive_ls.py
```diff
@@ -267,6 +266,176 @@ def __init__(self, model, params, filter_results, cov_type='opg',
...
+            # Rescale the covariance parameters based on the scale estiamted
+            # as of observation t
+            d = max(self.loglikelihood_burn, self.nobs_diffuse)
+            nmissing = np.sum(self.filter_results.nmissing[d:t + 1])
+            n = (t + 1 - d) - nmissing
+            scale = (np.sum(self.filter_results.scale_obs[d:t + 1]) / n)
+            cov_p *= scale / self.scale
+        else:
+            d = max(self.loglikelihood_burn, self.nobs_diffuse)
+            t = self.model.nobs
```
jbrockmendel: is `t` used?

statsmodels/tsa/statespace/mlemodel.py
```diff
@@ -1126,7 +1126,7 @@ def score_obs(self, params, method='approx', transformed=True,
 
         if method == 'harvey':
             score = self._score_obs_harvey(
-                params, transformed=transformed,
+                params,
```
jbrockmendel: Is this related to the rest of the PR?

ChadFulton: Nope!

statsmodels/regression/recursive_ls.py
ChadFulton: Down below in the `nobs_effective` computation.

statsmodels/regression/recursive_ls.py
ChadFulton: Yeah, this is basically copied from `model.py::LikelihoodModelResults::t_test`. But the `loc` parameter is added, and the `df_resid` computation below is different. Maybe there's a better way to handle it. Two options:

- Refactor the base `t_test` to allow passing more things in as parameters. The problem is that I can't see many other use cases that would require that.
- In this `t_test`, temporarily override `self.params` and `self.df_resid`, call the parent method, and then set them back to were they were. This seems like a bummer to me since it's not very "clean", but I suppose it might be better?

Other options?

statsmodels/regression/recursive_ls.py
ChadFulton: Not directly since we added a parameter. There might be something to do I guess.

statsmodels/regression/recursive_ls.py
jbrockmendel: My suggestion would be to make repeated code into helper method/functions.  This isn't a hill I'd make a stand on, but copy/pasted code is big concern so there might be some upside to encouraging/forming good habits/norms.

ChadFulton:
Harvey's book suggests Harvey and Collier (1977) as a reference for recursive t-tests

PR #4768, opened 2018-06-29
Running flake8 tests show 101 undefined names

cclauss:
In Python 3 there will be even more than Python 2.

Each __undefined name__ has the possibility of raising _NameError_ at runtime.

PR #4668, opened 2018-05-21
BLD: Switch to versioneer

bashtage:
Simply release by using versioneer to handle versioning
Add a v0.10.0dev tag to generate correct build numbers from master

PR #4773, opened 2018-07-03
implement setup.cfg, flake8 fixups in __init__ and info.py

jbrockmendel:
Implements setup.cfg and tells flake8 to ignore _almost_ everything.  The idea is that we can clean things up file-by-file, making it stricter as we go.

Related:
#4768 which applies high-priority checks to _all_ files.  I think this can be run in combination with that by running a plain `flake8` after the more specific command currently in that PR.  (@cclauss is that correct?
#4668 also implements setup.cfg (and should be merged)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,38 @@
+[flake8]
+max-line-length = 80
+max-complexity = 10
+
+exclude =
+    .tox,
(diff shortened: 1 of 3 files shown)


jbrockmendel:
Superseded, closing.

PR #4778, opened 2018-07-05
MAINT: specify which build

thequackdaddy:
An attempt to fix the python 2.7/numpy=1.11 issue. I am locking the build which should do the trick, I think.


thequackdaddy:
Looks like this worked. I think you might want to wait for conda to get fixed, or just merge this temporarily and revert later.

thequackdaddy:
Sounds good. Closing.

PR #4815, opened 2018-07-27
ENH/BUG: Improve SARIMAX for time series close to non-stationary

ChadFulton:
This fixes some issues that can arise for close to non-stationary time series. Previously these models weren't even considered, but with e.g. #4739 these models may crop up more often. This:

- Fixes a problem with the `self.polynomial_*` variables got filled with exactly zeros (for example when non-stationary starting parameters were replaced with zeros)
- Bounds the initial variance estimate away from zero (again, can happen when non-stationary starting parameters were replaced with zeros)


ChadFulton:
Green, merging.

Issue #4796, opened 2018-07-13
BUG: ExponentialSmoothing fails with ZeroDivisionError


abhishekkaranth:
For the model parameters
`ExponentialSmoothing(arr, trend='add', damped=False, seasonal='add', seasonal_periods=5)`
where` len(arr) = 12`, the fit of model fails with ZeroDivisionError.

The issue is caused while calculating aicc of the model since it is calculated as:
```
k = m * seasoning + 2 * trending + 2 + 1 * damped
aicc = aic + (2 * (k + 2) * (k + 3)) / (self.nobs - k - 3)
```
In this case:
```
self.nobs = 12
k = 9
```
This can happen with other dataset as well.
Should the model fail in such cases? Even if yes, we can throw better errors I feel.

ChadFulton:
Thanks for the report! Yes, I think that in this case the model should run, and we can set the aicc to `np.inf`.

PR #4825, opened 2018-08-02
BUG: ExponentialSmoothing AICC Divide By Zero Error

andrewcking:
ExponentialSmoothing can fail in some circumstances due to AICC having zero in the denominator when calculating. This simply catches a ZeroDivisionError when calculating AICC and sets AICC to np.inf in those cases. Custom build appears to be working (building and running yes but also giving good results) where it would simply fail before.

See #4796

### DIFF
--- a/statsmodels/tsa/holtwinters.py
+++ b/statsmodels/tsa/holtwinters.py
@@ -732,7 +732,10 @@ class ExponentialSmoothing(TimeSeriesModel):
         # (s0 + gamma) + (b0 + beta) + (l0 + alpha) + phi
         k = m * seasoning + 2 * trending + 2 + 1 * damped
         aic = self.nobs * np.log(sse / self.nobs) + (k) * 2
-        aicc = aic + (2 * (k + 2) * (k + 3)) / (self.nobs - k - 3)
+        try: 
+            aicc = aic + (2 * (k + 2) * (k + 3)) / (self.nobs - k - 3)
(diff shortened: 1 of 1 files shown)


statsmodels/tsa/holtwinters.py
```diff
@@ -732,7 +732,10 @@ def _predict(self, h=None, smoothing_level=None, smoothing_slope=None,
         # (s0 + gamma) + (b0 + beta) + (l0 + alpha) + phi
         k = m * seasoning + 2 * trending + 2 + 1 * damped
         aic = self.nobs * np.log(sse / self.nobs) + (k) * 2
-        aicc = aic + (2 * (k + 2) * (k + 3)) / (self.nobs - k - 3)
+        try: 
+            aicc = aic + (2 * (k + 2) * (k + 3)) / (self.nobs - k - 3)
+        except ZeroDivisionError:
```
bashtage: Better to check that self.nobs - k - 3 > 0 since will get wrong result if self.nobs - k - 3 is negative.

lingyaw: Hi,
I tried to edit holtwinter.py. Unfortunately, I can't change it... do you know how to fix it ?

bashtage:
This is fixed in master.

lingyaw:
Thanks. I used sudo pip3 to install it directly. I will download from
github.

PR #4827, opened 2018-08-03
TST: Add OpenBlas testing for Windows  

bashtage:
Add pip runs using pip wheels so that Windows tests also use OpenBlas
Failing since the HoltWinters tests are not valid since the optimizer does not converge.


bashtage:
Not needed

PR #4834, opened 2018-08-04
Implement full flake8 checking for a subset of files in good condition

jbrockmendel:


ChadFulton:
I think this sounds like a good idea, and we can hopefully progressively increase the list.

jbrockmendel:
Yep, and #4835 does that “progressive” thing from the opposite direction.

jbrockmendel:
@ChadFulton once the button is pushed here we can make a push to get all of statespace over the fully-passing line.

ChadFulton:
Will merge this tomorrow, unless there are concerns.

ChadFulton:
Thanks @jbrockmendel

Issue #4832, opened 2018-08-04
qqplot_2samples cannot handle samples of differing length


dsaxton:
In most applications where one is interested in comparing two distributions the distributions will not have the same size, however `qqplot_2samples` seems to require that this be the case.  I think this function should handle the general case where `x` and `y` have arbitrary lengths.

bashtage:
Fixed in #5673

PR #4840, opened 2018-08-07
Allow qqplot_2samples to accept samples of unequal size

dsaxton:
This tries to address issue #4832 where `qqplot_2samples` requires that both samples have the same size.  Here if the function encounters samples of differing length it calculates the quantile function for the larger sample and then takes the estimated quantiles corresponding to the smaller sample using linear interpolation.  These two samples (which now are of equal length) are then plotted in the usual way.


bashtage:
Need a minimial test that shows that the code executes when samples have different sizes.

statsmodels/graphics/gofplots.py
```diff
@@ -308,10 +309,37 @@ def qqplot(self, xlabel=None, ylabel=None, line=None, other=None,
             if not check_other:
                 other = ProbPlot(other)
 
-            fig, ax = _do_plot(other.sample_quantiles,
-                               self.sample_quantiles,
-                               self.dist, ax=ax, line=line,
-                               **plotkwargs)
+            if self.nobs < other.nobs:
+                finv = interp1d([(i+1) / other.nobs for i in range(other.nobs)], other.sorted_data)
```
bashtage: This problem is completely symmetric.  I should be possible to always fix the long series and then just call a single version of _do_plot

statsmodels/graphics/gofplots.py
```diff
@@ -308,10 +309,37 @@ def qqplot(self, xlabel=None, ylabel=None, line=None, other=None,
...
+                                  distargs=self.distargs, a=self.a, 
+                                  loc=self.loc, scale=self.scale)
+
+                fig, ax = _do_plot(other.sample_percentiles,
+                                   p_plot.sample_percentiles,
+                                   p_plot.dist, ax=ax, line=line,
+                                   **plotkwargs)
+
+            else:
+                fig, ax = _do_plot(other.sample_percentiles,
```
bashtage: Just use a single call here fo the forms

```
fig, ax = _do_plot(other_percentiles,
                                  self_percentiles,
                                   dist, ax=ax, line=line,
                                   **plotkwargs)
```

and then define `other_percentiles`, `self_percentiles` and `dist` correctly depending on the sample lengths.

dsaxton: Yes, agreed.  Will work on making that change.

Since within the `if` and `elif` statements we've already computed the corresponding quantiles for the larger sample in sorted order, would you recommend passing these directly to `_do_plot` instead of initializing a new `ProbPlot` and recomputing the percentiles?

Regarding the `dist` parameter, my understanding from looking over the code is that this is a theoretical distribution which wouldn't change depending on sample size and so we could leave this alone, but please let me know if I'm mistaken on this.

I'll also put together a minimal test for the proposed change.

dsaxton:
It seems my tests are failing and I haven't been able to identify the issue.  Any ideas as to what the cause might be?

bashtage:
Here is the fit.

1. Add `from __future__ import division` as the first line in the file.  This is just good practice.
2.  You need to use 1.0 instead on 1.  Using 1.0 will promote everything to a float rather than doin integer math, which was producing lots of 0s.  This only affects Python 2.7, not 3, and so you only failed there

```python
            if self.nobs < other.nobs:
                finv = interp1d([(i+ 1.0) / other.nobs for i in range(other.nobs)],
                                other.sorted_data)
                q = finv([(i+1.0) / self.nobs for i in range(self.nobs)])
                self_percentiles = self.sample_percentiles
                other_percentiles = q

            elif other.nobs < self.nobs:
                finv = interp1d([(i+1.0) / self.nobs for i in range(self.nobs)],
                                self.sorted_data)
                q = finv([(i+1.0) / other.nobs for i in range(other.nobs)])
                self_percentiles = q
                other_percentiles = other.sample_percentiles

```

dsaxton:
Great, thanks so much. `from __future__ import division` always struck me as a strange incantation.  Quick question, but should I be concerned that the LGTM analysis always fails?

bashtage:
LGTM is broken and should be ignored.  Only appveyor and travis really matter.

dsaxton:
@bashtage I noticed that there seem to be merge conflicts for this pull request, is this something that I should resolve (I am still very new to making pull requests, this was actually my first ever)?

dsaxton:
@bashtage It seems that the merge conflict stems from a couple aesthetic / readability changes I made to `test_gofplots.py` related to alignment in spacing.  In the future I won't make these modifications as part of an unrelated set of changes and will instead offer them as separate pull requests.

bashtage:
You need to rebase your commit ontop of master.

It can be a bit tricky.  You can do 

git add remote upstream https://github.com/statsmodels/statsmodels
git fetch --all
git rebase upstream/master
<Here you fix the merge conflicts so that test_gofplots looks right>
git add statsmodels/graphics/tests/test_gofplots.py
git rebase --continue
git push -f

If something goes wrong you can do git rebase --abort instead of --continue

bashtage:
You you think you need help I or @josef-pkt could do the rebase.

dsaxton:
@bashtage Ok, I will try this.  Should it be `git remote add upstream...` rather than `git add remote upstream...` by chance?  I get an error with the latter command.

bashtage:
Yes, you are right.

dsaxton:
Thanks, hopefully I didn't botch that.

dsaxton:
@bashtage It looks like the checks failed because of a test `test_transform_dataframe `.  I'm assuming this is unrelated to this particular PR?

bashtage:
These are unrelated to your code changes.

@josef-pkt I think this is ready.

dsaxton:
@josef-pkt Are there any additional changes I should make to this pull request?

bashtage:
@dsaxton Did you mean to close this?  I think it is ready and will get merged.  Merging can be very slow around here, but it doesn't mean that it isn't welcome.

josef-pkt:
I would like to get this or something like this into statsmodels in the next week.
Probability plots are on my current schedule, when I'm not getting distracted.

dsaxton:
It was an accident.  I reset my local master branch to resemble upstream master when creating a new branch for another pull request, but forgot that it would also reset these changes.

PR #4842, opened 2018-08-07
CLN: Flake8 fixups and linting for statespace files (but not tests)

jbrockmendel:
@ChadFulton The statespace files themselves are in pretty good shape, there are are a couple of points I'll comment on inline where your input will be needed in a follow-up.

After this `statespace/tests/results/*.py` will be an easy PR, then the test files will be split into a few steps because they're in less-good condition than the rest.

### DIFF
--- a/statsmodels/tsa/statespace/initialization.py
+++ b/statsmodels/tsa/statespace/initialization.py
@@ -602,11 +602,12 @@ class Initialization(object):
         # the blocks and return those
         else:
             for block_index, init in self.blocks.items():
-                out = init(
-                    index=tuple(np.array(index)[block_index, ]),
-                    model=model, initial_state_mean=initial_state_mean,
-                    initial_diffuse_state_cov=initial_diffuse_state_cov,
-                    initial_stationary_state_cov=initial_stationary_state_cov)
+                # TODO: Should something be done with the object returned
+                # by init?
+                init(index=tuple(np.array(index)[block_index, ]),
+                     model=model, initial_state_mean=initial_state_mean,
+                     initial_diffuse_state_cov=initial_diffuse_state_cov,
+                     initial_stationary_state_cov=initial_stationary_state_cov)
 
         return (initial_state_mean, initial_diffuse_state_cov,
                 initial_stationary_state_cov)
--- a/statsmodels/tsa/statespace/initialization.py
+++ b/statsmodels/tsa/statespace/initialization.py
@@ -598,12 +598,10 @@ class Initialization(object):
                     tools.solve_discrete_lyapunov(transition,
                                                   selected_state_cov,
                                                   complex_step=complex_step))
-        # Otherwise, if using blocks, combine the elements returned by each of
-        # the blocks and return those
         else:
+            # Otherwise, if using blocks, recursively initialize
+            # them (values will be set in-place)
             for block_index, init in self.blocks.items():
-                # TODO: Should something be done with the object returned
-                # by init?
                 init(index=tuple(np.array(index)[block_index, ]),
                      model=model, initial_state_mean=initial_state_mean,
                      initial_diffuse_state_cov=initial_diffuse_state_cov,
--- a/statsmodels/tsa/statespace/mlemodel.py
+++ b/statsmodels/tsa/statespace/mlemodel.py
@@ -6,29 +6,35 @@ Author: Chad Fulton
 License: Simplified-BSD
 """
 from __future__ import division, absolute_import, print_function
-from statsmodels.compat.python import long
-
-try: unicode
-except NameError: unicode = str
+import warnings
 
 import numpy as np
 import pandas as pd
 from scipy.stats import norm
 
-from .simulation_smoother import SimulationSmoother
-from .kalman_smoother import SmootherResults
-from .kalman_filter import (INVERT_UNIVARIATE, SOLVE_LU)
-import statsmodels.tsa.base.tsa_model as tsbase
-import statsmodels.base.wrapper as wrap
+from statsmodels.compat.python import long
+
+from statsmodels.tools.tools import pinv_extended, Bunch
+from statsmodels.tools.sm_exceptions import PrecisionWarning
 from statsmodels.tools.numdiff import (_get_epsilon, approx_hess_cs,
                                        approx_fprime_cs, approx_fprime)
 from statsmodels.tools.decorators import cache_readonly, resettable_cache
 from statsmodels.tools.eval_measures import aic, bic, hqic
-from statsmodels.tools.tools import pinv_extended, Bunch
-from statsmodels.tools.sm_exceptions import PrecisionWarning
+
+import statsmodels.base.wrapper as wrap
+
 import statsmodels.genmod._prediction as pred
 from statsmodels.genmod.families.links import identity
-import warnings
+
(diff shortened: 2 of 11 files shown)


statsmodels/tsa/statespace/varmax.py
```diff
@@ -860,7 +862,7 @@ def get_prediction(self, start=None, end=None, dynamic=False, index=None,
 
         # If we had exog, then the last predicted_state has been set to NaN
         # since we didn't have the appropriate exog to create it. Then, if
-        # we are purely forecasting, 
+        # we are purely forecasting, TODO: Finish this sentence?
```
jbrockmendel: @Chad this probably needs your input.

statsmodels/tsa/statespace/tools.py
```diff
@@ -1617,7 +1619,7 @@ def copy_missing_matrix(A, B, missing, missing_rows=False, missing_cols=False,
     try:
         if not A.is_f_contig():
             raise ValueError()
-    except:
+    except (AttributeError, ValueError):
```
jbrockmendel: @ChadFulton I'm fairly confident this is what should be caught, but this merits another pair of eyes.

statsmodels/tsa/statespace/mlemodel.py
```diff
@@ -2711,15 +2719,15 @@ def summary(self, alpha=.05, start=None, title=None, model_name=None,
...
             het = np.array([[np.nan]*2])
         try:
             lb = self.test_serial_correlation(method='ljungbox')
-        except:
+        except Exception:  # FIXME: catch something specific
             lb = np.array([[np.nan]*2]).reshape(1, 2, 1)
         try:
             jb = self.test_normality(method='jarquebera')
-        except:
+        except Exception:  # FIXME: catch something specific
```
jbrockmendel: @ChadFulton This is the next-to-last thing that needs your input.

statsmodels/tsa/statespace/initialization.py
```diff
@@ -602,11 +602,12 @@ def __call__(self, index=None, model=None, initial_state_mean=None,
         # the blocks and return those
         else:
             for block_index, init in self.blocks.items():
-                out = init(
-                    index=tuple(np.array(index)[block_index, ]),
-                    model=model, initial_state_mean=initial_state_mean,
-                    initial_diffuse_state_cov=initial_diffuse_state_cov,
-                    initial_stationary_state_cov=initial_stationary_state_cov)
+                # TODO: Should something be done with the object returned
+                # by init?
```
jbrockmendel: @ChadFulton last thing.

statsmodels/tsa/statespace/mlemodel.py
ChadFulton: What do you think about just `# noqa` on these? Two reasons:

- These are in the summary, and I think that we want to show the summary regardless of any errors here.
- There are a whole host of exceptions that could arise from the class methods or from the functions that they call, so it might be tough to narrow down all of the possible exceptions.

statsmodels/tsa/statespace/initialization.py
ChadFulton: Ah, no, I think they're set in place at each recursion. I guess maybe I was originally planning something different. If you don't mind, can you replace:

```python
# Otherwise, if using blocks, combine the elements returned by each of
# the blocks and return those
```

with:

```python
# Otherwise, if using blocks, recursively initialize them (values will be set in-place)
```

statsmodels/tsa/statespace/representation.py
```diff
@@ -237,19 +237,22 @@ class Representation(object):
     """
     transition = MatrixWrapper('transition', 'transition')
     r"""
-    (array) Transition matrix: :math:`T~(k\_states \times k\_states \times nobs)`
+    (array) Transition matrix:
+    :math:`T~(k\_states \times k\_states \times nobs)`
```
ChadFulton: I think this has to be on one line for Sphinx, not sure if they fixed it though.

statsmodels/tsa/statespace/representation.py
```diff
@@ -237,19 +237,22 @@ class Representation(object):
...
     """
     state_intercept = MatrixWrapper('state intercept', 'state_intercept')
     r"""
     (array) State intercept: :math:`c~(k\_states \times nobs)`
     """
     selection = MatrixWrapper('selection', 'selection')
     r"""
-    (array) Selection matrix: :math:`R~(k\_states \times k\_posdef \times nobs)`
+    (array) Selection matrix:
+    :math:`R~(k\_states \times k\_posdef \times nobs)`
```
ChadFulton: Same here, one line I think.

statsmodels/tsa/statespace/representation.py
```diff
@@ -237,19 +237,22 @@ class Representation(object):
...
     """
     state_cov = MatrixWrapper('state covariance matrix', 'state_cov')
     r"""
-    (array) State covariance matrix: :math:`Q~(k\_posdef \times k\_posdef \times nobs)`
+    (array) State covariance matrix:
+    :math:`Q~(k\_posdef \times k\_posdef \times nobs)`
```
ChadFulton: And same line here too.

ChadFulton:
Other than those comments, LGTM. Thanks!

statsmodels/tsa/statespace/representation.py
jbrockmendel: Other entries are on two lines, e.g. line 234-237 above:

```
     r"""
     (array) Observation covariance matrix:
     :math:`H~(k\_endog \times k\_endog \times nobs)`
     """
```

statsmodels/tsa/statespace/mlemodel.py
jbrockmendel: At the very least it should be `exception Exception:`.  With just `except:` we will e.g. catch `KeyboardInterrupt` which is definitely undesired.

I'd be OK with leaving the "FIXME" here for the time being.  It's a low enough priority that we don't need to fix it immediately, but high enough that it should be hard to ignore on the next pass.

statsmodels/tsa/statespace/representation.py
jbrockmendel: Is there a conclusion on what the right thing to do is here?

statsmodels/tsa/statespace/mlemodel.py
ChadFulton: > At the very least it should be exception Exception:. With just except: we will e.g. catch KeyboardInterrupt which is definitely undesired.

Good point.

> I'd be OK with leaving the "FIXME" here for the time being. It's a low enough priority that we don't need to fix it immediately, but high enough that it should be hard to ignore on the next pass.

Okay, we can discuss later.

ChadFulton:
@jbrockmendel happy to merge this, but I think there are still a few outstanding changes required.

statsmodels/tsa/statespace/representation.py
ChadFulton: (my comment here can be ignored)

jbrockmendel:
> but I think there are still a few outstanding changes required.

Just implemented what I think are the remaining ones.  Let me know if I missed anything.

ChadFulton:
Thanks @jbrockmendel !

PR #4844, opened 2018-08-07
CLN: Fully lint regime_switching

jbrockmendel:
@ChadFulton this one was a breeze.  Only thing that could use your attention in a follow-up is an `except Exception:` that should probably be more specific.


ChadFulton:
LGTM, thanks! (I'm going to leave the FIXME comment in for now, since I'm not sure at the moment what the range of possible exceptions are).

ChadFulton:
Thanks @jbrockmendel

PR #4845, opened 2018-08-07
ENH: Improve speed in Exponential Smoothing

bashtage:
Add Cython version of all smoothers, providing gains of ~ 50x for relevant models sizes
Fixed a bug that affected multiplicative trend initial values
Add ability to set starting values
Add ability to fix all model parameters or any subset
Improve model initialization options to speed up fitting if desired
Improve tests
Added summary
Added `params_formatted` which returns a DataFrame.  I did not want deprecation headaches so left `params` as it is -- a dictionary.

xref #4842


bashtage:
@tvanzyl If you are still around it would be helpful if you could look at this. In particular, the change to the initialization of the trend when it is multiplicative.

tvanzyl:
I will take a look.

statsmodels/tsa/holtwinters.py
```diff
@@ -221,13 +217,34 @@ def _holt_win_mul_add_dam(x, xi, p, y, l, b, s, m, n, max_seen):
         return max_seen
     for i in range(1, n):
         l[i] = (y_alpha[i - 1]) - (alpha * s[i - 1]) + \
-            (alphac * (l[i - 1] * b[i - 1]**phi))
-        b[i] = (beta * (l[i] / l[i - 1])) + (betac * b[i - 1]**phi)
+               (alphac * (l[i - 1] * b[i - 1] ** phi))
+        b[i] = (beta * (l[i] / l[i - 1])) + (betac * b[i - 1] ** phi)
         s[i + m - 1] = y_gamma[i - 1] - \
-            (gamma * (l[i - 1] * b[i - 1]**phi)) + (gammac * s[i - 1])
+            (gamma * (l[i - 1] * b[i - 1] ** phi)) + (gammac * s[i - 1])
```
josef-pkt: In general we didn't use spaces around power `**`

josef-pkt:
I only did a partial skimming of the changes and didn't see any problems.

ChadFulton:
Great addition to the exponential smoothing models!

josef-pkt:
I only did a very coarse skimming.
I'm merging this given there are no problems pointed out with it.
(I didn't check for backwards incompatible changes given that this was only recently added to statsmodels.)

@bashtage Thank you

bashtage:
It is 100% compatible with the 0.9 release. The keyword arguments really
need rearrangement but this can wait until 3.5 is the minimum so it can
become keyword argument only.

PR #4846, opened 2018-08-07
CLN: Fully lint compat

jbrockmendel:


statsmodels/compat/__init__.py
```diff
@@ -1 +1,23 @@
-from .python import *
```
bashtage: Can't you just leave this in and noqa it?  Or do you get other flake errors?

jbrockmendel: That would be allowed.  In this case changing it surfaced the fact that `urlretrieve` was defined in the py3 section but not the py2 section.  Not that we actually need `urlretrieve`, but the explicitness can be useful.

jbrockmendel:
@josef-pkt ping.  You can do this; I believe in you.

jbrockmendel:
Looks like some unwanted stuff got in here in a rebase.  Since #5000 will make much of this unnecessary, closing.  Will re-do after #5000.

PR #4849, opened 2018-08-08
Fully lint statsmodels/formula

jbrockmendel:
One by one...

### DIFF
--- a/statsmodels/formula/tests/test_formula.py
+++ b/statsmodels/formula/tests/test_formula.py
@@ -1,18 +1,21 @@
-from statsmodels.compat.python import iteritems, StringIO
 import warnings
 
+import numpy.testing as npt
+import numpy as np
+import pandas as pd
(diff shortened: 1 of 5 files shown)


statsmodels/formula/tests/test_formula.py
```diff
@@ -74,25 +77,68 @@ def test_tests():
...
-    from pandas import read_table
-    dta = read_table(dta, sep=" ")
+    dta = StringIO('type" "income" "education" "prestige"\n'
```
bashtage: You lost a quote here. SHould be '"type"

jbrockmendel: Thanks.

jbrockmendel:
rebased

jbrockmendel:
@josef-pkt one push of the button away from never being bothered about this directory again.  (And notice I even rebased the way you prefer)

PR #4857, opened 2018-08-08
Fixed some typos and broken maths in `mixed_linear.rst`

amirmasoudabdol:
I've unified the use of `j`th in the documents by replacing them ``:math:`j`th`` with `math:`j^\rm{th}`` as this is the preferred format in `numpy` documentation as well. Also, Sphinx cannot process the current format really well, check the live documentation.


amirmasoudabdol:
I'm not sure what's the problem here, if someone guides me I can maybe fix it. I've only changed some docs, and I'm confused about the build error.

amirmasoudabdol:
I've run the `make html` and tested them. Here is the screenshot of one of them. Notice the `k`th.

PR #4865, opened 2018-08-09
Update contingency_tables.rst

amirmasoudabdol:
Fix broken `.. math:`s by adding the missing tab before each equation.

### DIFF
--- a/docs/source/contingency_tables.rst
+++ b/docs/source/contingency_tables.rst
@@ -73,7 +73,7 @@ product of the row and column marginal distributions:
 
 .. math::
 
-    P_{ij} = \sum_k P_{ij} \cdot \sum_k P_{kj} \forall i, j
+    P_{ij} = \sum_k P_{ij} \cdot \sum_k P_{kj} \quad \text{for all} \quad  i, j
 
 We can obtain the best-fitting independent distribution for our
 observed data, and then view residuals which identify particular cells
--- a/docs/source/contingency_tables.rst
+++ b/docs/source/contingency_tables.rst
@@ -73,7 +73,7 @@ product of the row and column marginal distributions:
 
 .. math::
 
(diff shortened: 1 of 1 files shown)


amirmasoudabdol:
In this case, the formula was completely broken, now it's been shown correctly.

josef-pkt:
a horizontal space in before `forall` would be useful
(in general I prefer to spell `for all` out in text except in math journals.)

amirmasoudabdol:
I agree. The spacing is rather off. We can do with these two, which one can we agree on?

screenshot 2018-08-10 at 00 53 32

or 

screenshot 2018-08-10 at 00 52 22

josef-pkt:
horizontal space is fine, but the first version is a hybrid
I meant, dropping the forall symbol to have something like
`math  ... text{for all}  i, j`

amirmasoudabdol:
I see. Except for the hybrid one, all is fine I guess. But I've tried to find a documentation on `numpy` or `scipy` to see what's their preferred method, but I couldn't recall anything with a similar formula. Do you have an example?

josef-pkt:
I don't find an example browsing around in docs and notebooks.
I don't remember where I might have used this in documentation or whether at all. For simplicity I also might drop the `for all i, j` completely or put it in the surrounding text.

I think using text `for all` is good a general rule for our documentation, when we actually want to add it to be "cleaner".

amirmasoudabdol:
I've honestly confused now what'd be the final version. This:

```
.. math::

    P_{ij} = \sum_k P_{ij} \cdot \sum_k P_{kj} \quad \text{for all  i, j}
```

amirmasoudabdol:
This should be good then 👍 

screenshot 2018-08-12 at 04 47 10

josef-pkt:
Sorry, this got lost.

Can you rebase? then I will merge. 
(If needed, I can do the rebase)

PR #4871, opened 2018-08-10
BLD: Refactor Setup

bashtage:
This version builds on #3195 and appears to work as required:

- [x] correct sdist
- [x] build with Cython present
- [x] build without Cython and with C sources
- [x] error if no Cython and no C
- [x] work in an empty environment from sdist or wheel

It tries to simplify as much as possible and to let other packages handle tasks they already handle, like letting Cython determine whether to recompile a pyx file. 

It adds versioneer since a lot of setup is a bespoke version of what versioneer already does.


setup.cfg
```diff
@@ -13,3 +13,11 @@ select=
     # E111: Indentation is not a multiple of four
     E114
     # E114: Indentation is not a multiple of four (comment)
+
+[versioneer]
```
jbrockmendel: Can we adopt a convention of putting two blank lines between blocks?  In particular I'm thinking `[coverage:run]` and `[coverage:report]` should have 1 blank line, so more-distinct blocks can have one more line.

Obviously, not a big deal.

setup.py
```diff
@@ -1,187 +1,71 @@
...
+    from Cython import Tempita
+    from Cython.Build import cythonize
+    from Cython.Distutils import build_ext
+
+    HAS_CYTHON = True
 except ImportError:
-    # no setuptools installed
-    from distutils.core import setup, Command
-    _have_setuptools = False
+    from setuptools.command.build_ext import build_ext
```
jbrockmendel: Is this still for the "no setuptools installed" case?  Is that still possible in this day and age?

setup.py
```diff
@@ -1,187 +1,71 @@
...
-    setuptools_kwargs = {}
-    if sys.version_info[0] >= 3:
-        sys.exit("Need setuptools to install statsmodels for Python 3.x")
+import versioneer
 
+###############################################################################
+# Key Values that Change Each Release
+###############################################################################
+SETUP_REQUIREMENTS = {'numpy': '1.11',  # released March 2016
+                      'scipy': '0.18',  # released July 2016
```
jbrockmendel: +1 for noting the release dates

jbrockmendel:
With a few comments, generally +1.

Does the new setup.py pass strict linting?  If so, please add it to the lint.sh file.

jbrockmendel:
Does this mean tools/cythonize.py can go?  (and possibly cythonize.dat, cythonize_exclusions.dat)?

setup.py
bashtage: There are only 2 cases, Cython installed or setuptools installed.

setup.py
```diff
@@ -191,387 +75,197 @@ def check_dependency_versions(min_versions):
...
+        with open(source_name, 'r') as templated:
+            pyx_template = templated.read()
+        pyx = Tempita.sub(pyx_template)
+        pyx_filename = source_name[:-3]
+        with open(pyx_filename, 'w') as pyx_file:
+            pyx_file.write(pyx)
+        file_stats = os.stat(source_name)
+        try:
+            os.utime(pyx_filename, ns=(file_stats.st_atime_ns, file_stats.st_mtime_ns))
+        except AttributeError:
```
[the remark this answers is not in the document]
bashtage: Yes

bashtage:
> Does the new setup.py pass strict linting? If so, please add it to the lint.sh file.

Yes

> Does this mean tools/cythonize.py can go? (and possibly cythonize.dat, cythonize_exclusions.dat)?

Yes, these are no longer required.  It is more-or-less as vanilla as it can get while handling the two important cases -- cython (or not) and deferred numpy.

setup.py
```diff
@@ -1,187 +1,71 @@
...
+                  'docs': ['sphinx', 'nbconvert', 'jupyter_client', 'ipykernel',
+                           'matplotlib', 'nbformat', 'numpydoc', 'pandas-datareader']}
+
+###############################################################################
+# Values that rarely change
+###############################################################################
 
 DISTNAME = 'statsmodels'
 DESCRIPTION = 'Statistical computations and models for Python'
+README = open(pjoin(os.getcwd(), 'README.rst')).read()
```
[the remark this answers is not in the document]
bashtage: Yes

setup.py
jbrockmendel: Yep, reading comprehension fail on my part, thanks.

josef-pkt:
Can this be made to stop the installation with an error message if numpy and scipy are older than the setup requirements?
That is, I still like to avoid that pip updates numpy and scipy if they are too old, which is the pip default now, AFAIR/AFAIU.

SETUP_REQUIREMENTS = {'numpy': '1.11',  # released March 2016
                      'scipy': '0.18',  # released July 2016

bashtage:
While I think it is simpler to leave this up to pip/end users, I have added checks for the SETUP_REQUIREMENTS as minimum versions.

josef-pkt:
last change looks good to me
I prefer to leave it to the user, but not to `pip`.
(Personally, I still have python versions on my computer that don't satisfy the minimum scipy requirement, and I don't want pip to touch my scipy if I forget about adding the no-dep)

I'd like to try this locally, and then merge when green, within a few days. That leaves us still some time before the next release to see if there are any problems.
This also fixes the problem that we had in 0.9 that scipy wasn't specified as build dependency, AFAIU.

jbrockmendel:
@bashtage needs rebase
@josef-pkt a lot of this is sorely-needed.  Would it be easier if done in smaller pieces?

josef-pkt:
PR is fine, I just need to try it out before merging to see whether it works for me.

bashtage:
@josef-pkt Have you had a chance to use this locally?  I think it is ready and there are no further changes expected.

josef-pkt:
first try py3.4 virtualenv compiler mingw/gcc

```
(try_py34) M:\josef_new\venvs\try_py34\Scripts>pip install -U --no-deps M:\josef
_new\eclipse_ws\statsmodels\statsmodels_py34_pr
Processing m:\josef_new\eclipse_ws\statsmodels\statsmodels_py34_pr
  Installing build dependencies ... done
Building wheels for collected packages: statsmodels
  Running setup.py bdist_wheel for statsmodels ... done
  Stored in directory: C:\Users\josef\AppData\Local\Temp\pip-ephem-wheel-cache-7
e1n1wu1\wheels\6d\53\6e\60ef11dfb6338dd1f6338b55574a345031ab9697e61a445a27
Successfully built statsmodels
Installing collected packages: statsmodels
  Found existing installation: statsmodels 0.9.0rc1
    Uninstalling statsmodels-0.9.0rc1:
      Successfully uninstalled statsmodels-0.9.0rc1
Successfully installed statsmodels-0.6.0rc1+3719.gf1ccdbd
You are using pip version 10.0.1, however version 18.0 is available.
You should consider upgrading via the 'python -m pip install --upgrade pip' comm
and.
```

segfault when running pytest statsmodels\regression

running just one module I get segfault in 
(try_py34) M:\josef_new\venvs\try_py34\Scripts>pytest -v "M:\josef_new\venvs\try
_py34\Lib\site-packages\statsmodels\regression\tests\test_regression.py"

`..\Lib\site-packages\statsmodels\regression\tests\test_regression.py::test_formula_missing_cat`


possibly user (my) errors and/or fragility of setup.py/pip
But AFAIR, I had used this venv for 0.9rc release testing

(got distracted with cleaning PRs
TBC)

jbrockmendel:
> first try py3.4 virtualenv compiler mingw/gcc

Is this tested in the CI?

bashtage:
If you manually install pip wheels in the same venv and then install
statsmodels using setup in master do you get a different result? A segfault
in a regression test indicates an issue in the numpy or scipy install,
rather than with statsmodels setup.

bashtage:
Also, did you manually install pytest in the venv? If not you might have
been running with a different python.

bashtage:
Could yo uprovide details on recreating your setup?  I haven't been able to get pandas correctly installed using Python 3.4 and any mingw compiler.

bashtage:
I finally got sm build with mingw.  It segfaults immediately.  I suspect that mingw isn't compatible with the Visual Studio 2010 built wheels on PyPI.  

This is likely unfixable, so shouldn't be a blocker.  IM) the relevant benchmark is that a vanilla installation works using the same compilers that were used to build the wheels for NumPy, Scipy and Pandas (2008/2010/2015 for the different versions of Python).

josef-pkt:
I will try again today.
The configuration itself is not a problem, I have been running it for years and I still have my development setup for statsmodels (inplace build) with this. It is based on the last WinPython that came with gcc which is python 3.4. (I didn't manage to get Microsoft compilers to work before summer and haven't tried again.)

However, yesterday I installed the git source checkout directly from the venv, while for testing before I had built the wheels in my checkout, but also I checked that the appveyor wheels work. So I'm not sure which version works with the venv.

josef-pkt:
The main or only reason that we still support python 3.4 is because this is currently the only working setup that I have.

bashtage:
> The main or only reason that we still support python 3.4 is because this is currently the only working setup that I have.

I always recommend conda.  It makes it very easy to setup/teardown venvs on Windows. 

If you don't want conda, I think 3.5 or 3.6 are also very easy to setup.  Install VS 2015 community. Then download form Python.org, update pip, and pip install NumPy, SciPy, Pandas, Matplotlib, ipython, pytest and plenty of other packages.

bashtage:
> The configuration itself is not a problem, I have been running it for years and I still have my development setup for statsmodels (inplace build) with this. It is based on the last WinPython that came with gcc which is python 3.4. (I didn't manage to get Microsoft compilers to work before summer and haven't tried again.)

Are you using the wheels from PyPI or binaries compiled with mingw?

Earlier today I did the following:

Install MinGW (gcc 7.3.0), then install python 3.4 from Python.org, then pip install numpy scipy pandas=0.21, and finally install statsmodels using mingw (setting distutils.cfg). Everything went fine, but when I try and import statmodels.api the memory explodes before a segfault. This strategy is mixing and matching compilers, so I'm not surprised that it doesn't work correctly.

bashtage:
I made a temporary commit to run appveyor for Python 2.7, 3.4, 3.5, and 3.6 for both x86 and x64.

Here are the logs:

https://ci.appveyor.com/project/bashtage/statsmodels/build/1.0.884

This commit does the following:

1. Install numpy, scipy, cython
2. Build a sdist
3. Remove numpy, scipy, cython
4. Install using the sdist using `pip install statsmodels.tar.gz` where I renamed the sdist to `statsmodels.tar.gz`

It was able to install sm into all of the empty environments and run the test suite.  There is 1 error in the test suite due to failing to copy some data files.  I submitted a PR to simplify this by moving the datafiles to a standard location (tests/results).

josef-pkt:
2nd try runs fine: create `setup.py bdist_wheel` in my checkout python, pip install wheel in venv
(most likely my venv is not set up for compiling correctly, or pip in there is to old to build correctly)

` 13 failed, 6957 passed, 130 skipped, 4 xfailed, 1 xpassed, 267 warnings, 4 error in 826.25 seconds`

I got the errors in VAR tests
there are some errors with my old version of cvxopt.
the rest of the log is not anymore in the small windows command shell (too many lines).

I will try something else later today, then it looks ready for merging.

pip install -U without --no-deps did not install numpy in the venv, i.e. uses the one from the shared parent python

bashtage:
The errors in vector ar are due to me missing some data files in a strange
place. There is a new pr which moves them to a more reasonable place so
they will be included.

.travis_coveragerc
```diff
@@ -24,6 +24,8 @@ omit =
     */tests/p*
     */tests/r*
     */tests/s*
+    # Versioneer
+    */_version.py
```
jbrockmendel: Should this also go in .coveragerc?  (Which I’d be +1 on merging into setup.cfg BTW)

setup.py
```diff
@@ -1,186 +1,94 @@
...
+                  'docs': ['sphinx', 'nbconvert', 'jupyter_client',
+                           'ipykernel', 'matplotlib', 'nbformat', 'numpydoc',
+                           'pandas-datareader']}
 
-curdir = os.path.abspath(os.path.dirname(__file__))
-README = open(pjoin(curdir, "README.rst")).read()
-CYTHON_EXCLUSION_FILE = 'cythonize_exclusions.dat'
+###############################################################################
+# Values that rarely change
+###############################################################################
```
jbrockmendel: Does flake8 not complain about this?  Usually I end up with one # and a bunch of dashes

setup.py
```diff
@@ -1,186 +1,94 @@
...
 
 DISTNAME = 'statsmodels'
 DESCRIPTION = 'Statistical computations and models for Python'
+with open(pjoin(os.getcwd(), 'README.rst')) as readme:
```
jbrockmendel: Is the getcwd() robust to running from outside this directory?

setup.py
```diff
@@ -1,186 +1,94 @@
...
-                    not numpy_incl in ext.include_dirs):
-                ext.include_dirs.append(numpy_incl)
-        _build_ext.build_extensions(self)
-
-
-def generate_cython():
-    cwd = os.path.abspath(os.path.dirname(__file__))
-    print("Cythonizing sources")
-    p = subprocess.call([sys.executable,
-                          os.path.join(cwd, 'tools', 'cythonize.py'),
```
jbrockmendel: I think this means we can get rid of tools.cythonize

setup.py
```diff
@@ -190,390 +98,217 @@ def check_dependency_versions(min_versions):
...
+]
+
+
+class CleanCommand(clean):
+    def run(self):
+        msg = """
 
-    if not ISRELEASED:
-        FULLVERSION += '.dev0+' + GIT_REVISION[:7]
+python setup.py clean is not supported.
```
jbrockmendel: Why not?

bashtage: It needs a custom clean to be useful, but this is fragile and adds a lot of LOC.  Git clean -xfd is a simple method to clean everything not tracked.

bashtage: It is also what NumPy (and possibly pandas) does.

setup.py
bashtage: Yes, in due course

setup.py
bashtage: Probably not

setup.py
bashtage: Nope

.travis_coveragerc
bashtage: Probably, these are vendored files with very low hit rates since they cover many continencies. Their coverage (or lack) does not reflect statsmodels.

setup.py
jbrockmendel: pandas supports it; not a big enough deal to risk de-railing this PR

josef-pkt:
I will merge this without any more checking (I cannot rebuild inplace right now because I have too many open notebooks and interpreters that I don't want to shut down. The rest worked fine.)

bashtage:
@josef-pkt This is ready now.

jbrockmendel:
Needs rebase.

josef-pkt:
all green, merging

@bashtage Thank you

jbrockmendel:
Way to go fellas, this is a big win.

bashtage:
Thanks - I'm glad this has gotten in, it will simplify some of the testing in my downstream projects

rtsio:
When is this going to be available? Simple installation of statsmodels is failing in our Docker build because it requires numpy to be installed first.

PR #4872, opened 2018-08-10
CLN: Remove nose and related references

bashtage:
Remove nose
Remove tests that were left in to accomodate nose
Remove documentation references to nose


bashtage:
closes #4867

statsmodels/compat/testing.py
```diff
@@ -22,12 +22,9 @@
     from unittest.case import SkipTest
 except ImportError:
     try:
-        from nose import SkipTest
+        from unittest2 import SkipTest
```
jbrockmendel: Shouldn’t we standardize on pytest.skip?

jbrockmendel:
@josef-pkt does this also mean we can go through and get rid of those PyTesterWhatever in all the `__init__` files?

bashtage:
> Will np.testing.assert_raises be an issue?

No, in the sense that legacy numpy depends and so must be installed as long as numpy depends on it.  sm just doesn't directly use nose, so that if/when upstream has no nose references then it can be totally dropped.

bashtage:
> does this also mean we can go through and get rid of those PyTesterWhatever in all the __init__ files?

I'd hope so, but that can wait for a future PR since it is easy. Slightly ironically, when testing the setup PR today I ran the test using `import statsmodels as sm; sm.test()` on a pip install using a sdist.

josef-pkt:
The PyTesterWhatever stay as I explained before

```
import statsmodels.regression as smr
smr.test()

....
```

bashtage:
> The PyTesterWhatever stay as I explained before

Isn't it easier to just run

```
pytest statsmodels/regression
```

?

jbrockmendel:
> I'd hope so, but that can wait for a future PR since it is easy.v

@bashtage totally fair.  This PR LGTM.

> The PyTesterWhatever stay as I explained before

@josef-pkt You "explained before" by pointing towards #4245 which discusses the proposition that there should be One True Way to invoke the test suite.  Jeepers buddy, cool those reflexes.

josef-pkt:
It's easier for users/developers that are used to a command line and are able to quickly open one that has the right python on the system/Windows path.

It's more difficult for users that just know how to open an interpreter, or have already an interpreter open and want to check if the installed package works correctly.
(This included me in the early days, when I mainly ran e.g. `from scipy import stats; stats.test()` )

expert developer usage: Given the size of statsmodels, I'm using now exclusively the command line because I ran either one test module or just one test class.

josef-pkt:
> Jeepers buddy, cool those reflexes.

AFAIR, I made the same arguments more explicitly already before that. 
I don't think I have to repeat the same arguments in the same elaborate form each time.

There is minimal cost to keeping it in, and there is a benefit to less experienced users, so it stays.
(And to emphasize again, less experienced user was my own experience when I started out with scipy, and I still don't like it that I cannot do `import pandas as pd, pd.test()` after updating the pandas version. I don't want to have to care about or know whether a package uses pytest or nosetest or whatevertest, I just want to check that the install is not screwed up on my machine.

statsmodels/base/tests/test_generic_methods.py
```diff
@@ -384,7 +384,7 @@ def setup(self):
 
     def test_zero_constrained(self):
         # override, we cannot remove the only regressor
-        raise SkipTest
+        pytest.skip(msg='Override since cannot remove the only regressor')
         pass
```
jbrockmendel: Is the `pass` here needed?

jbrockmendel:
> AFAIR, I made the same arguments more explicitly already before that.

Maybe to someone... regardless, let's avoid further hijacking this thread.  #4245 is the right place to discuss the sentimental value of this code.

> CLN: Remove SkipTest

@bashtage nice!

josef-pkt:
>  discuss the sentimental value of this code.

Are you listening to what I'm saying?
I know that your view for what constitutes clean and user friendly code does not coincide with my view, but if you don't watch your language, I will ignore all your comments.

Whether it has "sentimental value" for me after ten years of scipy and statsmodels and scipy development is completely irrelevant to the point here, which was raised in this issue and not by me.
The test functions stay as user convenience, the "sentimenal value" is because of my experience as a "newbie".

statsmodels/nonparametric/tests/test_kde.py
```diff
@@ -208,7 +207,7 @@ def t_est_density(self):
 
     def test_evaluate(self):
         if self.kernel_name == 'cos':
-            raise SkipTest("Cosine kernel fails against Stata")
+            pytest.skip(msg="Cosine kernel fails against Stata")
```
josef-pkt: based on pytest documentattion keyword name: msg -> reason

statsmodels/graphics/tests/test_functional.py
```diff
@@ -21,7 +22,7 @@
 data = data.raw_data[:, 1:]
 
 
-@skipif(not have_matplotlib, reason='matplotlib not available')
+@pytest.mark.skipif(not have_matplotlib, reason='matplotlib not available')
```
josef-pkt: just import the skipif name would be simpler
`from pytest.mark import skipif`
or whatever the correct import is

bashtage: I tried to be coherent in all marks, either, skip, skipif, or example.  Probably a few others to clean up, e.g., @slow

bashtage: I have made all marks consistent now using `@pytest.mark.`_marktype_

josef-pkt:
@bashtage  Don't squash this. I would like to see what works and how things need to by changed to be pytest compatible.

I will go over it when it is green (or as green as it can get)

bashtage:
I only squashed some experiments that were trying to work through a pytest
bug that affects 2.7.

josef-pkt:
The travis runtime went up for the first two python version, compared to travis run of master. However this is most likely mostly noise. The actual runtime reported for the unit tests went up only a bit (maybe close to 10 %, the number of unit tests reported is now larger)

bashtage:
I think Travis is having network issues. There have been random Failures on webuse tests today.

bashtage:
I think this one is finished.

docs/source/dev/test_notes.rst
```diff
@@ -80,12 +80,12 @@ tested. The tests look something like
             assert_almost_equal(self.res1.foo, self.res2.foo, 4)
 
     if __name__ == "__main__":
```
josef-pkt: I think we should remove this from the documentation know.
We don't want any `if __name__ == "__main__"` parts anymore.

Adding a section about running individual test(classes) on the command line would be better and reflect more modern usage.
However, that can be a separate PR. This PR is for the transition to pytest

statsmodels/examples/tests/test_notebooks.py
```diff
@@ -3,21 +3,21 @@
 import os
 import sys
 
-from statsmodels.compat.testing import SkipTest, example
+import pytest
 
 try:
-    import pytest
+    print('In try')
```
josef-pkt: Are the two print here left intentionally in the code

josef-pkt:
explanation for running pytest on commandline sounds very good.

I didn't check details for individual tests

as green as it gets, no merge conflicts, so merging

@bashtage Thank you

PR #4875, opened 2018-08-10
TST: remove npt.dec, change many assert_raises to pytest.raises

jbrockmendel:
Still a bunch of `assert_raises` to get rid of, but this gets a good fraction of the way there.


jbrockmendel:
breaking into smaller pieces; closing.

PR #4907, opened 2018-08-15
TST: use pytest.warns instead of npt.assert_warns

jbrockmendel:
Also uses more readable contextmanager syntax.


jbrockmendel:
rebased @josef-pkt ping

jbrockmendel:
@josef-pkt green, single commit, super easy

PR #4909, opened 2018-08-15
ENH: QIC for GEE

kshedden:
This is an implementation of Pan's QIC (Akaike-like information criterion for GEE).

### DIFF
--- a/statsmodels/genmod/tests/test_gee.py
+++ b/statsmodels/genmod/tests/test_gee.py
@@ -10,12 +10,11 @@ exactly.
 """
 
 from statsmodels.compat import lrange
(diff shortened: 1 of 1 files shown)


statsmodels/genmod/generalized_estimating_equations.py
```diff
@@ -1324,6 +1324,92 @@ def _derivative_exog(self, params, exog=None, transform='dydx',
         return margeff
 
 
+    def QIC(self, params, scale, cov_params):
```
bashtage: should functions have lower case names?

jbrockmendel: +1

Issue #4635, opened 2018-05-15
Failed to install statsmodels 0.9.0 if numpy is not installed before


tamasgal:
Our CI server discovered this issue in the fresh install stage. An example how to reproduce it is given below.

The workaround is obviously installing numpy first.

The main issue is, when running CI, people often use a single file (requirements.txt) to install the dependencies. I am not sure what's the exact way to define the order of the packages to install, but statsmodels seem to come before numpy (in our case), although numpy is at the top of the file.

```
░ tamasgal@greybox:~
░ 07:52:49 > pyenv virtualenv 3.6.4 statsmodels_test
Using base prefix '/Users/tamasgal/.pyenv/versions/3.6.4'
New python executable in /Users/tamasgal/.pyenv/versions/3.6.4/envs/statsmodels_test/bin/python3.6
Also creating executable in /Users/tamasgal/.pyenv/versions/3.6.4/envs/statsmodels_test/bin/python
Installing setuptools, pip, wheel...done.
Requirement already satisfied: setuptools in /Users/tamasgal/.pyenv/versions/3.6.4/envs/statsmodels_test/lib/python3.6/site-packages
Requirement already satisfied: pip in /Users/tamasgal/.pyenv/versions/3.6.4/envs/statsmodels_test/lib/python3.6/site-packages
░ tamasgal@greybox:~
░ 07:54:51 > pyenv shell statsmodels_test
░ tamasgal@greybox:~
░ 07:55:01 > pip install statsmodels
Collecting statsmodels
  Using cached https://files.pythonhosted.org/packages/67/68/eb3ec6ab61f97216c257edddb853cc174cd76ea44b365cf4adaedcd44482/statsmodels-0.9.0.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/private/var/folders/n2/fpv312vd5xg91ncw_s5f1m5m0000gn/T/pip-install-azgq3_k_/statsmodels/setup.py", line 347, in <module>
        from numpy.distutils.misc_util import get_info
    ModuleNotFoundError: No module named 'numpy'

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /private/var/folders/n2/fpv312vd5xg91ncw_s5f1m5m0000gn/T/pip-install-azgq3_k_/statsmodels/
```

elboerto:
[...]

PR #4912, opened 2018-08-15
BLD: Rebased ci improvements 

bashtage:
Rebased and slightly modified version of @matthew-brett 's CI improvements that add pip testing and pre-release testing. The latter is pretty useful and I submitted a PR with fixes for 24 fails that are coming. It also adds pip for appveyor to avoid the issues around 0.9's release. 

I made some slight configuration changes to the travis builds to provide coverage.  Currently

NumPy: 1.11, 1.12, 1.13, 1.14, 1.15
SciPy: 0.18, 0.19, 1, 1.1
Pandas: 0.19, 0.20, 0.22, 0.23 
Matplotlib: 1.5, 2, 2.2

This is at least as good as it was before.   This only required 1 additional job rather than 2 that were in the original PR. 

closes #4652


tools/ci/docbuild_install.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
```
jbrockmendel: Coach is complaining about these she bangs being misssing

ChadFulton:
+1 on improving CI, although sorry to say that I don't understand the details

bashtage:
The main point of this PR is to add testing against pip wheels.  This will avoid release problems where conda packages pass but pip wheels fail.   The other gain from this is testing against pre-releases of NumPy , SciPy and pandas.

tools/ci/docbuild_install.sh
bashtage: These are only for use with source and no standalone scripts, so no shebang.

statsmodels/regression/tests/test_recursive_ls.py
```diff
@@ -160,7 +160,7 @@ def test_glm(constraints=None):
 
     # Regression coefficients, standard errors, and estimated scale
     assert_allclose(res.params, res_glm.params)
-    assert_allclose(res.bse, res_glm.bse)
+    assert_allclose(res.bse, res_glm.bse, atol=1e-6)
```
jbrockmendel: Is this not passing at the moment?

bashtage: This PR adds new testing environments using PyPI packages which use different compilers and blas.

josef-pkt:
based on the time in the PR testing lists:
change to appveyor is a bit expensive, goes from around 40 min to 1h20min which can make the CI testing queue quite a bit slower and longer when there is more action.

travis goes from just below 3h to around 3h15min, which is not too large an increase (travis runs more simultaneous jobs)

matthew-brett:
For the time of the Appveyor checks - maybe you can drop the conda builds on the basis that conda-forge is testing those for you?    I think y'all do need to test 32-bit Windows, it's significantly different from 64-bit, and still fairly widely used.

matthew-brett:
For a long time the 32 bit installer was the default on Windows.  I am sure
that you can give up on 32 bit in a a couple of years.

bashtage:
I marked a few more tests as slow and added --skip-slow to pytest.  The set of 7 runs finish in about 1 hour now. It woudl probably be best to let the slow tests run on 1 or 2 platforms, probably 1 2.7 and 1 3.6 just to be a bit more careful.

appveyor.yml
```diff
@@ -4,9 +4,18 @@ clone_depth: 50
 os: Visual Studio 2015
 
 environment:
+  PYTEST_DIRECTIVES: --skip-slow
```
josef-pkt: question: to disable skip-slow temporarily e.g. in a PR, can we just delete `--skip-slow` or does there need to be something there so it's not empty

bashtage:
Empty works.

josef-pkt:
I will merge this with just a surface understanding of the changes.
We will need to turnoff `--skip-slow` in release preparation and maybe some PRs that touch skipped unit tests, 
e.g. We need more work on ZeroInflated countmodels that has now a ZIP marked as slow.
(slowness is largely because we don't have good start_params yet)

josef-pkt:
green enough, merging

@bashtage @matthew-brett  Thank you

josef-pkt:
commit 6b0302fd0f7d298fd22c8948fd3dadfd939726bf closed too many issues, i.e. issues that require the setup.py changes for handling build/install dependencies like #4635

PR #4932, opened 2018-08-17
Move misplaced results files into tests/results/ directories

jbrockmendel:
Update relevant imports.

A couple of files that appear orphaned are moved to sandbox/unsorted/


jbrockmendel:
This can be split into smaller pieces if easier for reviewers

PR #4934, opened 2018-08-17
Replace copy/paste with imports

jbrockmendel:


jbrockmendel:
@josef-pkt sandbox.tests.test_formula tests are failing here because the non-sandbox Contrast expects np.array args while the sandbox version takes Formula objects.  Troubleshooting these tests is a pain because of namespace/frame hijinks.

I'm guessing sandbox.formula is entirely superseded by patsy.  Is formula in the "may one day be useful" category?

jbrockmendel:
Closing in favor of #4940.

PR #4940, opened 2018-08-19
remove sandbox.formula; long-since superseded by patsy

jbrockmendel:
Since sandbox.formula is the only thing that uses sandbox.contrast_old (for which there are newer, harder, better, faster, stronger versions in `stats.contrast`), that gets removed too.

One test file and two examples files go.


jbrockmendel:

PR #4944, opened 2018-08-19
BLD: Add pinned numpy on appveyor

bashtage:
Pin numpy against old SciPy

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,6 +10,7 @@ environment:
     - PY_MAJOR_VER: 2
       PYTHON_ARCH: "x86_64"
       SCIPY: "0.18"
+      NUMPY: "1.11"
     - PY_MAJOR_VER: 3
       PYTHON_ARCH: "x86_64"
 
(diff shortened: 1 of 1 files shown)


josef-pkt:
didn't help, same error even with numpy 1.11.3
something else wrong than my guess in #4939

```
statsmodels\compat\scipy.py:259: in <module>
    from scipy.special import gammaln
c:\py\lib\site-packages\scipy\special\__init__.py:636: in <module>
    from ._ufuncs import *
E   ImportError: DLL load failed: The specified module could not be found.
```

josef-pkt:
I don't know much about conda
e.g. it looks like old numpy version got reuploaded/rebuilt 2 days ago, our appveyor also started to fail 2 days ago
https://anaconda.org/anaconda/numpy/files?version=1.11.3

I have no idea what that implies or where to get more info

bashtage:
I suspect that icc_rt is required.  It disappeared across the successful and the failed builds.  I've added this as a requirement so we'll see if that fixes the issue.

josef-pkt:
just as background info: what is icc_rt?

Thanks for figuring this out.
Merging, so we have one ci green again

bashtage:
Intel c compiler runtime

Issue #4748, opened 2018-06-16
plot_pacf vs pacf nlags default behavior 


gargashish11:
functions in question are considering different default values for nlags when it (and all other parameters)   is omitted. pacf is defaulting to 40 and plot_pacf to the length of input array.

bashtage:
This is addressed in #4949

PR #4949, opened 2018-08-21
BUG: Set default lags for acf/pacf plots

### LINKED ISSUE #4663: ENH: pacf needs to limit nlag to ensure that it is well defined.
```
plot_pacf(np.array([ -6.,   8.,   4.,  -2.,   4.,   1.,  -2.,  -3.,  -5.,   2.,   4.,
        12.,  24., -17.,  -3.,  -4.,  -8.,  -2.,   4.,  -1.,   3.,   5.,
         7.,  -6., -26.,  17.,  -4.,  -5.,   8.,   4., -10.,  -3.,  -1.,
        -2.,  -7.,  18.,  -6.,  14.,   6.,  -4.,  13.,  -9.,  11.,   2.,
         0.,  -3.,  10., -17.,  -4.,  -6.,   3.,  -3.,  -7.,  13.,  -7.,
        -5.,   0.,  -5.,  -5.,  -6.,  25., -16.,  -1.,   9., -12.,   7.,
         2.,   2.,  -1.,  -2.,  -4.,   4., -23.,  16.,  -1.,   4.,   6.,
        -1.,   6.,   0.,  -7.,  -4.,  10.,  -7.,   1.,   5.,  -3.,  -4.,
        -1.,   4.,   1.,  -4.,   9.,  -3.,  -4.,  37.,  -5.,   4.,   1.,
         2.,  -2.,  -4.,   1.,  -1.,   3.,  -1.,  -3., -27.,  21., -13.,
        -8.,  -3.,   3.,   3.,  -4.,  -2.,  -9.,  -7.,   2., -10., -21.,
        19.,   5.,  -2.,  -7.,  -2.,  -5.,  -7.,   6.,  13.,   3.,  10.,
        -2.,   1.,  -2.,   8.,  11.,   4.,   3.,   1., -10.,  -8.,  -2.,
        -5.,  -2.,  -1.,  -2.,  -4.,  -7.,  -2.,   3.,  11.,  13.,   9.,
         6.])) 
```

returns

> .../python3.6/site-packages/statsmodels/regression/linear_model.py:1283: RuntimeWarning: invalid value encountered in sqrt
>   return rho, np.sqrt(sigmasq)

`plot_acf(np.array(...))` - on the same data - works as expected.
Digging into the source code, it appears that `statsmodels.tsa.stattools.pacf(np.array(...))` works for the default `nlags=40` but results in the same error for nlags returned by  `_prepare_data_corr_plot(x, lags, zero)` in statsmodels.graphics.tsaplots

bashtage:
Set a sensible default value for acf and pacf plot lags when not provided
using a method common to R based on the log of the number of observations.

closes #4663

### DIFF
--- a/statsmodels/graphics/tsaplots.py
+++ b/statsmodels/graphics/tsaplots.py
@@ -3,7 +3,6 @@
 
 import numpy as np
 
-from statsmodels.compat.pandas import sort_values
 from statsmodels.graphics import utils
 from statsmodels.tsa.stattools import acf, pacf
 
@@ -12,7 +11,10 @@ def _prepare_data_corr_plot(x, lags, zero):
     zero = bool(zero)
     irregular = False if zero else True
     if lags is None:
-        lags = np.arange(not zero, len(x))
+        # GH 4663 - use a sensible default value
+        nobs = x.shape[0]
+        lim = min(int(np.ceil(10 * np.log10(nobs))), nobs - 1)
+        lags = np.arange(not zero, lim + 1)
     elif np.isscalar(lags):
         lags = np.arange(not zero, int(lags) + 1)  # +1 for zero lag
     else:
@@ -48,7 +50,11 @@ def _plot_corr(ax, title, acf_x, confint, lags, irregular, use_vlines,
             lags = lags[1:]
             confint = confint[1:]
             acf_x = acf_x[1:]
-        ax.fill_between(lags, confint[:, 0] - acf_x, confint[:, 1] - acf_x, alpha=.25)
+        lags = lags.astype(np.float)
+        lags[0] -= 0.5
+        lags[-1] += 0.5
+        ax.fill_between(lags, confint[:, 0] - acf_x,
+                        confint[:, 1] - acf_x, alpha=.25)
 
 
 def plot_acf(x, ax=None, lags=None, alpha=.05, use_vlines=True, unbiased=False,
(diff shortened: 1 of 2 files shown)


bashtage:
Log is a sensible choice since it is probably maximum rate with which standard confidence regions are plausible.

bashtage:
closes #4748

statsmodels/graphics/tsaplots.py
```diff
@@ -48,6 +52,9 @@ def _plot_corr(ax, title, acf_x, confint, lags, irregular, use_vlines,
             lags = lags[1:]
             confint = confint[1:]
             acf_x = acf_x[1:]
+        lags = lags.astype(np.float)
+        lags[0] -= 0.5
+        lags[-1] += 0.5
```
josef-pkt: what's the reason for this?
Doesn't this keep the fill_between too far away from the plot left and right edges?

bashtage: Nope, it fills right to the point and so 1/2 the dot is outside the CI on the edges.

This is from master:

image

statsmodels/graphics/tsaplots.py
```diff
@@ -12,7 +12,11 @@ def _prepare_data_corr_plot(x, lags, zero):
     zero = bool(zero)
     irregular = False if zero else True
     if lags is None:
-        lags = np.arange(not zero, len(x))
+        # GH 4663 - use a sensible default value
+        nobs = x.shape[0]
+        lim = min(int(np.ceil(10 * np.log10(nobs))), nobs-1)
+        lim = lim if lim <= nobs - 1 else nobs - 1
```
josef-pkt: isn't this redundant with the previous line?

pep-8 missing spaces in `nobs-1` on previous line

josef-pkt:
LGTM, I don't have any clear opinion about the default, so following R is sensible.

statsmodels/graphics/tsaplots.py
bashtage: Not needed

josef-pkt:
all green, merging

@bashtage Thank you

PR #4954, opened 2018-08-21
BUG: disallow use_self=False

jbrockmendel:


statsmodels/base/model.py
```diff
@@ -1223,6 +1223,9 @@ def normalized_cov_params(self):
 
     def _get_robustcov_results(self, cov_type='nonrobust', use_self=True,
                                use_t=None, **cov_kwds):
+        if use_self is False:
```
bashtage: Why not just remove `use_self` and any references to it?

jbrockmendel: Because I bayessume the maintainer would object to the idea of dead code being removed.

(bayessume: a Bayesian inference that to the naive observer looks like an assumption)

jbrockmendel:
@josef defacto green, no merge conflicts, bug-adjacent

jbrockmendel:
woot! less than 10% of outstanding PRs are mine now

PR #4988, opened 2018-08-24
[TST/MAINT] de-duplicate get_correction_factor code

jbrockmendel:
ATM the `get_robust_clu` classmethod calculates a correction factor and then a bunch of classes have that code copy/pasted into their `setup_class`.  A bunch of _other_ classes have a slightly-changed variant of that calculation, with no documentation as to why the slight change was made (IIRC the answer is "to make it match Stata/R").

This de-duplicates all that code by implementing `get_correction_factor`.

### DIFF
--- a/statsmodels/discrete/discrete_model.py
+++ b/statsmodels/discrete/discrete_model.py
@@ -3743,7 +3743,6 @@ class L1CountResults(DiscreteResults):
     __doc__ = _discrete_results_docs % {"one_line_description" :
             "A results class for count data fit by l1 regularization",
             "extra_attr" : _l1_results_attr}
-        #discretefit = CountResults(self, cntfit)
 
     def __init__(self, model, cntfit):
         super(L1CountResults, self).__init__(model, cntfit)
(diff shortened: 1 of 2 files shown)


jbrockmendel:
Take a look at the affected code. No information is lost.

jbrockmendel:
> I should clarify that information is in fact gained, will do so when at a real keyboard.

Consider the perspective of a new reader.  Unless you know ex-ante that there are only two versions of this code, the natural assumption is that the similar-looking code is duplicated for a reason, i.e. that there are small differences between the many versions.  This eliminates that confusion and the accompanying cognitive overhead.

The big thing that needs to be documented in the affected code is "Why does class A use implementation X while class B uses implementation Y?"  This PR doesn't answer that question, but puts a big TODO note in place, and minimizes the number of places where the question needs answering.

Another non-obvious thing that makes the two versions look more different than they are is the inversion:
```
        corr_fact = (nobs) / float(nobs - 1.)
        cls.corr_fact = np.sqrt(1./corr_fact)    # <-- inverting is not done in the other version
```

which this PR much more clearly writes as

```
        corr_fact = (nobs - 1.) / nobs
        cls.corr_fact = np.sqrt(corr_fact)
```

making it clear that the only difference between the two versions is the subtraction of `k_params` in the denominator of the other version.

Last, reducing code duplication is about forward-looking bug-prevention. As we see all over the code-base, one version of duplicated code get missed, and by the time someone realizes other versions were missed, its an archaeological adventure to sort out "are these slightly different on purpose?  It must be on purpose right?"

@bashtage care to weigh in?

jbrockmendel:
@josef-pkt besides being Goos code hygiene, this is a move toward parameterization/fixturization, which will allow us to systematically cover “method” etc args.

PR #5016, opened 2018-08-25
ENH: Burgs algorithm

bashtage:
Add Burg's algorithm as an alternative to Yule-Walker for PACF, AR, and ACF estimation.

### DIFF
--- a/statsmodels/tsa/stattools.py
+++ b/statsmodels/tsa/stattools.py
@@ -612,8 +612,8 @@ def pacf_burg(x, nlags=None, demean=True):
     if p > nobs - 1:
         raise ValueError('nlags must be smaller than nobs - 1')
     d = np.zeros(p + 1)
-    d[0] = x.dot(x)
-    phi = np.ones(p + 1)
+    d[0] = 2 * x.dot(x)
+    phi = np.zeros(p + 1)
     u = x[::-1].copy()
     v = x[::-1].copy()
     d[1] = u[:-1].dot(u[:-1]) + v[1:].dot(v[1:])
@@ -627,7 +627,8 @@ def pacf_burg(x, nlags=None, demean=True):
         v[1:] = last_v[1:] - phi[i] * last_u[:-1]
         d[i + 1] = (1 - phi[i] ** 2) * d[i] - v[i] ** 2 - u[-1] ** 2
         phi[i + 1] = 2 / d[i + 1] * v[i + 1:].dot(u[i:-1])
-    sigma2 = (1 - phi ** 2) * d / (2 * (nobs - np.arange(0, p + 1)))
+    sigma2 = (1 - phi ** 2) * d / (2. * (nobs - np.arange(0, p + 1)))
+    phi[0] = 1  # Insert the 0 lag partial autocorrel
 
     return phi, sigma2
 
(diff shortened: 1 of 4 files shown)


jbrockmendel:
Based on a quick pass this looks good.  Only question: why does burg belong in linearmodel.  (I have the same question about yulewalker...)

bashtage:
Burg is an alternative to yule_walker since both estimate the parameters of
an AR1 and so should go in the same module.

statsmodels/tsa/stattools.py
```diff
@@ -574,6 +574,65 @@ def pacf_yw(x, nlags=40, method='unbiased'):
     return np.array(pacf)
 
 
+def pacf_burg(x, nlags=None, demean=True):
```
ChadFulton: This name (`pacf_burg`) has a different structure from `levinson_durbin_pacf`, but of course they do different things w.r.t. the pacf. Just wanted to check that this was desired.

statsmodels/tsa/stattools.py
```diff
@@ -574,6 +574,65 @@ def pacf_yw(x, nlags=40, method='unbiased'):
...
+    """
```
ChadFulton: Perhaps a one-line description?

statsmodels/tsa/stattools.py
```diff
@@ -574,6 +574,65 @@ def pacf_yw(x, nlags=40, method='unbiased'):
...
+        raise ValueError('x must be 1-d or squeezable to 1-d.')
+    if demean:
+        x = x - x.mean()
+    nobs = x.shape[0]
+    p = nlags if nlags is not None else min(int(10 * np.log10(nobs)), nobs - 1)
+    if p > nobs - 1:
+        raise ValueError('nlags must be smaller than nobs - 1')
+    d = np.zeros(p + 1)
+    d[0] = 2 * x.dot(x)
+    phi = np.zeros(p + 1)
```
ChadFulton: I don't know if we're still trying for descriptive naming, but could be `arcoefs` or something if we are.

ChadFulton:
Looks like a nice addition to me. Couple of small comments.

statsmodels/tsa/stattools.py
bashtage: The most comparable is `pacf_yw` which has signature `def pacf_yw(x, nlags=40, method='unbiased')` vs `def pacf_burg(x, nlags=None, demean=True):`

There is no sense of alternative methods here so that parameter makes no sense. `nlags=None` is the future of acf/pacf estimations (I have another PR that fixes this), and then `10*log10(nobs)`.

As for `levinson_durbin` vs `lebinson_durbin_pacf`, the interpretation of `x` depending another input is IMO a mistake. It should have one job -- convert acovf to arcoefs (and other stuff).  This makes it clear that the user can choose their preferred pacf estimator before running LD, and LD_pacf doesn't need to know about all of the pacf estimators available.

statsmodels/tsa/stattools.py
bashtage: Yes

statsmodels/tsa/stattools.py
bashtage: Good idea.

ChadFulton:
Thanks! If you think it's ready to merge, I will do so tomorrow, unless there are concerns.

ChadFulton:
@bashtage ready for merge?

bashtage:
Yes

ChadFulton:
Thanks @bashtage !

PR #5018, opened 2018-08-25
MAINT: fix typos in discrete_margins

jbrockmendel:
maringal --> marginal
an scalar --> a scalar
constant is or None --> constant is 0 or None
Only overall has any effect here.you --> Only overall has any effect here
DataFrames --> DataFrame

and incidentally some other fixups in the vicinity of these typos


jbrockmendel:

PR #5042, opened 2018-08-28
ENH: Add innovations algorithm

bashtage:
Add innovations algorithm to convert acov to MA
Add filter to filter innovations usign output of ia

### DIFF
--- a/statsmodels/compat/pandas.py
+++ b/statsmodels/compat/pandas.py
@@ -29,3 +29,12 @@ else:
 
     data_klasses = (pandas.Series, pandas.DataFrame, pandas.Panel,
                     pandas.WidePanel)
+
+try:
+    import pandas.testing as testing
+except ImportError:
+    import pandas.util.testing as testing
+
+assert_frame_equal = testing.assert_frame_equal
+assert_index_equal = testing.assert_index_equal
+assert_series_equal = testing.assert_series_equal
--- a/statsmodels/tsa/tests/test_stattools.py
+++ b/statsmodels/tsa/tests/test_stattools.py
@@ -1,4 +1,4 @@
-from statsmodels.compat.python import lrange, PY3
+from statsmodels.compat.python import lrange
 
 import os
 import warnings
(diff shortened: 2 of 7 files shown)


josef-pkt:
looks good based on a quick skimming
closes #2593

josef-pkt:
sorry, pushed the wrong button (not enough coffee yet

statsmodels/compat/pandas.py
```diff
@@ -29,3 +29,12 @@ def sort_values(df, *args, **kwargs):
 
     data_klasses = (pandas.Series, pandas.DataFrame, pandas.Panel,
                     pandas.WidePanel)
+
+try:
+    import pandas.testing as testing
```
jbrockmendel: What is the version number cutoff at which this becomes unnecessary?

bashtage:
0.20

statsmodels/tsa/tests/test_stattools.py
```diff
@@ -714,3 +714,78 @@ def test_pacf_burg_error():
...
+    ma = -0.9
+    acovf = np.array([1 + ma ** 2, ma])
+    theta, _ = innovations_algo(acovf, nobs=4)
+    e = np.random.randn(5)
+    endog = e[1:] + ma * e[:-1]
+    resid = innovations_filter(endog, theta)
+    expected = [endog[0]]
+    for i in range(1, 4):
+        expected.append(endog[i] + theta[i, 0] * expected[-1])
+    print(expected)
```
josef-pkt: there is still a print left

josef-pkt:
separated into two commits looks good/useful
I only superficially skimmed the algorithm, it might be a bit slow for large nobs and n=nobs

LGTM, AFAICS

bashtage:
print is finally gone.

bashtage:
It is not that bad for fairly long time series - I tested numba and it could be trivially sped up by a factor of 250-400 with a simple @numba.jit.

josef-pkt:
I think we should add numba as optional requirement with several cases to gain experience with it in 0.11, it's a bit late for 0.10.

josef-pkt:
merge when green (ETA 12 to 13 hours, i.e. not today)

bashtage:
@josef-pkt green.

josef-pkt:
@bashtage Thank you

PR #5043, opened 2018-08-28
BUG: Initialization when simulating

ChadFulton:
With the introduction of the new `Initialization` objects for state space initialization, the `simulate` method was never updated to initialize simulations appropriately.

(No backport required, since the new `Initialization` objects were introduced after v0.9 and haven't yet been in a release).


jbrockmendel:
test?

ChadFulton:
> test?

Good idea!

ChadFulton:
Added test for "known initialization" which handily tests that `Initialization` objects are being used and also tests for the cast of non-zero state initialization.

PR #5068, opened 2018-08-30
MAINT: strict linting for test_varmax

jbrockmendel:
@ChadFulton 

also updated usages of `assert_raises` to more readable `with pytest.raises`

### DIFF
--- a/lint.sh
+++ b/lint.sh
@@ -29,6 +29,7 @@ if [ "$LINT" == true ]; then
         statsmodels/tsa/vector_ar/dynamic.py \
         statsmodels/tsa/statespace/*.py \
         statsmodels/tsa/statespace/tests/results/ \
+        statsmodels/tsa/statespace/tests/test_varmax.py \
         statsmodels/tsa/statespace/tests/test_var.py \
         conftest.py
     if [ $? -ne "0" ]; then
--- a/statsmodels/tsa/statespace/tests/test_varmax.py
+++ b/statsmodels/tsa/statespace/tests/test_varmax.py
@@ -5,27 +5,27 @@ Author: Chad Fulton
 License: Simplified-BSD
 """
 from __future__ import division, absolute_import, print_function
-
-import numpy as np
-import pandas as pd
-import pytest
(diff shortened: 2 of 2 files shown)


ChadFulton:
Why is `assert_equal` bad?

jbrockmendel:
> Why is assert_equal bad?

Take a look at how it is used at the moment: `assert_equal(re.search('Results for equation %s' % name, table) is None, False)`.  This is obviously a relic being forced to avoid `assert ` statements;  a reader will find that _much_ less intuitive than the pythonic replacement `assert re.search('Results for equation %s' % name, table)`

ChadFulton:
> a reader will find that much less intuitive than the pythonic replacement assert re.search('Results for equation %s' % name, table)

This seems like a dubious empirical claim :)

jbrockmendel:
> This seems like a dubious empirical claim :)

Are you serious?  Because I expect I compose a large portion of the non-you/josef readership of these files and I as such I am an expert on the empirics of this claim.

jbrockmendel:
Can we get this over with.  The regexp part in particular should be obvious after a good night's sleep.

jbrockmendel:
>> a reader will find that much less intuitive than the pythonic replacement` assert re.search('Results for equation %s' % name, table)`

> This seems like a dubious empirical claim :)

Consider a thought experiment about internal consistency. `tsa.x13` has the code:

```
    if re.search("Mean is not significant", results.results):
        include_mean = False
    elif re.search("Constant", results.results):
        include_mean = True
    else:
        include_mean = False
```

I contend that this is much more readable than the alternative:

```
    if np.equal(re.search("Mean is not significant", results.results) is None, False):
        include_mean = False
    elif np.equal(re.search("Constant", results.results) is None, False):
        include_mean = True
    else:
        include_mean = False
```

The notion that `assert_equal(re.search(...) is None, False)` is clearer than `assert re.search(...)` is analogous to the notion that the latter `if np.equal(...)` version above is clearer than the existing code in `tsa.x13`.

PR #5071, opened 2018-08-30
MAINT: remove example file corresponding to removed module

jbrockmendel:


jbrockmendel:
@josef-pkt this is an example file for a file that has been removed

jbrockmendel:
@josef-pkt line 7 of this example file imports from `survival2`, which was removed a while ago.  This file is defunct.

PR #5086, opened 2018-09-02
TST: parametrize tests instead of using for loops

jbrockmendel:
For any skeptics, recall the advantage of parametrizing is that if a particular case fails, a) the parameter value that fails is included in the test output and b) subsequent tests are still run, so we get information on all failing cases.


jbrockmendel:
Travis fail is from before #5080.  Can avoid rebase/CI clog?

bashtage:
Looks OK to go in. Squash and rebase.

bashtage:
Assuming it passes.

PR #5088, opened 2018-09-02
DOC: Add javascript to link to other doc versions

thequackdaddy:
This is just some (not so) fancy javascript/sphinx/css that will add a dropdown so you can select older versions of the docs... Requires a file I called `versions.json` which is a json array of all the folders you want to add links to. 

See my gh-pages branch here:

https://thequackdaddy.github.io/statsmodels.github.io/devel/dev/index.html

https://github.com/thequackdaddy/statsmodels.github.io

### DIFF
--- a/docs/themes/statsmodels/layout.html
+++ b/docs/themes/statsmodels/layout.html
@@ -16,6 +16,23 @@
 $.facebox.settings.closeImage = "{{ pathto('_static/closelabel.png', 1) }}"
 $.facebox.settings.loadingImage = "{{ pathto('_static/loading.gif', 1) }}"
 </script>
+
+<script>
+$(document).ready(function() {
+  $.getJSON("{{ pathto('../versions.json', 1) }}", function(data) {
+    $(".header").prepend(`
+    <div class="dropdown">
+      <button class="dropbtn">Other Versions</button>
+      <div class="dropdown-content">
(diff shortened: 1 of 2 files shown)


jbrockmendel:
I don’t know enough about JavaScript to comment on the implementation, but this is a neat idea.

Is it feasible to put the same drop down menu on the older version pages so readers can get back to the newest version?

thequackdaddy:
> Is it feasible to put the same drop down menu on the older version pages so readers can get back to the newest version?

Playing with this, there's a few instances that you would need to cherry-pick old commits because packages have been updated and the location of some items on the internet have moved (Rdatasets, especially)... not impossible, but definitely annoying. 

So feasible yes, but non-trivial. 

It might be easier to find a few common pages (like just the the index.html) and edit them directly in the statsmodels.github.io project. Running `make html` on the old, out-of-date tags, is just going to be a big waste of time.

thequackdaddy:
Also, I added a couple of css fixes that I think improves the way the search box looks...

Before:
screen shot 2018-09-02 at 3 53 40 pm

After:
screen shot 2018-09-02 at 3 53 28 pm

josef-pkt:
merging without trying to check any details.

@thequackdaddy Thank you for taking care of this

PR #5118, opened 2018-09-06
MAINT: remove file attempting to automatically generate an api.py file

jbrockmendel:
Looks like a valiant experiment, but one that is never going to be useful.


jbrockmendel:
> actually, I think this looks interesting. I haven't seen it before.

While this isn't my first-choice outcome for this PR, this is a big improvement over the status quo.

PR #5120, opened 2018-09-06
BUG: Ensure internal tester exits with error if needed

bashtage:
Ensure that the internal tester errors when an internal error was encountered

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -39,6 +39,5 @@ build_script:
   - python setup.py develop
 
 test_script:
-  # Remove xdist until warning reporting is working
   - echo pytest -n auto -r s statsmodels --skip-examples %PYTEST_DIRECTIVES%
   - pytest -n auto -r s statsmodels --skip-examples %PYTEST_DIRECTIVES%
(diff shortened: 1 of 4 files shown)


josef-pkt:
this exits the interpreter after a call to statsmodels.test(), doesn't it?

This is not intended behavior, i.e. we want to be able to use repeated calls in the same python session.
```
statsmodels.regression.test()
statsmodels.genmode.test()
...
```

bashtage:
It appears to be needed to get a correct status when running on a CI

python -c "import statsmodels;statsmodels.test()"

I'll just make it optional with a kwarg exit=False by default.

josef-pkt:
kwarg sounds good

In an old version I had
`sudo $PYTHON -c "import statsmodels as sm; a=sm.test(); import sys; sys.exit((len(a.failures)+len(a.errors))>0)"`

e.g. https://github.com/statsmodels/statsmodels/pull/1082/files#diff-354f30a63fb0907d4ad57269548329e3L49

but adding an exit kwarg is simpler

josef-pkt:
looks good to me as far as I can tell

green enough, merging

@bashtage Thank you

PR #5123, opened 2018-09-06
CLN: Move results files to a location that is copied to install

bashtage:
Ensure results are copied to install


jbrockmendel:
@josef-pkt Travis fails are spurious; should merge so #4871 can move forward.

PR #5124, opened 2018-09-06
MAINT: fix generated double whitespace

jbrockmendel:
R2nparray.R and estmat2nparray.ado currently add an extra space before an `=`.  This removes that extra space in the generating files and all of the affected files.


jbrockmendel:
@josef-pkt I know this isnt something you care about, but it bothers others, is really easy to fix, and this fixes it at the source, so as to never come up again.

PR #5129, opened 2018-09-06
TST: avoid thousands of DeprecationWarnings in qsturng_

jbrockmendel:
https://travis-ci.org/statsmodels/statsmodels/jobs/425375334

CI jobs are failing because of zillions of warnings being issued.  This PR addresses:

```
/home/travis/build/statsmodels/statsmodels/statsmodels/stats/libqsturng/qsturng_.py:480: DeprecationWarning: In future, it will be an error for 'np.bool_' scalars to be interpreted as an index
  f += 1./(191.*(v,1e38)[np.isinf(v)])
```

### DIFF
--- a/statsmodels/stats/libqsturng/qsturng_.py
+++ b/statsmodels/stats/libqsturng/qsturng_.py
@@ -459,6 +459,7 @@ def _ptransform(p):
     """function for p-value abcissa transformation"""
     return -1. / (1. + 1.5 * _phi((1. + p)/2.))
 
+
 def _func(a, p, r, v):
     """
     calculates f-hat for the coefficients in a, probability p,
@@ -474,12 +475,17 @@ def _func(a, p, r, v):
     if r == 3:
         f += -0.002 / (1. + 12. * _phi(p)**2)
 
+        if np.isinf(v):
+            # Clip infinite v values
+            v = 1e38
+
         if v <= 4.364:
-            f += 1./517. - 1./(312.*(v,1e38)[np.isinf(v)])
+            f += 1./517. - 1./(312.*v)
         else:
-            f += 1./(191.*(v,1e38)[np.isinf(v)])
+            f += 1./(191.*v)
 
     return -f
+
 
 def _select_ps(p):
     # There are more generic ways of doing this but profiling
@@ -511,6 +517,7 @@ def _select_ps(p):
         return .500, .675, .750
     else:
         return .100, .500, .675
+
 
 def _interpolate_p(p, r, v):
     """


jbrockmendel:
Another failure coming from zillions of DeprecationWarnings in patsy importing from pandas:
```
/home/travis/miniconda/envs/statsmodels-test/lib/python3.4/site-packages/patsy/util.py:652: DeprecationWarning: pandas.core.common.is_categorical_dtype is deprecated. import from the public API: pandas.api.types.is_categorical_dtype instead
  return safe_is_pandas_categorical_dtype(data.dtype)
```

https://travis-ci.org/statsmodels/statsmodels/jobs/425375338

jbrockmendel:
Not clear where this one is coming from, looks like upstream somewhere:

```
/home/travis/miniconda/envs/statsmodels-test/lib/python3.7/site-packages/numpy/matrixlib/defmatrix.py:68: PendingDeprecationWarning: the matrix subclass is not the recommended way to represent matrices or deal with linear algebra (see https://docs.scipy.org/doc/numpy/user/numpy-for-matlab-users.html). Please adjust your code to use regular ndarray.
```
https://travis-ci.org/statsmodels/statsmodels/jobs/425375334

statsmodels/stats/libqsturng/qsturng_.py
```diff
@@ -474,13 +475,18 @@ def _func(a, p, r, v):
...
 
+        if np.isinf(v):
+            # Clip infinite v values
+            v = 1e38
+
         if v <= 4.364:
-            f += 1./517. - 1./(312.*(v,1e38)[np.isinf(v)])
+            f += 1./517. - 1./(312.*v)
         else:
-            f += 1./(191.*(v,1e38)[np.isinf(v)])
```
josef-pkt: that's not the correct fix
I think we need to convert numpy bool to python bool which should work for indexing a tuple
(or astype(int) should also work)

josef-pkt: correction: the current fix is correct just not literally the same
still LGTM

jbrockmendel: Glad we're on the same page.  I'll ping on de-facto green.

Couple others similar to this coming up.

statsmodels/stats/libqsturng/qsturng_.py
```diff
@@ -474,13 +475,18 @@ def _func(a, p, r, v):
     if r == 3:
         f += -0.002 / (1. + 12. * _phi(p)**2)
 
+        if np.isinf(v):
```
bashtage: v=min(v,1e38)?

statsmodels/stats/libqsturng/qsturng_.py
bashtage: The method used in the original code is an abomination to decent code.

There are lots more places to fix this bug in this library.  I think I got them all in mine

statsmodels/stats/libqsturng/qsturng_.py
bashtage: Or maybe a ternary v = v is not isinf(v) else 1e38

jbrockmendel: I tend to like the two-line version here because I can see in coverage reports which cases are hit.  Failing that I think the `v = min(v, 1e38)` would be my preference, which I think is what you used.

statsmodels/stats/libqsturng/qsturng_.py
jbrockmendel: > The method used in the original code is an abomination to decent code.

Seriously.

> There are lots more places to fix this bug in this library. I think I got them all in mine

Yah.  Since I just re-pushed, yours will finish CI first, so I'll rebase and just use your version for this file.

josef-pkt: > The method used in the original code is an abomination to decent code.

Nope, it's standard advertised python, that numpy doesn't want to follow anymore.
Python's bool behaves like an int, and indexing with bool was a "cute" trick before the conditional assignment was added to Python.

(https://github.com/rogerlew/qsturng-py doesn't seem to be maintained anymore, so I guess we maintain our code without foreign compatibility requirements.)

bashtage:
It is still an unreadable abomination of and if then. I wonder if Python
generates the correct ast of an if then assignment.

statsmodels/stats/libqsturng/qsturng_.py
jbrockmendel: > Nope, it's standard advertised python,

Really not worth arguing about, but consider:
- it was not immediately obvious to you that the version I implemented was functionally equivalent
- one person whose opinion you respect and one person who has two thumbs and strong opinions about the Star Wars Expanded Universe had a visceral reaction to this pattern.
- cute tricks and good code are overlapping but far-from-identical sets

josef-pkt:
> I wonder if Python generates the correct ast of an if then assignment.

I guess not, Python book is a 0-1 integer, i.e. it should correspond to a indexing `(a, b)[int(bool(x)]`, only numpy uses elementwise array indexing which defines different behavior.

another one: boolean d  as dummy variable 
```
>>> True * 5
5
>>> False * 5
0
```

> it was not immediately obvious to you that the version I implemented was functionally equivalent

I had no problems reading the old version, I misunderstood your new version.

`f += 1./(191.*(v,1e38)[np.isinf(v)])`
corresponds to
```
if not np.isinf(v):
    f += 1./(191. * v)
else:
    f += 1./(191. * 1e38)
```

i.e. we are not changing `v` which might have side effects later on.

(maybe I'm too old)

PR #5132, opened 2018-09-06
MAINT/TST: Fix/Silence DeprecationWarnings

jbrockmendel:
Each commit addresses a single source of these warnings.

Kept peripheral cleanup to a near minimum (the case that strayed farthest from this goal was that last commit, which touches regression._prediction and genmod._prediction).

It's conceivable that the way _penalties currently works is intentional, but it looks like `grad` was deprecated but internal usage was not updated to reflect that.  If this perception is wrong then this commit will have to be reverted and those warnings addressed separately.

Made an executive decision to stop passing through numpy's `RuntimeWarning`s when `tvalues` includes a division by zero.  On balance those tend to occur when other convergence-type problems occur, and an additional warning isn't all that helpful.

`patsy` change to make `design_info.builder` not need the `.builder` is effective as of patsy4.0, which is the required version according to setup.py


statsmodels/base/data.py
```diff
@@ -475,14 +474,16 @@ def _convert_endog_exog(self, endog, exog=None):
 
     @classmethod
     def _drop_nans(cls, x, nan_mask):
-        if hasattr(x, 'ix'):
+        if hasattr(x, 'loc'):
```
bashtage: cleaner to check for Series, DataFrame

statsmodels/base/data.py
```diff
@@ -475,14 +474,16 @@ def _convert_endog_exog(self, endog, exog=None):
...
+            # i.e. is pandas Series/DataFrame
             return x.loc[nan_mask]
         else:  # extra arguments could be plain ndarrays
             return super(PandasData, cls)._drop_nans(x, nan_mask)
 
     @classmethod
     def _drop_nans_2d(cls, x, nan_mask):
-        if hasattr(x, 'ix'):
+        if hasattr(x, 'loc'):
```
bashtage: same

statsmodels/base/model.py
```diff
@@ -1257,7 +1257,10 @@ def tvalues(self):
         """
         Return the t-statistic for a given parameter estimate.
         """
-        return self.params / self.bse
+        with np.errstate(divide="ignore"):
```
bashtage: Should be shown to end users.  Probably don't suppress here or in tests.

statsmodels/nonparametric/kernel_regression.py
```diff
@@ -767,9 +767,9 @@ def _compute_se_lambda(self, Y, X):
         but slows down computation significantly.
         """
         n = np.shape(Y)[0]
-        lam = np.empty(shape=(self.nres, ))
+        lam = np.empty(shape=(self.nres,))
         for i in range(self.nres):
-            ind = np.random.random_integers(0, n-1, size=(n,1))
+            ind = np.random.randint(0, n-1, size=(n, 1))
```
bashtage: Needs to be (0,n,size=(n,1)) -> not the deprecation warning saying +1. Former was closed.  This is open.

statsmodels/nonparametric/kernel_regression.py
```diff
@@ -793,12 +793,12 @@ def _compute_sig(self):
...
         # Calculate the restricted mean. See p. 372 in [8]
         M = KernelReg(Y, X, self.var_type, self.model.reg_type, self.bw,
-                      defaults = EstimatorSettings(efficient=False)).fit()[0]
+                      defaults=EstimatorSettings(efficient=False)).fit()[0]
         M = np.reshape(M, (n, 1))
         e = Y - M
         e = e - np.mean(e)  # recenter residuals
         for i in range(self.nboot):
-            ind = np.random.random_integers(0, n-1, size=(n,1))
+            ind = np.random.randint(0, n-1, size=(n, 1))
```
bashtage: Same, (0,n)

statsmodels/sandbox/nonparametric/kernel_extras.py
```diff
@@ -75,7 +75,7 @@ class TestFForm(object):
     References
     ----------
     See Racine, J.: "Consistent Significance Testing for Nonparametric
-    Regression" Journal of Business \& Economics Statistics.
+    Regression" Journal of Business \\& Economics Statistics.
```
bashtage: Why does this need any escaping?

statsmodels/sandbox/regression/penalized.py
```diff
@@ -59,8 +59,8 @@ class TheilGLS(GLS):
 
     .. math:: q = R \beta + v
```
bashtage: Better to make the docstring raw and leave single escapes than to mix and match escaped-escapes and non-escape sequences (e.g., \\ and \b)

statsmodels/sandbox/regression/penalized.py
```diff
@@ -59,8 +59,8 @@ class TheilGLS(GLS):
...
 
-    :math:`E(u) = 0`, :math:`cov(u) = \Sigma`
-    :math:`cov(u, v) = \Sigma_p`, with full rank.
+    :math:`E(u) = 0`, :math:`cov(u) = \\Sigma`
+    :math:`cov(u, v) = \\Sigma_p`, with full rank.
```
bashtage: raw docstrings are just better with LaTeX math

bashtage:
I have quite a few that you don't have. but we have overlap, so will have conflicts  Feel free to take them from mine into this and I'll close mine.

bashtage:
I should add the dead simplest way to avoid these warnings (temporarily) is to 

pip install pytest==3.7

which will pin to the last version that worked.

jbrockmendel:
> Only the massive set of new warnings should be suppressed. 0 div should still be allowed since these are really errors in test design.

Fair enough.

> so will have conflicts Feel free to take them from mine into this and I'll close mine.

I was thinking that you're more likely to get this past the maintainer in a timely manner than I am, so we should go the other way.

statsmodels/base/model.py
jbrockmendel: Will revert here; merits discussion elsewhere.

josef-pkt:
Can your make a PR with just this:
`pip install pytest==3.7`

I'm not sure whether there is not also a bug in pytest 3.8, and I would rather not change a lot of things right away, and let other packages figure this out first.
I'm a bit suspicious whether all this backslashes are really incorrect.

statsmodels/sandbox/nonparametric/kernel_extras.py
jbrockmendel: I'd be happy without; assumed previous author had a reason.  Will change.

statsmodels/sandbox/regression/penalized.py
jbrockmendel: Yah, wasn't super-happy with this mixed case thing.

jbrockmendel:
> I'm not sure whether there is not also a bug in pytest 3.8

What bug are you referring to?

> I'm a bit suspicious whether all this backslashes are really incorrect.

Referring to regexes or docstrings or both?

> and I would rather not change a lot of things 

Totally out of character.

bashtage and I have unfortunate timing when it comes to trying to be helpful.  I think the thing to do is merge #5135 when feasible and I'll rebase+repush with just the non-overlapping changes.

bashtage:
All of the back slashes I looked at were incorrect. Mostly missing raw
string flags.

josef-pkt:
what about the regexes like `\[` ?
If you switch back to older pytest, then we see whether the unit tests actually still pass, instead of trying to get results through the warnings avalanche.

(Even if everything is fine, then I don't want to see the same deprecation warnings a thousand times each time pandas or numpy changes something that affects patsy or statsmodels.)

jbrockmendel:
Rebased following #5135.

bashtage:
LGTM

jbrockmendel:
@josef-pkt should I bother rebasing or is that just going to clog the CI?

bashtage:
Should be rebased and merged.  Solves, among other issues, statsmodels calling code that statsmodels has deprecated.

jbrockmendel:
> Should be rebased and merged. Solves, among other issues, statsmodels calling code that statsmodels has deprecated.

I very much agree.  But my understanding is that my PRs are currently being actively ignored, so until that is over pushing will only serve to clog the CI.

jbrockmendel:
@josef-pkt this actually includes a non-trivial thing that does need your attention

PR #5139, opened 2018-09-07
MAINT: remove tsa/mlemodel.py

jbrockmendel:
Imports numdifftools (no longer a requirement) and just passes on failure, meaning the one class it defines will fail if ever used.

There is nothing in this class that isn't better accomplished by tsa.base.tsa_model.TimeSeriesModel


jbrockmendel:

Issue #5092, opened 2018-09-03
DEPR: Deprecations from prior versions


jbrockmendel:
There is a lot of code marked as deprecated or to be removed.  This issue is intended to keep track of them and improve the odds those deprecations are followed up on.

Short-run, ~four things are actionable:

a) things that were marked as to-be-removed in version 0.8 or 0.9 should _actually_ be removed in version 0.10.
b) things that were marked as removed "in the future" should be updated to be more specific.
c) decide which if any of these messages are fine being kept indefinitely, e.g. perhaps the "nobs is deprecated in favor of lags"

----

- [ ] `stats._lilliefors.lillifors` "will be removed in 0.9"
- [ ] `stats.diagnostic.acorr_breush_godfrey` "will be removed in 0.9"
- [ ] `stats.diagnostic.het_breushpagan` "will be removed in 0.9"
- [ ] `stats.diagnostic.StatTestMC` "delete when testing is finished", "this has been copied to tools/mctools.py, with improvements"
- [ ] `stats.multicomp.fdrcorrection_bak` "Old version, to be deleted"
- [ ] `stats.multicomp.maxzero` "Not used anymore"
- [ ] `stats.multicomp.maxzerodown` "Not used anymore"
- [ ] `stats.multicomp.rejectionline` "Not used anymore"
- [ ] `stats.runs.symmetry_bowker` "Deprecated, use stats.TableSymmetry instead" <--at least 3 years like this
- [ ] `stats.runs.mcnemar`  "Deprecated, use stats.TableSymmetry instead" <--at least 3 years like this
- [ ] `stats.runs.cochrans_q` "Deprecated, use stats.TableSymmetry instead" <--at least 3 years like this
- [ ] `GLMResults.resid_anscombe` "Anscombe residuals currently unscaled. In a future release they will be scaled"
- [ ] `genmod.families.family.Family` "Calling Family(..) with a link class as argument is deprecated"
- [ ] `base._penalties._check_wts` "`wts` method is deprecated. Use `weights` instead"
- [ ] `base._penalties.Penalty.grad` "grad method is deprecated. Use `deriv` instead"
- [ ] `base._penalties.CovariancePenalty.grad` "grad method is deprecated. Use `deriv` instead"
- [ ] `datasets.utils.as_numpy_dataset`: "load will return datasets containing pandas DataFrames and Series in the future"
[...]

PR #5170, opened 2018-09-10
DEPR: Remove items deprecated in previous versions

jbrockmendel:
Ref #5092.

Also usages of `np_matrix_rank`, which is no longer necessary.

### DIFF
--- a/examples/notebooks/regression_diagnostics.ipynb
+++ b/examples/notebooks/regression_diagnostics.ipynb
@@ -174,9 +174,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "name = ['Lagrange multiplier statistic', 'p-value', \n",
+    "name = ['Lagrange multiplier statistic', 'p-value',\n",
     "        'f-value', 'f p-value']\n",
-    "test = sms.het_breushpagan(results.resid, results.model.exog)\n",
+    "test = sms.het_breuschpagan(results.resid, results.model.exog)\n",
     "lzip(name, test)"
    ]
   },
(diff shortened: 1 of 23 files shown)


bashtage:
Also seems completely coherent and straightforward.

jbrockmendel:
@josef-pkt ping, CI fail is false positive

jbrockmendel:
@josef-pkt these should be removed before 0.10.0

jbrockmendel:
@josef-pkt appvryor queue is empty, good time to do some merging

jbrockmendel:
@ChadFulton I know you were interested in remove deprecated functions at one point

jbrockmendel:
@ChadFulton (last one for today) getting this up-to-date before 0.10.0 would be really nice, could use your help making it a reality (ok and also #5157)

ChadFulton:
In principle LGTM, two questions:

- Can you remind me why `np_matrix_rank` is no longer necessary? Is it because our minimum numpy has moved up?

- This actually implements some of the warned deprecations (`get_logdet`, `het_breushpagan`, `acorr_breush_godfrey`, `lillifors`).  I am happy to do this, since they are all marked as being removed in v0.9, and so removal in v0.10 seems like no big deal, but @josef-pkt please let me know if we should not do this for some reason.

jbrockmendel:
> Can you remind me why np_matrix_rank is no longer necessary? Is it because our minimum numpy has moved up?

Yes.

ChadFulton:
Thanks @jbrockmendel

PR #5176, opened 2018-09-11
 TST: Fix random generation issue

bashtage:
Set seed so that results are identical across runs

See 
https://travis-ci.org/bashtage/statsmodels/jobs/427055102

### DIFF
--- a/statsmodels/tsa/tests/test_arima.py
+++ b/statsmodels/tsa/tests/test_arima.py
@@ -2,7 +2,6 @@ from statsmodels.compat.python import lrange, BytesIO, cPickle
 
 import os
 import warnings
-from distutils.version import LooseVersion
 
 import numpy as np
 from numpy.testing import (assert_almost_equal, assert_, assert_allclose,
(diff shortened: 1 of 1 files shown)


jbrockmendel:
appveyor fail in test_corpssd, haven’t seen that one in a while.

bashtage:
@josef-pkt Causing failures 

https://travis-ci.org/bashtage/statsmodels/jobs/427700066

josef-pkt:
@bashtage  looks like another missing random seed

about appveyor: I have been checking it for several days now.
32-bit Python 2.7 is fragile, but has no consistent pattern that I could see. There are occasional failures or sequences of failing test runs, but which unit test fails, changes across runs (It might repeat a few times before changing again).
I don't really want to lower test precision in the hope that the fragile Win 32bit Python 2.7 environment will go away soon.

josef-pkt:
merged (I hadn't checked this PR yet, and didn't realize it fixes the random issue)

@bashtage Thank you

statsmodels/tsa/tests/test_arima.py
```diff
@@ -1916,6 +1914,7 @@ def test_arima_1123():
 
     dates = pd.date_range(start='1980', periods=nobs, freq='A')
 
+    np.random.seed(12345)
```
jbrockmendel: This is already done on 1906.  Is there a reason it is added here again?

bashtage:
No. Just overly aggressive.  These should ask us the reset random state
decorator ideally. In the sordid of one way.

PR #5179, opened 2018-09-11
BLD: Include pyx.in in sdist

bashtage:
Ensure cython files are included to avoid issue when Cython is availble
Defers choice of recompileing Cython or using C to Cython


bashtage:
setup would not work from a sdist without these files if Cython was already installed.  This is the simplest fix and defers any decisions about which source file to use -- pyx, pyx.in, or c -- to Cython.

bashtage:
Yes, MANIFEST.in can be cleaned up a bit

bashtage:
Closing since redundant with #5188.

Issue #5082, opened 2018-09-01
TST: Codecov False-Positives Caused by Noise in mixed_linear_model tests?


jbrockmendel:
codecov report for #5080 (which touches two lines in the sandbox, should have zero effect on coverage):

https://codecov.io/gh/statsmodels/statsmodels/compare/f324895b1ef9a15006649928e14263bb56cf0a28...d32903a1d75dd5f2a7f63929662f43f9984c79a3/changes

It is registering a decrease in coverage because of some lines not hit in `mixed_linear_model`.  I speculate that this is related to the flaky test described in #4414.  If correct, then this is _good_ news, since it is something we can fix and not a problem in codecov.

cc: @bashtage @kshedden

PR #5210, opened 2018-09-14
BLD: Ensure master is available when linting

bashtage:
Travis only clones a single branch by default. Manually clone master

### DIFF
--- a/lint.sh
+++ b/lint.sh
@@ -47,7 +47,10 @@ if [ "$LINT" == true ]; then
     fi
 
     # Tests any new python files
-    NEW_FILES=$(git diff master --name-status -u -- "*.py" | grep ^A | cut -c 3- | paste -sd " " -)
+    git fetch --unshallow --quiet
+    git config remote.origin.fetch "+refs/heads/*:refs/remotes/origin/*"
+    git fetch origin --quiet
+    NEW_FILES=$(git diff origin/master --name-status -u -- "*.py" | grep ^A | cut -c 3- | paste -sd " " -)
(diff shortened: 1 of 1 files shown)


jbrockmendel:
The coveralls false positive is annoying.  Is the config in the repo?  Or in josef’s account?

bashtage:
This doesn't produce a fp.  Instead the modifications I put in place to strictly lint new files aren't being run since master isn't available on travis.   This will allow new files to be strictly linted.  At least that is that I think is happening.

bashtage:
The coveralls is either due to a bug in the way they aggregate coverage
from multiple runs or a noisy test that sometime covers a branch and other
time doesn't.

jbrockmendel:
> The coveralls is either due to a bug in the way they aggregate coverage

Yah, I think it may be in mixed_linear_model (#5082).  I was thinking of the config because rgommers mentioned a few threads back that scipy sets the failure cutoff at 1% to avoid this issue.

bashtage:
This may not be needed.

josef-pkt:
coveralls: I had mixed up percent and fraction, the threshold was at 0.011 percent, now I changed it to 0.11 percent which should cover the -0.03 percent.

bashtage:
This is needed to get the new file lint to work correctly and should be correct now.

bashtage:
@josef-pkt small build issue about linting

josef-pkt:
merging, (I assume it's fine, I don't know enough to tell whether a partial, depth limited would be enough. looks fast enough) 

@bashtage thank you

PR #5211, opened 2018-09-14
MAINT: Import instead of copy/pasting utils

jbrockmendel:
Particularly the tools.tools functions are strictly better maintained in the non-sandbox versions.

Of the three remaining functions, `rank`, `fullrank`, `ECDF`, there are _almost_ identical non-sandbox versions, and I think it very likely that the differences are unintentional.  But until that is confirmed, leaving those in place.


jbrockmendel:
@josef-pkt I know this doesn’t interest you, but it’s green and one less thing to do later.

PR #5221, opened 2018-09-18
MAINT: remove never-needed callable and never-used compat functions

jbrockmendel:
Broken off from #5181.

The only things outside of compat.python that are changed are imports of `callable`, which are no longer necessary since 3.1,3.2 are not supported.  Some other unused imports in `datetools` are cleaned up, nothing major.

In compat.python, `callable` is removed, along with a handful of functions that are never even imported.


statsmodels/compat/python.py
```diff
@@ -20,7 +20,7 @@
     pickle = cPickle
     import urllib.request
     import urllib.parse
-    from urllib.request import HTTPError, urlretrieve, URLError
```
jbrockmendel: ATM urlretrieve is defined twice

bashtage:
LGTM.  callable should definitely go.  The others are likely overly aggressive copying from six.

PR #5223, opened 2018-09-18
MAINT: fully lint compat

jbrockmendel:
Broken off from #5181 (should be it for the time being)


bashtage:
Looks normal, and inconsequential.  LGTM

bashtage:
That one test is still random failing on Win32/mkl.

bashtage:
linting failure

PR #5229, opened 2018-09-19
MAINT: remove unnecessary usages of iterkeys

jbrockmendel:
For these purposes I'm defining "unnecessary" as those where the call to iterkeys is immediately wrapped in sorted/list/tuple, or where the newer notation `for key in d:` is preferred to `for key in iterkeys(d):`.  Turns out this constitutes all existing uses of `iterkeys`.


jbrockmendel:
OK.  Lucky there are plenty of other perfectly nice PRs ready to go.

jbrockmendel:
CI fail false-positive

Issue #5144, opened 2018-09-07
CI: Use CircleCI to take some of the pressure off Travis


jbrockmendel:

jbrockmendel:
Closed by #5617 using azure

PR #5238, opened 2018-09-20
TST: circle ci

thequackdaddy:
Here's an attempt to use circle ci. 

xref #5144
xref #5154

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,61 @@
+version: 2
+jobs:
+
+  python37:
+    docker:
+      - image: continuumio/miniconda3
+    environment:
+      PYTHON: 3.7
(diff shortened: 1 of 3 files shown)


bashtage:
Should add [skip appveyor] to any commit that only requires travis (osx, docs, etc) or circle ci.  I should also do the same.

jbrockmendel:
+1

thequackdaddy:
Yes, its a little slow. 

I was thinking if feasible though to offload some of the test cases from travis here. Maybe we only use 2-3 jobs? 

That might free up time so #5237 is not as annoying.

bashtage:
Going to close.  I think azure dominates since this isn't setup.

PR #5239, opened 2018-09-20
BLD: Add platform-specific skips

bashtage:
Add the ability to manually skip each platform using the syntax below anywhere in the commit message

[skip appveyor]
[appveyor skip]
[skip travis]
[travis skip]

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -106,6 +106,10 @@ notifications:
     on_success: always
 
 before_install:
+  # Skip if commit message contains [skip travis] or [travis skip]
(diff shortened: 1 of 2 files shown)


bashtage:
Can reduce pressure on CI when only working on docbuild issues or appveyor debugging.

bashtage:
Could add [skip circle] or [circle skip] for #5238

bashtage:
If you look at the builds, they are all skipped shortly after starting. Not as nice as [skip ci] which is still preferred, but appveyor has gotten strange about skip ci/skip appveyor (needs to be in commit subject now) and sometimes you want to skip one but not the other.

jbrockmendel:
Seems like an easy win.  Needs rebase

PR #5241, opened 2018-09-20
REF: Implement get_start_params_l1

jbrockmendel:
`fit_regularized` analogue to #5205.  Clarifies what parts of the namespace are needed where.

With a little more work, subclasses can override `_get_start_params_l1` and not need to override the `DiscreteModel` implementation of fit_regularized.

Also, the `_get_start_params_l1` implementations have some duplicated code that may be shareable.

Still needs docstrings.

### DIFF
--- a/statsmodels/discrete/count_model.py
+++ b/statsmodels/discrete/count_model.py
@@ -192,6 +192,31 @@ class GenericZeroInflated(CountModel):
                                       use_self=True, use_t=use_t, **cov_kwds)
         return result
 
+    def _get_start_params_l1(self, method='l1',
+            maxiter='defined_by_method', full_output=1, callback=None,
+            alpha=0, trim_mode='auto', auto_trim_tol=0.01, size_trim_tol=1e-4,
+            qc_tol=0.03, **kwargs):
+
+        alpha_p = alpha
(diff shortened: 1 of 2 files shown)


jbrockmendel:
@bashtage appveyor fail was in rdataset, was URLError (not HTTPError)

bashtage:
TBH I'm not that keen on this PR. It adds more lines than it removes and doesn't lead to any simplifications.  It also introduces some complex functions w/o docstrings. 

Can you take the next step with it and unify the fit call?

jbrockmendel:
> It also introduces some complex functions w/o docstrings.

Totally on board with amending this PR with docstrings.

> Can you take the next step with it and unify the fit call?

Sure.

jbrockmendel:
> Can you take the next step with it and unify the fit call?

Doing this now.  To actually de-duplicate the code, we need some variant of #5678.  For now I'm going to add just enough of that in for demonstration purposes.

jbrockmendel:
Actually, doing the de-duplication of fit_regularized becomes a lot less messy (and merge-conflicty) after #5678, #5235, and #5234 (the least trivial of the bunch)

PR #5250, opened 2018-09-23
BUG/TST: TSA models: _get_index_label_loc failed when using row labels.

ChadFulton:
The bug causes problems with `predict`. For example:

```python
import numpy as np
import pandas as pd
import statsmodels.api as sm

endog = pd.Series([1, 2, 3, 4], index=['a', 'b', 'c', 'd'])
mod = sm.tsa.SARIMAX(endog)
res = mod.filter([0., 1])
res.predict(end='c')
```

would yield a:

```
KeyError: 'The `end` argument could not be matched to a
           location related to the index of the data.'
```

### DIFF
--- a/statsmodels/tsa/base/tsa_model.py
+++ b/statsmodels/tsa/base/tsa_model.py
@@ -243,7 +243,9 @@ class TimeSeriesModel(base.LikelihoodModel):
         Parameters
         ----------
         key : label
-            The key for which to find the location
+            The key for which to find the location if the underlying index is
+            a DateIndex or a location if the underlying index is a RangeIndex
+            or an Int64Index.
         base_index : pd.Index, optional
             Optionally the base index to search. If None, the model's index is
             searched.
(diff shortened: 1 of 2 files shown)


statsmodels/tsa/base/tsa_model.py
```diff
@@ -408,7 +414,32 @@ def _get_index_label_loc(self, key, base_index=None):
...
+                        loc = np.argmax(loc)
+                    # Finally, Pandas may return an integer array of
+                    # locations that match the given value, for e.g.
+                    # pd.DatetimeIndex(['2001-02', '2001-01']).get_loc('2001')
+                    # (this appears to be slightly undocumented behavior, since
+                    # only int, slice, and mask are mentioned in docs for
+                    # pandas.Index.get_loc as of 0.23.4)
+                    else:
+                        loc = loc[0]
+                if not isinstance(loc, numbers.Integral):
```
jbrockmendel: Will this get e.g. np.int64(2)?  (And is that a relevant case?)

ChadFulton: It will get it, and it is a relevant case!

ChadFulton:
Green, will merge tomorrow unless there are concerns.

statsmodels/tsa/base/tsa_model.py
jbrockmendel: Good.  Here's a corner case that surprised me:

```
t = np.timedelta64(3, 'ns')
isinstance(t, numbers.Integral)
True
```

Not really worth worrying about here.

PR #5263, opened 2018-09-27
BLD: Only unshallow when required

bashtage:
Only run unshallow when required, which is usually on master


jbrockmendel:
This works as advertised for me.

LGTM

PR #5299, opened 2018-10-04
Dynamic regression

AliHeggie:
Implements a dynamic regression model of the form

$y_t = x_t \beta_t + \epsilon_t$

x and \beta can be vectors to support multiple exogenous regressors.

Each exogenous regressor in $\beta$ evolves according to a univariate unobserved components model.

At present $\beta$ may have an irregular, local level and linear trend component. The local level and trend can be deterministic or stochastic.

It is intended to extend the modelling options for $\beta$ to include the full suite of terms routinely found in unobserved components models including dummy seasonal terms, harmonic seasonal terms and autoregressive terms.

This is my fist contribution to an open source project so I don't expect the code to be acceptable in it's current state. In particular tests and documentation are required.

I'd appreciate any and all pointers for how to get this to a state where it is a valuable addition to Statsmodels.

### DIFF
--- a/statsmodels/tsa/statespace/dynamic_regression.py
+++ b/statsmodels/tsa/statespace/dynamic_regression.py
@@ -7,66 +7,101 @@ License: Simplified-BSD
 import pdb
 import numpy as np
 from scipy.signal import lfilter
+from scipy.linalg import block_diag
 import statsmodels.api as sm
 import pandas as pd
 from .tools import (constrain_stationary_univariate,
     unconstrain_stationary_univariate)
+import warnings
 
 # Construct the model
-class SDR(sm.tsa.statespace.MLEModel):
+class dynamic_regression(sm.tsa.statespace.MLEModel):
     """
     Univariate time-varying coefficient regression models
     
     A model in which the coefficients of the exogonous covariates are 
-    described by a univariate structural model
+    described by a univariate unobserved components model. At present
+    only local level and trend models are available for the coefficinets.
+    Ulitmately it is intended to add the full suite of components that
+    are standard in unobserved components models: 
 
     Parameters
     ----------
     exog : array_like or None, optional
         Exogenous variables.
-    freq_seasonal: list of dicts or None, optional.
-        Whether (and how) to model seasonal component(s) with trig. functions.
-        If specified, there is one dictionary for each frequency-domain
-        seasonal component.  Each dictionary must have the key, value pair for
-        'period' -- integer and may have a key, value pair for
-        'harmonics' -- integer. If 'harmonics' is not specified in any of the
-        dictionaries, it defaults to the floor of period/2.
-    arma_order : tuple
-        A tuple giving the order of the autoregressive and moving average
-        components of the error term respectively. (0,0) by default.
+    exog_models : dict or list of dicts, optional
+        Either a dict, a list containing a single dict, or a list of dict
+        s, one for each exogonous regressor.
+        Admisble keys are the supported component types: 'irregular', 
+        'local_level', and 'trend'. If a key is included in the dict and 
+        it's value is not False, then it is included in the model. If the 
+        corresponding value is "deterministic" then the state 
+        corresponding to that component will not have a disturbance term.
     dynamic_regression : bool
-        Whether or not the regression coefficient updates over time. If False
-        then the states corresponding to the regression coefficients have 0
-        variance.
+        Whether or not the regression coefficient updates over time. If 
+        False then the states corresponding to the regression 
+        coefficients have 0 variance.
     """
-    def __init__(self, endog, exog, freq_seasonal,arma_order=(0,0),
-        dynamic_regression=True):
-        self.arma_order = arma_order
-        self.freq_seasonal_periods = [d['period'] for d in freq_seasonal]
-        self.freq_seasonal_harmonics = [d.get(
-            'harmonics', int(np.floor(d['period'] / 2))) for
-            d in freq_seasonal]
-        self.ar_order = self.arma_order[0]
-        self.ma_order = self.arma_order[1]
-        self.r = max(self.ar_order,self.ma_order+1)
-        self.enforce_stationarity = True
-        self.enforce_invertibility = True
-        self.dynamic_regression = dynamic_regression
-
+    def __init__(self, endog, exog,exog_models={"local_level":True}):
         #Get k_exog from size of exog data
+        valid_components = set(['irregular','local_level','trend'])
         try:
             self.k_exog = exog.shape[1]
         except IndexError:
             exog = np.expand_dims(exog,axis=1)
             self.k_exog = exog.shape[1]
-        k_states = (self.r + self.k_exog +
-            sum(self.freq_seasonal_harmonics)*2)
-        k_state_cov = (self.dynamic_regression*self.k_exog * 3 + 1)
+        if isinstance(exog_models,list):
+            if len(exog_models)!=self.k_exog and len(exog_models)!=1:
+                raise ValueError('We should either recieve no exog_model'
+                    ' ,a single exog_model or a list of k_exog exog_models ')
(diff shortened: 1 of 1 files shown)


statsmodels/tsa/statespace/dynamic_regression.py
```diff
@@ -0,0 +1,321 @@
...
+from scipy.signal import lfilter
+from scipy.linalg import block_diag
+import statsmodels.api as sm
+import pandas as pd
+from .tools import (constrain_stationary_univariate,
+    unconstrain_stationary_univariate)
+import warnings
+
+# Construct the model
+class dynamic_regression(sm.tsa.statespace.MLEModel):
```
bashtage: Needs to follow standard Python name rules, e.g. DynamicRegression

statsmodels/tsa/statespace/dynamic_regression.py
```diff
@@ -0,0 +1,321 @@
...
+        Exogenous variables.
+    exog_models : dict or list of dicts, optional
+        Either a dict, a list containing a single dict, or a list of dict
+        s, one for each exogonous regressor.
+        Admisble keys are the supported component types: 'irregular', 
+        'local_level', and 'trend'. If a key is included in the dict and 
+        it's value is not False, then it is included in the model. If the 
+        corresponding value is "deterministic" then the state 
+        corresponding to that component will not have a disturbance term.
+    dynamic_regression : bool
```
bashtage: Should be just `dynamic` or possibly `update`

statsmodels/tsa/statespace/dynamic_regression.py
```diff
@@ -0,0 +1,321 @@
...
+        Transform unconstrained parameters used by the optimizer to constrained
+        parameters used in likelihood evaluation.
+
+        Used primarily to enforce stationarity of the autoregressive lag
+        polynomial, invertibility of the moving average lag polynomial, and
+        positive variance parameters.
+
+        Parameters
+        ----------
+        unconstrained : array_like
```
bashtage: array_like is for inputs that could be various classes, e.g. ndarray, Series, DataFrame.  Should be ndarray or (I don't like) array (since array is a python type, and confuses some introspection)

statsmodels/tsa/statespace/dynamic_regression.py
```diff
@@ -0,0 +1,321 @@
...
+        """
+        unconstrained = np.array(unconstrained, ndmin=1)
+        constrained = np.zeros(unconstrained.shape, unconstrained.dtype)
+        start = 0
+        # Transform the covariance params to be positive
+        constrained[start:] = unconstrained[start:]**2
+        return constrained
+    
+    def untransform_params(self, constrained):
+        """
```
bashtage: Are these useful publically or should they be private?

statsmodels/tsa/statespace/dynamic_regression.py
```diff
@@ -0,0 +1,321 @@
...
+        constrained = np.array(constrained, ndmin=1)
+        unconstrained = np.zeros(constrained.shape, constrained.dtype)
+        # Transform the AR parameters (phi) to be stationary
+        start = 0
+        # Untransform the covariance params which have been transformed
+        # to be positive
+        unconstrained[start:] = np.sqrt(constrained[start:])
+        return unconstrained
+
+def plot_dynamic_regression(results,which="smoothed",figsize=None,
```
bashtage: Need spaces.  Should run

`flake8 --isolated statsmodels/tsa/statespace/dynamic_regression.py`

and fix any issues.

statsmodels/tsa/statespace/dynamic_regression.py
ChadFulton: In this case, I think array_like is okay since we might actually have `pd.Series` (e.g. if you were to pass in the `params` from a previously estimated results object).

statsmodels/tsa/statespace/dynamic_regression.py
ChadFulton: `transform_params` and `untransform_params` are part of the typical state space class structure, so they're public (for better or worse)

statsmodels/tsa/statespace/dynamic_regression.py
```diff
@@ -0,0 +1,321 @@
...
+
+# Construct the model
+class dynamic_regression(sm.tsa.statespace.MLEModel):
+    """
+    Univariate time-varying coefficient regression models
+    
+    A model in which the coefficients of the exogonous covariates are 
+    described by a univariate unobserved components model. At present
+    only local level and trend models are available for the coefficinets.
+    Ulitmately it is intended to add the full suite of components that
```
ChadFulton: typo in "Ultimately"

statsmodels/tsa/statespace/dynamic_regression.py
```diff
@@ -0,0 +1,321 @@
...
+    are standard in unobserved components models: 
+
```
ChadFulton: Can you include the latex definition of the model here?

statsmodels/tsa/statespace/dynamic_regression.py
```diff
@@ -0,0 +1,321 @@
...
+    Parameters
+    ----------
+    exog : array_like or None, optional
+        Exogenous variables.
+    exog_models : dict or list of dicts, optional
+        Either a dict, a list containing a single dict, or a list of dict
+        s, one for each exogonous regressor.
+        Admisble keys are the supported component types: 'irregular',
```
ChadFulton: typo in "Admissible"

statsmodels/tsa/statespace/dynamic_regression.py
```diff
@@ -0,0 +1,321 @@
...
+        self.obs_intecept = np.zeros(1)
+        self.ssm['obs_intercept'] = self.obs_intecept
+        self.state_intercept = np.zeros((self.k_states,1))
+        self.ssm['state_intercept'] = self.state_intercept
+        #construct covariance matrices
+        self.obs_cov =  np.zeros((1,1))
+        self.ssm['obs_cov'] = self.obs_cov
+        self.state_cov = np.zeros((self.k_state_cov,self.k_state_cov))
+        self.ssm['state_cov'] = self.state_cov
+
```
ChadFulton: Would be great if you could implement `param_names`

ChadFulton:
I've added some specific comments, but there are some more general ones:

- We need to find a good reference for at least the basic TVP class of models
- As Kevin said, implementing unit tests are important - at the most basic level, these can be testing that the state space model matrices are set up correctly, but there should be a package somewhere (e.g. in R) that we can test results against too.

statsmodels/tsa/statespace/dynamic_regression.py
```diff
@@ -0,0 +1,321 @@
...
+    Parameters
+    ----------
+    exog : array_like or None, optional
+        Exogenous variables.
+    exog_models : dict or list of dicts, optional
+        Either a dict, a list containing a single dict, or a list of dict
+        s, one for each exogonous regressor.
+        Admisble keys are the supported component types: 'irregular', 
+        'local_level', and 'trend'. If a key is included in the dict and
```
ChadFulton: I think it should be `level` and `trend`, to match what `UnobservedComponents` has.

statsmodels/tsa/statespace/dynamic_regression.py
bashtage: I think I was looking at the output, which is always a ndarray. I take your point on the input.

ChadFulton: Good point - I should change that in the base class too.

AliHeggie:
Just pushed a bunch of commits to address some of the above.

The major outstanding issue is tests and documentation improvement I think.

According to the Codacy report there is a syntax problem on line 186 but I'm not sure what this is - seems like a perfectly valid f-string to me.

PR #5301, opened 2018-10-04
BUG: MICEData sets initial imputation incorrectly

### LINKED ISSUE #5254: BUG : MICEData creates new values after imputation
This imputation model is supposed to impute values from other(pre-existing) observations, as I know.

It does, however, create a value that did never exist in other observations. 

I used this imputation after applying LabelEnconder in Scikit-Learn. 

In this process, NaN values were replaced with 0(if NaN exists) and after they were encoded by LabelEncoder, I again replaced 0 with NaN values. And then I did imputation as uploaded here.

Any suggestions about this problem?

jtweeder:
Previous code only imputed value as the distance from the mean for the closest observed value, and not actually the observed value when MICEData is initialized.

closes #5254

### DIFF
--- a/statsmodels/imputation/mice.py
+++ b/statsmodels/imputation/mice.py
@@ -293,7 +293,7 @@ class MICEData(object):
             di = self.data[col] - self.data[col].mean()
             di = np.abs(di)
             ix = di.idxmin()
-            imp = di.loc[ix]
+            imp = self.data[col].loc[ix]
             self.data[col].fillna(imp, inplace=True)
 
     def _split_indices(self, vec):
--- a/statsmodels/imputation/tests/test_mice.py
+++ b/statsmodels/imputation/tests/test_mice.py
@@ -81,12 +81,16 @@ class TestMICEData(object):
                         np.concatenate((np.arange(10),
                                         np.arange(11, 30, 2),
                                         np.arange(30, 200))))
+        assert_equal([set(imp_data.data[col]) for col in imp_data.data],
+                     [set(df[col].dropna()) for col in df])
 
         for k in range(3):
             imp_data.update_all()
             assert_equal(imp_data.data.shape[0], nrow)
             assert_equal(imp_data.data.shape[1], ncol)
             assert_allclose(orig[mx], imp_data.data[mx])
+            assert_equal([set(imp_data.data[col]) for col in imp_data.data],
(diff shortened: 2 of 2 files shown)


jtweeder:
Failling this test in [statsmodels.imputation.test.test_mice.py](https://github.com/statsmodels/statsmodels/blob/master/statsmodels/imputation/tests/test_mice.py)

```Python
def test_combine(self):

    np.random.seed(3897)
    x1 = np.random.normal(size=300)
    x2 = np.random.normal(size=300)
    y = x1 + x2 + np.random.normal(size=300)
    x1[0:100] = np.nan
    x2[250:] = np.nan
    df = pd.DataFrame({"x1": x1, "x2": x2, "y": y})
    idata = mice.MICEData(df)
    mi = mice.MICE("y ~ x1 + x2", sm.OLS, idata, n_skip=20)
    result = mi.fit(10, 20)

    fmi = np.asarray([ 0.1920533 ,  0.1587287 ,  0.33174032])
    assert_allclose(result.frac_miss_info, fmi, atol=1e-5)

    params = np.asarray([-0.05397474,  0.97273307,  1.01652293])
    assert_allclose(result.params, params, atol=1e-5)

    tvalues = np.asarray([ -0.84781698,  15.10491582,  13.59998039])
    assert_allclose(result.tvalues, tvalues, atol=1e-5)
```
I have not confirmed yet by manually calculating, but I am guessing the numbers in hard coded arrarys are based on the MICEData code bug that assigns distance from the mean, and not actual mean of a column...

kshedden:
@jtweeder Thanks, this looks correct to me.

The currently failing test is a regression test, so you should change the hard-coded constants to whatever you are getting with the new code.

Would it be possible to write a test that confirms that the set of unique values in each column does not change when doing mice imputation?  That test would have caught this issue.

thequackdaddy:
Just some experimenting around shows that testpath=0.4.1 from conda-forge is the culprit for the docbuild failure.  

https://github.com/conda-forge/testpath-feedstock/issues/7

Thats fixed in version 0.4.2 which is coming soon to a travis build near you:

https://github.com/conda-forge/testpath-feedstock/pull/8

I'd suggest waiting a day a two and re-pushing once conda-forge is pushing the newer testpath.

kshedden:
Thanks @jtweeder, I am merging now

PR #5304, opened 2018-10-06
ENH: conditional logistic and Poisson regression

kshedden:


josef-pkt:
If it is just Logit, and eventually Poisson, and it doesn't use the GLM/family framework, then `discrete` would be the better location

jbrockmendel:
> then discrete would be the better location

Ha, I was just about to ask why this was in gemod too

josef-pkt:
needs rebase to get test results, last run failed with a pytest version comparison but that I worked around in master.

test runs were green before, except for linting/pep-8

kshedden:
I rebased but it is still failing.  I don't think it's a linting issue, but
maybe I am reading the output wrong.

josef-pkt:
The problems on travis look unrelated to this PR. AFAICS, this PR is green.

one travis 3.6 job has a problem with linking to numpy in the statespace extension. numpy and cython version look the same as in previous successfull runs.

The only difference I can see is that the run in this PR mentions downgrades to python 3.6.6 from python 3.6.7
https://travis-ci.org/statsmodels/statsmodels/jobs/461367637

(the python version sequence looks a bit complicated, python is several times upgraded and downgraded)

```
The following packages will be DOWNGRADED:
    cryptography:         2.4.1-py36h1ba5d50_0                 --> 2.3.1-py36hdffb7b8_0  conda-forge
    openssl:              1.1.1a-h7b6447c_0                    --> 1.0.2p-h470a237_1     conda-forge
    python:               3.6.7-h0371630_0                     --> 3.6.6-h5001a0f_3      conda-forge
```

the previous run only installed python 3.6.6
https://travis-ci.org/statsmodels/statsmodels/jobs/461067669

I didn't see anything else that looks different and we can hope that the build error goes away with another condaforge update.

PR #5308, opened 2018-10-06
REF: Collect covtype descriptions, de-duplicate normalization func

jbrockmendel:


bashtage:
Rebase and this should get in.

PR #5309, opened 2018-10-07
DOC: Add demos with plots to graphics

### LINKED ISSUE #834: graphics missing plots in documentation
several graphics functions don't have examples that create plots

jtweeder:
Here is a first draft to close #834 

I tried to use the existing samples that may have been in the docstring, but not producing a plot.  If an example was not present I tried to create a simple one that was easy to follow, and showed the intent of the function while following the format of other examples.  I am open to any suggestions you may have for better use cases in any of the plots.


bashtage:
New code files have to pass strict linting.  If you are going to assign a lambda, might as well write a local function.

PR #5310, opened 2018-10-08
add sigma in stats.DescrStatsW.ztest_mean

guofei9987:
add sigma in stats.DescrStatsW.ztest_mean  
now we can use `DescrStatsW.ztest_mean(value=0, sigma=1.0, alternative='two-sided')`  to do ztest  with the sigma(or std) we assign. 
when we use `DescrStatsW.ztest_mean(value=0, alternative='two-sided')` where we do not assign sigma, the function goes just the same as before-modified.

ps: I have compared the new ztest_mean to Excel z.test, and the result is the same.
pps: If we want to compare means and use ttest, we can not assign sigma. This is why sigma matters in ztest.


guofei9987:
Test cases need to be updated,and this pr needs some revision

PR #5318, opened 2018-10-10
REF: implement blas_dot

jbrockmendel:
@ChadFulton related #5283.


ChadFulton:
Looks neat! I'm pretty busy at the moment so having trouble finding time to review things.

Basically my only question is on the fused types.

- The thing I like about #5283 is that it's kind of isolated, so that if it is merged, we can see how we like fused types without impacting the core state space code. My main concern about fused types is adding another layer of complexity without making any particular improvement (except possibly to readability).

- On the other hand, this is a nice PR because moving the `?dot` code out allows handling the weirdness in one place, but it uses fused types in the code code, which is less desirable from my point of view for its first use.

My first choice would be a revised version of this PR that continues to use Tempita. I don't anticipate merging a fused types PR for v0.10. I know you have had negative views on watering down your PRs in the past, so either way, we can keep talking about this :)

jbrockmendel:
>  so that if it is merged, we can see how we like fused types without impacting the core state space code

Totally on board with doing things incrementally.

> a revised version of this PR that continues to use Tempita.

I'm not totally clear on what this would look like.  Can you expand on this?

> (except possibly to readability).

 I view this as central.  Put yourself in the position of a 3rd-party reader.  Statespace code is well-written, but _hard_.  A marginal readability improvement can lower the barrier to entry quite a bit.

> I don't anticipate merging a fused types PR for v0.10

Totally OK, as this is not at all trivial.  It would, however, be awesome if the many trivial PRs out there got merged.  The large majority of those are outside your purview though.

jbrockmendel:
@ChadFulton want to move forward with this for 0.11.0?

ChadFulton:
> @ChadFulton want to move forward with this for 0.11.0?

Yes, I think so, but slowly at least to start. How about a version of #5283 that implements only the fused types `copy` and then implements that only in `_copy_missing_*` and `_reorder_missing_*` functions (and to keep things straightforward, no pep8 / cleaning in the same PR). Then we can merge that, and see how things go? How does that sound?

jbrockmendel:
I'll start by reverting the lint fixes to 5283 and see how close that gets us there.

Really, I think this is likely to be lower-handing fruit.

ChadFulton:
I want to leave this PR here for now, to see how #5283 works out.

PR #5322, opened 2018-10-12
DOC: Move magic

thequackdaddy:
I think we can revert #5306 now. However, the magic in the jupyter notebooks need to be moved to a separate cell because that now counts against `from __future__` being at the "top" of a file.

### DIFF
--- a/examples/notebooks/discrete_choice_example.ipynb
+++ b/examples/notebooks/discrete_choice_example.ipynb
@@ -25,12 +25,19 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {
-    "collapsed": false
+    "collapsed": true
    },
    "outputs": [],
    "source": [
-    "%matplotlib inline\n",
-    "\n",
+    "%matplotlib inline"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
     "from __future__ import print_function\n",
     "import numpy as np\n",
     "import pandas as pd\n",
@@ -43,9 +50,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "collapsed": false
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "print(sm.datasets.fair.SOURCE)"
@@ -54,9 +59,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "collapsed": false
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "print( sm.datasets.fair.NOTE)"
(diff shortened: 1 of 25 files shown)


thequackdaddy:
Hmm... I think either sphinx or ipython 7.0.1 is breaking the docbuild. I'll take a closer look.

thequackdaddy:
xref https://github.com/ipython/ipython/issues/11394

examples/notebooks/discrete_choice_example.ipynb
```diff
@@ -99,9 +96,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "collapsed": false
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "affair_mod = logit(\"affair ~ occupation + educ + occupation_husb\" \n",
```
jbrockmendel: Can you trim trailing whitespace while in the neighborhood

thequackdaddy:
Sorry I’ll take another look in a bit. Hopefully by this weekend.

examples/notebooks/discrete_choice_example.ipynb
thequackdaddy: Thanks for pointing out. Fixed.

thequackdaddy:
This is still broken. ipython 7.2 still isn't compatible. I'm fighting with it. 

Latest issue is here...

https://github.com/ipython/ipython/issues/11362

thequackdaddy:
xref ipython/ipython#11362

thequackdaddy:
Unless there's a rush, I'm going to hold off looking at this for a week or so. 
1. The big difference is that IPython is not so smart about knowing when to break lines. I submitted a PR with IPython to make it a little smarter, but my gut is that is a pretty radical change and I don't know if they are ready for it just yet. I don't want to break more people's docs. 

2. There is a change in nbconvert/tornado/notebook that hasn't made its way yet to condo-forge/label/gcc7 yet. https://github.com/jupyter/notebook/issues/4311#issuecomment-470110129 Once that is updated, I'll try again.

thequackdaddy:
Well since we pinned ipython a few major updates happened so, there's a lot to be fixed here...

* Sphinx is now 2.0. There's also some deprecation error that I'm not sure if it applies to us. 
* ipython is up to 7.5 (or something like that)
* The xenial Ubuntu seems to work a little better.

I also probably have a few commits here that I'll squash when I was going back and forth between different settings. Most of the commits will stay, but I'll get rid of a few revisions in the next day or so. Let me know if anyone has thoughts/improvements/objections.

thequackdaddy:
Fail is #5457 . 

Otherwise, I think this is good. Thoughts or should we merge?

thequackdaddy:
merging then

PR #5352, opened 2018-10-28
CLN: Fix F901 and E306 mixups

jbrockmendel:
Ref #5325

### DIFF
--- a/statsmodels/stats/_lilliefors.py
+++ b/statsmodels/stats/_lilliefors.py
@@ -166,6 +166,7 @@ def get_lilliefors_table(dist='norm'):
 
 
         # also build a table for larger sample sizes
+
         def f(n):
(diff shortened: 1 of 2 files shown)


jbrockmendel:
@ChadFulton raising NotImplemented is egregious enough that it should actually be fixed.

ChadFulton:
Thanks @jbrockmendel

statsmodels/stats/_lilliefors.py
```diff
@@ -166,6 +166,7 @@ def get_lilliefors_table(dist='norm'):
 
 
         # also build a table for larger sample sizes
+
```
thequackdaddy: @jbrockmendel I think we actually should've removed lines here. flake8 can be confusing/finicky

thequackdaddy: I made a quick fix is #5335 ... let's see if I know what I'm doing.

PR #5353, opened 2018-10-28
CLN: W605 fixups in vector_ar

jbrockmendel:
There are a bunch more outside vector_ar, but no point doing it now since this is likely to be ignored for weeks on end.

Ref #5335.


jbrockmendel:
@josef-pkt ping, can we get these things taken care of in a timely manner?

jbrockmendel:
@josef-pkt ping.  Baby steps.

ChadFulton:
I see - so the various regex work for now because the various `\<character>` combinations actually used don't happen to be Python escape sequences, so in these specific cases adding the "r" flag doesn't make a different, but the version without the "r" is fragile in case someone tried to use a regex special character that collided with a Python escape sequence?

LGTM, will merge tonight if no concerns.

ChadFulton:
Thanks @jbrockmendel!

PR #5435, opened 2018-12-23
ENH Score test enhancements for GEE

kshedden:
Allow the submodel to be provided as a fitted GEEResults object, instead of via a system of linear constraints.

### DIFF
--- a/statsmodels/genmod/generalized_estimating_equations.py
+++ b/statsmodels/genmod/generalized_estimating_equations.py
@@ -734,7 +734,7 @@ class GEE(base.Model):
             return [np.array(array[self.group_indices[k], :])
                     for k in self.group_labels]
 
-    def compare_score_test(self, submodel, tol=1e-6):
+    def compare_score_test(self, submodel):
         """
         Perform a score test for the given submodel against this model.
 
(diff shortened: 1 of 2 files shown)


statsmodels/genmod/generalized_estimating_equations.py
```diff
@@ -728,6 +733,137 @@ def cluster_list(self, array):
             return [np.array(array[self.group_indices[k], :])
                     for k in self.group_labels]
 
+    def score_test(self, submodel, tol=1e-6):
```
josef-pkt: maybe rename to `compare_score_test`

I haven't looked at any details yet (still recovering from moving to a new house), but we should get consistent signature and interpretation for score_test.

I'm not sure whether `score_test` should be attached to the restricted model, the unrestricted model or both, depending on the options. (eg. I have an option for variable addition when the restricted model is available.)

for LR test we use `compare_lr_test` that takes a nested model.

josef-pkt:
IIUC, then the `ii = _score_test_submodel(self, submodel.model, tol)` only works for "zeros constraint", i.e. dropping/adding variables but not for general linear restrictions.

Kevin added a subspace check to OLS/RegressionModel compare_lm_test.
(In my version of the score_tests I don't check and assume that the user provides a nested model.)

kshedden:
I was thinking about generalizing this so that it would work as long as the
columnspace of the submodel is nested in the columnspace of the parent
model.  I will look into that.

Issue #5438, opened 2018-12-28
Exponential smoothing with seasonality initial Level value is weird


aelavender:
For Exponential Smoothing with seasonality, the initial Level (if not provided by the user) is set as follows:
`y[np.arange(self.nobs) % m == 0].mean()`
https://github.com/statsmodels/statsmodels/blob/da9d7e9431095025d6bd66d25cb87d46220769fe/statsmodels/tsa/holtwinters.py#L769

In other words, take the mean of all the points corresponding to the first season.  

This is an unusual way to initialize, and produces strange results for time series with a strongly non-zero slope: you'll start with a Level close to the mean of the series (which might not be anywhere close to the level at the start of the series).  Then you'll have huge initial seasonality to compensate.  The result is a forecast that looks fine in total but whose components are weird.  Ideally, the mean of the seasonality time series should be zero (for addititive seasonality) or one (for multiplicative), but this method results in values far from that.

Hyndman (whose text is mentioned in the documentation) [provides some suggestions](https://robjhyndman.com/hyndsight/hw-initialization/).  His simplest method is to take the mean of the first period (not the first season of each period, as is currently implemented above).  

[NIST's method](https://web.archive.org/web/20180312182644/http://www.itl.nist.gov:80/div898/handbook/pmc/section4/pmc435.htm) is even simpler: just use y[0], identical to the non-seasonal version.  (They don't explicitly mention it, but the example on the next page shows that result.)

I think we should use Hyndman's 1998 method, and change L769 to be
```
l0 = y[:m].mean() if l0 is None else l0
```

I can put up a PR if you wish.  

Thanks!

ChadFulton:
Thanks for this! I completely agree, and a PR would be very welcome.

aelavender:
Here's an example.  The source time series is generated from a random walk with drift, with weekly seasonality overlaid.  I'm using additive trend and seasonality.  

![image](https://user-images.githubusercontent.com/15818906/50498538-b7b1d300-09f6-11e9-9080-d607079e4aee.png)

Note how, while the fitted data looks good, the trend is much higher than the data, and the seasonality is very negative to compensate.

aelavender:
Thanks for the quick response, @ChadFulton!  I'll put up a PR within a day or so.

bashtage:
[...]

PR #5439, opened 2018-12-30
ENH: Holt Winters - set initial level to mean of first period

aelavender:
See description in https://github.com/statsmodels/statsmodels/issues/5438

### DIFF
--- a/statsmodels/tsa/holtwinters.py
+++ b/statsmodels/tsa/holtwinters.py
@@ -766,7 +766,7 @@ class ExponentialSmoothing(TimeSeriesModel):
         l0 = self._l0
         b0 = self._b0
         if seasoning:
-            l0 = y[np.arange(self.nobs) % m == 0].mean() if l0 is None else l0
+            l0 = y[:m].mean() if l0 is None else l0
             if b0 is None and trending:
                 lead, lag = y[m:m + m], y[:m]
                 if trend == 'mul':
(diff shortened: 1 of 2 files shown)


aelavender:
Hmm... seems like there's some non-determinism in the relevant unit test.  After 017ece2 I got this failure: https://travis-ci.org/statsmodels/statsmodels/jobs/473466369, which was unexpected -- I believe the change here should not affect the forecast, just the individual components.  Here's the run after changing the unit test: https://travis-ci.org/statsmodels/statsmodels/jobs/473487877 -- the forecast is back to what it was originally.  So I'm confused.

aelavender:
Hi @ChadFulton, thanks for taking a look.

Re: normalising initial seasonality.  This should already be taken care of: if we are setting `l0` equal to `mean(y[:m]`, then `y[:m] / l0` will have a mean of 1 and `y[:m] - l0` will have a mean of 0 already.  

Re: enforcing the constraint at every update step.  Doing so is not part of the standard Holt-Winters process, but has been proposed as a modification.  See, for example, [this paper](https://pubsonline.informs.org/doi/pdf/10.1287/opre.34.1.174), which I still need to read through.  

Re: fitting.  Yes, I agree that we should be optimising `m-1` seasonals instead of `m`.  This was not in my direct use case (I'm running my own optimiser) so I overlooked it.  I'll work on it when I get the chance.

bashtage:
This has been superseded by other improvements to initialization.

PR #5440, opened 2019-01-02
ENH: Use GLM to get starting values for GEE

kshedden:
This resolves many of the convergence problems I have seen people encounter with GEE.  Making it the default will avoid a lot of frustration.  To get back the old behavior, just pass in `start_params = np.zeros(...)`.


statsmodels/genmod/generalized_estimating_equations.py
```diff
@@ -1086,8 +1087,10 @@ def predict(self, params, exog=None, offset=None,
 
     def _starting_params(self):
 
-        # TODO: use GLM to get Poisson starting values
-        return np.zeros(self.exog.shape[1])
+        model = GLM(self.endog, self.exog, family=self.family,
+                       offset=self._offset_exposure, weights=self.weights)
```
thequackdaddy: Haven't verified this, but I don't think `weights=self.weights` does anything here.

Would it make sense to use `freq_weights` or `var_weights` instead? Using either will get you the same `params`. `cos_params` would be different, but I don't think you need that.

kshedden: Thanks, we should probably unify the notation so that GEE also uses freq_weights and var_weights.  But for now I will just use freq_weights because I think that is what we have in GEE.

PR #5448, opened 2019-01-03
BUG: Remove $$ in as_latex in summary and summary2

khs:
$$ breaks some implementations of LaTeX and is always meaningless.


bashtage:
Would be happy to merge this if you could write a test. Something that would produce a $$ in the old version, and then make sure that it does not after your patch.

bashtage:
This has been fixed in master in 98f9032220bb563c5a836cbd8d8e996fe6102c44

PR #5501, opened 2019-02-15
BUG: Correct error in VAR ACF

bashtage:
VAR ACF misses a transpose and produces incorrect values when number
of values if larger than model order


ChadFulton:
Thanks & LGTM, sorry about the delay.

PR #5510, opened 2019-02-21
ENH Conditional multinomial logit

kshedden:
Equivalent to femlogit in Stata


kshedden:
#5305  impacts this (can't call cov_params)

PR #5515, opened 2019-02-28
DOC: AR(2) example corrected

vriesa:
The example for the AR(2) model has a mistake. Inserting the state space model into the AR(2) equation :math: `y_t + \phi_1 y_{t-1} + \phi_2 y_{t-2}` must yield :math:`\epsilon_t`. This does not work with the previous version. Cf. Durbin and Koopman (2012), pp. 54.

### DIFF
--- a/docs/source/statespace.rst
+++ b/docs/source/statespace.rst
@@ -70,11 +70,11 @@ This can be put into state space form in the following way:
 
    y_t & = \begin{bmatrix} 1 & 0 \end{bmatrix} \alpha_t \\
    \alpha_t & = \begin{bmatrix}
-      \phi_1 & \phi_2 \\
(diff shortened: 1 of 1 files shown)


ChadFulton:
There are multiple ways to put an AR(p) model in state space form. The form that Durbin and Koopman show is sometimes attributed to Andrew Harvey, while the form that we show in the documentation is sometimes attributed to Jim Hamilton, and it is available in his Time Series Analysis book.

The clarification about the definition of eta_t is a good one, though.

ChadFulton:
> The clarification about the definition of eta_t is a good one, though.

Although in your edit, the timing convention is not right - it should be eta_t = \epsilon_t

Issue #3670, opened 2017-05-16
fit_regularized().summary() shows as None


Iceman121:
Hi,
I am running the following regression code:

```
Kara = sm.OLS(train.ix[:,0], train.ix[:, col_for_coeff]).fit_regularized(alpha=0.2, L1_wt=0.5)
print(Kara.summary())
```

The output is shown as 

```
None
```

There was hardly any incidents on Stackoverflow, except one small mention in the comments section on [this question.](http://stackoverflow.com/questions/40072870/statistical-summary-table-in-sklearn-linear-model-ridge) Any help?

System info:
OS: Windows 8.1
Python: 3.5.3
Statsmodels: 0.8.0

Iceman121:
That's what I used as a workaround, but I figured people would like to know that this is happening, and that I am not the only one facing this issue.

josef-pkt:
related #3626 There is currently no cov_params and inference for elastic net penalization.

josef-pkt:
Thanks for the stackoverflow link.
To the answer: using the standard cov_params, `normalized_cov_params` ignores the penalization (and I'm not sure it uses the scale estimate, but maybe it does that.)

Focanocio:
Hi guys. Sorry, It was not very clear to me if there is a solution - and could not find anything about Kara.params. Any way to find the parameters of the regularized regression?

Iceman121:
'Kara' was just the name I gave to my object. Basically, if you do ```sm.OLS().fit_regularized()```, the object has an attribute called ```params```.

You can call it in the following way:
```
supercool_godawesome_model = sm.OLS(exog, endog).fit_regularized(alpha=0.2, L1_wt=0.5)
regularized_regression_parameters = supercool_godawesome_model.params
print(regularized_regression_parameters)
```

Does that help?

Focanocio:
Dang! I have not checked that the name of your object was Kara, sorry for that. Now I get it.

I was just wondering if there was a way to get ohter goodness-of-fit parameters, because I wanted to test backward elimination, but I guess I will have to resort to conventional OLS without regularization to do so =/

Thank you for the help!

owlas:
Thanks @Chroner2 for the temporary fix, I'm also using `.params` for now. 

However, there is still a lot of missing information. Do we know why?

kshedden:
Regularization is a work in progress, not just in terms of our
implementation, but also in terms of methods that are available.  For
[...]

PR #5521, opened 2019-03-04
ENH: Tweedie log-likelihood (+ridge regression by gradient for all GLM)

kshedden:
The quasi-likelihood allows fit_regularized to be used with Tweedie family GLM's.

Also bypasses coordinate-descent when calling fit_regularized on a GLM with L1_wt=0.  This is ridge regression and can be fit more efficiently and accurately using gradients optimization.

#3670

### DIFF
--- a/statsmodels/genmod/families/family.py
+++ b/statsmodels/genmod/families/family.py
@@ -202,7 +202,7 @@ class Family(object):
 
         .. math::
            resid\_dev_i = sign(y_i-\mu_i) \sqrt{D_i}
-           
+
         D_i is calculated from the _resid_dev method in each family.
         Distribution-specific documentation of the calculation is available
         there.
@@ -1166,7 +1166,7 @@ class NegativeBinomial(Family):
     Parameterization for :math:`y=0, 1, 2, \ldots` is
 
     .. math::
-      
+
        f(y) = \frac{\Gamma(y+\frac{1}{\alpha})}{y!\Gamma(\frac{1}{\alpha})}
               \left(\frac{1}{1+\alpha\mu}\right)^{\frac{1}{\alpha}}
               \left(\frac{\alpha\mu}{1+\alpha\mu}\right)^y
@@ -1329,6 +1329,9 @@ class Tweedie(Family):
         See statsmodels.genmod.families.links for more information.
     var_power : float, optional
         The variance power. The default is 1.
+    eql : bool
+        If True, the Extended Quasi-Likelihood is used, else the
+        likelihood is used (however the latter is not implemented).
 
     Attributes
     ----------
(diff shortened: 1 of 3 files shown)


statsmodels/genmod/families/family.py
```diff
@@ -1451,10 +1455,28 @@ def loglike_obs(self, endog, mu, var_weights=1., scale=1.):
...
+        if self.eql == False:
+            # We have not yet implemented the actual likelihood
+            return np.nan
+
+        p = self.var_power
+        llf = np.log(2 * np.pi * scale) + p * np.log(endog) - np.log(var_weights)
+        llf *= -1 / 2
+
+        u = endog ** (2 - p) - (2 - p) * endog * mu ** (1 - p) + (1 - p) * mu ** (2 - p)
+        u *= var_weights / (scale * (1 - p)**2)
```
LudvicLaberge: When testing with sas's formula I realize they actually have a typo it is (1-p)*(2-p) not (1-p)**2

kshedden: Thanks, I will fix it.  It you have a separate PR with better tests we can go with yours.

thequackdaddy: So I actually liked a paper by Kaas when I was studying this... 

https://lstat.kuleuven.be/research/seminars_events/files/3afmd/Kaas.PDF

Page 8 has his formula for EQL (formula 4). The deviance on page 7 (formula 3) should pretty much be our `_resid_dev` method. You could call the existing method instead of re-writing it. You just have to add on the `var_weights` piece and the 2 * pi * scale adjustment.

You might also want to note in the doctoring that I think this EQL function is only valid when `1 < p < 2`.

statsmodels/genmod/families/family.py
```diff
@@ -1451,10 +1455,28 @@ def loglike_obs(self, endog, mu, var_weights=1., scale=1.):
...
+        If eql is True, the Extended Quasi-Likelihood is used.  At present,
+        this method returns NaN if eql is False.  When the actual likelihood
+        is implemented, it will be accessible by setting eql to False.
+
+        References
+        ----------
+        http://support.sas.com/documentation/cdl/en/stathpug/67524/HTML/default/viewer.htm#stathpug_hpgenselect_details16.htm
         """
-        return np.nan
+        if self.eql == False:
```
thequackdaddy: Also has this been PEP8'ed? I recall that its preferred just to do a `if not self.eql:` but I have no clue what the rule is... But I feel like I've seen it somewhere...

statsmodels/genmod/families/family.py
LudvicLaberge: Yes exactly, only valid between 1 < p < 2 because at 1 you would use poisson and at 2 gamma.

LudvicLaberge:
@kshedden I've been using your modified version and really like the new tweedie regularized fit! I was wondering, in general, how often are pull requests released to master on this repo and also, secifically, do we have an idea of when this PR will be released? Thank you for the improvements!

kshedden:
@LudvicLaberge Thanks for the feedback.  We are in a bit of limbo now with broken master.  But I will merge anyway in a week or so if things don't clear up.

thequackdaddy:
@kshedden Looking at the test suite results, it looks like your new test is failing on the python 2.7 travis-ci instances. Want to take a look at that?

Issue #5546, opened 2019-03-26
Convergence problem in documentation for MixLM


el-hult:
I'm trying to learn mixed linear models in statmodels. 
Was checking out this documentation that should be generated on stable build. In older versions, there was no problems, but now it raises convergence problems in some estimations. 

http://www.statsmodels.org/stable/examples/notebooks/generated/mixed_lm_example.html

I reproduced it with the code:
```python
import statsmodels.api as sm
data = sm.datasets.get_rdataset('dietox','geepack').data
model = smf.mixedlm('Weight ~ Time', data, groups=data["Pig"]) # Random Intercept
model = sm.MixedLM.from_formula("Weight ~ Time", data, re_formula="Time", groups=data["Pig"]) # Random time-slope
result = model.fit()
```

I suggest the documentation should be flagged as deprecated. Possibly there is also some bug underneath the hood somewhere...

kshedden:
I can follow up with more detail later, but here is a quick response to your question.  I think it is helpful to discuss this by comparing to R.

I used the Python code below, which is a lightly modified version of what you posted above:

```
import statsmodels.api as sm                                                                                                    
import statsmodels.formula.api as smf                                                                                           
data = sm.datasets.get_rdataset('dietox','geepack').data                                                                        
data.to_csv("dietox.csv")   # for comparison to R                                                                                                    
model1 = smf.mixedlm('Weight ~ Time', data, groups=data["Pig"]) # Random Intercept                                              
result1 = model1.fit()                                                                                                          
                                                                                                                                
model2 = sm.MixedLM.from_formula("Weight ~ Time", data, re_formula="Time", groups=data["Pig"]) # Random time-slope              
result2 = model2.fit()  
```

Here is the R code that I used for comparison:

```
library(lme4)                                                                                                                                          
da = read.csv("dietox.csv")                                                                                                     
                                                                                                                                
model1 = lmer(Weight ~ Time + (1 | Pig), data=da)                                                                                                                                               
model2 = lmer(Weight ~ Time + (Time | Pig), data=da)   
```

For the first model, in statsmodels we get:

```
In [14]: result1.summary()                                                                                                      
Out[14]: 
<class 'statsmodels.iolib.summary2.Summary'>
"""
         Mixed Linear Model Regression Results
========================================================
Model:            MixedLM Dependent Variable: Weight    
No. Observations: 861     Method:             REML      
No. Groups:       72      Scale:              11.3669   
[...]

PR #5551, opened 2019-04-01
ENH: extend mixedlm optimizer attempts

kshedden:
Try cg after bfgs and lbfgs fail

#5546


bashtage:
Passes travis, good enough for a simple one liner.

PR #5553, opened 2019-04-02
Fixes issue #2129: Fail more gracefully on bad exog for ARMAX

al1190:
Fixes issue #2129 

Hello, this is my first time contributing to an open source project, so I apologize in advance if this pull request is not up to standards.

I have been trying to work on a solution (with the help of @Juanlu001) to better explain the error in the following [example](https://stackoverflow.com/questions/27057445/how-to-fit-an-armax-model-using-statsmodels/). The statsmodels version I am using is: 'v0.10.0.dev0+1066.gdfc0b24c4'

I added an extra-condition(starting from line 408) that checks whether there is at least a constant in exog, in which case, the code raises a (hopefully) clearer error. I think this is the best place where to raise the error (and also the least invasive), because it is the part of the code that adds the trend to exog and as it is it adds one, whether exog has already a constant or not, which later leads to the current error as in the above example.

I would be grateful for any feedback you can give me on how to improve this solution.

### DIFF
--- a/statsmodels/tsa/arima_model.py
+++ b/statsmodels/tsa/arima_model.py
@@ -912,7 +912,7 @@ class ARMA(tsbase.TimeSeriesModel):
 
         # (re)set trend and handle exogenous variables
         # always pass original exog
-        k_trend, exog = _make_arma_exog(endog, self.exog, trend)
+        k_trend, exog = _make_arma_exog(endog, self.data.orig_exog, trend)
 
         # Check has something to estimate
         if k_ar == 0 and k_ma == 0 and k_trend == 0 and k_exog == 0:
--- a/statsmodels/tsa/arima_model.py
+++ b/statsmodels/tsa/arima_model.py
@@ -405,7 +405,12 @@ def _make_arma_exog(endog, exog, trend):
     if exog is None and trend == 'c':   # constant only
         exog = np.ones((len(endog), 1))
     elif exog is not None and trend == 'c':  # constant plus exogenous
-        exog = add_trend(exog, trend='c', prepend=True)
+        if len(exog.var(axis=0)) > np.count_nonzero(exog.var(axis=0)): # if true, it means there is at least a constant
+            # column in the exog.
+            raise ValueError("Exog should not contain a constant or trend. These should be added in the fit method."
+                             "See: https://www.statsmodels.org/dev/generated/statsmodels.tsa.arima_model.ARMA.html")
+        else:
+            exog = add_trend(exog, trend='c', prepend=True, has_constant='skip')
     elif exog is not None and trend == 'nc':
         # make sure it's not holding constant from last run
         if exog.var() == 0:
--- a/statsmodels/tsa/tests/test_arima.py
+++ b/statsmodels/tsa/tests/test_arima.py
@@ -2432,3 +2432,13 @@ def test_endog_int():
     resf = ARIMA(yf.cumsum(), order=(1, 1, 1)).fit(disp=0)
     assert_allclose(res.params, resf.params, rtol=1e-6, atol=1e-5)
     assert_allclose(res.bse, resf.bse, rtol=1e-6, atol=1e-5)
+
+
+def test_constant_exog_raises_proper_error():
+    # issue 2129
+    with pytest.raises(ValueError) as excinfo:
+        np.random.seed(13)
+        y = np.random.randn(100)
+        u = np.ones((100, 2))
+        ARMA(y, order=(1, 1), exog=u).fit()
+    assert "Exog should not contain a constant or trend" in str(excinfo.value)
--- a/statsmodels/tsa/tests/test_arima.py
+++ b/statsmodels/tsa/tests/test_arima.py
@@ -2327,7 +2327,7 @@ def test_arima_exog_predict():
         res_002.forecast(steps=h, exog=np.empty(20))
 
 
-def test_arima_fit_mutliple_calls():
+def test_arima_fit_multiple_calls():
     y = [-1214.360173, -1848.209905, -2100.918158, -3647.483678, -4711.186773]
     mod = ARIMA(y, (1, 0, 2))
     # Make multiple calls to fit


Juanlu001:
There are three tests failing:

https://github.com/statsmodels/statsmodels/blob/a251e4cd4b689781916e1001d7e8ca2c0395ef1f/statsmodels/tsa/tests/test_arima.py#L2358-L2373

https://github.com/statsmodels/statsmodels/blob/a251e4cd4b689781916e1001d7e8ca2c0395ef1f/statsmodels/tsa/tests/test_arima.py#L2330-L2338

And lastly:

https://github.com/statsmodels/statsmodels/blob/a251e4cd4b689781916e1001d7e8ca2c0395ef1f/statsmodels/tsa/innovations/tests/test_arma_innovations.py#L22-L24

I suspect that the reason might be that, if an `ARMA` is created without `exog`, a new one is generated which contains a constant:

https://github.com/statsmodels/statsmodels/blob/a251e4cd4b689781916e1001d7e8ca2c0395ef1f/statsmodels/tsa/arima_model.py#L908-L910

https://github.com/statsmodels/statsmodels/blob/a251e4cd4b689781916e1001d7e8ca2c0395ef1f/statsmodels/tsa/arima_model.py#L403-L406

However, I don't understand enough about the problem to decide what to do here.

josef-pkt:
@al1190 @Juanlu001 Thanks you, and welcome to open source

This shouldn't raise an exception based on the unit test failure.
The old issue #2129 might be outdated.

Did you see whether the exception in #2129 is still being raised in the master version?

I think #5538 is a similar issue of having already a constant in the data, but not as part of exog, so not the same source of the constant.

Juanlu001:
Thanks @josef-pkt for your quick answer! I'm almost certain that #2129 is still present in `master` (as @al1190 said, `gdfc0b24c4`). Perhaps the best thing would be to add a regression/unit test for it? @al1190 just in case, can you paste your terminal output of

```
$ pip list | grep statsmodels
$ python
...
>>> import statsmodels as an
>>> sm.__version__
...
>>> [failing example]
```

?

al1190:
Hi @josef-pkt  and @Juanlu001 , this is the terminal output:
```
$  pip list | grep statsmodels
statsmodels        0.10.0.dev0+1065.g0c3f6ce60 /home/al1190/Desktop/FOSS/statsmodels
$ python
Python 3.6.8 |Anaconda, Inc.| (default, Dec 30 2018, 01:22:34) 
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import statsmodels as sm
>>> sm.__version__
'v0.10.0.dev0+1066.gdfc0b24c4'
>>>#failing example
>>> import statsmodels.api as sm
>>> import numpy as np
>>> np.random.seed(13)
>>> y = np.random.randn(100)
>>> u = np.ones((100, 2))
>>> armax1 = sm.tsa.ARMA(y, order=(1, 1), exog=u).fit()
  File "/home/al1190/Desktop/FOSS/statsmodels/statsmodels/tsa/arima_model.py", line 915, in fit
    k_trend, exog = _make_arma_exog(endog, self.exog, trend)
  File "/home/al1190/Desktop/FOSS/statsmodels/statsmodels/tsa/arima_model.py", line 410, in _make_arma_exog
    raise ValueError("Exog should not contain a constant or trend. These should be added in the fit method."
ValueError: Exog should not contain a constant or trend. These should be added in the fit method.See: https://www.statsmodels.org/dev/generated/statsmodels.tsa.arima_model.ARMA.html
```
Also, the non-clear error is still present in the non-development version (0.9.0) of statsmodels:
```
>>>import statsmodels.api as sm
>>> import numpy as np
>>> np.random.seed(13)
>>> y=np.random.randn(100)
>>> u=np.ones((100,2))
>>> aramx1=sm.tsa.ARMA(y,order=(1,1),exog=u).fit()
  File "/home/al1190/anaconda3/lib/python3.7/site-packages/statsmodels/tsa/arima_model.py", line 946, in fit
    start_ar_lags)
  File "/home/al1190/anaconda3/lib/python3.7/site-packages/statsmodels/tsa/arima_model.py", line 562, in _fit_start_params
    start_params = self._fit_start_params_hr(order, start_ar_lags)
  File "/home/al1190/anaconda3/lib/python3.7/site-packages/statsmodels/tsa/arima_model.py", line 491, in _fit_start_params_hr
    start_params[:k] = ols_params
ValueError: could not broadcast input array from shape (2) into shape (3)
```

al1190:
Hello @josef-pkt , 
@Juanlu001 and I added a test (```test_constant_exog_raises_proper_error```) in test_arima.py, which raises exception #2129 in the current master version, proving the issue is still present. 

Then, we tried to look into two of the tests (```test_arima_fit_multiple_calls``` and ```test_long_ar_start_params```) which failed after the introduction of my fix to issue #2129. 
Turns out that these two tests failed because they fit multiple models, and after the first fit, the exog matrix is not reset to its original value (which is None), but is kept as the array of constants created by the function ```make_arma_exog``` in the first model fit, causing the fail in the subsequent fit. Albeit in line 914 of ```arima_model.py``` the comment clearly states to reset the exog to its original value, the current code did not do that. Therefore, we changed the code to be consistent with the comment. Doing so makes these two tests pass, but unfortunately it also makes 9 other tests fail. 

Before proceeding any further and looking into the other failing tests, we would love to have some feedback on our current approach. Is it ok or should we change it? 

Thank you! :)

Juanlu001:
On second inspection, I think some tests failing with `TypeError: object of type 'numpy.float64' has no len()` do so because of the way you compute the number of constants in your fix, because it seems to assume that `exog` is a column vector (and according to the pytest logs, many times it's a row).

al1190:
Hi Juan,

Thank you!🙂

Yes, I fix the error treating exog as a matrix and if one dimensional, a
column vector. I'll look into it. 🤔

Juanlu001:
This PR now has conflicts - with all the frenetic activity in the repository, I guess we will have to review the approach after the release :)

Juanlu001:
Issue #2129 was fixed couple of weeks ago in https://github.com/statsmodels/statsmodels/commit/119fcf90b218b95e48897a6fcaff6ef64dfd314c, part of https://github.com/statsmodels/statsmodels/pull/5833. If only we had realized it was that easy :)

Still, I think that @al1190 's comment https://github.com/statsmodels/statsmodels/pull/5553#issuecomment-483677435 highlights some possible weird behavior that deserves closer inspection. Pinging @jbrockmendel and @bashtage in case they want to take a look.

bashtage:
Let's leave this open for now.

bashtage:
Closing since this class is deprecated.

PR #5556, opened 2019-04-04
Update _smoothers_lowess.pyx

r-barnes:
The user is supposed to supply a value in the range [0,1] to the argument frac in the Lowess smoother, but this is not enforced! As a result, if the user mistakenly supplies a value such as frac=40 (for 40%) the software silently defaults to frac=1. Instead, the user should be alerted to this mistake.

### DIFF
--- a/statsmodels/nonparametric/_smoothers_lowess.pyx
+++ b/statsmodels/nonparametric/_smoothers_lowess.pyx
@@ -138,6 +138,9 @@ def lowess(np.ndarray[DTYPE_t, ndim = 1] endog,
 
     y = endog   # now just alias
     x = exog
+           
+    if not 0<=frac<=1:
+           raise ValueError("Lowess `frac` must be in the range [0,1]!")
 
     n = x.shape[0]
 


r-barnes:
Alternatively, this patch could go in the Python code for the Lowess smoother [here](https://github.com/statsmodels/statsmodels/blob/master/statsmodels/nonparametric/smoothers_lowess.py); however, it seems as though the check should be placed as closely to the usage as possible to prevent other code paths from inadvertently supplying the bad values this is supposed to exclude.

bashtage:
Happy to merge this PR if you can write a test that shows that it works for frac <0 and frac > 1.

r-barnes:
I'm afraid I'm not sure where or how to do that.

bashtage:
You need to add a function to 

https://github.com/statsmodels/statsmodels/blob/master/statsmodels/nonparametric/tests/test_lowess.py

which has the name patters `test_`<SOMTHING_MEANINGFUL> that will hit this issue and check that the error is raised.  To check that an error is raised, you can follow the pattern here:

https://github.com/statsmodels/statsmodels/blob/b8be04d124466ce3f1f03ab40add693dc598f244/statsmodels/tsa/tests/test_stattools.py#L224

bashtage:
Thanks. While this really should have a test, thanks for helping -- all improvements are welcome.

PR #5569, opened 2019-04-17
ENH: Improve numerical stability of _ar_transparams, _ar_invtransparams

jcdang:
(1 - e^-x)/(1 + e^-x) = tanh(x/2) 

I modified the code to use numpy's tanh since it doesn't have the issue with large scale numbers.

### DIFF
--- a/statsmodels/tsa/tests/test_tsa_tools.py
+++ b/statsmodels/tsa/tests/test_tsa_tools.py
@@ -104,6 +104,11 @@ def test_vech():
     assert (np.array_equal(vech(arr), [1, 4, 7, 5, 8, 9]))
 
 
+def test_ar_transparams():
+    arr = np.array([-1000.0, -100.0, -10.0, 1.0, 0.0, 1.0, 10.0, 100.0, 1000.0])
+    assert (not np.isnan(tools._ar_transparams(arr)).any())
+
+
 class TestLagmat(object):
     @classmethod
     def setup_class(cls):
--- a/statsmodels/tsa/tsatools.py
+++ b/statsmodels/tsa/tsatools.py
@@ -609,17 +609,15 @@ def _ar_transparams(params):
 
     Parameters
     ----------
-    params : array
+    params : array_like
         The AR coefficients
 
     Reference
     ---------
     Jones(1980)
     """
-    newparams = ((1-np.exp(-params))/
-                (1+np.exp(-params))).copy()
-    tmp = ((1-np.exp(-params))/
-               (1+np.exp(-params))).copy()
+    newparams = np.tanh(params/2)
+    tmp = np.tanh(params/2)
     for j in range(1,len(params)):
         a = newparams[j]
         for kiter in range(j):
--- a/statsmodels/tsa/tsatools.py
+++ b/statsmodels/tsa/tsatools.py
@@ -632,10 +632,10 @@ def _ar_invtransparams(params):
 
     Parameters
     ----------
-    params : array
+    params : array_like
         The transformed AR coefficients
     """
-    # AR coeffs
+    params = params.copy()
     tmp = params.copy()
     for j in range(len(params)-1,0,-1):
         a = params[j]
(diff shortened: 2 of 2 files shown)


ChadFulton:
Thanks! Would you mind also updating `_ar_invtransparams` to use arctanh?

jcdang:
> Thanks! Would you mind also updating `_ar_invtransparams` to use arctanh?

I can update that specific part of the code but I don't think `_ar_invtransparams` looks correct as a whole.  [This](https://github.com/statsmodels/statsmodels/blob/599c76251354958bb72625c80eb9a3152167b680/statsmodels/tsa/tsatools.py#L645-L646) is questionable.

@jseabold Can you help?

This my proposed change:
```
    invarcoefs = -np.log((1-params)/(1+params))
```
to
```
    invarcoefs = 2*np.arctanh(params)
```

ChadFulton:
That looks good to me. The code you referenced is a little less clear than it could be, but it seems to be computing the partial autocorrelations just fine, which is what it is intended to do.

jcdang:
> That looks good to me. The code you referenced is a little less clear than it could be, but it seems to be computing the partial autocorrelations just fine, which is what it is intended to do.

I can tell you for a fact that `_ar_transparams` and `_ar_invtransparams` are not inverses of each other (both before and after the hyperbolic tangent changes).  I was unable to write a unit test that covers the inverse transparams because I'm uncertain how it is suppose to work.

ChadFulton:
> I can tell you for a fact that _ar_transparams and _ar_invtransparams are not inverses of each other (both before and after the hyperbolic tangent changes). 

Can you expand on this? In practice, they seem to work as inverses.

```python
import numpy as np
from numpy.testing import assert_allclose
from statsmodels.tsa.tsatools import _ar_transparams, _ar_invtransparams

ar_params = np.r_[0.3, 0.2123, -0.1]
assert_allclose(ar_params, _ar_invtransparams(_ar_transparams(ar_params)))
```

jcdang:
Here's one of the param inputs I caught while running ARIMA:
```
ar_params = np.r_[[63.07415068, -4.5933353, -6.1342963, -7.82067222, -9.73510476, -12.0269079, -14.95639924]]
```

It seems to be okay for params between 0 and 1 but larger numbers will throw it off

Maybe what I'm using is not a valid test scenario?
```
import statsmodels.api as sm
import numpy as np

values = np.array([x / 100.0 for x in range(0, 20)])
mod = sm.tsa.ARIMA(values, order=(7, 0, 0))
res = mod.fit()
```

ChadFulton:
I see, thanks for the example. After applying `_ar_transparams(ar_params)`, the transformed parameters are right at the boundary of the transformation space (i.e. the roots of the associated lag polynomial are all very close to 1).

```python
ar_params = np.r_[63.07415068, -4.5933353, -6.1342963, -7.82067222, -9.73510476, -12.0269079, -14.95639924]
transformed = _ar_transparams(ar_params.copy())
p = np.polynomial.Polynomial(np.r_[1, -transformed])
print(np.abs(1 / p.roots()))
```

yields:

    [0.9999994298 0.9999999656 0.9999999656 1.           1.           1.           1.          ]

I think probably the issue is that we're losing some precision somewhere.

It's not surprising that it pops up in your example, since the transformation is to require the parameters to correspond to a stationary process, while the process that you are fitting is definitely non-stationary.

The version of the transformation in the state space library (`sm.tsa.statespace.tools.constrain_stationary_univariate` and `sm.tsa.statespace.tools.unconstrain_stationary_univariate`) work somewhat better in this particular example, but they're still not perfect.

bashtage:
@jcdang Any chance you could implement the other direction and then we can get this merged for the upcoming release?

jcdang:
@bashtage It's implemented already.  I didn't write a test for it.  Do you want me to add the  composed inverse test @ChadFulton gave?

bashtage:
@jcdang Nope, goo to go then.

PR #5584, opened 2019-04-30
MAINT: Remov deprecation issues

bashtage:
Remove date and period-related issues
Escape strings


bashtage:
Passed here: https://travis-ci.org/bashtage/statsmodels/builds/526411112

Issue #55, opened 2011-08-25
perfect fit (in rlm)


wesm:
Original Launchpad bug 790770: https://bugs.launchpad.net/statsmodels/+bug/790770
Reported by: josef-pktd (joep).

look at what happens in the models, model results when we have a perfect fit, and decide what to do in this cases
example for rlm, but there might be similar problems in other cases

mailinglist 2011-05-31
for rlm having a zero residual doesn't cause an exception anymore, but having a perfect fit can produce nans in some results (scale=0 and 0/0 division)

```
import numpy as np
import scikits.statsmodels.api as sm
x = np.array([0, 0.96, 2.18])
levels = np.array([2.8, 2.8, 2.8])
res1 = sm.RLM(endog=levels,exog=sm.add_constant(x),M=sm.robust.norms.HuberT()).fit()
print res1.params
print res1.sresid
print res1.scale

>>> print res1.params
[  2.77555756e-17   2.80000000e+00]
>>> print res1.sresid
[ nan  nan  nan]
>>> print res1.scale
0.0
```

PR #5585, opened 2019-04-30
Add check for convergence

jpfeiffg:
Currently in robust_linear_model, we do one iteration, rescale, and then enter the iterations loop.

This can be a problem if the first iteration is a perfect fit or nearly a perfect fit. When dividing by scale, we get either NaN or an ill-conditioned problem.

### DIFF
--- a/statsmodels/robust/robust_linear_model.py
+++ b/statsmodels/robust/robust_linear_model.py
@@ -272,7 +272,7 @@ class RLM(base.LikelihoodModel):
         # done one iteration so update
         history = self._update_history(wls_results, history, conv)
         iteration = 1
-        converged = 0
+        converged = _check_convergence(criterion, iteration, tol, maxiter)
         while not converged:
             self.weights = self.M.weights(wls_results.resid/self.scale)
             wls_results = reg_tools._MinimalWLS(self.endog, self.exog,


bashtage:
Needs a test that exits here.

jpfeiffg:
> Needs a test that exits here.

Thanks, let me add one and get back to you. I see the tests in the robust/tests directory; I need to figure out their data format and then add a test for x ~ x.

bashtage:
You can simulate the data as part of the test, something like 

```python
rs = np.random.RandomState(0)
x = rs.standard_normal(100)
y = x + 1e-16*rs.standard_normal(100)  # or just y = x
```

It is best to use a dedicated RandomState so that the test is the same on every run.

bashtage:
I don't think this can ever stop at this point since the initial criterial is always `np.inf`, see

https://github.com/statsmodels/statsmodels/blob/3284ca5c120b22787117a517fe06f248e32b51ed/statsmodels/robust/robust_linear_model.py#L259

I think you always have to run a couple of iterations to know a model has converged. 

In particular on the first iteration, 

`not (np.any(np.fabs(criterion[iteration] - criterion[iteration-1]) > tol) and iteration < maxiter)`

is 

`not (np.any(np.fabs(criterion[iteration] - np.inf) > tol) and iteration < maxiter)`

which is always a nonstop.  **AFAICT The only way it could stop here would be is `maxiter` is 1 or less.**

thequackdaddy:
For funsies, I remember dealing with this in GLM a long, long time ago. 

https://github.com/statsmodels/statsmodels/commit/05afd28dff7ea975789862c283108db97f752dc2

bashtage:
I think to beleive it has converged you need to estimate the model at least twice, so that is OK.  An looking again, if maxiter == 0, then it will ignore this and always run one, so this check seems to be OK here.

@jpfeiffg The only thing the test needs to do is to fit(maxiter=0) and verify that it exits immediately with iteration=1

jpfeiffg:
Hey, thank you for the response. I'm sorry for the delay, I have been working on other stuff and found it hard to write the test :(.

Are you sure the change is always a no-op? I added some prints to my test case (which is, fitting "a ~ b - 1" when a and b are both == range(10)).

The _check_convergence call in this case is:
_check_convergence([inf, nan], 1, 1e-08, 50).

It evaluates to True:
_check_convergence([inf, nan], 1, 1e-08, 50) = not (np.any(np.fabs(criterion[1] - criterion[0]) > 1e-08) and 1 < 50)
                                             = not (np.any(np.fabs(nan - inf) > 1e-08) and 1 < 50)


>>> not (np.any(np.fabs(np.nan - np.inf) > 1e-08) and 1 < 50)
True
>>> np.nan > 1e-08
False

So, I think if we believe that exiting after 1 iteration is a sensible thing to do here, then the code change in this pull request, is sensible.

As to whether we should exit after 1 iteration, I don't know the module well enough. I see that we're attempting to re-weight the LS in the iteration which seems fine. The problem is, in my example where we are fitting X to itself, we divide by 0 in the wls_results.resid/self.scale line, because self.scale can be 0.

Question for those who know the module well - does self.scale == 0 imply that we have in fact converged?

bashtage:
I hadn't considered the Nan case, althiugh I would describe that as a
different bug, in particular this case should be explicitly exited rather
than relying a not (Nan comparison to an number)

bashtage:
Yes, scale=0 would always be converged since all errors are 0

jpfeiffg:
OK. We could explicitly check for scale==0. I think what's happening in this case is that deviance is very large, NaN in my example when compiled with -copt=-mavx, or just large numbers when compiled normally.

I think it makes sense to treat scale == 0 or scale ~ 0 identically?

josef-pkt:
No, if the initial estimate of scale=0, then we don't necessarily have converged.

based on what I remember
The initial estimate uses MAD which has scale = 0 if at least half of the observattions have perfect fit. However, the estimate for the scale in the iteration is based on Huber or other methods and not MAD.
original ancient issue #55

Is perfect fit to 50% of the data a feature or a bug?
There should be some more recent issues on this. The "hardcore" robust literature considers it a feature, I think it's most often not desired, and Huber estimator doesn't have that feature.

jpfeiffg:
@josef-pkt , thanks for the links to older bugs, the history / context is useful.

I would agree that users usually don't want to get NaN back if they passed in reasonable data such as a perfect fit or 50%+ perfect fit. Can we reach agreement on what to do in this case, at an API level? If we agree on that I think it will be straightforward to make the code changes involved.

I propose a good thing for RLM.fit(partial_perfect_fit) to do, is to proceed until we hit a scale of 0 since it's not meaningful. I would be love to hear other approaches since I'm not familiar with robust regression. I claim in general that we shouldn't return NaN here but would also be open to understanding why this is reasonable.

Is there anyone else who should be brought in to comment? Besides @bashtage and @josef-pkt I see that @jseabold has done a lot of work in robust/ over time.

josef-pkt:
When I cannot make up my mind, which happened in this case with 50% perfect fit, then I like to push it off to the user.

So, I think we could add a keyword option to stop the iteration if estimated scale = 0. That would depend on the breakdown point of the chosen scale estimator or (maybe) initial scale estimator, around 50% for MAD and something higher for iqr.
Huber scale is always nonzero, IIRC, but might have strange behavior if a large fraction of observations has zero residuals and the others have residuals with large magnitude, i.e. the distribution of the data is far away from a normal distribution.

If scale=0, then we still have params, but I think all standard errors will be nan or zero. (I don't remember if I looked at that part, or if the references that I had used say anything about inference in the perfect fit case.)

**Correction** after reading the docstring of RLM.fit again.
The default scale is MAD (and not HuberScale as I remembered). In the past I was looking mainly at the case when initial scale estimate is MAD and iteration scale estimate is HuberScale. IIRC 

If user specified MAD scale_est, then I also think it will have converged at perfect fit if the scale=0.

IIRC, I didn't look much at that case because MAD has low efficiency as scale estimator, and I was more interested in cases where we don't have a very high breakdown point. (Default RLM is not really designed for high breakdown point, i.e. something close to 50%)

so
- if scale_est == 'mad' and scale==0, then stop as converged with perfect fit.
- if scale_est == HuberScale and scale==0, then make continuation depend on user option 
   default is to continue to convergence and handle initial estimated scale=0 with something nonzero

bashtage:
Another option would be to mimic what R or Stata does in this case,
especially if they do the same thing.

josef-pkt:
terminology: "exact fit property"
used in
Ricardo A. Maronna, Douglas R. Martin, Victor J. Yohai 2006 Robust Statistics: Theory and Methods

I was rereading the sections in the book. It is defined as having at least the number of observations corresponding to the breakdown point in a linear subspace.

AFAIR, M-estimator with Huber loss function does not have a breakdown point < 1.
But breakdown point of MAD for scale is 0.5, IIRC.
I don't remember what determines the breakdown point of redescending loss functions.

jpfeiffg:
I saw this was referenced as a release blocker - I can work on this at some point but I don't have time to do anything in the near term. Happy to hand it off if that helps; if not, I may come back to it later.

bashtage:
Would be nice to have the right fix, but not a blocker.

bashtage:
What does R do in this data?  Could just follow their lead.

PR #5596, opened 2019-05-01
TST: Explicitly set seed in basinhopping

bashtage:
Set seed in basinhopping optimizers to improve reproducibility


bashtage:
I see random failures in this, so trying to tighten it up with how it uses random values in case something else is going on.

PR #5609, opened 2019-05-02
CI: Setup Azure-Pipelines

jbrockmendel:
Closes #5423 
Makes progress towards closing #5231.

This implementation is pretty bare-bones, just takes the default template Azure provides and tweaks it enough for the tests to run as expected.  In follow-ups we can configure Windows and MacOS builds.

4 builds run in parallel, 3 of them taking about 21 minutes.  The py27 build is taking another 8 minutes because it apparently is building pandas instead of getting a wheel.  Will look into whats causing this.

cc: @datapythonista if you have any ideas on how to implement this more effectively/gracefully.


bashtage:
Closing in favor of #5617

PR #5617, opened 2019-05-04
CI: Implement azure-pipelines with multi-platform support

jbrockmendel:
Alternative to (builds on) #5609 enabling Windows builds in addition to Linux builds.  Mac builds are easy to enable, but are currently failing.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,9 +10,7 @@ environment:
     # Pip builds
     - PYTHON: C:\Python27
       PYTEST_DIRECTIVES:
-    - PYTHON: C:\Python27-x64
     - PYTHON: C:\Python36
-    - PYTHON: C:\Python36-x64
       PYTEST_DIRECTIVES:
     # Conda builds
     - PY_MAJOR_VER: 2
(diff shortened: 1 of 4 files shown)


bashtage:
This looks really good.  What do you know about customizing version of dependencies?  Appveyor runs only modern packages, so this can replace all of the x64 appveyor build. Travis runs a lot of legacy configs that are needed to make sure that things don't get broken.  Is there a way to get legacy dependencies?

bashtage:
Seems to dominate #5609

bashtage:
If you would disable the x64 appveyor builds for Windows, then I'll merge this now.  It can continue to be tweaked, but this is a strict improvement.   

Up next would be getting the OSX and pip builds out of travis.

jbrockmendel:
> What do you know about customizing version of dependencies?

I know pandas does it, will look into this in a follow-up.

> If you would disable the x64 appveyor builds for Windows, then I'll merge this now.

OK, I'm guessing for now that means removing lines 13 and 15 from appveyor.yml, but am leaving in place 29-30.

bashtage:
Really need to fix that random failure, which is pretty common. 

The next step will be to get some pressure off Travis.  Ideally would get travis down to 5.  Might make sense to leave the most difficult (complex, docbuild, etc) ones on travis.

jbrockmendel:
> Really need to fix that random failure, which is pretty common.

Yah, that'd be nice.  Does that need to be part of this PR?

bashtage:
No.

PR #5621, opened 2019-05-05
CLN: simplify lint configuration, fix some invalid escapes

jbrockmendel:
Adds a couple of lint checks that are passing but not currently specified.

Collects lint checks to be less verbose.

AFAICT the vast majority of remaining invalid escapes are caused by windows paths in docstring'd-out code (usually tracebacks) that needs to be either made into documentation or removed.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,14 +33,8 @@ parentdir_prefix = statsmodels-
 
 
 [flake8]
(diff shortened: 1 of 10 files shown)


bashtage:
Could you squash to one commit? And fix the issue, probably by removing the
new rule.

PR #5622, opened 2019-05-05
DOC: Restore import

bashtage:
Restore import for notebook

### DIFF
--- a/examples/notebooks/tsa_arma_0.ipynb
+++ b/examples/notebooks/tsa_arma_0.ipynb
@@ -300,7 +300,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from statsmodels.tsa.arima_process ArmaProcess"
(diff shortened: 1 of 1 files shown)


bashtage:
xref #5005

thequackdaddy:
Do you want to try the "set -e" setting again? 

Seeing as some of the notebooks also attempt to download stuff and those downloads fail randomly, is there a way to auto-retry building a notebook 2 or 3 times before failing?

thequackdaddy:
Wait I'm not remembering the whole story right. We do have "set -e", but it only runs before `make html`. 

https://github.com/statsmodels/statsmodels/blob/ef65a1757a9740c71df71a154174a923963bb264/tools/ci/docbuild.sh#L9

bashtage:
If it was above pytest then this would have failed.

PR #5628, opened 2019-05-06
MAINT: Relax tolerance on OSX only

bashtage:
Relax tolerance on test that randomly fails on OpenBlas/OSX

- [X] closes #5158
- [X] code/documentation is well formatted.  
- [X] properly formatted commit message. See 
      [NumPy's guide](https://docs.scipy.org/doc/numpy/dev/gitwash/development_workflow.html#writing-the-commit-message).

PR #5633, opened 2019-05-06
TST: change skip to xfail for test_compare_numdiff on OSX

jbrockmendel:
xref #5628

### DIFF
--- a/statsmodels/regression/tests/test_lme.py
+++ b/statsmodels/regression/tests/test_lme.py
@@ -88,8 +88,6 @@ def loglike_function(model, profile_fe, has_fe):
 class TestMixedLM(object):
 
     # Test analytic scores and Hessian using numeric differentiation
-    @pytest.mark.skipif(PLATFORM_OSX, reason='known to fail on OSX due to '
-                                             'numerical differences')
     @pytest.mark.slow
     @pytest.mark.parametrize("use_sqrt", [False, True])
     @pytest.mark.parametrize("reml", [False, True])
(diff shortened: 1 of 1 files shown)


bashtage:
This is not quite right.  

The correct solution is to run the test on all platforms in a Try, except, and then if except AssertionError is raised an and platform is OSX, then then to pytest.xfail(reason='') or .skip('').  This test passes on some OSX versions (see Travis, where is has been passing using Conda/MKL I think). 

So something like

```
# GH XXXX, YYYY explains reason why this is used
try: 
    buggy_on_osx_maybe
except AssertionError as err:
   if PLATFORM_OSX:
        pytest.xfail(reason="Known to fail on OSX")
   raise err
```

bashtage:
https://travis-ci.org/statsmodels/statsmodels/builds/528831936 has a passing OSX 3.7

statsmodels/tools/_testing.py
```diff
@@ -183,3 +185,24 @@ def check_predict_types(results):
         cls = pd.Series if predicted.ndim == 1 else pd.DataFrame
         predicted_expected = cls(predicted, index=exog_index)
         assert_equal(predicted_expected, predicted_pandas)
+
+
+def try_and_xfailif(condition, reason):
```
bashtage: could you rename to `allowfailif` (to be testing-framework-like) of `allow_fail_if` to be Python-like.

jbrockmendel:
@bashtage same duration permissions problem

bashtage:
Could you rebase, squash (or skip the commit with azure changes) and remove the azure template changes from the PR.  Not sure how they ended up there, but they are no longer current and I don't want to change it.

PR #5637, opened 2019-05-07
CLN: strict linting for tools directory

jbrockmendel:
At some point we should figure out what parts of this directory are actually needed/used.


bashtage:
fold toc and github stats are used.  The others are questionable.

PR #5652, opened 2019-05-08
ENH/BUG: Use intercept form of trend / exog in VARMAX start params (not mean form)

### LINKED ISSUE #5651: BUG: VARMAX start parameters use mean form for trend / exog, when state space formulation expects intercept form
VARMAX start parameters use mean form for trend / exog, when state space formulation expects intercept form. Need to left-multiply by (I - A_1  - ... - A_p).

ChadFulton:
- [x] closes #5651 
- [x] tests added / passed. 

Not adding a unit test for start params changes, since we don't really guarantee anything here.

### DIFF
--- a/statsmodels/tsa/statespace/varmax.py
+++ b/statsmodels/tsa/statespace/varmax.py
@@ -330,9 +330,9 @@ class VARMAX(MLEModel):
             trendexog_params = np.linalg.pinv(exog).dot(endog)
             endog -= np.dot(exog, trendexog_params)
             if self.k_trend > 0:
-                trend_params = trendexog_params[:self.k_trend].T.ravel()
+                trend_params = trendexog_params[:self.k_trend].T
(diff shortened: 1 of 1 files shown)


ChadFulton:
Honestly, the whole starting parameters routine could be done better, but not for this PR which just fixes a problem: that the mean and intercept can be quite different, so using the mean can imply pretty bad starting parameters and non-convergence issues.

Examples of improvements needed: not sure why I estimated trend / exog separate from VAR, and probably should use a longer initial VAR to get residuals for the VMA estimation in analogy to univariate Hannan-Rissanen.

ChadFulton:
Failures on Azure / Python 3.5 seem unrelated (not sure what that test log is showing) so I will merge later today, I guess.

bashtage:
Yes. That is an unrelated issue that has been fixed in master.

PR #5663, opened 2019-05-08
MAINT: Add test runners

bashtage:
Add test runners to all __init__ that have test directories below
reformat affected __init__ modules

xref #4245

### DIFF
--- a/statsmodels/__init__.py
+++ b/statsmodels/__init__.py
@@ -1,14 +1,14 @@
 from __future__ import print_function
-
-__docformat__ = 'restructuredtext'
-
-from distutils.version import LooseVersion
-import os
-import sys
+from .compat import PY3
 
 from warnings import simplefilter
-from statsmodels.tools.sm_exceptions import (ConvergenceWarning, CacheWriteWarning,
-                                             IterationLimitWarning, InvalidTestWarning)
+
+from .tools._testing import PytestTester
+from .tools.sm_exceptions import (ConvergenceWarning, CacheWriteWarning,
+                                  IterationLimitWarning, InvalidTestWarning)
+from ._version import get_versions
+
+__docformat__ = 'restructuredtext'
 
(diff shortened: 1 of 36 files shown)


statsmodels/__init__.py
```diff
@@ -1,14 +1,14 @@
...
-import sys
+from ._test_runner import PytestTester
+from .tools.sm_exceptions import (ConvergenceWarning, CacheWriteWarning,
+                                  IterationLimitWarning, InvalidTestWarning)
+from ._version import get_versions
 
-from warnings import simplefilter
-from statsmodels.tools.sm_exceptions import (ConvergenceWarning, CacheWriteWarning,
-                                             IterationLimitWarning, InvalidTestWarning)
+__docformat__ = 'restructuredtext'
```
jbrockmendel: Not for this PR, but are these lines still necessary?  They pop up in a bunch of places.

statsmodels/_test_runner.py
```diff
@@ -0,0 +1,28 @@
+import os
+import sys
+from distutils.version import LooseVersion
+
+
+class PytestTester(object):
```
jbrockmendel: We already have both tools.testing and tools._testing.  Can't this go in one of those?

statsmodels/formula/__init__.py
```diff
@@ -1,5 +1,6 @@
 __all__ = ["handle_formula_data"]
```
jbrockmendel: "test"?

statsmodels/_test_runner.py
bashtage: Some care is needed to avoid circular imports, so maybe

statsmodels/__init__.py
bashtage: Absolutely no idea.  I suspect not but don't need to fix everything now.

PR #5667, opened 2019-05-08
TST: Adjust tol and reset random state

bashtage:
Adjust tolerance on test that randomly fails on Windows
Reset randomstate to ensure that the test is identical in repeated runs


jbrockmendel:
Something I've wondered about a lot of tests that are subject to floating point randomness: would rounding the random inputs to, say, 6 decimal points avoid the conditions that making floating point noise an issue?  i.e. could we test for equality instead of allclose?

bashtage:
I think you still have numerical precision issues and so have a finite tolerance.  This might make it easier to work out what to use.

PR #5675, opened 2019-05-09
CLN: strict linting for emplike/tests

jbrockmendel:


bashtage:
I'm also somewhat skeptical of large straight cleans -- they are better to get in as part of a useful refactoring.  I'll merge this since it is mostly in one file that no one cares about since it only contains resutls.

PR #5680, opened 2019-05-10
ENH: Allow const idx to be found

bashtage:
Check const_idx when has_const is set to True

- [x] tests added / passed. 

### DIFF
--- a/statsmodels/base/data.py
+++ b/statsmodels/base/data.py
@@ -114,16 +114,9 @@ class ModelData(object):
         self.__dict__.update(d)
 
     def _handle_constant(self, hasconst):
-        if hasconst is not None:
-            if hasconst:
-                self.k_constant = 1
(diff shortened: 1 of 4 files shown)


josef-pkt:
This removes the short-circuiting to avoid checking for where the constant is.
It's a performance regression when the user knows that there is an implicit constant, i.e. if `has_const=True`

josef-pkt:
A file search does not find any usage of hasconst outside of the related unit tests.
I'm pretty sure I used it somewhere, maybe a notebook or PR, as computational shortcut. (saves two matrix_rank computation in implicit constant case, but only one ptp in the explicit constant case.)

bashtage:
It is imo a minor cost in the usual case of a columns of 1s. It should skip
the implicit const check is has const is 1 too, and I'll update that since
we can't find the const column anyway.

PR #5686, opened 2019-05-10
DOC: Fix parameter description in weightstats

### LINKED ISSUE #5618: DOCS: parameters of same type are displayed incorrectly
#### Describe the bug

The documentation for function parameters of the same type are not displayed correctly. An example of the current documentation (for the function `ttest_ind`) is shown in this image:
current_docs. 

There is a similar problem for other functions.

The expected documentation is shown in this image:
expected_docs

The second argument is shown as the type in the first image.

The section of the docstring for the current documentation is 

```
x1, x2 : array_like, 1-D or 2-D
        two independent samples, see notes for 2-D case
```

If the space between `x1` and `x2` is removed then the expected documentation is generated.

The current format follows the numpy docstring standard so I expect the problem is somewhere in the generation of the html.

huizew:
As suggested in https://github.com/statsmodels/statsmodels/issues/5618#issuecomment-489373679. This PR will close #5618.

x1 and x2 are in separate lines now, and keep original descriptions.

### DIFF
--- a/statsmodels/stats/weightstats.py
+++ b/statsmodels/stats/weightstats.py
@@ -858,8 +858,10 @@ class CompareMeans(object):
 
         Parameters
         ----------
-        x1, x2 : array_like, 1-D or 2-D
-            two independent samples, see notes for 2-D case
+        x1 : array_like, 1-D or 2-D
+            first of the two independent samples, see notes for 2-D case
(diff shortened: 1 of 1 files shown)


bashtage:
LGTM. Thanks.

PR #5691, opened 2019-05-11
MAINT: Near-duplicate example file, remove dominated version

jbrockmendel:
Update reference in non-dominated version, avoiding use of invalid escapes.


bashtage:
These both clearly don't work so I would prefer to get them both out, but in the spirit of compromise, I'm happy getting rid of the worse one.

PR #5695, opened 2019-05-11
MAINT: remove NameError-having version of levinson_durbin, just keep …

jbrockmendel:
…copied-and-fixed version in stattools.

There's not much left in this example file; if someone else wanted to get rid of it I wouldn't lose any sleep.

### DIFF
--- a/statsmodels/tsa/stattools.py
+++ b/statsmodels/tsa/stattools.py
@@ -903,8 +903,8 @@ def periodogram(X):
     return pergr
 
 
-#copied from nitime and statsmodels\sandbox\tsa\examples\try_ld_nitime.py
-#TODO: check what to return, for testing and trying out returns everything
+# moved from sandbox.tsa.examples.try_ld_nitime, via nitime
+# TODO: check what to return, for testing and trying out returns everything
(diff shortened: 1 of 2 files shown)


statsmodels/tsa/stattools.py
```diff
@@ -903,8 +903,8 @@ def periodogram(X):
     return pergr
 
 
-#copied from nitime and statsmodels\sandbox\tsa\examples\try_ld_nitime.py
```
bashtage: retain orig location, and say "an moved from " (possibly replace \ with  /)

statsmodels/tsa/stattools.py
```diff
@@ -971,6 +971,7 @@ def levinson_durbin(s, nlags=10, isacov=False):
     return sigma_v, arcoefs, pacf_, sig, phi  # return everything
 
 
+# moved from sandbox.tsa.examples.try_ld_nitime, via nitime
```
jbrockmendel: Whoops, this comment was supposed to go on line 906, will update.

bashtage:
Has references to history location.

PR #5696, opened 2019-05-11
CLN: remove identical functions from garch

jbrockmendel:
Functions from garch have been copied to non-sandbox modules, then better-maintained there.  This imports the non-sandbox versions and removes the old ones.

Really, I'd be OK with replacing the entire contents of that module with a message pointing users to `arch`, but that's for another day.

### DIFF
--- a/statsmodels/tsa/filters/filtertools.py
+++ b/statsmodels/tsa/filters/filtertools.py
@@ -52,6 +52,7 @@ def _pad_nans(x, head=None, tail=None):
 #from numpy import product,array
 
 
+# previous location in sandbox.tsa.try_var_convolve
 def fftconvolveinv(in1, in2, mode="full"):
     """Convolve two N-dimensional arrays using FFT. See convolve.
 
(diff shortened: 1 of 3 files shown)


statsmodels/tsa/filters/filtertools.py
```diff
@@ -308,8 +308,7 @@ def convolution_filter(x, filt, nsides=2):
     return result
 
 
-#copied from sandbox.tsa.garch
```
bashtage: Format and leave the comment.  Should someone ever want to see history, tells them to look at this files history.

bashtage: more generally this is probably the strategy that should be used to move functions if/when ready.  A clean comment back to its original location so that someone could find the history, If they need to see what went wrong.

bashtage:
Sorry to be a pain, but you add a comment in 

statsmodels.tsa.filters near fftconvolvinv that points to a previous location in statsmodels/sandbox/tsa/try_var_convolve.py since the forward reference is being deleted.  

Also please squash to 1 well-formatted commit message.

jbrockmendel:
De facto green

bashtage:
Has references to original location.

PR #5697, opened 2019-05-11
CLN: strict linting for examples/

jbrockmendel:
It's only the examples/unfinished directory that required any edits, everything else was already OK.


jbrockmendel:
@bashtage another URlError causing false positive

bashtage:
Small, no one looks at this code.

bashtage:
I wonder if the URLError that we are importing is different from the one raised

PR #5701, opened 2019-05-12
MAINT: missing imports, typos, fixes several NameErrors

jbrockmendel:


statsmodels/distributions/empirical_distribution.py
```diff
@@ -132,15 +132,13 @@ class ECDF(StepFunction):
     array([ 0.75,  1.  ,  0.  ,  0.25])
     """
     def __init__(self, x, side='right'):
-        step = True
```
bashtage: This is a lossy change since it loses what to do it step is added as an option. Could leave as a well-formatted comment.

jbrockmendel: The other branch hides a NameError since `ival` isn't defined (besides which it has a `return` in it).  So the alternative is to replace line 143 with raising NotImplementedError.  I prefer this route because it isn't hard-coding a branching.

jbrockmendel: I’m open to ideas if there’s a third option

bashtage:
You could make a well formatted comment that explains what would happen if
step was allowed to vary including that the path should follow.

jbrockmendel:
> that explains what would happen if step was allowed to vary including that the path should follow.

Someone is going to have to explain that to me first, because at the moment what would happen is a NameError.  In the parent class `ival` is in the `__init__` signature and defaults to zero, so maybe the author here intended for `ival` to always be zero here?  Or for it to be returned to the signature?

statsmodels/distributions/empirical_distribution.py
bashtage: Something like
```
# TODO: make step an arg and have a linear interpolation option?
# This is the path for step is True
# If step is False, interp1d(x,y,drop_errors=False,fill_values=ival)
# which currently raises a NameError, and so would need to be fixed
```

jbrockmendel: Made the suggested change.

This is one of the occasions where I'd be on board with the "thats what git history is for" approach to keeping the old version accessible (or just a reference back to this thread).

jbrockmendel: De facto green

bashtage: Thanks.  These types of comments would help should someone every try and rescue this code. They make it much easier to merge changes like this.

PR #5704, opened 2019-05-12
ENH: expose innovations computation method to API.

ChadFulton:
- [x] tests added / passed. 

This exposes the "innovations filter", which is used for computing the one-step-ahead prediction errors for one more multiple series, assuming a given ARMA process, via the innovations algorithm.

In particular, this can be useful for generalized least squares estimation of regression with ARMA errors, since it can compute the transformed variables on which regression can be run (like the Prais–Winsten approach but for arbitrary ARMA processes).


ChadFulton:
> This exposes the "innovations filter"

i.e. the filter was essentially already there in the Cython code, but there was no way to call it from Python.

ChadFulton:
Green. Will merge as this doesn't really affect anything else. If there do happen to be any concerns, we can always make changes until release.

PR #5709, opened 2019-05-12
TST: Fix bad bash

thequackdaddy:
- [x] tests added / passed. 

**Notes**:

So my bad push earlier had a couple bash errors I think that is going to prevent travis from running. Fixed them here.


thequackdaddy:
Ok. This works. Mac builds on travis now about 10 minutes faster.

PR #5712, opened 2019-05-13
CLN: fix invalid escapes in test_stattools introduced in #5707

jbrockmendel:
We're close to having all the invalid escapes gone (at which point we can add them to lint.sh and never worry about them again).  This fixes a handful that were recently introduced; all remaining cases are in either examples or sandbox


bashtage:
Thanks.

PR #5716, opened 2019-05-14
MAINT: Fix incorrect specification of loglike arg

jbrockmendel:
Replace no-longer-failing traceback for `res.conf_int()` with the array that is now produced.

I think there's only one more after this.


bashtage:
Thanks.

PR #5717, opened 2019-05-14
MAINT: fix non-working example ex_pandas

jbrockmendel:
A couple of usages were out of date:

- an incorrect frequency passed to `VAR` raised, so this now passes the correct frequency.  
- `order` was passed to ARMA.fit instead of to the ARMA constructor, which raises, so this passes it in the up-to-date signature.

A couple of things were commented out, and one docstring'd-out traceback for things that used to fail or not exist.  These are now printed like everything else as they now work.


jbrockmendel:
Closes #5690

PR #5721, opened 2019-05-14
CLN: strict linting for dimred, processreg, and their tests.

jbrockmendel:
These files are newly introduced; should have been passing strict-linting already.  Not sure why they weren't.


bashtage:
LGMT

PR #5728, opened 2019-05-15
CLN: remove and lint for unused imports

jbrockmendel:
In a few files, moved module-level docstrings from below imports to the correct location


bashtage:
A bit of a \*snap\*, but these should not be present.

PR #5730, opened 2019-05-15
CLN: strict linting for discrete.tests.results

jbrockmendel:
Results files aren't going to get cleaned up naturally, need dedicated passes


bashtage:
Cleaning read-only code for machines.

PR #5733, opened 2019-05-16
MAINT: unmaintained __main__ sections, addresses #5128

jbrockmendel:
In the case of `denton`, the `__main__` section has already been translated into `test_denton`


bashtage:
I don't think I will be able to merge this one.  Despite not liking mains in modules.

PR #5734, opened 2019-05-16
CLN: codes with only a few violations apiece

jbrockmendel:
Each commit is for a separate code.

### DIFF
--- a/statsmodels/sandbox/regression/tools.py
+++ b/statsmodels/sandbox/regression/tools.py
@@ -150,16 +150,6 @@ def norm_dlldy(y):
     '''
     return -y
 
-def ts_dlldy(y, df):
-    '''derivative of log pdf of standardized (?) t with respect to y
-
-    Notes
-    -----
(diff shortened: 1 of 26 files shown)


statsmodels/sandbox/regression/tools.py
```diff
@@ -150,16 +150,6 @@ def norm_dlldy(y):
...
-def ts_dlldy(y, df):
-    '''derivative of log pdf of standardized (?) t with respect to y
-
-    Notes
-    -----
-    parameterized for garch, with mean 0 and variance 1
-    '''
-    #(df+1)/2. / (1 + y**2/(df-2.)) * 2.*y/(df-2.)
-    #return -(df+1)/(df-2.) / (1 + y**2/(df-2.)) * y
-    return -(df+1)/(df) / (1 + y**2/(df)) * y
```
jbrockmendel: the same function is defined again below, but with a strictly better docstring

PR #5736, opened 2019-05-16
CLN: strict linting for regression/tests/results

jbrockmendel:


bashtage:
Files for machines only.  Would be good to have a single canonical point for Bunch.

bashtage:
It is only repeated 35 times.

PR #5744, opened 2019-05-17
CLN: Fix/lint for dangerious redefinitions and comparisons

jbrockmendel:
F812, E711, and E712 are the three least-common remaining warnings.  E711 and E712 check for comparison operations with None, True, and False.  F812 checks for e.g.

```
x = 4
foo = [x**2 for x in range(10)]
```

which can be really easy to not notice and become a PITA to debug.

### DIFF
--- a/statsmodels/base/l1_solvers_common.py
+++ b/statsmodels/base/l1_solvers_common.py
@@ -58,7 +58,7 @@ def qc_results(params, alpha, score, qc_tol, qc_verbose=False):
         fprime=fprime, alpha=alpha, params=params, passed_array=passed_array)
     passed = passed_array.min()
     if not passed:
-        num_failed = (passed_array == False).sum()
+        num_failed = (~passed_array).sum()
         message = 'QC check did not pass for %d out of %d parameters' % (
             num_failed, k_params)
         message += '\nTry increasing solver accuracy or number of iterations'\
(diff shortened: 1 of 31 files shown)


statsmodels/discrete/discrete_model.py
```diff
@@ -400,8 +400,8 @@ def cov_params_func_l1(self, likelihood_model, xopt, retvals):
         """
         H = likelihood_model.hessian(xopt)
         trimmed = retvals['trimmed']
-        nz_idx = np.nonzero(trimmed == False)[0]
```
bashtage: If these are bool arrays then this comparrison is more meaningful than comparing to int 0, and IMO should remain. Same goes for other numpy usage.  This is a bad pep8 rule probably, at least in some application.

bashtage: If these are bool arrays, you can just do np.nonzero(~trimmed)

statsmodels/base/l1_solvers_common.py
```diff
@@ -58,7 +58,7 @@ def qc_results(params, alpha, score, qc_tol, qc_verbose=False):
         fprime=fprime, alpha=alpha, params=params, passed_array=passed_array)
     passed = passed_array.min()
     if not passed:
-        num_failed = (passed_array == False).sum()
```
bashtage: (~passed_array).sum()

PR #5746, opened 2019-05-18
MAINT: fix missing or redundant imports

jbrockmendel:
`scipy.stats` is missing, which would cause NameErrors if the relevant code paths were ever hit.

numpy as imported both as `numpy` and a `np`.  Standardize on `np`.

Move all module-level imports to the top of the file.


bashtage:
Should really be BUG: Import missing module.

PR #5751, opened 2019-05-19
CLN: lint for ambiguous variable names

jbrockmendel:
One of the codes with a small number of remaining violations.  One-letter names are a PITA in general, are a hassle to track down on the places they are used.

### DIFF
--- a/statsmodels/base/l1_cvxopt.py
+++ b/statsmodels/base/l1_cvxopt.py
@@ -169,9 +169,9 @@ def _get_G(k_params):
     """
     The linear inequality constraint matrix.
     """
-    I = np.eye(k_params)
-    A = np.concatenate((-I, -I), axis=1)
-    B = np.concatenate((I, -I), axis=1)
+    I_k = np.eye(k_params)
+    A = np.concatenate((-I_k, -I_k), axis=1)
+    B = np.concatenate((I_k, -I_k), axis=1)
     C = np.concatenate((A, B), axis=0)
     # Return
     return matrix(C)
--- a/statsmodels/base/tests/test_shrink_pickle.py
+++ b/statsmodels/base/tests/test_shrink_pickle.py
@@ -57,7 +57,7 @@ class RemoveDataPickle(object):
 
         #check pickle unpickle works on full results
         #TODO: drop of load save is tested
-        res, l = check_pickle(results._results)
+        res, lbytes = check_pickle(results._results)
 
         #remove data arrays, check predict still works
         with warnings.catch_warnings(record=True) as w:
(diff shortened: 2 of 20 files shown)


bashtage:
I can't go for this.  While there are plenty of poorly names variables, some one letters make a lot of sense to people with domain knowledge and improve the code readabililty.  This one falls into unnecessary code churn, at least right now.

statsmodels/base/l1_cvxopt.py
```diff
@@ -169,9 +169,9 @@ def _get_G(k_params):
     """
     The linear inequality constraint matrix.
     """
-    I = np.eye(k_params)
```
bashtage: This one is sill.  I has bad but A and B are ok? I has clear meaning to anyone with linear algebra knowledge, and I_k isn't an improvement.

bashtage: Oh, wrong rule,  But I is still clear, at least with any decent fixed width font.

statsmodels/base/tests/test_shrink_pickle.py
```diff
@@ -57,7 +57,7 @@ def test_remove_data_pickle(self):
 
         #check pickle unpickle works on full results
         #TODO: drop of load save is tested
-        res, l = check_pickle(results._results)
+        res, lbytes = check_pickle(results._results)
```
bashtage: I'm happier with l changes, although not sure worth it either.  Especially when there isn't a specific meaning to the variable.

statsmodels/genmod/tests/test_glm.py
```diff
@@ -1692,26 +1692,26 @@ def test_wtd_patsy_missing():
 
 class CheckTweedie(object):
     def test_resid(self):
-        l = len(self.res1.resid_response) - 1
+        idx1 = len(self.res1.resid_response) - 1
```
bashtage: This one changes the flow.  One len names idx1 and the other l2?

bashtage:
If you want to change the lower l to a meaningful name and the noqa the
others, that will let the rule in.

statsmodels/nonparametric/kernel_regression.py
```diff
@@ -875,15 +875,15 @@ def _compute_test_stat(self, Y, X):
 
         m0 = model.fit(data_predict=X1)[0]
         m0 = np.reshape(m0, (n, 1))
-        I = np.zeros((n, 1))
+        I_n = np.zeros((n, 1))
```
jbrockmendel: Does `I` have a domain-specific meaning here I'm not aware of?  The other cases are identity matrices.  Unless this has a specific meaning, I might call this `zvec`

statsmodels/tsa/interp/denton.py
```diff
@@ -215,14 +215,14 @@ def dentonm(indicator, benchmark, freq="aq", **kwargs):
 
     # make partitioned matrices
     #TODO: break this out so that we can simplify the linalg?
-    I = zeros((n+m,n+m))
```
jbrockmendel: `I` isnt an identity matrix here.  Is it the standard usage in the literature?

bashtage: It is not standard but given the other notation in this function I strongly suspect they were following the author.

PR #5765, opened 2019-05-21
BUG/TST: Fix+test pieces of code that would raise NameError

jbrockmendel:
Subset of #5757 for which tests are feasible.


bashtage:
Improvements, and no losses.

PR #5774, opened 2019-05-22
PERF: use np.sum(...) instead of sum(...)

### LINKED ISSUE #4471: sigma_est_kpss sum(foo) vs foo.sum()
tsa.stattools._sigma_est_kpss starts off with `s_hat = sum(resids**2)`.  I expect that the author didn't realize this is different from `(resids**2).sum()`

```
In [2]: resids = np.random.randn(10**5)

In [3]: sum(resids**2)
Out[3]: 99483.85416165346

In [4]: (resids**2).sum()
Out[4]: 99483.85416165383

In [5]: r2 = resids**2

In [6]: %timeit sum(r2)
100 loops, best of 3: 10.6 ms per loop

In [7]: %timeit r2.sum()
10000 loops, best of 3: 47.2 µs per loop
```

jbrockmendel:
- [x] closes #4471

See #4471 for timing results.

### DIFF
--- a/statsmodels/tsa/stattools.py
+++ b/statsmodels/tsa/stattools.py
@@ -1691,11 +1691,12 @@ def kpss(x, regression='c', lags=None, store=False):
         lags = int(lags)
 
     if lags > nobs:
-        raise ValueError("lags ({}) must be <= number of observations ({})".format(lags, nobs))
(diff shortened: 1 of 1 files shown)


bashtage:
Thanks, some small fixes for recently introduced functions.

PR #5778, opened 2019-05-22
CLN: strict linting for test_varmax

jbrockmendel:
@ChadFulton


ChadFulton:
Very nice, thanks very much!

ChadFulton:
Can you remind me about `assert` versus `assert_` from `numpy.testing`?

bashtage:
In nose you needed assert_ to get any meaningful output. This isn't needed
with pytest, which shows you values of the variables compared, and so plain
assert is fine.

PR #5779, opened 2019-05-22
MAINT: statespace cleanup, alphabetically

jbrockmendel:
@ChadFulton this is the second of three statespace branches that have been sitting ready to go for a while, then I'll stop creating more work for you.

### DIFF
--- a/lint.sh
+++ b/lint.sh
@@ -85,6 +85,13 @@ if [ "$LINT" == true ]; then
         statsmodels/tsa/vector_ar/tests/results/ \
         statsmodels/tsa/statespace/*.py \
         statsmodels/tsa/statespace/tests/results/ \
+        statsmodels/tsa/statespace/tests/test_collapsed.py \
+        statsmodels/tsa/statespace/tests/test_concentrated.py \
+        statsmodels/tsa/statespace/tests/test_dynamic_factor.py \
+        statsmodels/tsa/statespace/tests/test_exact_diffuse_filtering.py \
+        statsmodels/tsa/statespace/tests/test_impulse_responses.py \
+        statsmodels/tsa/statespace/tests/test_initialization.py \
+        statsmodels/tsa/statespace/tests/test_kalman.py \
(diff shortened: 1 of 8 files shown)


jbrockmendel:
@bashtage additional failure mode `ssl.SSLError` for `get_rdataset`.

bashtage:
TBH I don't really see the point of switching existing assert_ to assert or assert_equal to assert A == B.  It creates a lot of churn and there is effectively 0 gain since NumPy will never drop these functions. 

I'm more positive about pytest.raises since this produces more natural function calls.

jbrockmendel:
I didn't count, but many of the asserts are for `assert foo is None` which is a strict (albeit small) improvement over an equality check.  Many others are for regex boolean checks which are really non-idiomatic in the status quo.  For the other cases, I'll keep your comment in mind for future branches.

ChadFulton:
@jbrockmendel sorry about the delay, but TBH you make it somewhat difficult to consider whether or not to accept some of these PRs because you mix things like easy-to-commit pep8 improvements with opinion-based code changes like using the `pytest.raises` context.

I'm undecided here.

jbrockmendel:
You do you buddy.  I just wanted to get this up so it doesn’t rot during my next leave of absence.

ChadFulton:
Just so you know where I'm coming from:

- I would have accepted a pep8 improvement patch and / or one that replaced `assert_equal` with `assert` or `assert_`
- I would have rejected a patch that removed `assert_raises`.

Since you don't seem to be in a compromising mood, we'll close this for now.

jbrockmendel:
Totally happy to compromise, will put together flake8-only version.

Out of curiosity, any idea why the `temp_filename` fixture in test_save is needed?  Doesnt NamedTemporaryFile do exactly the same thing?

jbrockmendel:
in test_mlemodel.test_score_analytic_ar1 approx_fd_centered is defined but never used.  im going to make the opinionated edit and decide it was intended to be part of the assertion in the following line

PR #5781, opened 2019-05-24
CLN: Replace #5779

jbrockmendel:
@ChadFulton the bare minimum to get linter to pass on statespace/tests.  There were a few places where changing assert_raises to pytest.raises was the best way to fix a linting complaint, but I didn't make that change in any _other_ places, so the usage is still plenty mixed.

### DIFF
--- a/statsmodels/tsa/statespace/tests/test_mlemodel.py
+++ b/statsmodels/tsa/statespace/tests/test_mlemodel.py
@@ -40,7 +40,7 @@ def get_dummy_mod(fit=True, pandas=False):
         endog = pd.Series(endog, index=index)
         exog = pd.Series(exog, index=index)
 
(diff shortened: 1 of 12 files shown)


bashtage:
Random fail.  Just merged patch to protect against this

statsmodels/tsa/statespace/tests/test_mlemodel.py
```diff
@@ -266,7 +277,7 @@ def partial_transform_sigma2(sigma2):
     approx_fd_centered = (
         mod.score(uparams, transformed=False, approx_complex_step=False,
                   approx_centered=True))
-    assert_allclose(approx_fd, analytic_score, atol=1e-5)
+    assert_allclose(approx_fd_centered, analytic_score, atol=1e-5)
```
ChadFulton: Good catch, thanks!

ChadFulton:
Thanks!

PR #5784, opened 2019-05-24
MAINT: implement parts of #5220, deprecate ancient aliases

jbrockmendel:


jbrockmendel:
@bashtage any thoughts here?  Among other things there are similar aliases in ar_model etc that should be treated symmetrically.

statsmodels/tools/decorators.py
```diff
@@ -2,7 +2,65 @@
 from statsmodels.tools.sm_exceptions import CacheWriteWarning
 import warnings
 
-__all__ = ['cache_readonly', 'cache_writable']
+__all__ = ['cache_readonly', 'cache_writable', 'deprecated_alias']
+
+
+def deprecated_alias(old_name, new_name, remove_version=None, msg=None,
```
bashtage: Is this vendored from somewhere? Might be goog to make it clear in the docstring if true, in case of changes.

jbrockmendel: I implemented the original in sm2, then you made some improvements in your PR (which I may have subsequently updated the original with)

jbrockmendel:
@bashtage does the current hold on non-doc/bug preclude deprecations?  Given the release cycle, getting these in under the wire can have a pretty big impact.

bashtage:
No, if they are essential.

bashtage:
Probably should have gotten this in for 0.10. Thanks. Will backport.

PR #5790, opened 2019-05-25
TST: Fix+test try_tukey_hsd

jbrockmendel:
Same track as #5787.

The only actual problems in the try_tukey_hsd file were caused by recarrays


jbrockmendel:
Looks like try_tukey_hsd.py and ex_pairwise.py have the same content.  try_tukey_hsd is closer to modern.

PR #5792, opened 2019-05-25
MAINT: matplotlib normed -> density

thequackdaddy:

Looks like matplotlib has changed the kwarg from `normed` to `density`. 

This is causing a fail in the pip_pre travis, and will likely cause fails elsewhere as time passes.

### DIFF
--- a/statsmodels/tsa/statespace/mlemodel.py
+++ b/statsmodels/tsa/statespace/mlemodel.py
@@ -2633,9 +2633,11 @@ class MLEResults(tsbase.TimeSeriesModelResults):
         # elements
         resid_nonmissing = resid[~(np.isnan(resid))]
         ax = fig.add_subplot(222)
-        # temporarily disable Deprecation warning, normed -> density
-        # hist needs to use `density` in future when minimum matplotlib has it
-        with warnings.catch_warnings(record=True):
+
+        # gh5792: Remove  except after support for matplotlib>2.1 required
+        try:
+            ax.hist(resid_nonmissing, density=True, label='Hist')
+        except AttributeError:
             ax.hist(resid_nonmissing, normed=True, label='Hist')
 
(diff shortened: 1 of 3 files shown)


jbrockmendel:
As long as you're doing this, can you update it in statsmodels/examples/* files?  That'll cut down the diff in #5787 et al

thequackdaddy:
I was unaware of that... maybe writing a compat layer would make more sense.

jbrockmendel:
I think for the example files is probably OK to assume an up-to-date-ish matplotlib.  Any idea how far back you have to go before the `density` kwarg doesnt work?

thequackdaddy:
Looking at their docs, I think it is was version 2.1 that switched. Its likely been raising a warning since then. 

In retrospect, writing a compat layer looks like its not worth it. Only a handful a places use this, and its a problem with a method, not a function. So I think I'll ignore my last comment.

statsmodels/tsa/statespace/mlemodel.py
```diff
@@ -2633,9 +2633,10 @@ def plot_diagnostics(self, variable=0, lags=10, fig=None, figsize=None):
         # elements
         resid_nonmissing = resid[~(np.isnan(resid))]
         ax = fig.add_subplot(222)
-        # temporarily disable Deprecation warning, normed -> density
-        # hist needs to use `density` in future when minimum matplotlib has it
-        with warnings.catch_warnings(record=True):
+
+        try:
+            ax.hist(resid_nonmissing, density=True, label='Hist')
```
jbrockmendel: comment describing the version numbers where this is relevant so this can eventually be removed?

thequackdaddy: done.

bashtage:
LGTM

PR #5796, opened 2019-05-27
CLN: test_data.py

jbrockmendel:
Getting this out of the way makes it easier to make progress on #5768.


bashtage:
Nothing lost in the clean.  Seems reasonable.

PR #5801, opened 2019-05-28
CI: Consolidate coveragerc spec

jbrockmendel:
The three .coveragerc files are already falling out of sync.  This consolidates them back to one file, then removes that file by moving the config into setup.cfg

The effect of this is that coverage.py will see the Cython.Coverage plugin even when running without cython coverage enabled.  But since we didn't _build_ the cython files with coverage enabled, this shouldn't affect anything.


bashtage:
You should not do this it means that heroic efforts will be needed to run
the test suite if built with cython coverage. Cython coverage should stay
separate and have its own coveragerc file.

jbrockmendel:
> You should not do this it means that heroic efforts will be needed to run
the test suite if built with cython coverage.

I believe you, but can you expand on this for my own edification?

bashtage:
If I have a coverage build but don't want to profile cython code at the
moment, then a full run will take an hour or so with this change and a
plain pytest statsmodels - - skip-slow instead of 8 minutes.

jbrockmendel:
> If I have a coverage build but don't want to profile cython code at the
moment

IIUC the workaround for that is to comment out one line in setup.cfg.  Since that's hardly heroic, I must be missing something.

I get that the cost is non-zero.  The flip-side is that we have a non-standard config that is harder to maintain, evidenced by the files already drifting apart.

FWIW pandas has the Cython.Coverage in its setup.cfg and a similar interface in setup.py for enabling coverage, doesn't seem to have caused any problems there.

bashtage:
Editing a setup.cfg is just non-idiomatic, and it is error-prone.  Considering that the CI setup is for automated reporting, and cython coverage can't be used with CI due to time limits (without a major refactor) and since most pyx aren't in git (only the pyx.in are), I don' see a case for removing a file that has 0 impact on CI or automation.  Suffice to say I'm -1 on this.   

The others can go/be consolidated, as long as Cyhotn coverage isn't on by default.

jbrockmendel:
Welp, I don't understand the concern well enough to totally get on board with it, but this is also clearly not a battle worth winning.  So: closing.  LMK if there is a de-duplication strategy that you would be on board with.

bashtage:
Exactly what you did except for adding the cython extension to setup and
removing the cython coverage rc file. Otherwise ok.

jbrockmendel:
I just did three test runs to see how the presence of that plugin in the config affects the timings:

```
$ python -m pytest statsmodels --skip-slow --cov=statsmodels/
598.02 seconds

$ git clean -xdf
$ export SM_CYTHON_COVERAGE=true
$ python setup.py build_ext --inplace
$ python -m pytest statsmodels/ --skip-slow --cov=statsmodels
1060.53 seconds

$ python -m pytest statsmodels/ --skip-slow --cov=statsmodels --cov-config=tools/coveragerc/.coveragerc_travis
1036.81 seconds
```

Conditional on the files having been built with cython coverage enabled, whether or not Cython.Coverage is present in the coveragerc doesn't seem to make a difference in the runtime.

bashtage:
20 minutes vs 10 minutes is a big difference when trying to verify a patch
didn't break something. Like like flushing $25 down the toilet for nothing.

jbrockmendel:
> 20 minutes vs 10 minutes is a big difference when trying to verify a patch

We're definitely talking about different things.  The relevant comparison is 1060s vs 1036s: conditional on SM_CYTHON_COVERAGE being enabled at build-time, the presence of the plugin in the coveragerc doesn't matter for the runtime.  Without SM_CYTHON_COVERAGE, the numbers are 598s vs 594s (the 594 is a run done since my previous comment)

bashtage:
I think your build was not fully enabled for Cython coverage.  You have to `git clean -xfd` before running setup since Cython needs to modify C source files.   I'm currently at 25 minutes on only at 68% and slogging through statespace, which is the really slow part.

bashtage:
```
7841 passed, 300 skipped, 53 xfailed, 8 xpassed, 192 warnings in 2772.75 seconds
```
**2772.75 seconds**

jbrockmendel:
Yep, after doing this in completely separate environments I get similar results.  I'll take a look at what went wrong with the not-completely-separate run

bashtage:
This might be back on.  I ran a standard coverage run w/o the plug-in and it was also dog slow, so it seems the instrumenting makes it slow no matter.

PR #5806, opened 2019-05-29
removed duplicates of commands done in 'update'

vss888:
The following is already done in 'update':
``` python
params = np.array(params, ndmin=1)
```
and
``` python
if not transformed:
    params = self.transform_params(params)
```
and so there is no need to do them twice.

The changes are very simple, but it would be nice to run whatever tests exists for the code (which I do not know how to run). I plan to create small and simple bite size pull requests, which are easy to create and easy to review.

### DIFF
--- a/statsmodels/tsa/statespace/mlemodel.py
+++ b/statsmodels/tsa/statespace/mlemodel.py
@@ -541,11 +541,7 @@ class MLEModel(tsbase.TimeSeriesModel):
             Additional keyword arguments to pass to the Kalman filter. See
             `KalmanFilter.filter` for more details.
         """
-        params = np.array(params, ndmin=1)
-
-        if not transformed:
-            params = self.transform_params(params)
-        self.update(params, transformed=True, complex_step=complex_step)
+        self.update(params, transformed=transformed, complex_step=complex_step)
 
         # Save the parameter names
         self.data.param_names = self.param_names
@@ -587,11 +583,7 @@ class MLEModel(tsbase.TimeSeriesModel):
             Additional keyword arguments to pass to the Kalman filter. See
             `KalmanFilter.filter` for more details.
         """
-        params = np.array(params, ndmin=1)
-
-        if not transformed:
-            params = self.transform_params(params)
-        self.update(params, transformed=True, complex_step=complex_step)
+        self.update(params, transformed=transformed, complex_step=complex_step)
 
         # Save the parameter names
         self.data.param_names = self.param_names
@@ -645,10 +637,7 @@ class MLEModel(tsbase.TimeSeriesModel):
             MLEModel._loglike_param_names, MLEModel._loglike_param_defaults,
             *args, **kwargs)
 
-        if not transformed:
-            params = self.transform_params(params)
-
-        self.update(params, transformed=True, complex_step=complex_step)
+        self.update(params, transformed=transformed, complex_step=complex_step)
 
         if complex_step:
             kwargs['inversion_method'] = INVERT_UNIVARIATE | SOLVE_LU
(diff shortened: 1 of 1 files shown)


bashtage:
Run tests:

```
python setup.py develop
pytest statsmodels # everything
pytest statsmodels/tsa # under statsmodels/tsa only
pytest statsmodels/tsa/statespace # under statsmodels/tsa/statespace only
```

statsmodels/tsa/statespace/mlemodel.py
```diff
@@ -691,15 +680,12 @@ def loglikeobs(self, params, transformed=True, complex_step=False,
...
         """
-        if not transformed:
-            params = self.transform_params(params)
-
         # If we're using complex-step differentiation, then we can't use
         # Cholesky factorization
         if complex_step:
             kwargs['inversion_method'] = INVERT_UNIVARIATE | SOLVE_LU
 
-        self.update(params, transformed=True, complex_step=complex_step)
```
jbrockmendel: @ChadFulton I think this is the main place that merits your attention

ChadFulton:
Thanks for the PR!

Unfortunately this change will break things because the transformed `params` vector needs to be passed to `_wrap_results` later in the method. That's why we do the transform at the top of the function (so that it's only done once) and then pass in the already transformed parameters to `update`.

vss888:
My bad: the line should say `transformed=transformed`. I fill fix the PR.

vss888:
I do not understand: in my version of the file this line is correct.

ChadFulton:
I always run the state space unit tests locally before pushing up commits, to make sure that my changes haven't broken anything. I encourage this practice for contributors, since it usually is a more efficient workflow if commits already at least work locally.

In your more recent commit I see you updated the code to set `params = self.update(...)`. This is better, but the thing is that we don't want to have to depend on `update` returning the parameter vector, since one of the goals of the state space models is to allow easy subclassing. So we want to minimize requirements that the subclasses must satisfy. As an example, while `SARIMAX` does return the vector at the end of `update`, `UnobservedComponents` does not.

Second `transform_params` is not a static method - generally it will depend on instance attributes. For example, in the `SARIMAX` class, the orders of the model instance are required for determining the appropriate transformation to apply.

Finally, the `update` calls that you removed in your more recent commit each do have an effect and must be kept - they cannot be removed.

> My bad: the line should say transformed=transformed. I fill fix the PR.

> I do not understand: in my version of the file this line is correct.

You may have discovered this already, but if you mean the line that @jbrockmendel pointed out, then yes, you did change that line from `transformed=True` to `transformed=transformed`. Github highlights deleted lines in red, and added lines in green.

ChadFulton:
I'll try to expand on things a little with some general comments about the layout of the state space models.

`MLEModel` is a base class that doesn't correspond to any particular state space model. Its basic job is to handle generic functionality related to models in which the state space matrices depend on parameters, particularly as this relates to maximum likelihood optimization. The `MLEResults` object is similar, and it also has the job of wrapping output in a user-friendly way, like with Pandas indexes, etc.

For this reason, many of the methods in `MLEModel` appear to do very little (like `transform_params` and `update`), but in fact they are just no-ops that need be overridden by subclasses that do implement specific state space models. For example, `SARIMAX` does quite a bit in its `transform_params` and `update` methods.

The last thing to mention is the way that Statsmodels feels about "state" in its model classes (here I'm referring to state as in computer science and not as in "state space models"). In general, model classes are not assumed to be stateful, which is why all of the model method require `params` as an argument. Combined with my comment in the previous paragraph, this is why there are so many `update` calls. Each time a model method is called, it cannot assume that the state space form matrices are what it wants, so it calls `update` so that the state space form matrices can be appropriately set with the given `params`.

bashtage:
Closed since this seems to be an infeasible approach.

PR #5820, opened 2019-06-04
ENH: Adding R-squared and Adj. R_squared to summary_col

gyassine:
- [ ] tests added / passed.  

I've only added both R-squared and Adj. R_squared at the end of the summary_col...  I tried my best via travis-ci, but i don't know where is the problem. Though it works in python 3.6 (in travis) and 3.7 (on my computer)


statsmodels/iolib/summary2.py
```diff
@@ -395,7 +395,13 @@ def _col_params(result, float_format='%.4f', stars=True):
         res.loc[idx, res.columns[0]] = res.loc[idx, res.columns[0]] + '*'
     # Stack Coefs and Std.Errors
     res = res.iloc[:, :2]
-    res = res.stack()
+    if not isinstance(result, MixedLMResultsWrapper):
```
bashtage: Why don't you just duck-type:

``` 
if hasattr(result,'rsquared'):
    # Do stuff here with rsquared
if hasattr(result,'rsquared)adj'):
    # Do stuff here with rsquared-adj
```

This is much cleaner than making ex exemption to a single model, and having to create a dependency in this module to an unrelated one.

gyassine:
Thanks for this suggestion @bashtage

statsmodels/iolib/summary2.py
```diff
@@ -395,6 +395,17 @@ def _col_params(result, float_format='%.4f', stars=True):
...
     res = res.iloc[:, :2]
+    res = res.iloc[:, :2]
+    rsquared = rsquared_adj = np.nan
+    if hasattr(result, 'rsquared'):
+        rsquared = result.rsquared
+    if hasattr(result, 'rsquared_adj'):
+        rsquared_adj = result.rsquared_adj
+    r_result = pd.DataFrame({'Basic': [rsquared], 'Adj.': [rsquared_adj]},
+                            index=['R-squared'])
+    for col in r_result:
```
bashtage: You should only append if at least one of the r2 is not null.

```
if np.any(r_result):
    for col in r_results:...
```

gyassine: True...
Thanks again (it's my first pull !)

bashtage:
You need to remove the unused import

```
statsmodels/iolib/summary2.py:4:1: F401 'statsmodels.regression.mixed_linear_model.MixedLMResultsWrapper' imported but unused
```

gyassine:
Is everything ok now ?

bashtage:
`test_summary_col` is failing.  Click on the links above.  You have changed the output and so the test needs to be updated.

gyassine:
I actually don't understand my mistake... 
When I changed the output (adding the R-squared line) one platform, it gave me another error **in the same function** telling me to remove it in another platform...

In the same platform also, it is not consistent. For example, in Travis, it works for everything but python 3.7... I cannot understand why  🤕 ...

bashtage:
Something is wrong with your code.  All of the R2 are nan in which case these shouldn’t be shown.

bashtage:
FWIW you are failing everything on Azure, so it is a bad test.

gyassine:
@bashtage Can you explain me why it doesn't work ?
- On azure & aapveyor, I don't understand where is the problem
- On travis, now some test ask me to put again what i took out

bashtage:
You are failing because the models you are using don't have R2, but your string is printing them

From travis

```
 \nR-squared nan       nan      \n          nan       nan      \nX   
```

Can you run the tests locally?

gyassine:
It works fine locally... I don't understand why it show this...

bashtage:
I have fixed this in #5995 and will use that one.  There was an issue in old pandas where np.all didn't work correctly on a dataframe and returned a series rather than a scalar.

bashtage:
Thanks.

PR #4621, opened 2018-05-12
automatic package created

abhijeetpanda12:
@ChadFulton I have only pushed the function definition.  We can discuss the improvements/ additions over here.

PR #5827, opened 2019-06-05
ENH: Add new version of ARIMA-type estimators (AR, ARIMA, SARIMAX)

ChadFulton:
Update: TODOs:

- [x] decide where to put the new `ARIMA` model class / current `ARIMA` deprecation strategy.
- [x] decide where to put the Brockwell and Davis datasets.

--------

This is very much a WIP, and basically none of the design / organizational things have been thought through (e.g. where to put it, what to call things, etc.). But is concrete enough already that hopefully there may be some useful discussion at this point about the general goals of such a project.

A couple of preliminaries:

- It is not done, and at this stage I'm most interested in soliciting very general design feedback.
- It is totally okay if this is not what we want / how we want it. Still, I think it is a good idea.

--------

My starting point for this project was:

1. We should have a wider range of estimator options for ARIMA-type models (e.g. Hannan-Rissanen, etc.), and these should be, as much as possible, stand-alone functions and be standardized.
    - In fact, we already have a lot of estimators in Statsmodels, particularly since @bashtage has been adding more, but they are scattered around and not as easy to use as they could be.

2. We should have a single model class that is the point of entry for all ARIMA-type models.
    - I think that this should ultimately be based on the state space setup, since it handles the most general case of all estimators.
    - Given a set of estimated parameters, I think that all postestimation stuff (forecasting, hypothesis tests, etc.) should be done through the state space framework.

3. It is not worth it to salvage the old AR / ARIMA models.

So this PR adds `tsa2.sarimax.estimators`, which has the following estimators for ARIMA-type models:

- `yule_walker` for AR(p) models (wrapper around `linear_model.yule_walker`)
- `burg` for AR(p) models (wrapper around `linear_model.burg`)
- `durbin_levinson` for fast computation of Yule-Walker across consecutive AR(p) orders (not sure why I didn't wrap `tsatools.levinson_durbin` - maybe it'll come to me if I look back at it later).
- `innovations` for MA(q) models (wrapper around `stattools.innovations_algo`)
- `hannan_rissanen` for ARMA models with a sum-of-squares approach (implementation based on Brockwell and Davis)
- `innovations_mle` for SARIMA models with numerical likelihood maximization (implementation based on Brockwell and Davis, ultimately wraps the `arma_innovations` code)
- `statespace` for SARIMAX model (wrapper around `statespace.SARIMAX`)
- `gls` for regression with ARIMA-type errors, where the ARIMA parameters for the error process can be estimated using any of the above (implementation based on Brockwell and Davis, uses innovation algorithm for fast transformation of endog / exog to induce homoskedasticity).

Input is standardized using the `SARIMAXSpecification` class, which handles a lot of things that ARIMA-type models need to worry about (handling different ways of specifying model orders, like non-consecutive lags, handling trends, getting parameter names, describing which of the above estimators are valid for a given set of model orders, etc.).

Output is standardized using the `SARIMAParams` class, which handles a lot of things that ARIMA-type parameters need to worry about (getting and settings parameters based on the usual econometric form vs lag polynomial form, checking for stationarity and invertibilty, etc.).

Finally, there is a new model class `ARIMA`, which is meant to (eventually) be a single point of entry for all ARIMA-type models (including SARIMAX models). It is a subclass of `statespace.SARIMAX` but allows fitting using any of the estimators above. It returns results that (will be) a subclass of `statespace.SARIMAXResults`, so that all postestimation can be done through the state space framework.

Anyway, I hope that this can be a step forward on `tsa`.

### DIFF
--- a/statsmodels/tsa/arima/estimators/gls.py
+++ b/statsmodels/tsa/arima/estimators/gls.py
@@ -1,0 +1,297 @@
+"""
+Feasible generalized least squares for regression with SARIMA errors.
+
+Author: Chad Fulton
+License: BSD-3
+"""
+import numpy as np
+import warnings
+
+from statsmodels.tools.tools import add_constant, Bunch
+from statsmodels.regression.linear_model import OLS
+from statsmodels.tsa.innovations import arma_innovations
+from statsmodels.tsa.statespace.tools import diff
+
+from statsmodels.tsa.arima.estimators.yule_walker import yule_walker
+from statsmodels.tsa.arima.estimators.burg import burg
+from statsmodels.tsa.arima.estimators.hannan_rissanen import hannan_rissanen
+from statsmodels.tsa.arima.estimators.innovations import (
+    innovations, innovations_mle)
+from statsmodels.tsa.arima.estimators.statespace import statespace
+
+from statsmodels.tsa.arima.specification import SARIMAXSpecification
+from statsmodels.tsa.arima.params import SARIMAXParams
+
+
+def gls(endog, exog=None, order=(0, 0, 0), seasonal_order=(0, 0, 0, 0),
+        include_constant=None, n_iter=None, max_iter=50, tolerance=1e-8,
+        arma_estimator='innovations_mle', arma_estimator_kwargs=None):
+    """
+    Estimate ARMAX parameters by GLS.
+
+    Parameters
+    ----------
+    endog : array_like
+        Input time series array.
+    exog : array_like, optional
+        Array of exogenous regressors. If not included, then `include_constant`
+        must be True, and then `exog` will only include the constant column.
+    order : tuple, optional
+        The (p,d,q) order of the ARIMA model. Default is (0, 0, 0).
+    seasonal_order : tuple, optional
+        The (P,D,Q,s) order of the seasonal ARIMA model.
+        Default is (0, 0, 0, 0).
+    include_constant : bool, optional
+        Whether to add a constant term in `exog` if it's not already there.
+        The estimate of the constant will then appear as one of the `exog`
+        parameters. If `exog` is None, then the constant will represent the
+        mean of the process. Default is True if the specified model does not
+        include integration and False otherwise.
+    n_iter : int, optional
+        Optionally iterate feasible GSL a specific number of times. Default is
+        to iterate to convergence. If set, this argument overrides the
+        `max_iter` and `tolerance` arguments.
+    max_iter : int, optional
+        Maximum number of feasible GLS iterations. Default is 50. If `n_iter`
+        is set, it overrides this argument.
+    tolerance : float, optional
+        Tolerance for determining convergence of feasible GSL iterations. If
+        `iter` is set, this argument has no effect.
+        Default is 1e-8.
+    arma_estimator : string, optional
+        The estimator used for estimating the ARMA model. This option should
+        not generally be used, unless the default method is failing or is
+        otherwise unsuitable. Not all values will be valid, depending on the
+        specified model orders (`order` and `seasonal_order`). Possible values
+        are:
+        * 'innovations_mle' - can be used with any specification
+        * 'statespace' - can be used with any specification
+        * 'hannan_rissanen' - can be used with any ARMA non-seasonal model
+        * 'yule_walker' - only non-seasonal consecutive
+          autoregressive (AR) models
+        * 'burg' - only non-seasonal, consecutive autoregressive (AR) models
+        * 'innovations' - only non-seasonal, consecutive moving
+          average (MA) models.
+        The default is 'innovations_mle'.
+    arma_estimator_kwargs : dict, optional
+        Arguments to pass to the ARMA estimator.
+
+    Returns
+    -------
+    parameters : SARIMAXParams object
+        Contains the parameter estimates from the final iteration.
+    other_results : Bunch
+        Includes eight components: `spec`, `params`, `converged`,
+        `differences`, `iterations`, `arma_estimator`, 'arma_estimator_kwargs',
+        and `arma_results`.
+
+    Notes
+    -----
+    The primary reference is [1]_, section 6.6. In particular, the
+    implementation follows the iterative procedure described in section 6.6.2.
+    Construction of the transformed variables used to compute the GLS estimator
+    described in section 6.6.1 is done via an application of the innovations
+    algorithm (rather than explicit construction of the transformation matrix).
+
+    Note that if the specified model includes integration, both the `endog` and
+    `exog` series will be differenced prior to estimation and a warning will
+    be issued to alert the user.
+
+    References
+    ----------
+    .. [1] Brockwell, Peter J., and Richard A. Davis. 2016.
+       Introduction to Time Series and Forecasting. Springer.
+
+    """
+    # Handle n_iter
+    if n_iter is not None:
+        max_iter = n_iter
+        tolerance = np.inf
+
+    # Default for include_constant is True if there is no integration and
+    # False otherwise
+    integrated = order[1] > 0 or seasonal_order[1] > 0
+    if include_constant is None:
+        include_constant = not integrated
+    elif include_constant and integrated:
+        raise ValueError('Cannot include a constant in an integrated model.')
+
+    # Handle including the constant (need to do it now so that the constant
+    # parameter can be included in the specification as part of `exog`.)
+    if include_constant:
+        exog = np.ones_like(endog) if exog is None else add_constant(exog)
+
+    # Create the SARIMAX specification
+    spec = SARIMAXSpecification(endog, exog=exog, order=order,
+                                seasonal_order=seasonal_order)
+    endog = spec.endog
+    exog = spec.exog
+
+    # Handle integration
+    if spec.is_integrated:
+        # TODO: this is the approach suggested by BD (see Remark 1 in
+        # section 6.6.2 and Example 6.6.3), but maybe there are some cases
+        # where we don't want to force this behavior on the user?
+        warnings.warn('Provided `endog` and `exog` series have been'
+                      ' differenced to eliminate integration prior to GLS'
+                      ' parameter estimation.')
+        endog = diff(endog, k_diff=spec.diff,
+                     k_seasonal_diff=spec.seasonal_diff,
+                     seasonal_periods=spec.seasonal_periods)
+        exog = diff(exog, k_diff=spec.diff,
+                    k_seasonal_diff=spec.seasonal_diff,
+                    seasonal_periods=spec.seasonal_periods)
+    augmented = np.c_[endog, exog]
+
+    # Validate arma_estimator
+    spec.validate_estimator(arma_estimator)
+    if arma_estimator_kwargs is None:
+        arma_estimator_kwargs = {}
+
+    # Step 1: OLS
+    mod_ols = OLS(endog, exog)
+    res_ols = mod_ols.fit()
+    exog_params = res_ols.params
+    resid = res_ols.resid
+
+    # 0th iteration parameters
+    p = SARIMAXParams(spec=spec)
+    p.exog_params = exog_params
+    if spec.max_ar_order > 0:
+        p.ar_params = np.zeros(spec.k_ar_params)
+    if spec.max_seasonal_ar_order > 0:
+        p.seasonal_ar_params = np.zeros(spec.k_seasonal_ar_params)
+    if spec.max_ma_order > 0:
+        p.ma_params = np.zeros(spec.k_ma_params)
+    if spec.max_seasonal_ma_order > 0:
+        p.seasonal_ma_params = np.zeros(spec.k_seasonal_ma_params)
+    p.sigma2 = res_ols.scale
+
+    ar_params = p.ar_params
+    seasonal_ar_params = p.seasonal_ar_params
+    ma_params = p.ma_params
+    seasonal_ma_params = p.seasonal_ma_params
+    sigma2 = p.sigma2
+
+    # Step 2 - 4: iterate feasible GLS to convergence
+    arma_results = [None]
+    differences = [None]
+    parameters = [p]
+    converged = False if n_iter is None else None
+    i = 0
+    for i in range(1, max_iter + 1):
+        prev = exog_params
+
+        # Step 2: ARMA
+        # TODO: allow estimator-specific kwargs?
+        if arma_estimator == 'yule_walker':
+            p_arma, res_arma = yule_walker(
+                resid, ar_order=spec.ar_order, demean=False,
+                **arma_estimator_kwargs)
+        elif arma_estimator == 'burg':
+            p_arma, res_arma = burg(resid, ar_order=spec.ar_order,
+                                    demean=False, **arma_estimator_kwargs)
+        elif arma_estimator == 'innovations':
+            out, res_arma = innovations(resid, ma_order=spec.ma_order,
+                                        demean=False, **arma_estimator_kwargs)
+            p_arma = out[-1]
+        elif arma_estimator == 'hannan_rissanen':
+            p_arma, res_arma = hannan_rissanen(
+                resid, ar_order=spec.ar_order, ma_order=spec.ma_order,
+                demean=False, **arma_estimator_kwargs)
+        else:
+            # For later iterations, use a "warm start" for parameter estimates
+            # (speeds up estimation and convergence)
+            start_params = (
+                None if i == 1 else np.r_[ar_params, ma_params,
+                                          seasonal_ar_params,
+                                          seasonal_ma_params, sigma2])
+            # Note: in each case, we do not pass in the order of integration
+            # since we have already differenced the series
+            tmp_order = (spec.order[0], 0, spec.order[2])
+            tmp_seasonal_order = (spec.seasonal_order[0], 0,
+                                  spec.seasonal_order[2],
+                                  spec.seasonal_order[3])
+            if arma_estimator == 'innovations_mle':
+                p_arma, res_arma = innovations_mle(
+                    resid, order=tmp_order, seasonal_order=tmp_seasonal_order,
+                    demean=False, start_params=start_params,
+                    **arma_estimator_kwargs)
+            else:
+                p_arma, res_arma = statespace(
+                    resid, order=tmp_order, seasonal_order=tmp_seasonal_order,
+                    include_constant=False, start_params=start_params,
+                    **arma_estimator_kwargs)
+
+        ar_params = p_arma.ar_params
+        seasonal_ar_params = p_arma.seasonal_ar_params
+        ma_params = p_arma.ma_params
+        seasonal_ma_params = p_arma.seasonal_ma_params
+        sigma2 = p_arma.sigma2
+        arma_results.append(res_arma)
+
+        # Step 3: GLS
+        # Compute transformed variables that satisfy OLS assumptions
+        # Note: In section 6.1.1 of Brockwell and Davis (2016), these
+        # transformations are developed as computed by left multiplcation
+        # by a matrix T. However, explicitly constructing T and then
+        # performing the left-multiplications does not scale well when nobs is
+        # large. Instead, we can retrieve the transformed variables as the
+        # residuals of the innovations algorithm (the `normalize=True`
+        # argument applies a Prais-Winsten-type normalization to the first few
+        # observations to ensure homoskedasticity). Brockwell and Davis
+        # mention that they also take this approach in practice.
+        tmp, _ = arma_innovations.arma_innovations(
+            augmented, ar_params=ar_params, ma_params=ma_params,
+            normalize=True)
+        u = tmp[:, 0]
+        x = tmp[:, 1:]
+
+        # OLS on transformed variables
+        mod_gls = OLS(u, x)
+        res_gls = mod_gls.fit()
+        exog_params = res_gls.params
+        resid = endog - np.dot(exog, exog_params)
+
+        # Construct the parameter vector for the iteration
+        p = SARIMAXParams(spec=spec)
+        p.exog_params = exog_params
+        if spec.max_ar_order > 0:
+            p.ar_params = ar_params
+        if spec.max_seasonal_ar_order > 0:
+            p.seasonal_ar_params = seasonal_ar_params
+        if spec.max_ma_order > 0:
+            p.ma_params = ma_params
+        if spec.max_seasonal_ma_order > 0:
+            p.seasonal_ma_params = seasonal_ma_params
+        p.sigma2 = sigma2
+        parameters.append(p)
+
+        # Check for convergence
+        difference = np.abs(exog_params - prev)
+        differences.append(difference)
+        if n_iter is None and np.all(difference < tolerance):
+            converged = True
+            break
+    else:
+        if n_iter is None:
+            warnings.warn('Feasible GLS failed to converge in %d iterations.'
+                          ' Consider increasing the maximum number of'
+                          ' iterations using the `max_iter` argument or'
+                          ' reducing the required tolerance using the'
+                          ' `tolerance` argument.' % max_iter)
+
+    # Construct final results
+    p = parameters[-1]
+    other_results = Bunch({
+        'spec': spec,
+        'params': parameters,
+        'converged': converged,
+        'differences': differences,
+        'iterations': i,
+        'arma_estimator': arma_estimator,
+        'arma_estimator_kwargs': arma_estimator_kwargs,
+        'arma_results': arma_results,
+    })
+
+    return p, other_results
--- a/statsmodels/tsa/arima/estimators/hannan_rissanen.py
+++ b/statsmodels/tsa/arima/estimators/hannan_rissanen.py
@@ -1,0 +1,190 @@
+"""
+Hannan-Rissanen procedure for estimating ARMA(p,q) model parameters.
+
+Author: Chad Fulton
+License: BSD-3
+"""
+import numpy as np
+
+from scipy.signal import lfilter
+from statsmodels.tools.tools import Bunch
+from statsmodels.regression.linear_model import OLS, yule_walker
+from statsmodels.tsa.tsatools import lagmat
+
+from statsmodels.tsa.arima.specification import SARIMAXSpecification
+from statsmodels.tsa.arima.params import SARIMAXParams
+
+
+def hannan_rissanen(endog, ar_order=0, ma_order=0, demean=True,
+                    initial_ar_order=None, unbiased=True):
+    """
+    Estimate ARMA parameters using Hannan-Rissanen procedure.
+
+    Parameters
+    ----------
+    endog : array_like
+        Input time series array, assumed to be stationary.
+    ar_order : int
+        Autoregressive order
+    ma_order : int
+        Moving average order
+    demean : bool, optional
+        Whether to estimate and remove the mean from the process prior to
+        fitting the ARMA coefficients. Default is True.
+    initial_ar_order : int, optional
+        Order of long autoregressive process used for initial computation of
+        residuals.
+    unbiased: bool, optional
+        Whether or not to apply the bias correction step. Default is True.
+
+    Returns
+    -------
+    parameters : SARIMAXParams object
+    other_results : Bunch
+        Includes three components: `spec`, containing the
+        `SARIMAXSpecification` instance corresponding to the input arguments;
+        `initial_ar_order`, containing the autoregressive lag order used in the
+        first step; and `resid`, which contains the computed residuals from the
+        last step.
+
+    Notes
+    -----
+    The primary reference is [1]_, section 5.1.4, which describes a three-step
+    procedure that we implement here.
+
+    1. Fit a large-order AR model via Yule-Walker to estimate residuals
+    2. Compute AR and MA estimates via least squares
+    3. (Unless `unbiased=False`) Perform bias correction
+
+    The order used for the AR model in the first step may be given as an
+    argument. If it is not, we compute it as suggested by [2]_.
+
+    The estimate of the variance that we use is computed from the residuals
+    of the least-squares regression and not from the innovations algorithm.
+    This is because our fast implementation of the innovations algorithm is
+    only valid for stationary processes, and the Hannan-Rissanen procedure may
+    produce estimates that imply non-stationary processes. To avoid
+    inconsistency, we never compute this latter variance here, even if it is
+    possible. See test_hannan_rissanen::test_brockwell_davis_example_517 for
+    an example of how to compute this variance manually.
+
+    This procedure assumes that the series is stationary.
+
+    References
+    ----------
+    .. [1] Brockwell, Peter J., and Richard A. Davis. 2016.
+       Introduction to Time Series and Forecasting. Springer.
+    .. [2] Gomez, Victor, and Agustin Maravall. 2001.
+       "Automatic Modeling Methods for Univariate Series."
+       A Course in Time Series Analysis, 171–201.
+
+    """
+    spec = SARIMAXSpecification(endog, ar_order=ar_order, ma_order=ma_order)
+    endog = spec.endog
+    if demean:
+        endog = endog - endog.mean()
+
+    p = SARIMAXParams(spec=spec)
+
+    nobs = len(endog)
+    max_ar_order = spec.max_ar_order
+    max_ma_order = spec.max_ma_order
+
+    # Default initial_ar_order is as suggested by Gomez and Maravall (2001)
+    if initial_ar_order is None:
+        initial_ar_order = max(np.floor(np.log(nobs)**2).astype(int),
+                               2 * max(max_ar_order, max_ma_order))
+    # Create a spec, just to validate the initial autoregressive order
+    _ = SARIMAXSpecification(endog, ar_order=initial_ar_order)
+
+    # Compute lagged endog
+    # (`ar_ix`, and `ma_ix` below, are to account for non-consecutive lags;
+    # for indexing purposes, must have dtype int)
+    ar_ix = np.array(spec.ar_lags, dtype=int) - 1
+    lagged_endog = lagmat(endog, max_ar_order, trim='both')[:, ar_ix]
+
+    # If no AR or MA components, this is just a variance computation
+    if max_ma_order == 0 and max_ar_order == 0:
+        p.sigma2 = np.var(endog, ddof=0)
+        resid = endog.copy()
+    # If no MA component, this is just CSS
+    elif max_ma_order == 0:
+        mod = OLS(endog[max_ar_order:], lagged_endog)
+        res = mod.fit()
+        resid = res.resid
+        p.ar_params = res.params
+        p.sigma2 = res.scale
+    # Otherwise ARMA model
+    else:
+        # Step 1: Compute long AR model via Yule-Walker, get residuals
+        initial_ar_params, _ = yule_walker(
+            endog, order=initial_ar_order, method='mle')
+        X = lagmat(endog, initial_ar_order, trim='both')
+        y = endog[initial_ar_order:]
+        resid = y - X.dot(initial_ar_params)
+
+        # Get lagged residuals for `exog` in least-squares regression
+        ma_ix = np.array(spec.ma_lags, dtype=int) - 1
+        lagged_resid = lagmat(resid, max_ma_order, trim='both')[:, ma_ix]
+
+        # Step 2: estimate ARMA model via least squares
+        ix = initial_ar_order + max_ma_order - max_ar_order
+        mod = OLS(endog[initial_ar_order + max_ma_order:],
+                  np.c_[lagged_endog[ix:], lagged_resid])
+        res = mod.fit()
+        p.ar_params = res.params[:spec.k_ar_params]
+        p.ma_params = res.params[spec.k_ar_params:]
+        resid = res.resid
+        p.sigma2 = res.scale
+
+        # Step 3: bias correction (if requested)
+        if unbiased:
+            Z = np.zeros_like(endog)
+            V = np.zeros_like(endog)
+            W = np.zeros_like(endog)
+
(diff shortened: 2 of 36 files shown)


statsmodels/tsa2/datasets/brockwell_davis_2002/__init__.py
```diff
@@ -0,0 +1,6 @@
+import pandas as pd
+
+from pytsa.datasets.brockwell_davis_2002.data.dowj import dowj
```
jbrockmendel: it looks like these files are implemented below.  are you intentionally importing from a different package?

statsmodels/tsa2/sarimax/estimators/tests/test_burg.py
```diff
@@ -0,0 +1,110 @@
...
+    # Difference and demean the series
+    endog = dowj.diff().iloc[1:]
+
+    # Burg
+    res, _ = burg(endog, ar_order=1, demean=True)
+    assert_allclose(res.ar_params, [0.4371], atol=1e-4)
+    assert_allclose(res.sigma2, 0.1423, atol=1e-4)
+
+
+@pytest.mark.low_precision
```
jbrockmendel: I like this idea.  How about taking it one step further and instead of putting the explanation in a comment, do:

```
@pytest.mark.low_precision("Test against Example 5.1.4 in Brockwell and Davis (2016) [...]")
```

statsmodels/tsa2/sarimax/estimators/tests/test_durbin_levinson.py
```diff
@@ -0,0 +1,93 @@
+import numpy as np
+
+from statsmodels.tsa.innovations.arma_innovations import arma_innovations
+from statsmodels.tsa2.datasets.brockwell_davis_2002 import dowj, lake
+from statsmodels.tsa2.sarimax.estimators.durbin_levinson import durbin_levinson
+
+import pytest
+from numpy.testing import assert_allclose, assert_equal, assert_raises
```
jbrockmendel: pls put third party imports together before first-party imports

statsmodels/tsa2/sarimax/estimators/tests/test_gls.py
```diff
@@ -0,0 +1,231 @@
...
+    res, _ = gls(endog, exog, order=(0, 0, 12), max_iter=2)
+
+    assert_allclose(res.exog_params, 328.45, atol=1e-2)
+    assert_allclose(res.ma_params, [.219, .098, .031, .064, .069, .111, .081,
+                                    .057, .092, -0.28, .183, -.672], atol=1e-3)
+    assert_allclose(res.sigma2, 12581, atol=1)
+
+
+def test_integrated():
+    # Get the lake data
```
jbrockmendel: comment unnecessary

statsmodels/tsa2/sarimax/tests/test_tools.py
```diff
@@ -0,0 +1,76 @@
...
+    assert_equal(validate_basic(0, 1), [0])
+    assert_equal(validate_basic([0], 1), [0])
+    assert_equal(validate_basic(np.array([1.2, 0.5 + 1j]), 2),
+                 np.array([1.2, 0.5 + 1j]))
+    assert_equal(
+        validate_basic([np.nan, -np.inf, np.inf], 3, allow_infnan=True),
+        [np.nan, -np.inf, np.inf])
+
+    # Invalid parameters
+    assert_raises(ValueError, validate_basic, [], 1, title='test')
```
jbrockmendel: pls use with `pytest.raises(ValueError, match="foo"):` syntax

statsmodels/tsa2/sarimax/tests/test_tools.py
```diff
@@ -0,0 +1,76 @@
...
+
+    # Non-list iterable input, lags
+    assert_equal(standardize_lag_order((1, 2)), 2)
+    assert_equal(standardize_lag_order((1, 3)), [1, 3])
+
+    assert_equal(standardize_lag_order(np.array([1, 2])), 2)
+    assert_equal(standardize_lag_order(np.array([1, 3])), [1, 3])
+
+    # List input, booleans
+    print(standardize_lag_order([0]))
```
jbrockmendel: why print?

statsmodels/tsa2/sarimax/tests/test_tools.py
```diff
@@ -0,0 +1,76 @@
...
+    # Integer input
+    assert_equal(standardize_lag_order(0, title='test'), 0)
+    assert_equal(standardize_lag_order(3), 3)
+
+    # List input, lags
+    assert_equal(standardize_lag_order([]), 0)
+    assert_equal(standardize_lag_order([1, 2]), 2)
+    assert_equal(standardize_lag_order([1, 3]), [1, 3])
+
+    # Non-list iterable input, lags
```
jbrockmendel: separate smaller-scoped tests

jbrockmendel: possibly collected in a test class

statsmodels/tsa2/sarimax/params.py
```diff
@@ -0,0 +1,402 @@
...
+        if self.k_seasonal_ma_params:
+            components.append('seasonal_ma=%s' %
+                              str(self.seasonal_ma_params))
+        if not self.spec.concentrate_scale:
+            components.append('sigma2=%s' % self.sigma2)
+        return 'SARIMAXParams(%s)' % ', '.join(components)
+
+    def __str__(self):
+        """Represent parameter array as a string."""
+        return str(self.params)
```
jbrockmendel: why not just alias `__repr__`?

statsmodels/tsa2/sarimax/model.py
```diff
@@ -0,0 +1,219 @@
...
+from statsmodels.tsa2.sarimax.estimators.innovations import (
+    innovations, innovations_mle)
+from statsmodels.tsa2.sarimax.estimators.gls import gls as estimate_gls
+from statsmodels.tsa2.sarimax.estimators.statespace import statespace
+
+from statsmodels.tsa2.sarimax.specification import SARIMAXSpecification
+from statsmodels.tsa2.sarimax.params import SARIMAXParams
+
+
+class ARIMA(sarimax.SARIMAX):
```
jbrockmendel: I see the appeal of implementing AR/ARMA/ARIMA in terms of statespace, but these should replace the existing models completely.

ChadFulton: Not sure I understand what you're getting at here. It sounds like you see some conflict between these goals?

statsmodels/tsa2/datasets/brockwell_davis_2002/__init__.py
ChadFulton: Just a typo, thanks

statsmodels/tsa2/sarimax/model.py
jbrockmendel: I don't want to see arima_model.py, ar_model.py etc become orphaned files and sit there forever slowly decaying (like arma_mle, descriptivestats, mlemodel, varma_process).  So I like the new implementations, just want to remove the old ones at the same time.

ChadFulton: Yes, that is also my goal, once the new implementations replace all of the features of the old models. I think that is true now, with the exception of model selection in the AR model. I anticipate that model selection will be replaced by #5151 (since #4621 has stalled).

ChadFulton:
A couple of additional implementation details that might be useful:

- Implementing this was based pretty much on Brockwell and Davis, and the basic line of testing is against their book or their software (ITSM or the R copycat package ITSMR).

- The convention here is that all intercept / regression effects (including trend effects) are estimated using the regression with SARIMA errors approach and so must either go through the `gls` estimator or the `statespace` estimator.
   - We do this in Statsmodels already, except that outside of state space models we just use the OLS estimator rather than the GLS estimator.
   - In the state space `SARIMAX`, the `trend` argument is estimated in intercept form. I now think for consistency, it is better for the basic ARIMA model to have all regression-type effects enter in the same way.
   - All of the estimators do include a `demean` argument that defaults to `True` (except `gls` and `statespace` which have `include_constant`). This is (a) since theoretically the SARIMA estimators should be applied to mean-zero processes, and (b) for consistency with other Statsmodels functions and other packages.

- In the estimator functions, integrated models are handled in one of three ways:
   1. The simpler estimators - `yule_walker`, `durbin_levinson`, `burg`, `innovations`, and `hannan_rissanen` - just don't allow specification of an integrated model (the have arguments like `ar_order`, so there just isn't a place to put an order of integration).
   2. The more complete estimators - `innovations_mle` and `gls` (since `gls` can use `innovations_mle` as part of the GLS estimation) - allows specification of an integrated model, but always directly difference the data before estimating the parameters.
   3. The state space estimator continues to incorporate integration directly into the state space form.

jbrockmendel:
Thanks for expanding on that.  Should I think of this PR as a Proof Of Concept, or something you would like to merge in this form?  After the release, are you hoping to get this done soonish (i.e. does this block other goals?) or can the review be more leisurely?

ChadFulton:
> Should I think of this PR as a Proof Of Concept, or something you would like to merge in this form? After the release, are you hoping to get this done soonish (i.e. does this block other goals?) or can the review be more leisurely?

Basically, I think that the setup here is good, and potentially mergeable after I clean up everything and add more tests. Maybe it should be merged soon after the next release so that other people besides me can start using it.

My main goal, actually, is that we can start some real documentation on how to use `tsa`, and I think that to do that we need to make the model interfaces more consistent. The biggest problem there is the signatures of the AR and ARIMA model methods, but also the state space models have some weird things too (e.g. the way `trend` is estimated), and so the goal of this PR is to figure out a consistent interface for all ARIMA-type models that we can confidently recommend to users as "the way".

But I think that as we write "good" documentation for this entry point to ARIMA-type models, we'll probably figure out that I didn't do some things quite right in this PR (like I set up some things in a way that turns out to be counter-intuitive, or hard to explain), and we'll need to change the implementation. I think that's okay for a new set of models that haven't been in a release yet, but it's something to keep in mind. And of course the more of those things that we can figure out in this PR already, the better.

statsmodels/tsa2/sarimax/estimators/tests/test_burg.py
ChadFulton: This is a good idea, thanks!

statsmodels/tsa2/sarimax/estimators/tests/test_durbin_levinson.py
ChadFulton: Thanks

statsmodels/tsa2/sarimax/estimators/tests/test_gls.py
ChadFulton: Are you sure you would have known what the next line was doing without that comment? :)

statsmodels/tsa2/sarimax/tests/test_tools.py
ChadFulton: I know we differ on this point, but thanks for the comment.

statsmodels/tsa2/sarimax/tests/test_tools.py
ChadFulton: Thanks!!

statsmodels/tsa2/sarimax/tests/test_tools.py
ChadFulton: good idea, thanks

statsmodels/tsa2/sarimax/params.py
ChadFulton: I think that aliasing `__repr__` is probably a better idea here, thanks.

ChadFulton:
I think that this PR is in good shape to be merged, even if we might expect some changes going forwards (and even though I continue to fight with the linter...).

**TODOs**: (list also added to top post in this issue)

(1) Decide where to put the new `ARIMA` model class / current `ARIMA` deprecation strategy.
(2) Decide where to put the Brockwell and Davis datasets.

bashtage:
Should add a lout FutureWarnign to ARMA and ARIMA `__init__` that these are deprecated, and statsmodels.NEWLOC.ARMA/ARIMA should be used. These functions will see no future development, although they won't go away anytime soon. 

Are you really planning on having `tsa2`?  I not a big fan of that.

statsmodels/tsa2/sarimax/estimators/gls.py
```diff
@@ -0,0 +1,305 @@
...
+from statsmodels.tsa2.sarimax.estimators.hannan_rissanen import hannan_rissanen
+from statsmodels.tsa2.sarimax.estimators.innovations import (
+    innovations, innovations_mle)
+from statsmodels.tsa2.sarimax.estimators.statespace import statespace
+
+from statsmodels.tsa2.sarimax.specification import SARIMAXSpecification
+from statsmodels.tsa2.sarimax.params import SARIMAXParams
+
+
+def gls(endog, exog=None, order=(0, 0, 0), seasonal_order=(0, 0, 0, 0),
```
bashtage: Should this function in in statsmodels.regression.linear_models?

jbrockmendel:
> Are you really planning on having tsa2? I not a big fan of that.

Agreed.  To the extent that these are replacements for existing models, we should deprecate the existing ones and put a plan in place to swap these in.  To the extent that this is new functionality, it should go in tsa where readers would expect to find it.

And to the extent that it is experimental, it should be private or otherwise marked so we don't need to worry about deprecation cycles for a while.

@ChadFulton to the extent that this can be broken into smaller pieces, it would make it easier on reviewers.  e.g.
- the small fixups in mlemodel.py could be totally independent
- burg.py merits a discussion on how it relates to the existing pacf_burg
- durbin_levinson.py merits a discussion on how it relates to the existing levinson_durbin

And definitely pls add it to the strict-linting in lint.sh

statsmodels/tsa2/sarimax/estimators/tests/test_yule_walker.py
```diff
@@ -0,0 +1,86 @@
...
+def test_invalid():
+    endog = np.arange(2)
+    assert_raises(ValueError, yule_walker, endog, ar_order=-1)
+    assert_raises(ValueError, yule_walker, endog, ar_order=1.5)
+
+    endog = np.arange(10)
+    assert_raises(ValueError, yule_walker, endog, ar_order=[1, 3])
+
+
+@pytest.mark.xfail
```
jbrockmendel: can you add a reason descrivbing why this fails

statsmodels/tsa2/sarimax/estimators/tests/test_hannan_rissanen.py
```diff
@@ -0,0 +1,105 @@
...
+    endog = np.arange(20)
+    # TODO: shouldn't allow initial_ar_order <= ar_order
+    hannan_rissanen(endog, ar_order=2, ma_order=0, initial_ar_order=1)
+    # TODO: shouldn't allow initial_ar_order <= ma_order
+    hannan_rissanen(endog, ar_order=0, ma_order=2, initial_ar_order=1)
+    # TODO: shouldn't allow initial_ar_order >= dataset
+    hannan_rissanen(endog, ar_order=0, ma_order=2, initial_ar_order=20)
+
+
+@pytest.mark.todo
```
jbrockmendel: I like this idea because it is very easy to grep for.  Can you do `mark.todo(explanation)`?

jbrockmendel: Looks like this also needs pytest.mark.smoke (at least for now)

statsmodels/tsa2/sarimax/estimators/tests/test_gls.py
```diff
@@ -0,0 +1,237 @@
...
+    # exog = pd.Series((sbl.index > '1983-01-01').astype(int),
+    #                  index=sbl.index).diff(12).iloc[12:]
+
+    # res, _ = gls(endog, exog, order=(0, 0, 12), max_iter=3)
+
+    # assert_allclose(res.exog_params, -328.45, atol=1e-2)
+    # assert_allclose(res.ma_params,
+    #                 [.219, .098, .031, .064, .069, .111, .081,
+    #                  .057, .092, -0.28, .183, -.672], atol=1e-3)
+    # assert_allclose(res.sigma2, 12581, atol=1)
```
jbrockmendel: don't leave commented-out code

statsmodels/tsa2/sarimax/estimators/tests/results/results_burg.R
```diff
@@ -0,0 +1,26 @@
+library(itsmr)
```
jbrockmendel: i wonder if we can smoke-test R scripts like this.  best-case would be to check their output is unchanged, but that would require more machinery.  I'm just thinking "this still runs correctly" tests

jbrockmendel: (not necessarily for this PR, just thinking out loud)

statsmodels/tsa2/sarimax/estimators/statespace.py
```diff
@@ -0,0 +1,123 @@
...
+                  enforce_invertibility=spec.enforce_invertibility,
+                  concentrate_scale=spec.concentrate_scale)
+    if fit_kwargs is None:
+        fit_kwargs = {}
+    fit_kwargs.setdefault('disp', 0)
+    res_ss = mod.fit(start_params=start_params, **fit_kwargs)
+
+    # Construct results
+    p.params = res_ss.params
+    res = Bunch({
```
jbrockmendel: we could consider using `types.SimpleNamespace` for these

ChadFulton:
> Should add a lout FutureWarnign to ARMA and ARIMA __init__ that these are deprecated

Yes definitely. 

> , and statsmodels.NEWLOC.ARMA/ARIMA should be used. These functions will see no future development, although they won't go away anytime soon.

> Are you really planning on having tsa2? I not a big fan of that.

Definitely not `tsa2`, but not sure what the correct answer is. It seems like there are a couple of options:

1. Put it in a `tsa` subpackage: called with `mod = sm.tsa.NEWLOC.ARIMA(...)`
2. Put it in `tsa` with a different name: called with `mod = sm.tsa.ARIMAModel(...)`
3. Put it somewhere that's not `tsa`: called with `mod = sm.NEWLOC.ARIMA(...)`
4. Put it in `tsa` and call it through a new argument in the existing class: `mod = sm.tsa.ARIMA(..., future=True)` or whatever the argument would be.

I like (4) the best from an ease of use standpoint: at some point, we can just change the default of `future` from `False` to `True` to force people to use the new model, and then eventually remove the `future` argument and the old models altogether. But I don't know if this is a headache in other ways (i.e. using the `__new__` method, etc.).

Personally, I would prefer not to do (2), since we already have so many of these model classes hanging out in `tsa`. I guess either (1) or (3) would work, but I can't think of a good name right now.

In any case, presumably the deprecation would look something like:

- v0.11 - add new class for optional use and `FutureWarning` noting that old classes/behavior will not be the default option in v0.12 (at least ARMA/ARIMA, if we're still keeping AR as in your new PR) 
- v0.12 - switch default to new class, change `FutureWarning` to note that old classes will be removed in v0.13
- v0.13 - remove warnings and old classes.

Or is that too fast? Do we need to wait to change the default in v0.13 and then drop in v0.14?

statsmodels/tsa2/sarimax/estimators/gls.py
ChadFulton: like a subclass of `GLS`, so `GLSARIMA` or something? I guess that might be a more natural place for it. I'll take a look to see if that causes any trouble.

statsmodels/tsa2/sarimax/estimators/tests/test_gls.py
ChadFulton: Good point, I'll remove it and make an issue instead.

statsmodels/tsa2/sarimax/estimators/tests/test_hannan_rissanen.py
ChadFulton: > Can you do mark.todo(explanation)?

Yes, good idea.

> Looks like this also needs pytest.mark.smoke (at least for now)

Why? It's not a smoke test AFAICT?

statsmodels/tsa2/sarimax/estimators/tests/test_yule_walker.py
ChadFulton: Good idea.

ChadFulton:
> the small fixups in mlemodel.py could be totally independent

True, and unfortunate.

> burg.py merits a discussion on how it relates to the existing pacf_burg

It's just a light wrapper around it (well around `linear_model.burg`), which is noted in the function docstring. Where does it need more discussion?

> durbin_levinson.py merits a discussion on how it relates to the existing levinson_durbin

This is very true, and probably should just be a wrapper around `levinson_durbin`.

> And definitely pls add it to the strict-linting in lint.sh

Sure, I'll take a look at that.

ChadFulton:
> @ChadFulton to the extent that this can be broken into smaller pieces, it would make it easier on reviewers.

I guess maybe `gls` and `hannan_rissanen` could have been broken off, but they still depend on `SARIMAXSpecification` and `SARIMAXParams`. The only other big piece then is `model.py`. I'm not sure I see a way to make particular PR substantially easier to review (but then I know that that's always the perspective of the person submitting the PR :)).

statsmodels/tsa2/sarimax/estimators/statespace.py
bashtage: Should make `Bunch = types.SimpleNamespace`

bashtage:
My Preferences:

1. Put it in tsa with a different name: called with mod = sm.tsa.ARIMAModel(...). **Edit**: Maybe someday once ARIMA is gone for a cycle it could take over the name. **Orig: 2**
2. Put it in a tsa subpackage: called with mod = sm.tsa.NEWLOC.ARIMA(...) **Orig: 1**
3. Put it somewhere that's not tsa: called with mod = sm.NEWLOC.ARIMA(...) **Edit**: This would probably be its actual location, with a pointer in the existing `ARMA` and `ARIMA` to use statsmodels.tsa.FULL.PATH.ARIMA. **Orig: 3**
4. Put it in tsa and call it through a new argument in the existing class: mod = sm.tsa.ARIMA(..., future=True) or whatever the argument would be. **Orig: 4**

1 and 2 are close, and I'm happy with either. 3 and 4 are both worse IMO. Gettign a good choice for `NEWLOC` is important -- short and clear ideally. 

> I like (4) the best from an ease of use standpoint: at some point, we can just change the default of `future` from `False` to `True` to force people to use the new model, and then eventually remove the `future` argument and the old models altogether. But I don't know if this is a headache in other ways (i.e. using the `__new__` method, etc.).

4 is a bit of a mess since it requires taking over `__new__` from the class and doing some magic to swap in one of the two ARIMA classes.  IMO not worth the complexity (see how `ARIMA` already does this with `ARMA` is d=0, and the long-lasting bug it created that I only recently fixed.

> Or is that too fast? Do we need to wait to change the default in v0.13 and then drop in v0.14?

Depends on how long the releases occur. I would like a release (0.11) this fall (October) to correspond to my teaching with this in it (and AutoReg).

bashtage:
> 3\. Put it somewhere that's not tsa: called with mod = sm.NEWLOC.ARIMA(...) **Edit**: This would probably be its actual location, with a pointer in the existing `ARMA` and `ARIMA` to use statsmodels.tsa.FULL.PATH.ARIMA. **Orig: 3**

This is one advantage -- the location exposed is the long-run location so it never gets duplicated.

statsmodels/tsa2/sarimax/estimators/gls.py
ChadFulton: Looked into this, but personally don't feel like it fits well in `linear_models` because it is so deeply integrated with the other ARIMA functions (e.g. it needs to call an ARIMA estimator).  We can move it later, or add a `GLSARIMA` wrapper later if it seems useful.

statsmodels/tsa2/sarimax/estimators/statespace.py
ChadFulton: Sounds like a good idea as a general principle, but will not do it in this PR.

ChadFulton:
To move forward on this:

- I think it's easiest to go with a `tsa` subpackage, and I think that should be `tsa.arima`. So that the main entry point would be `sm.tsa.arima.ARIMA`.

- I will keep Brockwell and Davis datasets in the current location (that will be `tsa.arima.datasets`), but not to make them part of the API (i.e. they will, for now, only be used as datasets for unit tests).  I will open an issue for moving them to `sm.datasets` (can be one of the rare "Good as first PR" issues).

Force pushed with these changes. It would be great to merge this soon. Once we see how things go, we can decide whether we need to put a more explicit "Experimental" warning on these models for the 0.11 release. Unfortunately, I doubt that these will have been in long enough to merit deprecating `sm.tsa.AR`, `sm.tsa.ARIMA`, etc. in favor of these new versions.

ChadFulton:
> I think it's easiest to go with a `tsa` subpackage, and I think that should be `tsa.arima`. So that the main entry point would be `sm.tsa.arima.ARIMA`.

If this is acceptable, then we could also use this for the new AR models of #6087, i.e. `sm.tsa.ar.AR`.

ChadFulton:
Unless there are any concerns, I will merge this once I get Windows to stop complaining about precision issues in tests (so hopefully tonight).

jbrockmendel:
I'll take another look this evening, if you're OK with "tonight" being Pacific Time

ChadFulton:
> I'll take another look this evening, if you're OK with "tonight" being Pacific Time

That would be much appreciated, thanks!  (I know it's a huge PR so whatever you have time for would be a great check).

statsmodels/tsa/arima/estimators/gls.py
```diff
@@ -0,0 +1,305 @@
...
+        `max_iter` and `tolerance` arguments.
+    max_iter : int, optional
+        Maximum number of feasible GLS iterations. Default is 50. If `n_iter`
+        is set, it overrides this argument.
+    tolerance : float, optional
+        Tolerance for determining convergence of feasible GSL iterations. If
+        `iter` is set, this argument has no effect.
+        Default is 1e-8.
+    arma_estimator : string, optional
+        The estimator used for estimating the ARMA model. This option should
```
jbrockmendel: This is a pretty big block of text.  Might want to format with sub-sections for the lists?

statsmodels/tsa/arima/estimators/gls.py
```diff
@@ -0,0 +1,305 @@
...
+    be issued to alert the user.
+
+    TODO: improve "Returns" documentation for `other_results`.
+
+    TODO: Add documentation for why we have e.g. `include_constant` but not
+    other trend specifications (e.g. that it is to maintain consistency with
+    estimation methods with assumptions that require demeaned series) or else
+    revisit this.
+
+    TODO: add support for fixed parameters
```
jbrockmendel: is this TODO for the docstring or the function?

statsmodels/tsa/arima/estimators/hannan_rissanen.py
```diff
@@ -0,0 +1,196 @@
...
+    possible. See test_hannan_rissanen::test_brockwell_davis_example_517 for
+    an example of how to compute this variance manually.
+
+    This procedure assumes that the series is stationary.
+
+    TODO: better warnings / errors when series are short relative to lag length
+
+    TODO: add support for fixed parameters
+
+    TODO: seems like we could add support for seasonal parameters here?
```
jbrockmendel: same question about placement of TODO comments.  (going to omit this if I see it below)

statsmodels/tsa/arima/estimators/hannan_rissanen.py
```diff
@@ -0,0 +1,196 @@
...
+
+    p = SARIMAXParams(spec=spec)
+
+    nobs = len(endog)
+    max_ar_order = spec.max_ar_order
+    max_ma_order = spec.max_ma_order
+
+    # Default initial_ar_order is as suggested by Gomez and Maravall (2001)
+    if initial_ar_order is None:
+        initial_ar_order = max(np.floor(np.log(nobs)**2).astype(int).item(),
```
jbrockmendel: why `.item()`

statsmodels/tsa/arima/estimators/tests/test_burg.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def test_invalid():
+    endog = np.arange(2) * 1.0
+    assert_raises(ValueError, burg, endog, ar_order=2)
+    assert_raises(ValueError, burg, endog, ar_order=-1)
+    assert_raises(ValueError, burg, endog, ar_order=1.5)
+
+    endog = np.arange(10) * 1.0
+    assert_raises(ValueError, burg, endog, ar_order=[1, 3])
```
jbrockmendel: not a big deal, but using `with pytest.raises(ValueError, match="whatever"):` is both more specific (bc of `match`) and more copy-paste friendly

statsmodels/tsa/arima/estimators/tests/test_durbin_levinson.py
```diff
@@ -0,0 +1,93 @@
+import numpy as np
+
+import pytest
+from numpy.testing import assert_allclose, assert_raises
+
+from statsmodels.tsa.innovations.arma_innovations import arma_innovations
+from statsmodels.tsa.arima.datasets.brockwell_davis_2002 import dowj, lake
+from statsmodels.tsa.arima.estimators.durbin_levinson import durbin_levinson
+
+@pytest.mark.low_precision('Test against Example 5.1.1 in Brockwell and Davis'
```
jbrockmendel: i expect strict linting would want another newline before this.  are these files not in the strict list?

statsmodels/tsa/arima/estimators/tests/test_durbin_levinson.py
```diff
@@ -0,0 +1,93 @@
...
+
+def test_nonstationary_series():
+    # Test against R stats::ar.yw; see results/results_yw_dl.R
+    endog = np.arange(1, 12) * 1.0
+    res, _ = durbin_levinson(endog, 2, demean=False)
+
+    desired_ar_params = [0.92318534179, -0.06166314306]
+    assert_allclose(res[2].ar_params, desired_ar_params)
+    # desired_sigma2 = 15.36526603
+    # assert_allclose(res[2].sigma2, desired_sigma2)
```
jbrockmendel: pls dont leave commented-out

statsmodels/tsa/arima/estimators/tests/test_hannan_rissanen.py
```diff
@@ -0,0 +1,105 @@
...
+    # TODO: shouldn't allow initial_ar_order <= ar_order
+    hannan_rissanen(endog, ar_order=2, ma_order=0, initial_ar_order=1)
+    # TODO: shouldn't allow initial_ar_order <= ma_order
+    hannan_rissanen(endog, ar_order=0, ma_order=2, initial_ar_order=1)
+    # TODO: shouldn't allow initial_ar_order >= dataset
+    hannan_rissanen(endog, ar_order=0, ma_order=2, initial_ar_order=20)
+
+
+@pytest.mark.todo('Improve checks on valid order parameters.')
+@pytest.mark.xfail
```
jbrockmendel: can the todo be made redundant by putting that info in the `reason=` for the xfail?

statsmodels/tsa/arima/estimators/tests/test_hannan_rissanen.py
```diff
@@ -0,0 +1,105 @@
...
+def test_invalid_orders():
+    endog = np.arange(2) * 1.0
+    # TODO: shouldn't allow ar_order >= dataset
+    hannan_rissanen(endog, ar_order=2)
+    # TODO: shouldn't allow ma_order >= dataset
+    hannan_rissanen(endog, ma_order=2)
+
+
+@pytest.mark.todo('Improve checks on valid order parameters.')
```
jbrockmendel: is the idea here that the calls below should raise?

statsmodels/tsa/arima/estimators/tests/test_hannan_rissanen.py
```diff
@@ -0,0 +1,105 @@
...
+    hannan_rissanen(endog, ar_order=[1, 4], ma_order=[1, 3])
+    hannan_rissanen(endog, ar_order=[0, 0, 1])
+    hannan_rissanen(endog, ma_order=[0, 0, 1])
+    hannan_rissanen(endog, ar_order=[0, 0, 1], ma_order=[0, 0, 1])
+
+    hannan_rissanen(endog, ar_order=0, ma_order=0)
+
+
+@pytest.mark.todo('Improve check/conversions on datatype.')
+@pytest.mark.xfail
```
jbrockmendel: ditto w/r/t todo being merged into xfail

statsmodels/tsa/arima/estimators/tests/test_innovations.py
```diff
@@ -0,0 +1,305 @@
...
+                           start_params=initial[1].params)
+
+    assert_allclose(p.ma_params, -0.818, atol=1e-3)
+
+    # TODO: the test for sigma2 fails; we get 2040.85 whereas BD reports
+    # 2040.75. Unclear if this is optimizers finding different maxima, or a
+    # reporting error by BD (i.e. typo where the 8 got reported as a 7). Should
+    # check this out with ITSM program. NB: state space also finds 2040.85 as
+    # the MLE value.
+    # assert_allclose(p.sigma2, 2040.75, atol=1e-2)
```
jbrockmendel: pls dont leave commented-out

jbrockmendel: maybe this one is OK given the preceeding comment

jbrockmendel:
Ran out of gas before getting through all of it.  A few comments about the pytest.mark.todo marker possibly being redundant, otherwise looks really good, particularly the documentation on where the test results are coming from.

Big picture I'd be more comfortable putting these in place if there was an agreed-upon gameplan for getting rid of the old versions (and potential overlap with #6087).  But you've been pretty patient about this so I can't in good conscience ask you to wait any more on my account.

statsmodels/tsa/arima/estimators/tests/test_hannan_rissanen.py
ChadFulton: Sure, we can remove the `todo` marks and just use the `xfail` marks.

statsmodels/tsa/arima/estimators/tests/test_durbin_levinson.py
ChadFulton: Thanks - I think they are missing it because of the decorator (i.e. when I run `pycodestyle` manually it doesn't pick up anything). I added the directory to the linter list - is that enough?

statsmodels/tsa/arima/estimators/hannan_rissanen.py
ChadFulton: Hmm good question. I wanted an `int` at some point, but can't remember why. Tests pass without it, so I'll remove.

statsmodels/tsa/arima/estimators/gls.py
ChadFulton: Good idea - lets see if I can get the RST formatting right...

statsmodels/tsa/arima/estimators/gls.py
ChadFulton: I'll move these out and make issues.

statsmodels/tsa/arima/estimators/hannan_rissanen.py
ChadFulton: Thanks.

statsmodels/tsa/arima/estimators/tests/test_durbin_levinson.py
ChadFulton: good call, thanks

statsmodels/tsa/arima/estimators/tests/test_hannan_rissanen.py
ChadFulton: Yep.

statsmodels/tsa/arima/estimators/tests/test_hannan_rissanen.py
ChadFulton: thanks

statsmodels/tsa/arima/estimators/tests/test_innovations.py
ChadFulton: I will move it to an xfail, thanks

ChadFulton:
> Ran out of gas before getting through all of it. A few comments about the pytest.mark.todo marker possibly being redundant, otherwise looks really good, particularly the documentation on where the test results are coming from.

Thanks

> Big picture I'd be more comfortable putting these in place if there was an agreed-upon gameplan for getting rid of the old versions (and potential overlap with #6087).

But I think there is a game plan for getting rid of older versions - once we are comfortable that this code works, we will follow this deprecation strategy:

In any case, presumably the deprecation will look like:

- v0.x - add new class for optional use and `FutureWarning` noting that the API endpoint `sm.tsa.ARIMA` will switch from `sm.tsa.arima_model.ARIMA` to `sm.tsa.arima.ARIMA` in v0.x+1.
- v0.x+1 - actually make the switch, change the `FutureWarning` to note that old classes (`sm.tsa.arima_model.ARIMA`) will be removed in v0.x+2
- v0.x+2 - remove warnings and old classes.

The only thing that's unclear to me is whether we can set x=11, since if we want to release in October that may not be enough time to make sure the new classes are robust. However, that is what I hope for.

> But you've been pretty patient about this so I can't in good conscience ask you to wait any more on my account.

:) this is going in tonight.

ChadFulton:
Merging when green.

ChadFulton:
Thanks @jbrockmendel and @bashtage!

Issue #5828, opened 2019-06-05
DimReductionRegression has no docstring


bashtage:
`DimReductionRegression` and related models have no documentations

@kshedden are these intended to be public, or are they private?

bashtage:
I suppose this is a base class and is actually private.  Are SIR and other classes meant to be public?

PR #5830, opened 2019-06-05
DOC: Add docstring for base class

kshedden:
#5828


bashtage:
How do you feel about renaming SIR to SlicedInverseReg or similar? It could
be quite hard to find SIR imo.

PR #5831, opened 2019-06-05
PERF: Import speed

jbrockmendel:
xref #3082 (does not close)

### DIFF
--- a/lint.sh
+++ b/lint.sh
@@ -79,6 +79,7 @@ if [ "$LINT" == true ]; then
         statsmodels/stats/tests/test_qsturng.py \
         statsmodels/stats/tests/test_regularized_covariance.py \
         statsmodels/stats/tests/results/ \
+        statsmodels/tests/ \
         statsmodels/tools/linalg.py \
         statsmodels/tools/web.py \
         statsmodels/tools/tests/test_linalg.py \
--- a/lint.sh
+++ b/lint.sh
@@ -95,6 +95,7 @@ if [ "$LINT" == true ]; then
         statsmodels/tsa/vector_ar/tests/results/ \
         statsmodels/tsa/statespace/ \
         statsmodels/tsa/tests/results/ \
+        statsmodels/__init__.py \
         statsmodels/conftest.py \
         statsmodels/tools/sm_exceptions.py \
         examples/ \
--- a/statsmodels/__init__.py
+++ b/statsmodels/__init__.py
@@ -18,7 +18,7 @@ if debug_warnings:
 
 def test(*args, **kwargs):
     from .tools._testing import PytestTester
-    tst = PytestTester()
+    tst = PytestTester(package_path=__file__)
     return tst(*args, **kwargs)
 
 
--- a/statsmodels/__init__.py
+++ b/statsmodels/__init__.py
@@ -1,22 +1,11 @@
-from __future__ import print_function
-from .compat import PY3
 
-from warnings import simplefilter
-
-from .tools._testing import PytestTester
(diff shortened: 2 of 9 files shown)


bashtage:
cmd on Windows needs " around the command, as in `python -c "import statsmodels"`

jbrockmendel:
Ended up just skipping on windows, since no variant of the command worked.

I think this will also make the test in #5235 pass.

jbrockmendel:
de-facto green

bashtage:
Will wait until after release.

bashtage:
I'll use this tag to remember PRs that are approved but deferred.

bashtage:
Any idea how much difference this makes?

jbrockmendel:
Running `python3 -X importtime -c "import statsmodels.api as sm"` on master takes 954861 microseconds, on this branch 853969.  So a little more than 10%.

bashtage:
Probably worth getting in before the next release then. 

Have you looked into how pandas is getting into the main namespace?  I think it happens though compat (which grabs a bunch of stuff)

jbrockmendel:
> Have you looked into how pandas is getting into the main namespace?

For `import statsmodels.api.as sm` my assumption is that it is imported in too many places to make lazifying it reasonable.  For `import statsmodels`, I think it came in through `tools._testing`, where IIRC it is just for assert_foo_equal and could totally be removed.

jbrockmendel:
Trying this out, lazifying pandas imports in a few places gets the import time for "import statsmodels" from 447082 microseconds down to 196531.  Down to 154490 if we get `tools.tools` imports out of `tools.__init__`, but that seems unlikely.

bashtage:
That seems like a worthy refactor.  One could consider moving tools._testing (or PyTestRunner in particular) to somewhere that prevents many imports. The privacy of this module allows for immediate refactoring w/o any issues.

jbrockmendel:
> One could consider moving tools._testing

I looked at something like this and it doesn't help because we also import from `statsmodels.tools.sm_exceptions` into `statsmodels.__init__`

bashtage:
Yes, that filtering is unfortunate.  Is it even needed?  Aren't warnings always shown when given, except things like deprecation?

bashtage:
No apparently.  I wonder if we could move that to base.__init__.  It seems like base.__init__ would always be imported before any model could be estimated that would produce those errors.

bashtage:
Actually, just move it to 

statsmodels.tools.sm_exceptions after the exceptions.

This file is always imported before any of these exceptions are given (obviously).  Also a natural place to have it.

jbrockmendel:
OK, will update.

In pandas there's a discussion about implementing asv for import times.  When that gets figured out I'll port whatever is implemented here.  (Will start putting together an asv suite here before that hopefully)

jbrockmendel:
Just pushed, for "import statsmodels" im seeing 84628 microseocnds vs 395499 on master.

bashtage:
I'm going to merge this for the next release since it has been a pain point.

bashtage:
Once all green.

bashtage:
You need to pass args and kwargs through to tst

jbrockmendel:
> You need to pass args and kwargs through to tst

Updated.

Any ballpark on how long the RC period will be?

bashtage:
One more week. PytestTester is having problems detecting the path.  Should modify it to take an argument `package_path=None` (which triggers teh current check), so that __file__ can be passed in from `statsmodels/__init__.py`

bashtage:
Thanks. Improving import perf is very useful.

PR #5837, opened 2019-06-07
CLN: invalid escapes in linear_model

jbrockmendel:
These must have been introduced recently, since at the time #5692 was made that fixed all remaining invalid escapes.  These will keep slipping through the cracks.


jbrockmendel:
@bashtage looking at the remaining complaints in this file, many of them are long TeX lines in docstrings.  Is there a preferred method for splitting those nicely?  The alternative is a ` # noqa:E501` at the end of the docstring, but that ignores the entire docstring and not just the small number of TeX lines.

PR #5847, opened 2019-06-08
BUG: Allow NumPy ints

### LINKED ISSUE #5570: BUG: SARIMAX does not accept NumPy integer periodicty
Passing a NumPy integer (instead of a Python integer) to the periodicty in the `seasonal_order` argument of `SARIMAX` raises the exception:

> TypeError: object of type 'numpy.int64' has no len()

The following minimal code snippet reproduces the problem:

```
from numpy import *
from statsmodels.tsa.statespace.sarimax import SARIMAX

t = linspace(0, 8.0*pi, num=40 + 1)
y = sin(t) + random.normal(scale=0.5, size=len(t))
model = SARIMAX(
    y,
    seasonal_order=(0, 1, 0, int64(10))).fit()
```

The problem is in fact related to `isinstance(int64(), int)` being `False`, see also https://github.com/numpy/numpy/issues/2951.
I am ussing statsmodels version 0.9.0, but looking at the master branch https://github.com/statsmodels/statsmodels/blob/master/statsmodels/tsa/statespace/tools.py#L181 I expect that this problem is not limited to my version only.
As a quick fix in my own code I do:
```
seasonal_order = tuple(int(i) for i in seasonal_order)
```
but it would be desirable if SARIMAX also works with NumPy integers.

bashtage:
Allow NumPy ints in companion matrix and other tsa functions

closes #5570

- [X] closes #5570
- [X] tests added / passed. 

### DIFF
--- a/statsmodels/tsa/statespace/initialization.py
+++ b/statsmodels/tsa/statespace/initialization.py
@@ -279,7 +279,8 @@ class Initialization(object):
         # Construct the index, using a slice object as an intermediate step
         # to enforce regularity
         if not isinstance(index, slice):
-            if isinstance(index, int):
+            if isinstance(index, (int, np.integer)):
+                index = int(index)
                 if index < 0 or index >= self.k_states:
                     raise ValueError('Invalid index.')
                 index = (index, index + 1)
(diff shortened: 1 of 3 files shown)


ChadFulton:
LGTM, thanks!!

PR #5852, opened 2019-06-08
ENH: Dimension reduction for covariance matrices

kshedden:
This doesn't need to go in right away.

### DIFF
--- a/statsmodels/regression/dimred.py
+++ b/statsmodels/regression/dimred.py
@@ -1,7 +1,8 @@
 import numpy as np
 from statsmodels.base import model
 import statsmodels.base.wrapper as wrap
-
+import pandas as pd
+import warnings
 
 class _DimReductionRegression(model.Model):
     """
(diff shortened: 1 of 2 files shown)


statsmodels/regression/dimred.py
```diff
@@ -1,6 +1,8 @@
 import numpy as np
 from statsmodels.base import model
 import statsmodels.base.wrapper as wrap
+import pandas as pd
+import warnings
```
jbrockmendel: Stdlib (which autocorrect thinks should be Stalin apparently), then third party, then first party, pls

statsmodels/regression/dimred.py
```diff
@@ -252,7 +254,201 @@ class DimReductionResultsWrapper(wrap.ResultsWrapper):
 wrap.populate_wrapper(DimReductionResultsWrapper,  # noqa:E305
                       DimReductionResults)
 
+
+class CovarianceReduction(_DimReductionRegression):
+    """
+    Dimension reduction for covariance matrices (CORE).).
```
jbrockmendel: Extra paren?

PR #5893, opened 2019-06-28
BUG: exponential smoothing - damped trend gives incorrect param, predictions

ChadFulton:
Exponential smoothing with a damped trend gives the wrong result for `res.params['initial_slope']` and gives wrong predictions.

The problem is the initial trend is accidentally multiplied by the damping parameter before the results object is created. This means that when predictions are made later, they will be based on the wrong initial trend.

This PR also fixes the problem that `sm.tsa.Holt` silently ignores the `initial_level` and `initial_slope` arguments.


ChadFulton:
Unrelated errors, merging.

PR #5894, opened 2019-06-29
MAINT: Bump dependencies.

ChadFulton:
Closes #5839 

(Note: still happy to change these if there are objections / concerns)

Drops support for Python 3.4 and updates dependencies to:

- Numpy >= 1.14 (January 2018)
- Scipy >= 1.0.0 (October 2017)
- Pandas >= 0.21 (October 2017)
- Cython >= 0.27 (September 2017)
- Patsy >= 0.5 (January 2018)

I think I updated all the relevant files, and I bumped some of the versions in Travis that we above the previous minimums to be above these new minimums too. Hope I didn't mess anything up.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,26 +37,26 @@ matrix:
     - python: 2.7
       env:
         - PYTHON=2.7
-        - NUMPY=1.13
+        - NUMPY=1.15
         - MATPLOTLIB=2.0
         - COVERAGE=true
       # Python 2.7 + baseline packages
     - python: 2.7
       env:
         - PYTHON=2.7
-        - NUMPY=1.12
+        - NUMPY=1.14
         - BLAS= # Do not specify blas in this config due to conflict
-        - SCIPY=0.19
-        - PANDAS=0.20
+        - SCIPY=1.0
+        - PANDAS=0.21
         - USEMPL=false
         - LINT=true
-    # Python 3.4 + baseline packages
-    - python: 3.4
+    # Python 3.5 + baseline packages
+    - python: 3.5
       env:
-        - PYTHON=3.4
-        - NUMPY=1.11
-        - SCIPY=0.18
-        - PANDAS=0.19
+        - PYTHON=3.5
+        - NUMPY=1.14
+        - SCIPY=1.0
+        - PANDAS=0.21
         - MATPLOTLIB=1.5
     # Python 3.7 + cutting edge packages
     - python: 3.7
@@ -78,8 +78,8 @@ matrix:
     - python: 3.5
       env:
         - PYTHON=3.5
-        - NUMPY=1.13
-        - SCIPY=1.0
+        - NUMPY=1.15
+        - SCIPY=1.1
         - PANDAS=0.22
         - MATPLOTLIB=2.0
         - LINT=true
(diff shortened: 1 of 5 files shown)


ChadFulton:
Obviously this does not need to be merged quickly. I will tentatively plan to merge it in a week, unless there are concerns.

thequackdaddy:
Out of curiosity, I used this Google Big Query:

```
#standardSQL
SELECT file.filename, COUNT(*) AS num_downloads
FROM `the-psf.pypi.downloads*`
WHERE file.project = 'statsmodels'
  -- Only query the last 30 days of history
  AND _TABLE_SUFFIX
    BETWEEN FORMAT_DATE(
      '%Y%m%d', DATE_SUB(CURRENT_DATE(), INTERVAL 30 DAY))
    AND FORMAT_DATE('%Y%m%d', CURRENT_DATE())
    AND file.version ='0.10.0'
GROUP BY file.filename
ORDER BY file.filename
```

https://bigquery.cloud.google.com/dataset/the-psf:pypi

And it gave me this table... 

https://docs.google.com/spreadsheets/d/1jkBYy12yOTReBywUx2DCu1HRmFj16zoFQoW8BOOZ_lk/edit?usp=sharing

Unless I'm doing something wrong (not unlikely) our #1 platform for downloads is....  cpython 2.7 on 64 bit linux.

Over 30% of wheels are python 2.7 wheels.

bashtage:
Most importantly Python should be >=3.5.  It is time for Python 2.7 to end, it has for all other projects I follow. This said, we shouldn't accept refactoring that remove 2.7 code until we are finished backporting to 0.10.1, since it could make backporting harder.  But we can drop it from CI and new code can be 3.5+. 

IMO Cython should be 0.29+.  This is a developer-only tool, and isn't needed to install or run.  IIRC It is also necessary to use Cython 0.29.10 (or later) for Python 3.8 alpha.

bashtage:
> Over 30% of wheels are python 2.7 wheels.

This must be dominated by CI runs.

ChadFulton:
> IMO Cython should be 0.29+. This is a developer-only tool, and isn't needed to install or run. IIRC It is also necessary to use Cython 0.29.10 (or later) for Python 3.8 alpha.

Thanks for pointing this out. I guess we should plan to potentially bump Cython again when Python 3.8 comes out in October (or whenever it finally does) in case of additional changes that require new Cython releases.

In the meantime, what should we do? Bump all the way to 0.29.10? That only came out this month, but I guess it's only an optional dependency.

> Most importantly Python should be >=3.5. It is time for Python 2.7 to end, it has for all other projects I follow. This said, we shouldn't accept refactoring that remove 2.7 code until we are finished backporting to 0.10.1, since it could make backporting harder. But we can drop it from CI and new code can be 3.5+.

This is a good point. I see that @josef-pkt raised the question of dropping support for Python 2.7 in the v0.11 release, and I think (but can't find it now) that we sort-of indicated that v0.10 was going to be the last Python 2.7 compatible release. If this is what we're doing, then I agree it would be best to make the switch now to allowing Python 3.5+ code, and drop Python 2.7 from CI.

Happy to update this PR to do that, but will wait a bit to see if there are other views.

bashtage:
Perhaps keep one 2.7 on Travis for now to make sure that all commits are 2.7 compatible for a little while.  This will help a lot with backports.

bashtage:
For Cython could just do >=0.29, which will result in 0.29.latest in most cases.

bashtage:
I think MPL needs a bump too.  Perhaps >=2.2 (2.2 is a LTS branch, which is a plus).

PR #5897, opened 2019-06-29
MAINT: Use pytest.raises to check error message

thequackdaddy:
- [X] tests added / passed. 

<details>

It looks like pytest 5 changed the way some errors are reported. Small syntax change and I think this should fix most of the failing tests.
</details>

### DIFF
--- a/statsmodels/regression/tests/test_tools.py
+++ b/statsmodels/regression/tests/test_tools.py
@@ -45,18 +45,18 @@ class TestMinimalWLS(object):
 
     @pytest.mark.parametrize('bad_value', [np.nan, np.inf])
     def test_inf_nan(self, bad_value):
-        with pytest.raises(ValueError) as err:
+        with pytest.raises(
+                ValueError,
+                match=r'detected in endog, estimation infeasible'):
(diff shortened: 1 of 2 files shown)


thequackdaddy:
Oh if anyone cares, you can fix the python 3.4 fail by adding a couple lines to the Conda builds...

However, we're likely to drop 3.4 soon, so I won't submit a PR.

https://github.com/statsmodels/statsmodels/compare/master...thequackdaddy:restore_free

statsmodels/regression/tests/test_tools.py
```diff
@@ -51,12 +51,12 @@ def test_inf_nan(self, bad_value):
             _MinimalWLS(endog, self.exog1, self.weights1,
                         check_endog=True, check_weights=True).fit()
         assert err.type is ValueError
-        assert 'endog' in str(err)
+        assert 'endog' in str(err.value)
```
bashtage: Would be better to use match as part of pytest.raises

thequackdaddy: done.

Issue #5844, opened 2019-06-07
Kalman filter as a service (question)


jaideeplamba:
I am trying to implement KalmanFilter as a service for our product. As a basic test scenario I implemented LocalLevelModel as give below. 

My example workflow scenario is as follows:
At time T=0, client will post 5 data points. Kalman will fit and respond with (a) predicted mean for data points (b) forecast for the next 5 data points. 
At time T=5, client will post next 5 data points. Now I want Kalman to use fit from previous run and use that to predict next 5 data points. 

Question:
What attributes from model or result should I save from first run and reuse in the second run. 

Thank-You for your help.

```python
class LocalLevel(sm.tsa.statespace.MLEModel):
    _start_params = [1., 1.]
    _param_names = ['var.level', 'var.irregular']

    def __init__(self, endog):
        super(LocalLevel, self).__init__(
            endog, k_states=1)  # , initialization='diffuse'

        self['design', 0, 0] = 1
        self['transition', 0, 0] = 1
        self['selection', 0, 0] = 1
        # Initialization
        k_states = 1
        initial_state = np.zeros((k_states,))
        initial_state_cov = np.eye(k_states) * 100
        self.initialize_known(initial_state, initial_state_cov)
        self.loglikelihood_burn = 1

    def transform_params(self, unconstrained):
        return unconstrained ** 2

    def untransform_params(self, unconstrained):
        return unconstrained ** 0.5

    def update(self, params, **kwargs):
        params = super(LocalLevel, self).update(params, **kwargs)
        self['state_cov', 0, 0] = params[0]
        self['obs_cov', 0, 0] = params[1]
```

ChadFulton:
Take a look at the example I posted here:

https://github.com/statsmodels/statsmodels/issues/5563#issuecomment-493619923

At time T=0 you want to do the part called "# Fit on the training sample"

At time T=5 you want to do the part called "# Now apply the filter to the next dataset"

and you can continue in a similar way as long as you like.

jaideeplamba:
Thank-You Chad for the quick response.
So the code below works in maintaining the state from one call to another. But why do I keep getting only one predicted value as an output in second pass. What am I missing ?

```python
complete_data_list = [59.040042,66.725715,40.399476,52.249948,48.609610,40.946166,52.468450,61.404242,35.637950,59.458336]
complete_data = pd.DataFrame(complete_data_list)

# Fit on the training sample
data1 = complete_data.iloc[0:5]
[...]

Issue #5665, opened 2019-05-08
Predictions for fitted AR-style models with new endogenous information


patricksheehan:
#### Is your feature request related to a problem? Please describe
I'm having trouble finding a concise way to generate predictions for a fitted SARIMAX-type model with new exogenous AND endogenous data.

Here's a concrete example:
- I make predictions every hour
- I'm able to retrain my models every day
- I get access to new endogenous information after every hour

#### Describe the solution you'd like
I'd like to be able to input an endogenous series in addition to the exogenous series I use for predictions between re-training of the model

#### Describe alternatives you've considered
I could in theory fit a model, pull the coefficients, and manually predict 1 step ahead with a model trained more steps backwards, but this seems like a good feature.

#### Additional context
The motivation is mostly for backtesting/cross-validation. Simulating the online-style environment requires frequent retraining which makes testing over many test periods quite computationally expensive

patricksheehan:
Also if someone has a better intermediate solution that would be welcome!

ChadFulton:
We recently added some new features to make this more convenient. Some examples are available here: https://www.statsmodels.org/devel/examples/notebooks/generated/statespace_forecasting.html

Based on the description here, I think this would be a case for the `extend` results method.

Issue #5563, opened 2019-04-10
SARIMAX: incremental Kalman filter (question)


vss888:
Could you please let me know if it is possible to use the current SARIMAX code to apply Kalman filter incrementally (to make it faster)?

Here is what I mean. Imagine, one fits an SARIMAX model (#1) using data with indices in `[iFit1, iFit2]`, then one wants to use such a model to make predictions for data with indices in `[iFit2+1, iDataEnd]`. To do that in the current code, it seems, one has to create a new SARIMAX model (#2) using data with indices in `[iFit1, iDataEnd]` and apply Kalman filter on the model #2 with parameters taken from model #1. After that one can request predictions from model #2. However, it is expensive computationally to do (for large models) if predictions are done on streaming data (**where new data arrives one data point at a time**) since every time one has to redo all Kalman filtering from `iFit1` to the index `iPred` (the one, at which a prediction is required). Mathematically, Kalman filter is applied incrementally, i.e. if Kalman filtering is done up to index `iPred`, then to do filtering up to index `iPred+1` one just hast to take the state at `iPred` and only apply one step of Kalman filter using a single new data point at `iPred+1`, and so it would be great if one could do it in the code incrementally as well.

Could you please let me know if there is a way to do such incremental Kalman filtering (i.e. one state update at a time) in the current SARIMAX implementation?

Thank you for your help!!!

ChadFulton:
We don't have this built in, but it is a feature that's been requested in the past and in principle wouldn't be too hard to add. I would be happy to look at a PR to add this feature.

vss888:
Thank you very much for your reply, Chad! Such a feature would be incredibly valuable for practical applications of your code. 

Our estimates indicate that the _Kalman filtering part_ of out-of-sample predictions would become from 100,000 to 10,000,000 times faster (in our use cases) if we could do the filtering incrementally. If the _prediction part_ is done independently from the filtering (for example, during the call of `SARIMAXResults.get_prediction`), then it would be only a small addition to the overall time required to make a prediction (about the same time scale as one step of Kalman filtering). Overall, the deployment of our models would become much easier and less computationally intensive, since we would be able to make multiple predictions sequentially instead of parallelizing them.

It would be really great if we could look into a PR to add this feature some time soon. If there is any part of the implementation, which I could help with, please, do let me know and I would be happy to contribute.

vss888:
[...]

PR #5915, opened 2019-07-04
ENH: state space: add methods to apply fitted parameters to new observations or new dataset

ChadFulton:
- [x] Add example notebook
- [x] Related to #5844, #5563, #5287, #3189, #5665, #5412
- [x] tests added / passed. 
- [x] properly formatted commit message

This PR adds three methods to the `MLEResults` class (and so the model subclasses `SARIMAX`, `UnobservedComponents`, `VARMAX`, and `DynamicFactor`):

- `append` : Recreate the results object with new data appended to the original data
- `extend` : Recreate the results object for new data that extends the original data
- `apply` : Apply the fitted parameters to new data unrelated to the original data

We get a number of questions about fitting model parameters on some training set, and then using those parameters to create expanding window h-step-ahead forecasts.

The `apply` method is for when you want to apply the current fitted parameters to an entirely different set of data (I think this is a less common case, but we have gotten some questions about it).

The more common cases use either `append` or `extend`.  The `append` method is straightforward, and basically just creates a new model object with the new observations appended to the end of the original model observations, and then creates a new results object using the original fitted parameters. The Kalman filter / smoother is run over the entire new dataset (both original observations and new observations).

The `extend` method is for the same use case as `append`, except that `extend` only runs the Kalman filter over the new observations. This can improve performance quite a bit if the original dataset is very large.

**Questions**:

- The main question is whether the method names make sense. Are there better names than `append`, `extend`, and `apply`?


ChadFulton:
Latest commit adds an example notebook to the docs, showing how these features can be used in the "cross-validation" (aka pseudo-real time forecast evaluation) type exercises that people have been asking about.

Rendered version of the notebook here: http://nbviewer.ipython.org/13d3b8118f798d56d8c9c976efa239d2

ChadFulton:
Basic outstanding question is API / naming. Currently, we have one new `MLEModel` method `clone` and the the three new `MLEResults` methods `append`, `extend`, and `apply`.

Unless there is any concern now, my plan is to merge tomorrow and assume that we can adjust API if necessary before release in v0.11.0.

PR #5918, opened 2019-07-06
MAINT: Backport "BUG: TVTP for Markov regression"

ChadFulton:
Backport of #5917. Not as straightforward as cherry-picking, since we merged #5826 after releasing v0.10, and that changed some of the filtering code.


ChadFulton:
@bashtage is this approach okay for backporting? will it cause problems somehow since I couldn't cherry-pick the same commits as in master?

bashtage:
I'm not dogmatic about the method, although I appreciate it if you cite the original PR in the 10.1 release doc.

Issue #5925, opened 2019-07-09
Off-by-one error / loose dtype checking in KPSS statistic


timstaley:
#### Describe the bug

Arguably the KPSS code has a couple of issues. Firstly, there's no check whether the input array has suitable dtype, which can result in some weird error messages.
Second, I _think_ there's an off-by-one error in the internal function `_sigma_est_kpss`, but it only actually causes an exception in a weird edge-case highlighted by the object dtype.

(If you look at the indexing here:
https://github.com/statsmodels/statsmodels/blob/master/statsmodels/tsa/stattools.py#L1733
Then you'll see that in the edge case `nobs==lags`, and for the last iteration in the loop we are computing `np.dot` on two empty arrays. The return value in this case is dtype dependent, it defaults to 0 for anything numeric, but if it happens to be an object-array we get None and the line of maths below breaks.)

NB, I don't have access to Kwiatkowski et al. (1992), so can't verify the original equation for myself.

#### Code Sample, a copy-pastable example if possible

```python
import numpy as np
from statsmodels.tsa.stattools import kpss

# Confusing error one:
boring_data = np.full(20, 0).astype(object)
kpss(boring_data, lags="auto")

# Confusing error two, hits possible off-by-one error in _sigma_est_kpss:
weird_data = array([0, 0, 0, 0, 0, 1, 1, 0, 0, 0, 0, 0, 1, 1, 0, 0, 0, 0, 0, 1, 1, 0,
       0, 0, 0, 0, 1, 1, 0, 0, 0, 0, 0, 1, 1, 0, 0, 0, 0, 0, 1, 1, 0, 0,
       0, 0, 0, 1, 1, 0, 0, 0, 0, 0, 1, 1, 0, 0, 0, 0, 0, 1, 1, 0, 0, 0,
       0, 0, 1, 1, 0, 0, 0, 0, 0, 1, 1, 0, 0, 0, 0, 0, 1, 1, 0, 0, 0, 0,
       0, 1, 1, 0, 0, 0, 0, 0, 1, 1, 0, 0, 0, 0, 0, 1, 1, 0, 0, 0, 0, 0,
       1, 1, 0, 0, 0, 0, 0, 1, 1, 0, 0, 0, 0, 0, 1, 1, 0, 0, 0, 0, 0, 1,
       1, 0, 0, 0, 0, 0, 1, 1, 0, 0, 0, 0, 0, 1, 1, 0, 0, 0, 0, 0, 1, 1,
[...]

PR #5928, opened 2019-07-11
ENH: Add array_like function to simplify input checking

bashtage:
Add a lightweight array subclass to simplify input validation and
standardization

- [X] tests added / passed. 

### DIFF
--- a/statsmodels/tsa/arima_model.py
+++ b/statsmodels/tsa/arima_model.py
@@ -17,6 +17,7 @@ from scipy.signal import lfilter
 from numpy import dot, log, zeros, pi
 from numpy.linalg import inv
 
+
 from statsmodels.tools.decorators import cache_readonly
 import statsmodels.tsa.base.tsa_model as tsbase
 import statsmodels.base.wrapper as wrap
@@ -29,6 +30,7 @@ from statsmodels.tsa.vector_ar import util
 from statsmodels.tsa.ar_model import AR
 from statsmodels.tsa.arima_process import arma2ma
 from statsmodels.tools.numdiff import approx_hess_cs, approx_fprime_cs
+from statsmodels.tools.validation import array_like
 from statsmodels.tsa.kalmanf import KalmanFilter
 
 _armax_notes = r"""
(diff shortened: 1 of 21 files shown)


bashtage:
Input validation is a real pain point for me, especially when writing light-weight user-focused functions that don't pass through base.Model.  This is the start of some validation classes that can be used as 1 liners to provide users with good information about requires shape. ndim, etc.  This method also allows other inputs types (beyond ndarray and pandas) to be added across the entire code base wherever it is used.

bashtage:
Probably closes #5925

jbrockmendel:
If it’s just/mostly for validation, what do we get from making a subclass instead of a function?  My understanding is ndarray subclass support can be fragile

bashtage:
I wrote a function version which is perfectly fine for unifying checks and conversion.  It didn't solve issues with handling pandas vs numpy and so I went for a very thin subclass which provides some pandas-focused services (1 for now).  This is so lightweight that I don't think the issues with subclasses manifest here.  For example, this subclass never propagates by construction (since it doesn't make sense).

bashtage:
You are probably right that it isn't worth the effort since I have found some of the gotchas that aren't really documented AFAICT.  Probably safer to use a function.

statsmodels/tsa/arima_model.py
```diff
@@ -421,9 +423,7 @@ def __init__(self, endog, order, exog=None, dates=None, freq=None,
                  missing='none'):
         super(ARMA, self).__init__(endog, exog, dates, freq, missing=missing)
         # GH 2575
-        _endog = endog if hasattr(endog, 'ndim') else np.asarray(endog)
-        if (_endog.ndim == 2 and _endog.shape[1] != 1) or _endog.ndim > 2:
-            raise ValueError('endog must be 1-d or 2-d with 1 column')
+        array_like(endog, 'endog')
```
jbrockmendel: should this be `endog = array_like(endog, 'endog')`?

statsmodels/tsa/arima_model.py
```diff
@@ -1535,7 +1534,7 @@ def forecast(self, steps=1, exog=None, alpha=.05):
         if exog is not None:
             #TODO: make a convenience function for this. we're using the
             # pattern elsewhere in the codebase
-            exog = np.asarray(exog)
+            exog = array_like(exog, 'exog', maxdim=2)
```
jbrockmendel: would it make sense to handle the TODO above along with this validation?

statsmodels/tsa/arima_model.py
bashtage: This is just a check. endog isn't used below

bashtage: All of the data handling is done somewhere in base.

PR #5937, opened 2019-07-13
DOC: Change some more links to https

rebecca-palmer:
More = not already done in 5615801e554063e1fb1c2355b99bab88265b0a9c

This is a set of text search-and-replaces [rpl](https://github.com/kcoyner/rpl/) -R http://domain.to.replace https://domain.to.replace ., after manually checking that each domain.to.replace supported https (I did _not_ test every individual link).

It shouldn't touch any actual code (only documentation), but has not been tested.


bashtage:
LGMT. Thanks.

PR #5939, opened 2019-07-14
DOC: Fix self-contradictory minimum dependency versions

rebecca-palmer:
Assuming that the more recently updated set is correct.

Given the amount of duplication with docs/install.rst, should INSTALL.txt instead be removed altogether?


bashtage:
Thanks for the careful checking.

PR #5943, opened 2019-07-15
ENH: Check int float scalar bool

bashtage:
Add's code to simplify input validation of non-array types

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>


statsmodels/tools/validation/validation.py
```diff
@@ -0,0 +1,178 @@
...
+        value converted to a int
+    """
+    if optional and value is None:
+        return None
+    is_bool = isinstance(value, bool)
+
+    if hasattr(value, 'squeeze') and callable(value.squeeze):
+        value = value.squeeze()
+
+    if isinstance(value, (int, np.integer)) and not is_bool:
```
jbrockmendel: Probably should exclude np.timedelta64, which is a subclass of np.integer

bashtage: Thanks.

PR #5950, opened 2019-07-15
MAINT: Drop redendant travis configs

bashtage:
Drop configs that are run on appveyor

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,15 +33,12 @@ env:
 matrix:
   fast_finish: true
   include:
+    # Python 3.7 + fixed pandas
     - python: 3.7
       env:
         - PYTHON=3.7
         - PANDAS=0.24
         - LINT=true
-    # Python 3.7 + cutting edge packages
-    - python: 3.7
-      env:
-        - PYTHON=3.7
         - COVERAGE=true
     # Documentation build (on Python 3.7 + cutting edge packages)
     - python: 3.7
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -74,14 +71,10 @@ matrix:
     - os: osx
       language: generic
       env:
-        - PYTHON=3.6.6
```
thequackdaddy: So the travis pip version uses pyenv to get the python version and it fails because "3.6" is too ambiguous. 

I noticed that the travis pip/openblas tended to be more delicate than the mkl version--especially for Mac. So I liked having separate ones. Not married to it, however.

.travis.yml
```diff
@@ -74,14 +71,10 @@ matrix:
...
+        - PYTHON=3.6
         - NUMPY=1.15
         - SCIPY=1.2
         - BUILD_INIT=tools/ci/travis_pip.sh
-    - os: osx
-      language: generic
-      env:
-        - PYTHON=3.7
```
thequackdaddy: And this is the conda/mkl version of python.

.travis.yml
bashtage: I think this can be dropped since python/pip is checked on azure for 3.5-3.7.

.travis.yml
bashtage: I restored this one since azure doesn't use conda.

.travis.yml
bashtage: I think this can be dropped since azure runs python/pip on OSX.  The only issue here is that OpenBlas is tied to the NumPy version, so different NumPy will get different OpenBlas.

PR #5962, opened 2019-07-16
CLN: cython cleanups

jbrockmendel:
trailing whitespace, `cimport numpy as cnp` instead of as `np` to disambiguate the namespaces

### DIFF
--- a/statsmodels/nonparametric/_smoothers_lowess.pyx
+++ b/statsmodels/nonparametric/_smoothers_lowess.pyx
@@ -138,8 +138,8 @@ def lowess(np.ndarray[DTYPE_t, ndim = 1] endog,
 
     y = endog   # now just alias
     x = exog
-           
-    if not 0<=frac<=1:
+
+    if not 0 <= frac <= 1:
            raise ValueError("Lowess `frac` must be in the range [0,1]!")
 
     n = x.shape[0]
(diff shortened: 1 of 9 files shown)


ChadFulton:
As I have mentioned before, I do not want to change `cimport numpy as cnp` even if Pandas does that. Cython's own documentation shows otherwise: https://cython.readthedocs.io/en/latest/src/tutorial/numpy.html.

bashtage:
I'm also -1 on cnp since it doesn't seem to be canonical Cython and I don't
really see any meaningful gain since cython automatically intercepts the
words that are in the pxd.

jbrockmendel:
> As I have mentioned before

[Here](https://github.com/statsmodels/statsmodels/pull/5360#discussion_r238082265) you said the exact opposite.

> since it doesn't seem to be canonical Cython

The benefit is for readers who are not familiar with the cythonc canon.

ChadFulton:
> Here you said the exact opposite.

Wow! Sorry about causing confusion about this!

Still I stand by this viewpoint here, I think we should follow the Cython docs. I can follow up with a PR to fix my mistake in _arma_innovations.

jbrockmendel:
> in _arma_innovations.

Also kalman_loglike.  Should grep for others (even if you're making a decision we'll all rue)

jbrockmendel:
reverted cnp, now mostly just removing trailing whitespace

bashtage:
LGMT. leave this for @ChadFulton.

ChadFulton:
Would merge but there seems to be a conflict?

bashtage:
Since @ChadFulton approved, in it goes.

bashtage:
Thanks.

PR #5972, opened 2019-07-18
MAINT: Remove string_types in favor of str

bashtage:
Remove old alias


jbrockmendel:
Looks like a complaint somewhere about `asstr2`

Issue #423, opened 2012-08-10
Handling of constant across models


jseabold:
We need to consistently handle constant across models. Results of linear models without constant need to be checked / need tests (or maybe we should just disallow it?), and we need an attribute (attached to data) that says whether a design as a constant and if so, what column it is in. I think `patsy` takes care of this for the formula cases now.

Issue #27, opened 2011-08-25
add hasconstant indicator for R-squared and df calculations


wesm:
Original Launchpad bug 574004: https://bugs.launchpad.net/statsmodels/+bug/574004
Reported by: josef-pktd (joep).

see also  Bug #440151: fvalue and mse_model are -inf if only 1 regressor

The R-squared and df calculations assume that there is a constant among the regressors.

todo: add a hasconstant indicator and adjust calculations for df and R-squared

Notes: pandas did recently a similar change

R-squared is not really well defined without constant and there are several competing definitions, but for the simple case of regression without constant we should use the total sum of squares instead of mean corrected ???

I would keep the R-squared with constant for transformed regressors that loses the constant for the regression, (example of heteroscedasticity).

PR #5973, opened 2019-07-18
BUG: variance_inflation_factor: resolve exog without constants

ZhiqingXiao:
Variance inflation factor (VIF) is used for testing collinearly of multiple features in a linear model, while the linear model may or may not have a constant term.

When the linear model does not have a constant term, users generally do not add constant to `exog`, which provides a misleading R^2 value, and this R^2 value has nothing to do with the collinearly of components in `exog`. In `OLS`, the parameter `hasconstant` resolves this and provide an alternative way to calculate R^2 in that case. Explicitly speaking, if we defined the R^2 in the form of R^2=1-SSE/SST, when there is a constant term, SST should be the sum of (y - mean)^2; when there are no constant term, SST should be the sum of y^2. The R^2 obtained by y^2 is what VIF actually wants. Therefore, it is necessary to provide the `hasconstant` parameter so users can obtain meaningful VIF result at that case.

When the linear model has a constant term, users are responsible for `add_constant`-ing to the `exog`, and everything works well. Explicityly listing out the `hasconstant` reminds the users of their responsibility.

Cf
https://github.com/statsmodels/statsmodels/issues/27
https://github.com/statsmodels/statsmodels/issues/423
https://github.com/statsmodels/statsmodels/pull/499

- [ ] closes #xxxx
- [ ] tests added / passed. 

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>

### DIFF
--- a/statsmodels/stats/outliers_influence.py
+++ b/statsmodels/stats/outliers_influence.py
@@ -146,7 +146,7 @@ def reset_ramsey(res, degree=5):
     return res_aux.f_test(r_matrix)  # , r_matrix, res_aux
 
 
-def variance_inflation_factor(exog, exog_idx):
+def variance_inflation_factor(exog, exog_idx, hasconst=None):
     """variance inflation factor, VIF, for one exogenous variable
 
     The variance inflation factor is a measure for the increase of the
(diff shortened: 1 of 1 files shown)


ZhiqingXiao:
> Hello @ZhiqingXiao! Thanks for updating this PR. We checked the lines you've touched for [PEP 8](https://www.python.org/dev/peps/pep-0008) issues, and found:
> 
> There are currently no PEP 8 issues detected in this Pull Request. Cheers! 🍻
> 
> ##### Comment last updated at 2019-07-18 13:05:53 UTC

Resolved.

bashtage:
Looks good, but needs a test to make sure it works.

ZhiqingXiao:
Things are more complex than I expected.... I will re-consider refining this pull request after I have added some tests for existing good use cases.

PR #5975, opened 2019-07-18
TST: Add a test for variance_inflation_factor

ZhiqingXiao:
Add tests to protect the existing (and released) functionality of `test_variance_inflation_factor`.
including the cases of perfect collinear, partial collinear, and no collinear.

Test case of perfect collinearly:
endog of OLS: [0, 1, 2, 3, 4, 5, 6, 7]
exog of OLS: [1, 2, 3, 4, 5, 6, 7, 8]
R^2 = 1
VIF = inf

Test case of partial collinearly:
endog of OLS: [0, 0, 1, 1, 2, 2, 3, 3]
exog of OLS: [0, 0, 0, 0, 1, 1, 1, 1]
R^2 = 0.8
VIF = 5

Test case of no collinearly:
endog of OLS: [0, 0, 0, 0, 2, 2, 2, 2]
exog of OLS: [0, 2, 0, 2, 0, 2, 0, 2]
R^2 = 0
VIF = 1

- [ ] closes #xxxx
- [ ] tests added / passed. 

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>

### DIFF
--- a/statsmodels/stats/tests/test_outliers_influence.py
+++ b/statsmodels/stats/tests/test_outliers_influence.py
@@ -18,9 +18,21 @@ def test_reset_stata():
     assert_almost_equal(stat.pvalue, 0.2221, decimal=4)
 
 
-def test_variance_inflation_factor():
+def test_variance_inflation_factor():    
     # Test Case: perfect collinear with constant component
-    exog = np.stack([np.arange(8), 10. - np.arange(8)]).T
+    exog = np.array([[0, 1, 2, 3, 4, 5, 6, 7], [1, 2, 3, 4, 5, 6, 7, 8]]).T
     exog = add_constant(exog)
-    vif = variance_inflation_factor(exog, 0)
-    assert_almost_equal(vif, 0., decimal=4)
+    vif = variance_inflation_factor(exog, 1)
+    assert_almost_equal(vif, float('+inf'), decimal=4)
+
+    # Test Case: partial collinear with constant component
+    exog = np.array([[0, 0, 1, 1, 2, 2, 3, 3], [0, 0, 0, 0, 1, 1, 1, 1]]).T
+    exog = add_constant(exog)
+    vif = variance_inflation_factor(exog, 1)
+    assert_almost_equal(vif, 5., decimal=4)
+
+    # Test Case: no collinear with constant component
+    exog = np.array([[0, 0, 0, 0, 2, 2, 2, 2], [0, 2, 0, 2, 0, 2, 0, 2]]).T
+    exog = add_constant(exog)
+    vif = variance_inflation_factor(exog, 1)
+    assert_almost_equal(vif, 1., decimal=4)
--- a/statsmodels/stats/tests/test_outliers_influence.py
+++ b/statsmodels/stats/tests/test_outliers_influence.py
@@ -18,7 +18,7 @@ def test_reset_stata():
     assert_almost_equal(stat.pvalue, 0.2221, decimal=4)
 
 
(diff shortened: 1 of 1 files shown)


bashtage:
Ideally the test should be in the same branch as the feature since they go together.

statsmodels/stats/tests/test_outliers_influence.py
```diff
@@ -14,3 +16,11 @@ def test_reset_stata():
...
     assert_almost_equal(stat.fvalue[0, 0], 1.52, decimal=2)
     assert_almost_equal(stat.pvalue, 0.2221, decimal=4)
+
+
+def test_variance_inflation_factor():
+    # Test Case: perfect collinear with constant component
+    exog = np.stack([np.arange(8), 10. - np.arange(8)]).T
+    exog = add_constant(exog)
+    vif = variance_inflation_factor(exog, 0)
+    assert_almost_equal(vif, 0., decimal=4)
```
bashtage: Is this different what one would get if has_constant=1 (or True)?

bashtage: If so, could you add that case as well.

ZhiqingXiao: This test does not relate to #5973 directly.
This test considers the case when `hasconstant=True`, which always worked correctly.
Setting `hasconstant=True` explicitly will return the same result.

ZhiqingXiao:
It seems that the original VIF function may have some other problems that refrain this test from being passed. (It produces NaN in some environments for this test case, while the correct result should be 0.)  Let me consider about this. I will re-open it after I sort out everything.

bashtage:
This looks good. Are you still planning on the enhancement?

statsmodels/stats/tests/test_outliers_influence.py
```diff
@@ -14,3 +16,23 @@ def test_reset_stata():
...
     assert_almost_equal(stat.fvalue[0, 0], 1.52, decimal=2)
     assert_almost_equal(stat.pvalue, 0.2221, decimal=4)
+
+
+def test_variance_inflation_factor():
+    # Test Case: perfect collinear with constant component
+    exog = np.array([[0, 1, 2, 3, 4, 5, 6, 7], [1, 2, 3, 4, 5, 6, 7, 8]]).T
+    exog = add_constant(exog)
+    vif = variance_inflation_factor(exog, 1)
+    assert_almost_equal(vif, float('+inf'), decimal=4)
```
bashtage: Better to use np.inf instead of float('+inf')

ZhiqingXiao:
@bashtage  Thanks for your attention, but I actually find the VIF function may not be perfectly fixed by merely adding a 'hasconstant' parameter. :(

**Detail:**
The following codes can reprise the bugs in the released codes, and even adding 'hasconstant' can not resolve it.

```
import numpy as np
from statsmodels.stats.outliers_influence import variance_inflation_factor

exog = np.array([[0, 1],] * 5)
vif = variance_inflation_factor(exog, 0)
print(vif)
# statsmodels reports np.nan (no matter hasconstant=True or False or None), 
# but the correct answer is np.inf.
```

The underlying problem is: Current implementation replies on the relationship VIF=1/(1-R^2). However VIF is not defined as such, and VIF should always be a number within [1, +inf], even when R^2 does not report a number within (0,1). In the above example, R^2 can be np.nan, then VIF becomes np.nan.

Maybe I need to think more about this. I may re-open the enhancement after I find out a perfect way to resolve this.

PR #5983, opened 2019-07-19
BUG: Fix return for RegressionResults

### LINKED ISSUE #4618: RegressionResults.__str__ returns None

bashtage:
RR str was missing the return keyword

closes #4618

- [X] closes #4618

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>

### DIFF
--- a/statsmodels/regression/linear_model.py
+++ b/statsmodels/regression/linear_model.py
@@ -1520,9 +1520,6 @@ class RegressionResults(base.LikelihoodModelResults):
         for key in kwargs:
             setattr(self, key, kwargs[key])
 
-    def __str__(self):
-        self.summary()
-
     def conf_int(self, alpha=.05, cols=None):
         """
         Returns the confidence interval of the fitted parameters.


statsmodels/regression/linear_model.py
```diff
@@ -1522,7 +1522,7 @@ def __init__(self, model, params, normalized_cov_params=None, scale=1.,
             setattr(self, key, kwargs[key])
 
     def __str__(self):
-        self.summary()
+        return self.summary()
```
jbrockmendel: should be `str(self.summary())`

jbrockmendel:
Needs test.  IIRC from when I tried to fix this, there was an issue with RegressionResultsWrapper returning not passing through `self._results.__str__` but using its own, which inherited from object

bashtage:
> Needs test. IIRC from when I tried to fix this, there was an issue with RegressionResultsWrapper returning not passing through `self._results.__str__` but using its own, which inherited from object

That's ok.  I'm not trying to change the functionality, only the remove an obvious bug in the simplest way.

jbrockmendel:
> That's ok. I'm not trying to change the functionality, only the remove an obvious bug in the simplest way.

I'm not sure I follow.  This seems like a "worth doing \Rightarrow worth doing right" situation

bashtage:
Better to remove it since this pattern isn't used.

jbrockmendel:
lgtm

PR #6059, opened 2019-08-02
REF: move garch to archive/

jbrockmendel:


bashtage:
I recently deprecated this.  I doubt anyone is using it, but I think it has to go through 1 release round before removal.   This assumes that it runs.  If it doesn't run at all, then it isn't being used and can be archived now.

jbrockmendel:
```
>>> from statsmodels.sandbox.tsa import garch
  File "statsmodels/statsmodels/sandbox/tsa/garch.py", line 84, in <module>
    import numdifftools as ndt
ModuleNotFoundError: No module named 'numdifftools'
```

bashtage:
Good enough for me.

Issue #6034, opened 2019-07-30
SARIMAX memory consumption


david-waterworth:
Right at the end of the fit process, SARIMAX is consuming almost 100% of my available memory. It runs for quite some time (using all CPU's) without any increase in memory then it starts to ramp up, and maintains 90% usage. 

Result is a crashed jupyter kernel (but the server stays up and the memory allocated). Restarting jupyter frees the memory.

Code is

```
model = sm.tsa.SARIMAX(x, order=(1,0,0), seasonal_order=(1,0,0,96))

results = model.fit()
print(results.summary())
```
 
x is an pandas array of length 34751

![image](https://user-images.githubusercontent.com/5028974/62091897-7815d700-b2b6-11e9-92f1-8161d5970773.png)

Also using `getsize` from https://stackoverflow.com/a/30316760/2981639 gives

![image](https://user-images.githubusercontent.com/5028974/62092442-836a0200-b2b8-11e9-9ad2-89e8b8a474b3.png)

```
import sys
from types import ModuleType, FunctionType
from gc import get_referents

# Custom objects know their class.
# Function objects seem to know way too much, including modules.
# Exclude modules as well.
BLACKLIST = type, ModuleType, FunctionType


def getsize(obj):
    """sum size of object & members."""
    if isinstance(obj, BLACKLIST):
        raise TypeError('getsize() does not take argument of type: '+ str(type(obj)))
    seen_ids = set()
    size = 0
    objects = [obj]
    while objects:
        need_referents = []
        for obj in objects:
            if not isinstance(obj, BLACKLIST) and id(obj) not in seen_ids:
                seen_ids.add(id(obj))
                size += sys.getsizeof(obj)
                need_referents.append(obj)
        objects = get_referents(*need_referents)
    return size
```

Edit:

This might give an indicator of where it's occurring. If I change the order / seasonal_order parameters SARIMAX can run out of memory before the model solves in which case it dumps the following Traceback.

> ---------------------------------------------------------------------------
> MemoryError                               Traceback (most recent call last)
> <ipython-input-24-d00c66b6c85e> in <module>
>       1 model = sm.tsa.SARIMAX(x, order=(1,1,0), seasonal_order=(1,1,0,96))
>       2 
> ----> 3 results = model.fit()
>       4 print(results.summary())
> 
[...]

PR #6071, opened 2019-08-04
ENH: state space: Improve low memory usability; allow in fit, loglike

ChadFulton:
- [x] helps improve #6034
- [x] tests added / passed. 

This PR improves the usability of memory conservation in state space models. In particular, we can now use maximum memory conservation in combination with `fit` and (since #6064 was merged) out-of-sample forecasting.


ChadFulton:
One thing to note: the default parameter covariance matrix in state space models is computed via OPG, but this is unavailable when the loglikelihood isn't stored for all observations.  So when the `memory_no_likelihood` option is used, the default parameter covariance matrix is switched to computation via complex-step approximation.

ChadFulton:
Final commit adds a `low_memory` argument to state space models' `fit` and `filter` methods that performs maximum memory conservation, which is mostly useful if all you want to do is fit the model and then perform out-of-sample forecasting.

ChadFulton:
Squashed. Will merge if green.

PR #6095, opened 2019-08-10
WIP: Oaxaca Variance/Other Models

aadams:
Hi everyone.

I should have added more ability for the end user to choose what type of non-discriminatory model they would like. It used to be a pooled only model, but now users can pass into two_fold() optionally what type of non-discriminatory model they wish to. I have checked these results with STATA and they are the same. I haven't written tests for them, but I will.

The real issue is that I cannot seem to get the variance/standard error calculations to work. I found what the R and Stata command cite as their method for calculating the variance, but I can't get the same numbers. This is what they cite: http://repec.org/dsug2005/oaxaca_se_handout.pdf. I would appreciate some help to figure out where I am going wrong. Var of the three_fold is much easier than the two_fold, so once the three_fold is figured out, I can move onto the two_fold. The output can be seen in the results by calling the .var on the results class. This calculation is the variance instead of the standard error.

I could also bootstrap the standard errors like the R command does, but it takes a while.

### DIFF
--- a/statsmodels/stats/oaxaca.py
+++ b/statsmodels/stats/oaxaca.py
@@ -1,6 +1,6 @@
-# TODO Variance can be calculated for the three_fold
-# TODO Group Size Effects can be accounted for
-# TODO Non-Linear Oaxaca-Blinder can be used
+# TODO Non-Linear Regressions can be used
+# TODO Further decomposition of the two_fold parameters i.e.
+# the delta method for further two_fold detail
 """
 Author: Austin Adams
 
@@ -8,7 +8,7 @@ This class implements Oaxaca-Blinder Decomposition. It returns
 a OaxacaResults Class:
 
 OaxacaBlinder:
-Two-Fold/Pooled (two_fold)
+Two-Fold (two_fold)
 Three-Fold (three_fold)
 
 OaxacaResults:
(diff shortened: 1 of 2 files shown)


AustinJAdams:
Oh, and also all the other packages from other languages call it endowments not characteristic, so I changed it.

bashtage:
You can make the test quite loose, say it should match to the first digit or as many as it does match.  You should also include the following in your test function

```
def test_standard_errors_bootstrap(reset_randomstate)
```

which ensures that `np.random.seed` is called with the same value before your test is run.

AustinJAdams:
@bashtage While testing my code against Stata, I actually found an error with their SE calculations. For large variance values (very large), Stata's bootstrap does not reliably converge. My SE calculations mimic R's numbers, but that package only allows discrete values for exog variables. I mimic when using their testing data-set. Should I upload their data-set (a 700 row x 5 column CPS data-set) or just test around values that I reliably get for a StatsModels data-set (which has a large variance of which does not fully converge for Stata)? I cannot force Stata's oaxaca to do more iterations.

bashtage:
You have to do the best testing that you can. Sounds like you have hit that bar.

statsmodels/stats/oaxaca.py
```diff
@@ -398,33 +495,59 @@ class OaxacaResults:
...
-            print(dedent("""\
-            Oaxaca-Blinder Two-fold Effects
-
-            Unexplained Effect: {:.5f}
-            Explained Effect: {:.5f}
-            Gap: {:.5f}""".format(
-                                self.params[0], self.params[1],
-                                self.params[2])))
-
+            if self.std is None:
```
bashtage: Ideally summary would return a Summary object which supports export to text and latex.  

Can be a different PR.

AustinJAdams: I can try to do that. Now that there are more columns, I can use the summary table tool.

AustinJAdams:
I was a bit afraid that the tests would run in parallel because np.random.seed is not thread-safe, but looks like it all passed.

statsmodels/stats/oaxaca.py
AustinJAdams: @bashtage Do you have any references so that I can look into how to do what you asked? I am trying to add non-linear models then I will look into summary models. I have started working on diff-in-diff as well, so these might take a backseat, as with this PR, this module is extremely usable.

PR #6096, opened 2019-08-12
DOC: Seasonality in SARIMAX Notebook

### LINKED ISSUE #6037: DOC/BUG: Seasonality in SARIMAX Notebook
Can you look at the code at the "**In[7]**" (below) in the documentation https://www.statsmodels.org/dev/examples/notebooks/generated/statespace_sarimax_stata.html
```python
# In[7]
#Fit the model
mod = sm.tsa.statespace.SARIMAX(data['ln_wpi'], trend='c', order=(1,1,1))
```
Prior to this code in the documentation, it is said that, since we want a polynomial that has terms for the 1st and 4th degrees of moving averages, we want to use `ma = (1,0,0,1)`

However, this `ma` part (i.e. `(1,0,0,1)`) is not included in the code above,. I think the order should be like that `order=(1,1,(1,0,0,1))` instead of `order=(1,1,1)` in the above code. So the final code in the box of `In[7]` have to be like that
```python
# In[7]
#Fit the model
mod = sm.tsa.statespace.SARIMAX(data['ln_wpi'], trend='c', order=(1,1,(1,0,0,1)))
```
Can you check it?

sameshl:
changed `order=(1,1,1)` to `order=(1,1,(1,0,0,1))`

closes #6037

### DIFF
--- a/examples/notebooks/statespace_sarimax_stata.ipynb
+++ b/examples/notebooks/statespace_sarimax_stata.ipynb
@@ -108,7 +108,7 @@
     "data.index = data.t\n",
     "\n",
     "# Fit the model\n",
-    "mod = sm.tsa.statespace.SARIMAX(data['wpi'], trend='c', order=(1,1,1))\n",
+    "mod = sm.tsa.statespace.SARIMAX(data['wpi'], trend='c', order=(1,1,(1,0,0,1)))\n",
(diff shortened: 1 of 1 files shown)


ChadFulton:
Thanks very much!

PR #6105, opened 2019-08-16
Update correlation_tools.py

astyonax:
if I understood well this function 
the variable `clipped` should report if any value was smaller than the threshold value not the hardcoded 0

Sorry that I cannot run the tests now. Later next week in case.

### DIFF
--- a/statsmodels/stats/correlation_tools.py
+++ b/statsmodels/stats/correlation_tools.py
@@ -20,7 +20,7 @@ from statsmodels.tools.sm_exceptions import (
 
 def clip_evals(x, value=0):  # threshold=0, value=0):
     evals, evecs = np.linalg.eigh(x)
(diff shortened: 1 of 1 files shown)


statsmodels/stats/correlation_tools.py
```diff
@@ -20,7 +20,7 @@
 
 def clip_evals(x, value=0):  # threshold=0, value=0):
     evals, evecs = np.linalg.eigh(x)
-    clipped = np.any(evals < 0)
+    clipped = np.any(evals < value)
```
dsaxton: Looks like this is a legitimate bug. @josef-pkt

bashtage:
Not tested, but fix is obviously correct.

Thanks.

astyonax:
Great! thank you!

PR #6124, opened 2019-08-22
Plot only unique censored points

sursu:
When one runs:

```python
sf = sm.SurvfuncRight(df["futime"], df["death"])
sf.plot()
```
The survival function is being plotted along with **all** the censored points.
Some of them are identical points.
I find it unnecessary to plot the same points multiple times. Fixed that.

### DIFF
--- a/statsmodels/duration/survfunc.py
+++ b/statsmodels/duration/survfunc.py
@@ -798,7 +798,7 @@ def plot_survfunc(survfuncs, ax=None):
 
         # Plot the censored points.
         ii = np.flatnonzero(np.logical_not(sf.status))
-        ti = sf.time[ii]
+        ti = np.unique(sf.time[ii])
         jj = np.searchsorted(surv_times, ti) - 1
         sp = surv_prob[jj]
         ax.plot(ti, sp, '+', ms=12, color=li.get_color(),


sursu:
??
How can that change be in conflict with anything?

bashtage:
There are no conflicts. The fail is not your fault.

bashtage:
Any way to test that this is the same plot?

sursu:
Some RGB pixel comparison can be done. But there is really no need for that:

The function I'm changing is here: https://github.com/statsmodels/statsmodels/blob/master/statsmodels/duration/survfunc.py.

When I run
```python
sf = sm.SurvfuncRight(df["futime"], df["death"])
```
(following the example [here](https://www.statsmodels.org/devel/duration.html))
the `sf` object is created.

This object has the `plot` method.
If we look in the docs, this method simply calls another function which is found in the same file: `return plot_survfunc(self, ax)`.

The lines where censored points are plotted in that function are:
```python
def plot_survfunc(survfuncs, ax=None):
    ...
        # Plot the censored points.
        ii = np.flatnonzero(np.logical_not(sf.status))
        ti = sf.time[ii]
        jj = np.searchsorted(surv_times, ti) - 1
        sp = surv_prob[jj]
        ax.plot(ti, sp, '+', ms=12, color=li.get_color(),
                label=label + " points")
```
Meaning of the lines in this function:
1. Finds the position of the censored points
2. Find the censoring time corresponding to these positions
3. Find the indices where those times are in `surv_times`
4. Get the survival probabilities for those indices

Clearly if the censoring times are the same, we will get the same indices and the same survival probabilities.

The thing is that matplotlib produces a bad plot there.
I have reproduced all the steps with `plotly` and got a similar, in terms of content, but better looking plot.
Noticed the issue because I had 30000+ points there to plot and it was a bit slow.

bashtage:
Looks right. I suppose this is small enough to not sweat the test.

bashtage:
Thanks.

PR #6144, opened 2019-08-30
BUG: use self.data consistently

comatrion:
The __init__ for MultiComparison defines self.data = np.asarray(data) where data is an input.

However on the line modified in this PR it calculates data.shape which fails for the case where data is, say, an array-like object without defined shape e.g. a list. 

The change is to use self.data.shape instead, which will  be defined.


bashtage:
Obviously correct.  Thanks.

PR #6162, opened 2019-09-15
DOC/SEC: Warn that only trusted files should be unpickled

rebecca-palmer:
The Python pickle format is intentionally a "can run arbitrary code" format:
https://docs.python.org/3/library/pickle.html#restricting-globals

### DIFF
--- a/statsmodels/iolib/smpickle.py
+++ b/statsmodels/iolib/smpickle.py
@@ -19,7 +19,9 @@ def save_pickle(obj, fname):
 
 def load_pickle(fname):
     """
-    Load a previously saved object from file
+    Load a previously saved object; **use only on trusted files**,
+    as unpickling can run arbitrary code.  (i.e. calling this on a
+    malicious file can wipe or take over your system.)
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


bashtage:
I'm happy to have these changes but the first unbroken line must fit in a single line and so be fewer than 72 characters (or fewer, depending on the level of indent). 


So something like

```
Loads a paickle

Load should only be used on trusted files since pickles can run arbitrary code.... 
```

rebecca-palmer:
Reformatted (keeping "use only on trusted files" in the first line to make it more likely to be seen).

statsmodels/iolib/smpickle.py
```diff
@@ -19,7 +19,12 @@ def save_pickle(obj, fname):
 
 def load_pickle(fname):
     """
-    Load a previously saved object from file
```
bashtage: Could you use the nonbold here too, as in the other two examples? Or is there some reason for bold here?

rebecca-palmer: The bold is to emphasize the warning: this one is the most important of the 3 because it's the only one that is listed in the documentation.

bashtage:
Thanks.

PR #6163, opened 2019-09-15
Improve the cvxopt not found error

rebecca-palmer:
The source contains what is obviously intended to be an error message, but was never actually output.


bashtage:
Lgtm

rebecca-palmer:
"0% of diff hit": you don't get here if cvxopt is installed, which it probably is on the test bot.

bashtage:
Thanks.

PR #6164, opened 2019-09-15
Be compatible with scipy 1.3

rebecca-palmer:
Scipy 1.3 removes deprecated functions from scipy.misc

There were 11 test failures (in Debian unstable) with this, #6162 and #6163, but all of them also occur in unmodified master:
Timeout in test_notebook statespace_seasonal.ipynb
9 failures to find pandas_datareader or seaborn (which are not bugs as these actually aren't installed)
Assumption that python is Python 3 in test_docstring_optimization_compat (see separate fix)


bashtage:
Lgtm. Will wait for green.

rebecca-palmer:
Assuming "0% of diff hit" means "you're modifying something with no test coverage", that's probably because the scipy.misc uses that do have test coverage have already been fixed, in
8233beb59d966e5808a1f5b59b2fa1f69e1fdd08
9aef30db576027aab39e328f5d2595efd500d3b0
190c961cdf89195911e3dfee8e8cce56ff94debf
7e60df8ece46707e4795e2617106894bace65843
and these are the last few left.

bashtage:
We test against 1.3 in Travis so these are all not run in the test by fiat.

bashtage:
Thanks.

PR #6165, opened 2019-09-15
Don't assume that 'python' is Python 3

rebecca-palmer:
On systems where 'python' is Python 2, test_docstring_optimization_compat tries to import statsmodels in Python 2, which does not work (currently with ImportError: No module named abc).


bashtage:
Lgtm. Will wait for green.

bashtage:
Green enough. Thanks.

PR #6179, opened 2019-10-02
ENH: Adds state space version of linear exponential smoothing models

ChadFulton:
- [x] tests added / passed. 

This adds a new model `sm.tsa.statespace.ExponentialSmoothing` that handles the linear class of exponential smoothing models by setting them up as a special case of the usual (linear Gaussian) state space framework.

The pros and cons of this model are given in the "Notes" section of the model class docstring:

> The parameters and states of this model are estimated by setting up the
exponential smoothing equations as a special case of a linear Gaussian
state space model and applying the Kalman filter. As such, it has slightly
worse performance than the dedicated exponential smoothing model,
`sm.tsa.ExponentialSmoothing`, and it does not support multiplicative
(nonlinear) exponential smoothing models.

> However, as a subclass of the state space models, this model class shares
a consistent set of functionality with those models, which can make it
easier to work with. In addition, it supports computing confidence
intervals for forecasts and it supports concentrating the initial
state out of the likelihood function.

I think we should include this in Statsmodels for the reasons given there, i.e.:

1. It is easier to work with in combination with state space models (e.g. for automatic forecasting / model selection)
2. It serves as a fully tested reference implementation for the linear exponential smoothing models, and so will likely be helpful in adding new features to the faster, dedicated exponential smoothing model class.
3. It has some new features: confidence intervals, and concentrating the initial states out of the likelihood function (especially useful for models with long seasonal cycles).

I have not added it to the `tsa` API, so right now it's location is `tsa.statespace.ExponentialSmoothing`.


ChadFulton:
Main question here is whether everyone is comfortable with a second version of exponential smoothing models.

I think it should be okay, since this one (which is slower) will be a level deeper into the API hierarchy, and so I don't think it will compete with the main models except when someone specifically wants it.

bashtage:
Seems fine to me to have some duplication now.  I think the direct Exp Smoothing need quite a bit of work to get to a real release quality.  Do this match R more closely?  That would be a useful feature.

ChadFulton:
> I think the direct Exp Smoothing need quite a bit of work to get to a real release quality.

I agree. I'm hoping we can use this model as a baseline to improve it, since this model is much better unit tested.

> Do this match R more closely? That would be a useful feature.

I don't know if it matches R more closely (it depends on what you have in mind). It is certainly much better tested against R (from the `fpp` and `forecast` packages), including the estimated states, forecasts, confidence intervals, and model output (mse, etc.).

It does not include some features that R has, like the specific parameter bounds that Hyndman et al. advocate, but it includes some features that R does not have, like the concentrated initial states.

I would say the main benefit is that it matches Statsmodels (especially state space) more closely, in terms of structure, model options, and results.

> Seems fine to me to have some duplication now. 

Great! Then I will plan to merge soon, unless there are other concerns that arise.

PR #6181, opened 2019-10-03
ENH: state space: add wrapped states and, where possible, named states

ChadFulton:
Adds a more convenient results attributes, `states`, for accessing predicted, filtered, and smoothed states and covariances.

For example:

```python
mod = sm.tsa.UnobservedComponents(...)
res = mod.fit()

# e.g. (assuming Pandas input was given)
res.states.smoothed['level'].plot()
res.states.smoothed_cov.loc['level', 'level'].plot() . # see below for a question on this approach
```

These are useful for several reasons:

- The filter output (like `res.predicted_state`) are always numpy arrays, so the main goal was to get a new attribute that would wrap the array with a Pandas date index when that's available
- `res.predicted_state` includes an additional element, the "initial" state: unlike e.g. `res.filtered_state` which has length `nobs`, `res.predicted_state` has length `nobs + 1`. This can be a gotcha if you expect the arrays to all line up. With Pandas output, we can index the initial state correctly.

Also adds a new property attribute `state_names` (similar to `param_names`) that can be implemented in state space models to give human-readable names to the state vector elements. This is more useful for some models than others:

- For `UnobservedComponents`, we can usefully name all of the states.
- For `DynamicFactor`, we can usefully name all the states (mostly just the factors and their lags, but also the error terms if those are in the state vector because they have an autoregressive component).
- For `SARIMAX`, I didn't give descriptive names any of the states (except for if regression is done via the sate vector). It's possible with a little bit of thinking we could come up with human-readable names that would describe what the state is capturing, but I don't know if that would be particularly helpful anyway, since I don't know of anything that uses intermediate states from these models.
- `VARMAX` has a similar situation to `SARIMAX`.

**Question**:

I don't know if we've settled on a way to set up the Pandas wrapped version of covariance matrices over time. I think we need a multi-index, but there are two ways to do it:

Option 1: 

```
                     level         trend
1959Q1 level  1.676490e-10  0.000000e+00
       trend  0.000000e+00  0.000000e+00
1959Q2 level  1.676490e-10  1.676490e-10
       trend  1.676490e-10  5.693389e-01
```

Option 2:

```
                     level         trend
level 1959Q1  1.676490e-10  0.000000e+00
trend 1959Q1  0.000000e+00  0.000000e+00
level 1959Q2  1.676490e-10  1.676490e-10
trend 1959Q2  1.676490e-10  5.693389e-01
```

I think Option 1 looks nicer when printed and it was what I was originally thinking of going with, but it is less good for selecting a time series of one of the variances. e.g., under Option 1 you might do:

```python
print(res.state.smoothed_cov.loc[np.s_[:, 'level'], 'level'])
```

and then you get:

```
1959Q1  level    1.676490e-10
1959Q2  level    1.676490e-10
```

Whereas under Option 2, you can do:

```python
print(res.states.filtered_cov.loc['level', 'level'])
```

and then you get:

```
1959Q1    1.676490e-10
1959Q2    1.676490e-10
```

So I am leaning towards Option 2, and this is what the PR currently has implemented. If anyone has any thoughts on this (or options I've missed!), please leave a comment.

### DIFF
--- a/statsmodels/tsa/statespace/dynamic_factor.py
+++ b/statsmodels/tsa/statespace/dynamic_factor.py
@@ -181,9 +181,13 @@ class DynamicFactor(MLEModel):
             k_states += self._error_order
             k_posdef += k_endog
 
+        # We can still estimate the model with no dynamic state (e.g. SUR), we
+        # just need to have one state that does nothing.
+        self._unused_state = False
         if k_states == 0:
             k_states = 1
             k_posdef = 1
+            self._unused_state = True
 
         # Test for non-multivariate endog
         if k_endog < 2:
@@ -535,6 +539,7 @@ class DynamicFactor(MLEModel):
             else:
                 # In the case of individual autoregressions, extract just the
                 # diagonal elements
+                # TODO: can lead to explosive parameterizations
                 params[self._params_error_transition] = (
                     res_errors.params.T[self._idx_error_diag])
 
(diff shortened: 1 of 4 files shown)


ChadFulton:
Failures are unrelated. This seems fairly innocuous, so I will merge tomorrow (with the current behavior for Pandas wrapping of the 3-dimensional array of covariance matrices) unless there are concerns.

ChadFulton:
Merging.

PR #6233, opened 2019-11-10
BUG: dynamic factor model use AR model for error start params if error_var=False

### LINKED ISSUE #6182: BUG: dynamic factor start_params may be explosive with autoregressive errors
Currently we just take a shortcut in `start_params` where if the error terms are individual ARs, rather than jointly as a VAR, we still estimate a VAR and then take the diagonal elements of each coefficient matrix to get the AR parameters. But these may not lead to stationary models.

ChadFulton:
- [x] closes #6182
- no tests added since this is just a change to start params computation


ChadFulton:
Green, merged.

PR #6242, opened 2019-11-12
BUG: Forecasts now ignore non-monotonic period index

### LINKED ISSUE #6177: BUG: non-monotonic date indexes for TSA models should be ignored
Currently, you can have an index like:

```python
tmp = pd.period_range(start=2000, end=2002, freq='A')
index = tmp.tolist() + tmp.tolist()
index
```

```
[Period('2000', 'A-DEC'),
 Period('2001', 'A-DEC'),
 Period('2002', 'A-DEC'),
 Period('2000', 'A-DEC'),
 Period('2001', 'A-DEC'),
 Period('2002', 'A-DEC')]
```

and the TSA models will try to use it, because they only check for date-type index that has an associated frequency. But for period indexes, we need to also check that they're monotonic (`is_monotonic`) and that they're full (`is_full`).

We should ignore these indexes for e.g. forecasting purposes, as we do with other types of date indexes that don't have associated frequencies.

ChadFulton:
- [x] closes #6177
- [x] tests added / passed. 

Note: Although we could potentially make non-full `PeriodIndex`'s work, currently they produce weird behavior and should be disallowed.

### DIFF
--- a/statsmodels/tsa/base/tests/test_tsa_indexes.py
+++ b/statsmodels/tsa/base/tests/test_tsa_indexes.py
@@ -16,8 +16,10 @@ import numpy as np
 import pandas as pd
 
 from numpy.testing import assert_equal, assert_raises
+import pytest
 
 from statsmodels.tsa.base import tsa_model
+from statsmodels.tools.sm_exceptions import ValueWarning
 
 nobs = 5
 base_dta = np.arange(nobs)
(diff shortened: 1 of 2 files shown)


ChadFulton:
I may have to eliminate the "non-full" part of this PR for now, because apparently `is_full` doesn't work well with frequency multipliers, e.g.:

```python
index = pd.period_range(start='2000Q1', periods=10, freq='2Q')
print(index.is_full)
```

yields `False`.

PR #6254, opened 2019-11-18
ENH: State space: Add finer-grained memory conserve settings

### LINKED ISSUE #6188: DESIGN: state space `low_memory`
Currently the `low_memory=True` argument to `fit` in state space models performs maximum memory conservation, so that (if I remember right) no Kalman filter output of length `nobs` is stored.

This is particularly useful when `nobs` and `k_states` are both large because the conditional covariance matrix storage has `nobs * k_states**2` elements.

Models fit this way are still useful for forecasting (and e.g. cross-validation-type model selection exercises), since that only depends on the final state estimates, but they can't do much else. For example, they done even have `fittedvalues` or `resid`.

The question here is: should we change `low_memory=True` to allow storing some or all of the `nobs * k_states` vectors. I guess the first thing to add back in would be `fittedvalues` and/or `resid`.

NB: regardless of what is decided for the `low_memory` option, users will still free to select custom memory conservation options directly (e.g. `mod.ssm.memory_no_filtered=True`), but this isn't well documented, and isn't flexible enough right now to allow keeping the `nobs * k_states` arrays while dropping the `nobs * k_states**2` arrays.

ChadFulton:
- [x] closes #6188
- [x] tests added / passed. 

This primarily does two things:

1. Adds finer-grained memory conservation control for state space models.
2. Sets the default for `MEMORY_CONSERVE` and `low_memory=True` to store all `k_endog x nobs` arrays (so that is `forecasts`, `forecast_errors`, `standardized_forecast_errors`) ~~and the two `k_states x nobs` arrays (`predicted_state`, and `filtered_state`)~~.

The benefits of (2) are the following:

- The `k_states**2 x nobs` arrays are not stored in this mode, and I think they are the main culprits for running out of memory.
- Storing `forecasts`, `forecast_errors`, and `standardized_forecast_errors` mean that most of the standard results (like `fittedvalues`, `resid`, and the state space diagnostics) will work even with the low memory option.


ChadFulton:
On second thought, I have changed the default to only store the `k_endog x nobs` arrays and **not** the `k_states x nobs` arrays.

ChadFulton:
I should note: there is a minor backwards incompatibility here, as I'm changing the values of the memory conservation constants. e.g. `MEMORY_NO_FORECAST` was previously `0x01` and it is now `0x03`. However, I think that the usage has always been to either use the boolean attributes (`mod.ssm.memory_no_forecast = True`) or to use the constant variables (`mod.ssm.set_conserve_memory(MEMORY_NO_FORECAST`), so I think that this change should not be a problem.

PR #6280, opened 2019-11-26
ENH: State space: Improve simulate, IRF, prediction

### LINKED ISSUE #2760: BUG/ENH: Statespace: Simulation / impulse response functions with time-varying state space matrices.

ChadFulton:
- [x] closes #2760
- [x] tests added / passed. 

Still need to do:

- [x] Update `MLEModel.impulse_responses` to either extend the model or else wrap `MLEModel.simulate`.
- [x] docstrings for new functions / `Appender` for subclass methods.
- [x] unit tests for `anchor` in the middle of the sample
- [x] unit tests for Pandas input (`endog` / `exog` and dates for `anchor`) and output (wrapping).

There are two primary purposes for this PR:

1. After fitting a model, allow simulating datasets starting from just after the last observation (i.e. conditional on the predicted state distribution after running the Kalman filter over the entire sample) - or more generally simulating datasets starting after any particular observation. This is via a new `anchor` argument.
2. Improve `simulate` and `impulse_responses` in time-varying state space models.

It also has makes other improvements:

- Centralizes the machinery for dealing with out-of-sample `exog` and trends, and so removes a fair amount of boilerplate in `get_prediction` methods.
- Adds a `repetitions` argument for generating multiple simulations.
- Adds methods for cloning and extending the `Representation` class and its subclasses.

### DIFF
--- a/statsmodels/tsa/statespace/dynamic_factor.py
+++ b/statsmodels/tsa/statespace/dynamic_factor.py
@@ -6,7 +6,6 @@ Author: Chad Fulton
 License: Simplified-BSD
 """
 
-from warnings import warn
 from collections import OrderedDict
 
 import numpy as np
@@ -24,7 +23,6 @@ from statsmodels.tools.tools import Bunch
 from statsmodels.tools.data import _is_using_pandas
 from statsmodels.tsa.tsatools import lagmat
 from statsmodels.tools.decorators import cache_readonly
-from statsmodels.tools.sm_exceptions import ValueWarning
 import statsmodels.base.wrapper as wrap
 from statsmodels.compat.pandas import Appender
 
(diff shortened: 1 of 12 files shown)


ChadFulton:
(this looks like a big commit, but 2/3 of the addition lines are just new unit tests)

ChadFulton:
Assuming this comes back green I will merge this PR, which has gotten a little larger than I originally planned, but is still over 50% new unit tests, and a lot of boilerplate removed.

PR #6361, opened 2019-12-21
Update dynamic_factor.py

dsgerbc:
I'm getting a lot of failing to converge errors in simple DFM setups, so I did a bit of digging. 
It would seem that the default factor normalization option in the PCA code doesn't transform res_pca.loadings along with res_pca.factors, so running with normalization=on results in non-zero mean endog in line 464 and throws off error process estimation. As a result either failing to converge or taking forever.

Turning off normalization in line 452 is one option to fix the issue, another is to replace res_pca.loadings.T in line 464 with res_pca.coeff
(Haven't tested that one).

I don't have a specific test to suggest.
Just put a breakpoint on line 464 and do endog.mean

- [ ] closes #xxxx
- [ ] tests added / passed. 

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>

### DIFF
--- a/statsmodels/tsa/statespace/dynamic_factor.py
+++ b/statsmodels/tsa/statespace/dynamic_factor.py
@@ -449,7 +449,7 @@ class DynamicFactor(MLEModel):
         # 1. Factor loadings (estimated via PCA)
         if self.k_factors > 0:
             # Use principal components + OLS as starting values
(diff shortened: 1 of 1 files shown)


bashtage:
Looks like this fix doesn't work.

PR #6375, opened 2019-12-28
ENH: Supporting custom kernel in local linear kernel regression

vsriram11:
Currently, KernelRegression supports only the Gaussian kernel. By adding the kernel as a parameter to the class, we can support custom kernels for the local linear mode of kernel regression.

This is useful for performing other nonparametric regressions. For example, this would now support the LOESS regression, which can be achieved by specifying local linear mode, and a tricube kernel. 

This PR also adds the tricube kernel to statsmodels library of kernels.

### DIFF
--- a/statsmodels/nonparametric/kernel_regression.py
+++ b/statsmodels/nonparametric/kernel_regression.py
@@ -99,7 +99,6 @@ class KernelReg(GenericKDE):
         self.ckertype = ckertype
         self.okertype = okertype
         self.ukertype = ukertype
-
         if not (self.ckertype in kernel_func and self.ukertype in kernel_func
                 and self.okertype in kernel_func):
             raise ValueError('user specified kernel must be a supported '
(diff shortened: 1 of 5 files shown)


statsmodels/nonparametric/kernel_regression.py
```diff
@@ -83,10 +91,14 @@ class KernelReg(GenericKDE):
         The bandwidth parameters.
     """
     def __init__(self, endog, exog, var_type, reg_type='ll', bw='cv_ls',
-                 defaults=None):
+                 ckertype='gaussian', okertype='wangryzin',
+                 ukertype='aitchisonaitken', defaults=None):
         self.var_type = var_type
         self.data_type = var_type
         self.reg_type = reg_type
+        self.ckertype = ckertype
```
bashtage: Can you add a check that these are value kernel names, and raise if not with a meaningful error.

vsriram11: Sure done, and added relevant unit test

vsriram11: @bashtage Any idea why the build checks below fail now though? Looks like it's failing when collecting tests, but it works fine locally..

vsriram11: Never mind, figured it out. Looks like a documented issue with pytest fixtures, that I managed to work around

bashtage:
Looks great, thanks.

PR #6408, opened 2020-01-13
ENH/CLN: Improve diagnostics

bashtage:
- [ ] closes #xxxx
- [ ] tests added / passed. 

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>


josef-pkt:
Don't merge this if the move doesn't preserve commit/blame history

bashtage:
It is not going to be possible to preserve this, nor is it needed. It can't
be preserved because the destination file exists already. There may be some
complicate way to preserve blame, but given how extensive these changes are
there isn't much point.

josef-pkt:
we move the module to a private `_diagnostic` and keep `diagnostic` as an api module. There are or will be several other private `_diagnostic_xxx` 

For bug hunting or finding the reason for some implementation choices a blame and history is very useful and needed.
Having it as a new file looses all references to the history.

bashtage:
#6410 retains history

PR #6411, opened 2020-01-14
ENH: state space: add Chandrasekhar recursions

ChadFulton:
- [x] Add unit tests
- [x] Add version for univariate approach
- [ ] Add doubling algorithm for stationary initialization

This PR adds the option to use Chandrasekhar recursions for computing the predicted state covariance matrix, as in Herbst (2015) (@eph). This has computational benefits when the observation dimension is much smaller than the state dimension, with the drawback that the stationary initialization has to be used and there can't be missing observations.

Not for v0.11, needs unit tests and implementation for the univariate case.

It's typically mentioned as useful for likelihood estimation in DSGE models, but I think it might help also with e.g. SARIMAX models with large seasonal effects, where the state dimension blows up.

Usage right now is e.g.:

```python
mod = sm.tsa.SARIMAX(endog, order=([1, 365], 0, 0))
mod.ssm.filter_chandrasekhar = True
res = mod.fit()
```

but at some point we may want to automatically select it as the default in some cases.


ChadFulton:
Latest commit adds support for this with the univariate filtering approach of Koopman and Durbin (2000) using periodic Chandrasekhar recursions ([Aknouche and Hamdi, 2007](https://arxiv.org/abs/0711.3857)).

Also adds unit tests.

ChadFulton:
This is now probably ready to merge after v0.11 is released, although it might be good to make an example notebook showing how to use this feature and maybe with some basic timing examples.

Might need more performance tuning for the DK univariate approach + Chandrasekhar recursions. So far on a brief test with a 3-variable VAR it doesn't seem to be giving much of an improvement, but maybe with a larger DFM model or something it might help.

ChadFulton:
@AKN-F2F this is a closed Pull Request, please create a new issue at https://github.com/statsmodels/statsmodels/issues/new/choose

PR #6415, opened 2020-01-15
BUG: RecursiveLS should not allow `fix_params` method.

ChadFulton:
- [x] tests added / passed. 

`RecursiveLS` can only allow parameter constraints via the constructor (because the way it is implemented changes the dimensions of the state vector, which is set in the constructor).

### DIFF
--- a/statsmodels/regression/recursive_ls.py
+++ b/statsmodels/regression/recursive_ls.py
@@ -136,6 +136,10 @@ class RecursiveLS(MLEModel):
         return super(MLEModel, cls).from_formula(formula, data, subset,
                                                  constraints=constraints)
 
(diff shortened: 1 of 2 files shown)


bashtage:
LGMT. Merging to include in rc2.

ChadFulton:
If it doesn't mess up your rc2 schedule, I have 1 more bug fix PR that I'd like to get in rc2, which I can push in a few hours.

bashtage:
I just did RC2 but a tested bug can always go in even after RC (features, like I did, should usually not go in after a RC).

ChadFulton:
Sounds great, thanks!

PR #6424, opened 2020-01-16
REF: More descriptive error message for recursive least squares parameter constraints.

ChadFulton:
Follow-up to #6415 that just improves the error message so that users know how to impose valid constraints in `RecursiveLS` models even though `fix_params` doesn't work.


ChadFulton:
Unrelated failure on Travis.

PR #6501, opened 2020-02-11
ENH: state space: add "Cholesky factor algorithm" simulation smoothing

ChadFulton:
- [x] add example notebook
- [x] add support for diffuse initialization
- [x] tests added / passed. 

This adds an alternative simulation smoother for (some) state space models, which basically is designed to take advantage of the sparse Cholesky algorithm for banded positive definite matrices that is available in LAPACK. When applicable, it can reduce both processing time and memory costs for simulation smoothing (most prominantly for use in Bayesian applications, for Gibbs sampling).

The basic citation is Chan and Jeliazkov (2009) "Efficient Simulation and Integrated Likelihood Estimation in State Space Models", although these methods have been used outside of time-series applications for a while.

As I put in the documentation:

    In particular, this simulation smoother computes the joint posterior mean
    and covariance matrix for the unobserved state vector all at once, rather
    than using the recursive computations of the Kalman filter and smoother. It
    then uses these posterior moments to sample directly from this joint
    posterior. For some models, it can be more computationally efficient than
    the simulation smoother based on the Kalman filter and smoother.

The caveat is that it requires that the state space model have non-degenerate innovations in both the observation and state equations. This is not compatible with our SARIMAX or VARMA model classes (although for the VAR and AR cases, I am planning to write a new model class for which it would be compatible).


ChadFulton:
I added an example notebook, describing (a bit) the CFA approach and showing how our simulation smoothers (either the usual one or the new CFA one) can be used for Gibbs sampling in a time-varying parameter VAR model.

(We don't have a TVP-VAR model built-in yet, but it's something I'm working towards)

ChadFulton:
A brief note on performance: at least for the TVP-VAR model, in the example notebook, there doesn't seem to be any improvement on computing time from using the CFA simulation smoother, while Chan and Jeliazkov (2009) say they experienced about 33% improvement relative to the usual Kalman filter / smoother-based simulation smoother.

This might mean there is more that could be done in our implementation, but I will note that our simulation smoothers are about twice as fast (they both are, since they both take about the same amount of time) as the Matlab CFA simulation smoother in Chan and Jeliazkov's replication code for the TVP-VAR model. So another possibility was that Chan and Jeliazkov were comparing their CFA version (which takes advantage of Matlab's sparse array functionality) against a non-compiled Kalman filter / smoother implementation, and so that was the reason they saw performance improvements to the CFA version.

ChadFulton:
CI is green, so I will merge soon unless there are concerns / comments.

PR #6508, opened 2020-02-14
ENH: Add option for original Breusch-Pagan heteroscedasticity test

eirki:
- [x] tests added / passed  

This adds as an option to run the original heteroscedasticity test described by Breusch-Pagan.

I recently spent a few hours researching why the results from `het_breuschpagan` did not match Stata's output, only to realize that the standard heteroscedasticity test used in Stata is not implemented in Statsmodels. Instead Statsmodels uses a modified version, described by Koenker (1981). The existing implementation is equivalent to the Stata commands `estat hettest, iid` and `estat hettest, fstat`. By setting the parameter `iid=False`, the first two return values are now identical to Stata's default (`estat hettest`).

### DIFF
--- a/statsmodels/stats/diagnostic.py
+++ b/statsmodels/stats/diagnostic.py
@@ -824,7 +824,7 @@ def acorr_breusch_godfrey(res, nlags=None, store=False):
         return lm, lmpval, fval, fpval
 
 
-def het_breuschpagan(resid, exog_het, iid=True):
+def het_breuschpagan(resid, exog_het, robust=False):
     r"""
     Breusch-Pagan Lagrange Multiplier test for heteroscedasticity
 
@@ -845,9 +845,10 @@ def het_breuschpagan(resid, exog_het, iid=True):
     exog_het : array_like
         This contains variables suspected of being related to
         heteroscedasticity in resid.
-    iid : bool, default True
+    robust : bool, default False
         Flag indicating whether to use the Koenker version of the
-        test (default) or the original Breusch–Pagan version.
+        test (default) which assumes independent and identically distributed
+        error terms, or the original Breusch–Pagan version.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


bashtage:
`iid` doesn't seem that clear to me.  Why not `koenker=falue` or `robust=false` (I like robust since this option robistifies the test to violations of some assumptions).

eirki:
I agree, `iid` is unclear. I've changed it to `robust`, which is set to `False` by default.

bashtage:
I fixed my comment above.  I think the n R2 version is robust, so the new version should be `robust=True` as default.  Sorry for the confusion.

eirki:
That makes sense, I've swapped the boolean values.

statsmodels/stats/diagnostic.py
```diff
@@ -845,12 +845,16 @@ def het_breuschpagan(resid, exog_het):
     exog_het : array_like
         This contains variables suspected of being related to
         heteroscedasticity in resid.
+    robust : bool, default False
```
bashtage: default True

statsmodels/stats/diagnostic.py
```diff
@@ -845,12 +845,16 @@ def het_breuschpagan(resid, exog_het):
...
+        Flag indicating whether to use the Koenker version of the
+        test (default) which assumes independent and identically distributed
+        error terms, or the original Breusch–Pagan version.
```
bashtage: Maybe clarify

, or the original Breusch-Pagan version which assumes residuals are normally distributed.

bashtage:
Thanks.  Will merge on green.

eirki:
> Thanks. Will merge on green.

Great! Thanks for handling this so quickly!

bashtage:
Thanks.

PR #6554, opened 2020-02-26
ENH: Add xvals param to lowess smoother

### LINKED ISSUE #6543: LOWESS regression should allow arbitrary x-values
Currently `statsmodels.nonparametric.smoothers_lowess.lowess` takes `endog` and `exog` arrays and returns fitted values at the same `exog` values as provided, or alternatively, at a subset of those values if `delta` is provided. However lowess regression is well-defined at every potential `x`-value and it should be easy for a user to perform lowess regression at a given set of `x`-values, just as say a linear model provides a `predict()` function for computing regressed values at arbitrary points.

One compelling use-case is providing boot-strap confidence bounds to a lowess regression. One might ressample their `exog` and `endog` data and rerun the lowess. However, then the output at each ressample will give the regressed values at a different set of `x`-values and cannot easily be compared to find confidence bounds.

The simplest change to support this would be to allow a `x` parameter to `statsmodels.nonparametric.smoothers_lowess.lowess` which determines where the regression is evaluated. If `x` is passed in, then `delta` is ignored and a one-dimensional `out` array is returned with the regression values at the corresponding values of `x`. If `return_sorted = True` then the `x` would be sorted, the function would return a two-dimensional `out` array with the sorted `x` values and their corresponding regressed values.

The proposed solution above, however, is inconsistent with the `predict()`-style interface of `OLS` and other models. However, the lowess fit interface is already inconsistent with those so I see no uniform way to fit both styles together, so I propose the simpler interface.

#### Additional context
A [Stack Overflow question](https://stackoverflow.com/questions/36252434/predicting-on-new-data-using-locally-weighted-regression-loess-lowess) addresses this problem but has no completely satisfactory answer. The top answer performs a lowess fit and then interpolates that to get values at the given `x`-coordinates which will be practically useful but unsatisfactory and more work when compared to the proposed change.

Would a pull request implementing this feature likely be accepted?

tgbrooks:
xvals allows evaluation of the lowess smoother at
x-values other than those used in the smoothing, for example
fitting from some points and using that to 'predict' others

- [x] closes #6543
- [x] tests added / passed. 

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>


statsmodels/nonparametric/_smoothers_lowess.pyx
```diff
@@ -24,12 +24,16 @@ cdef inline double fmax(double x, double y): return x if x >= y else y
 
 DTYPE = np.double
 ctypedef np.double_t DTYPE_t
+cdef double NAN = float("NaN")
```
bashtage: You could just use `cdef double NAN = 0.0 / 0.0` or 

```
from cython.math cimport nan
cdef double NAN = nan("")
```

statsmodels/nonparametric/_smoothers_lowess.pyx
```diff
@@ -42,6 +46,8 @@ def lowess(np.ndarray[DTYPE_t, ndim = 1] endog,
         The y-values of the observed points
     exog : 1-D numpy array
         The x-values of the observed points. exog has to be increasing.
+    weights : 1-D numpy array
```
bashtage: weights or resid_weights?

statsmodels/nonparametric/_smoothers_lowess.pyx
```diff
@@ -51,13 +57,19 @@ def lowess(np.ndarray[DTYPE_t, ndim = 1] endog,
...
+        Whether xvals was provided as a an argument or whether
+        we are just using xvals = exog
 
     Returns
     -------
     out : numpy array
         A numpy array with two columns. The first column
         is the sorted x values and the second column the
         associated estimated y-values.
+    resid_weights: numpy array
```
bashtage: Are there different from weights?  Also (throughout ideally, return type should be `ndarray` as in `resid_weights: ndarray`.

statsmodels/nonparametric/tests/test_lowess.py
```diff
@@ -120,6 +120,16 @@ def test_options(self):
         assert_almost_equal(actual_lowess, actual_lowess1, decimal=13)
         assert_(actual_lowess.dtype is np.dtype(float))
 
+        # Test specifying xvals explicitly
```
bashtage: Could these be added as new tests of the feature, rather than to an existing test?

statsmodels/nonparametric/smoothers_lowess.py
```diff
@@ -32,10 +32,14 @@ def lowess(endog, exog, frac=2.0/3.0, it=3, delta=0.0, is_sorted=False,
     delta : float
         Distance within which to use linear-interpolation
         instead of weighted regression.
+    xvals: 1-D numpy array
```
bashtage: `xvals : ndarray, 1d`

statsmodels/nonparametric/smoothers_lowess.py
```diff
@@ -10,7 +10,7 @@
 import numpy as np
 from ._smoothers_lowess import lowess as _lowess
 
-def lowess(endog, exog, frac=2.0/3.0, it=3, delta=0.0, is_sorted=False,
+def lowess(endog, exog, frac=2.0/3.0, it=3, delta=0.0, xvals=None, is_sorted=False,
```
bashtage: The new optional should, unfortunately, go at the end.  I cold also live with making everything after (and including) `it=3` keyword only, so that `frac=2.0/3.0, *, it=3,...`.

bashtage: This breaks code but in a way that is loud and obvious.  Leaving it were it is could like to hard to trace breaks.

josef-pkt:
I think the name in the lowess function should be `exog_predict` instead of xvals.
In internal code `xvals` is ok.

statsmodels/nonparametric/_smoothers_lowess.pyx
tgbrooks: 'weights' are the data point weights for regressing at a specific point, while `resid_weights` is used for the overall down-weighting of outliers produced from multiple iterations and so don't depend upon the x values.

tgbrooks:
I've now made all the changes I wanted to make. Thanks for the feedback and let me know if anything else should be changed or added.

bashtage:
I checked coverage in the pyx files -- all of the new code is covered by your tests.

bashtage:
Any chance you could add the use of this to a notebook (or create a new one) to advertise this feature?  Ideally to produce confidence intervals as you intended in the first place.

tgbrooks:
That sounds like a good idea. It looks like the repo has .ipynb and .py files in examples/ but that the .py autogenerated from the .ipynb. How will I autogenerate it? Just export manually from jupyter?

josef-pkt:
Only commit notebooks without the output to save on repo size, i.e. clear output before committing

The py files are generated by manual export, so it does not always happen.
One advantage of also having the py-files is that they provides less noise changesets if we need to check the history (which doesn't happen very often for notebooks).

bashtage:
There is a script in tools that generates the py files.

tgbrooks:
I added the ipython notebook with a simple example. As well, I changed the `delta` behavior I mentioned above: now `delta` is not used on the `exog_predict` points, only on the `exog` and `endog` points. This means that the value of the fit at a point of `exog_predict` does not depend upon the other values in `exog_predict`. I think this reflects the most typical use-case best.

Let me know if any other changes are needed.

bashtage:
Looks great. Many thanks; this will be in the next release.

Issue #6359, opened 2019-12-20
ENH: Add Prediction Intervals to Holt-Winters class


samtriolo:
#### Is your feature request related to a problem? Please describe
I'd like for statsmodels holt-winters (HW) class to calculate prediction intervals (PI).

#### Describe the solution you'd like
Please include a parameter (or method, etc) in the [holt winters class](https://www.statsmodels.org/dev/generated/statsmodels.tsa.holtwinters.ExponentialSmoothing.html) that calculates prediction intervals for the user, including eg upper and lower x / y coordinates for various (and preferably customizable) confidence levels (eg 80%, 95%, etc).

#### Describe alternatives you have considered
It seems there are very few resources available regarding HW PI calculations.  Some only cover certain use cases - eg only additive, but not multiplicative, trend.  So performing the calculations myself in python seemed impractical and unreliable.  The only alternatives I know of are to use the R forecast library, which does perform HW PI calculations.  

#### Additional context
According to one of the more [commonly cited](https://otexts.com/fpp2/ets-forecasting.html) resources on the internet on this topic, HW PI calculations are more complex than other, more common PI calculations.  The same resource (and a couple others I found) mostly point to the [text book](http://www.exponentialsmoothing.net/) (specifically, chapter 6) written by the author of the R library that performs these calculations, to see further details about HW PI calculations.  

####  Additional References
- [The R library](https://www.rdocumentation.org/packages/stats/versions/3.6.1/topics/predict.HoltWinters) that implements holt-winters, including prediction interval calculations.
- Some academic papers that discuss HW PI calculations: [one](https://www.sciencedirect.com/science/article/abs/pii/016920709090103I) and [two](https://core.ac.uk/download/pdf/48532500.pdf)

wesleylongsworth:
This feature is the only reason my team hasn't fully migrated our HW forecasting app from R to Python . We've been successful with R for ~15 months, but have had to spend countless hours working around vague errors from R's forecast package. 

The PI feature is the only piece of code preventing us from fully migrating our enterprise forecasting tool from R to Python and benefiting from Python's much friendlier debugging experience.

ChadFulton:
Thanks for letting us know!

I'll just mention for the pure additive cases, v0.11 has a version of the exponential smoothing models that will allow for prediction intervals, via the model at `sm.tsa.statespace.ExponentialSmoothing`.

[...]

PR #6560, opened 2020-02-28
ENH: Holt-Winters simulations

s-scherrer:
- [x] tests added / passed. 

Implementation of `HoltWintersResults.simulate` as discussed in #6359, as a first step towards prediction intervals.

This is still work in progress, but I would like to get some feedback about the general direction/coding style/implementation details.

I didn't manage to reproduce the results from R's forecast package exactly in the current test, but the difference is low. How exact does it have to be, @ChadFulton?


s-scherrer:
>     * [``](https://github.com/statsmodels/statsmodels/blob/aed6a666980fbb78d146927f89b299b53f7dd3ba/statsmodels/tsa/holtwinters.py):

This is on purpose to follow the mathematical notation in FPP2. Should I change this to somewhat else?

statsmodels/tsa/holtwinters.py
```diff
@@ -430,6 +430,227 @@ def _fmt(x):
...
+            phi = 1
+        if seasonal is None:
+            s[:,:] = neutral_s
+        if not damped:
+            phi = 1
+
+        if error == 'add':
+            # estimate error variance from residuals
+            sigma = np.sqrt(np.sum(self.resid**2) / (len(self.resid) - n_params))
+            eps = np.random.randn(steps, nsim) * sigma
```
bashtage: Need to let a user (optionally) pass in a RandomState rather than using the global generator.

bashtage:
You need to check if matplotib is available, and if not skip that part of the test.  Ideally split the matplotlib and non-matpltlib then you can `@pytest.mark.matplotlib` the test that uses MPL to skip it when not available.

statsmodels/tsa/tests/test_holtwinters.py
```diff
@@ -358,6 +378,56 @@ def test_hw_seasonal_buggy(self):
...
+            self.austourists, seasonal_periods=4, trend="add", seasonal="mul",
+            damped=False
+        ).fit()
+
+        np.random.seed(0)
+        nsim = 100
+        steps = 20
+        sim = fit.simulate(steps, nsim=nsim, error="mul")
+
+        import matplotlib.pyplot as plt
```
bashtage: Make a plotting only test.

statsmodels/tsa/holtwinters.py
ChadFulton: In addition to supporting RandomState, I think we should also allow a few other cases:

1. User passes in the `eps` they want to use directly (helpful for replication / unit testing)
2. User passes in a scipy distribution function they want to use (allows non-Gaussian errors)
3. (Optional, we can leave this as a TODO) User specifies bootstrapping, in which case we re-sample the in-sample errors.

statsmodels/tsa/tests/test_holtwinters.py
s-scherrer: Sorry, I should have removed this before. This was just for myself as a first visual sanity check, not a real test. I removed it now.

s-scherrer:
I made the following changes:
* Added keyword arguments for passing a random state/seed, a scipy distribution function, fixed errors, or bootstrap
* Moved tests to the end of the test file, using pytest fixtures for the data instead of the test class
* The tests set up a ``HoltWintersResults`` object with exactly the same state as is obtained from R's forecast package for all stable models (the ones that forecast allows). With this, predictions match up to 5 decimals.
* Added a test that checks whether the different keyword options are accepted and run (no comparison of outputs)

statsmodels/tsa/tests/test_holtwinters.py
s-scherrer: Sorry, I forgot to remove this, this was not thought of as a real test, but a first visual sanity check for me. I removed it now.

statsmodels/tsa/holtwinters.py
s-scherrer: I implemented all these with two new keyword arguments:
* `seed` accepts a `RandomState` or an integer
* `random_errors` can be None, an array of values that are used directly, a scipy distribution function, or `"bootstrap"`.

ChadFulton:
Thanks for this! I should have time for a review this weekend.

s-scherrer:
@ChadFulton I just found that there's still an issue when the fit was performed with `use_boxcox=True`. I am working on it.

statsmodels/tsa/holtwinters.py
```diff
@@ -944,12 +1250,12 @@ def _predict(self, h=None, smoothing_level=None, smoothing_slope=None,
         level = lvls[1:nobs + 1].copy()
         if use_boxcox or use_boxcox == 'log' or isinstance(use_boxcox, float):
             fitted = inv_boxcox(fitted, lamda)
-            level = inv_boxcox(level, lamda)
```
s-scherrer: I don't really understand why `level`, `slope` and `season` are "back-transformed" here. For the simulations I need the values in the transformed space, therefore I commented them for now.

The tests are still running fine, does this mean we can remove this?

ChadFulton: That is interesting. I can't see why the back-transformation would be sensible, and I notice that R's forecast package does not back-transform the states.

The only question here is about backwards compatibility - if people are relying on the current behavior. But it looks like the current behavior might be best classified as a bug?

statsmodels/tsa/holtwinters.py
```diff
@@ -430,6 +433,309 @@ def _fmt(x):
...
+            neutral_s = 0
+
+        # set initial values
+        # (notation as in https://otexts.com/fpp2/ets.html)
+        y = np.empty((steps, nsim))
+        lvl = np.empty((steps + 1, nsim)) # lvl instead of l because of E741
+        b = np.empty((steps + 1, nsim))
+        s = np.empty((steps + m, nsim))
+        # the following uses python's index wrapping
+        lvl[-1,:] = self.level[start]
```
ChadFulton: space after comma

statsmodels/tsa/holtwinters.py
```diff
@@ -430,6 +433,309 @@ def _fmt(x):
...
+        b[-1,:] = self.slope[start]
+        if start == -1:
+            s[-m:,:] = np.tile(self.season[-m:], (nsim, 1)).T
+        elif 0 <= start and start < m:
+            initial_seasons = self.params["initial_seasons"]
+            _s = np.concatenate((initial_seasons[start+1:],
```
ChadFulton: space between operators like `+`

ChadFulton:
This is looking great, thanks! I just have a couple of comments.

The only other thing is that the updates should be cleaned up to match up with PEP8, I noted a couple of them (like spaces after commas, or spaces between operators like `+`).

statsmodels/tsa/holtwinters.py
```diff
@@ -430,6 +433,309 @@ def _fmt(x):
...
+    def simulate(self, steps, start=-1, nsim=1, error="add",
+                 random_errors=None, random_state=None):
+        r"""
+        Random simulations using the state space formulation.
+
+        Parameters
+        ----------
+        steps : int
+            The number of simulation steps.
+        start : int, optional
```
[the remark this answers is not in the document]
s-scherrer: Done, I very liberally copied the documentation and code regarding `anchor` from `MLEModel.simulate`.

statsmodels/tsa/holtwinters.py
s-scherrer: Yes, I would also classifiy this as a bug. 
What would be the best way to handle this? I could, for a start, store the original values in a private attribute and use these for the simulations. Then we could open a separate issue and PR to discuss backwards compability issues.

ChadFulton: What you suggest is definitely the ideal way forward, but a little annoying for you. If you don't mind doing that, that would be great.

ChadFulton:
I think this is ready to merge, except for updating the back-transformation to store the original values. If you want to do that, we can merge directly afterwards, assuming CI is still green.

Thanks again!

ChadFulton:
This is great, thanks very much for this PR.

s-scherrer:
Thanks a lot for your assistance. I had a lot of fun and learned quite a bit.

PR #6579, opened 2020-03-12
ENH: state space: compute smoothed state autocovariance matrices for arbitrary lags

ChadFulton:
- [x] tests added / passed. 

We already compute the `lag=1` smoothed autocovariance matrix for the state vector (i.e. Cov(t, t-1)) in the Cython recursions (because it is required for the EM algorithm applied to the transition equation, and so it's best if it's pretty fast).

This PR adds a new `smoothed_state_autocovariance` method to the `SmootherResults` class to allow computing autocovariances with any `lag = 0, +-1, +-2, ...`. One application of this is computing the "news" for nowcasting models, described in Banbura and Modugo (2014). Not in Cython since as far as I know, this is just for postestimation results, and won't be in an inner loop.


ChadFulton:
Squashed.

ChadFulton:
Also, this is still a semi-private method, since this PR only puts a new method in `SmootherResults` and doesn't make it available as a method in `MLEResults`, which I think is good to start with.

PR #6580, opened 2020-03-12
BUG: `MLEResults.states.predicted` has wrong index

ChadFulton:
- [x] tests added / passed. 

The `states` attribute added in v0.11 allows accessing the predicted, filtered, and smoothed states (and covariance matrices) as Pandas objects with proper indexes / columns.

Unfortunately, I set the index for the predicted states as going from e.g. `-1, 0, ..., nobs` instead of `0, ..., nobs, nobs + 1`. This fixes the problem and adds unit test.

We may want to issue a bugfix release for this.


ChadFulton:
Failure unrelated (Azure brownout for Mac OSX 10.13), will merge.

ChadFulton:
@bashtage sorry to bother you since I should know this - but is the right step now for me to backport this into maintenance/0.11.x?

PR #6590, opened 2020-03-17
ENH: Include figsize as parameter for IRF plot

cimarieta:
Include figsize as parameter for IRF-related plots, since it is available for [`irf_grid_plot`](https://github.com/statsmodels/statsmodels/blob/master/statsmodels/tsa/vector_ar/plotting.py#L188). Without this, it is still possible to change the figure size, but it has to be done afterwards through `fig.set_size_inches(x, y)`, so this PR brings ease to the customization of figure size.

- [ ] closes #xxxx
- [ ] tests added / passed. 

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>

### DIFF
--- a/statsmodels/tsa/vector_ar/irf.py
+++ b/statsmodels/tsa/vector_ar/irf.py
@@ -109,7 +109,7 @@ class BaseIRAnalysis(object):
             pass {'fontsize' : 8} or some number to your taste.
         plot_params : dict
 
-        figsize: (int, int), default (10, 10)
+        figsize: (float, float), default (10, 10)
             Figure size (width, height in inches)
         plot_stderr: bool, default True
             Plot standard impulse response error bands
@@ -195,7 +195,7 @@ class BaseIRAnalysis(object):
             pass {'fontsize' : 8} or some number to your taste.
         plot_params : dict
 
-        figsize: (int, int), default (10, 10)
+        figsize: (float, float), default (10, 10)
             Figure size (width, height in inches)
         plot_stderr: bool, default True
             Plot standard impulse response error bands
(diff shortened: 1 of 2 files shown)


statsmodels/tsa/vector_ar/irf.py
```diff
@@ -88,9 +88,9 @@ def cum_effect_cov(self, *args, **kwargs):
         raise NotImplementedError
 
     def plot(self, orth=False, impulse=None, response=None,
-             signif=0.05, plot_params=None, subplot_params=None,
-             plot_stderr=True, stderr_type='asym', repl=1000,
-             seed=None, component=None):
+             signif=0.05, plot_params=None, figsize=(10, 10),
```
bashtage: Need a `,*,` to change order.  I would place the `,*,` after orth so that all of the parameters are keyword only except orth.

statsmodels/tsa/vector_ar/irf.py
```diff
@@ -228,6 +233,7 @@ def plot_cum_effects(self, orth=False, impulse=None, response=None,
                                      hlines=lr_effects,
                                      subplot_params=subplot_params,
                                      plot_params=plot_params,
+                                     figsize=figsize,
```
bashtage: Need a test.

statsmodels/tsa/vector_ar/irf.py
cimarieta: Thanks, I have included it now.

statsmodels/tsa/vector_ar/irf.py
cimarieta: Finally, I have included one test specifically for the figure size. Please let me know if there is any other changes I can make.

bashtage:
On last change needed.  Need to decorate the test with @pytest.mark.matplotlib since it requires MPL

cimarieta:
As requested, I added the decorator to the test. Thank you for reviewing it :)

bashtage:
Thanks. It should pass now and then will go in.

bashtage:
Thanks!

PR #6601, opened 2020-03-24
Update interactions_anova.ipynb

seanlane:
Fix URL that has been changed.

Nothing material has changed, it is trivial to verify that the old URL results in a 404 error while the new link returns content from the aforementioned course that this example is based on.


bashtage:
Thanks.

PR #6599, opened 2020-03-21
DOC/ENH: example notebook: stationarity and detrending (ADF/KPSS)

manmeetchaudhuri:
Hi,

I prepared an example notebook containing the following features:
- test (adf and kpss) for stationarity of a time series
- detrending a time series by differencing

I am motivated to contribute more towards the growth of "statsmodels" library. This is my first and a small contribution to begin with. 

I am looking forward to feedback and guidance from the senior members. 

Thanks and kind regards,
Manmeet Kumar Chaudhuri.

PR #6614, opened 2020-03-30
DOC/ENH: example notebook (r): stationarity and detrending (ADF/KPSS)

manmeetchaudhuri:
Hi @ChadFulton 
 
Reference: #6599 

I had to close the pull request because of some issues at my end. I have taken a note of all your comments and have accounted for them in this pull request. Moreover, I have tried to reply to all your comments on #6599. The replies are available at #6599. I shall be happy to address any more comments and make suitable changes in the documents. 

I would like to contribute more example notebooks and would appreciate if you would like me to focus on some specific topics.

Thanks!


ChadFulton:
This looks good to me, thanks very much! I think this will be helpful to users.

(Just FYI, I think that there may still be some language in the notebook that we might want to adjust later - i.e. I prefer describing hypothesis test results as providing evidence rather than in absolute terms - but we can do that in a follow up PR).

PR #6620, opened 2020-04-04
ENH: add rsquared for statespace model #4734

### LINKED ISSUE #4734: ENH: Add rsquared and variants to state space models
Currently the state space models don't compute any kind of R^2 type value, but it might be handy / desirable in some cases. "Forecasting, structural time series models and the Kalman filter" (Harvey, 1989) section 5.5.5 describes three variants:

- R^2 comparing fit to a model with just the mean (i.e. usual R^2): 1 - SSE / SSM
- R^2 comparing fit to a random walk + drift model: 1 - SSE  / SSDM (SSDM = sum of squares of first differences around the mean)
- R^2 comparing fit to a random walk + drift + seasonal dummies: 1 - SSE / SSDSM (SSDSM = sum of squares of first differences around the seasonal means)

This should not be too difficult to add (and also not urgent), so it might be good for someone who is interested in contributing a first PR.

BenjaminLiuPenrose:
- [x] closes #4734
- [ ] tests added / passed. 

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>

### DIFF
--- a/statsmodels/tsa/statespace/mlemodel.py
+++ b/statsmodels/tsa/statespace/mlemodel.py
@@ -2894,6 +2894,25 @@ class MLEResults(tsbase.TimeSeriesModelResults):
         """
         return self.params / self.bse
 
+    @cache_readonly
+    def rsquared(self):
+        """
+        (float) conventional R-squared, 1 - sse/ssm
+        """
+        ssm = np.var(self.endog) * len(self.endog)
+        return 1. - self.sse / ssm
+
+    @ cache_readonly
+    def rsquared_difference(self)：
+        """
+        (float) R-squared, 1 - SSE / SSDM
+        SSDM = sum of squares of first differences around mean
+        see eq. 5.5.14 in "Forecasting, structural time series
+        models and the Kalman filter" (Harvey, 1989)
+        """
+        ssdm = np.var(np.diff(self.endog)) * (len(self.endog) - 1)
+        return 1. - self.sse / ssdm
+
     def test_normality(self, method):
         """
         Test for normality of standardized residuals.


BenjaminLiuPenrose:
where should I write the test case to pass the `codecov/patch `

ChadFulton:
Thanks very much for submitting this PR, @BenjaminLiuPenrose! This will be a really nice feature to have.

I have a couple of suggestions:

1. For computing the SSE, you're currently using the large-sample approximation of (5.5.3), because `self.resid` corresponds to v_t and not to \tilde v_t (which is `self.standardized_forecasts_error`). Instead, why don't we use the finite prediction error variance defined in (5.5.2), so that you would compute the SSE as in (5.5.13):

    ```python
    d = np.maximum(self.loglikelihood_burn, self.nobs_diffuse)
    srss = np.sum(self.standardized_forecasts_error[0, d:]**2)
    f_T = self.forecasts_error_cov[0, 0, -1] * srss
    sse = f_T * rss
    ```

2. Since there are not multiple ways to define the R^2 depending on the baseline model used as a comparison, how about defining one new method, `get_rsquared` that accepts one argument `baseline`. `baseline` could accept either `"mean"`, which is what you currently have as the `rsquared` attribute, but could also accept `"rwdrift"`, which is what you currently have as `rsquared_difference`. I think the default value should be `"rwdrift"`. Eventually (or in this PR if you like), we could add a `seasonal` argument to allow specifying the R^2 that Harvey has in (5.5.17).

3. Then the two properties that you added would just call that method with the argument `baseline="mean"`, and `baseline="rwdrift"`. I also suggest that you rename the attributes in `MLEResuilts` to be `rsquared_mean`, `rsquared_rwdrift`.

4. I would suggest having the `rsquared` property raise a `NotImplementedError`, with a message letting users know that in state space models there is not a single comparison model that will always be useful, and suggesting that they use `rsquared_rwdrift`. You can also mention that `rsquared_mean` exists, but is not generally recommended. I think this will be helpful, since the basic `rsquared_mean` is not usually so helpful in time series contexts.

5. We need to make sure that these don't return erroneous numbers for multivariate models. For this PR, if you want to just support univariate models, that's okay. In that case, you should raise a `NotImplementedError` if `self.model.k_endog > 1`. If you want to try to support multivariate models, 

Thanks again!

BenjaminLiuPenrose:
> Thanks very much for submitting this PR, @BenjaminLiuPenrose! This will be a really nice feature to have.
> 
> I have a couple of suggestions:
> 
> 1. For computing the SSE, you're currently using the large-sample approximation of (5.5.3), because `self.resid` corresponds to v_t and not to \tilde v_t (which is `self.standardized_forecasts_error`). Instead, why don't we use the finite prediction error variance defined in (5.5.2), so that you would compute the SSE as in (5.5.13):
>    ```python
>    d = np.maximum(self.loglikelihood_burn, self.nobs_diffuse)
>    srss = np.sum(self.standardized_forecasts_error[0, d:]**2)
>    f_T = self.forecasts_error_cov[0, 0, -1] * srss
>    sse = f_T * rss
>    ```
> 2. Since there are multiple ways to define the R^2 depending on the baseline model used as a comparison, how about defining one new method, `get_rsquared` that accepts one argument `baseline`. `baseline` could accept either `"mean"`, which is what you currently have as the `rsquared` attribute, but could also accept `"rwdrift"`, which is what you currently have as `rsquared_difference`. I think the default value should be `"rwdrift"`. Eventually (or in this PR if you like), we could add a `seasonal` argument to allow specifying the R^2 that Harvey has in (5.5.17).
> 3. Then the two properties that you added would just call that method with the argument `baseline="mean"`, and `baseline="rwdrift"`. I also suggest that you rename the attributes in `MLEResults` to be `rsquared_mean`, `rsquared_rwdrift`.
> 4. I would suggest having the `rsquared` property raise a `NotImplementedError`, with a message letting users know that in state space models there is not a single comparison model that will always be useful, and suggesting that they use `rsquared_rwdrift`. You can also mention that `rsquared_mean` exists, but is not generally recommended. I think this will be helpful, since the basic `rsquared_mean` is not usually so helpful in time series contexts.
> 5. We need to make sure that these don't return erroneous numbers for multivariate models. For this PR, if you want to just support univariate models, that's okay. In that case, you should raise a `NotImplementedError` if `self.model.k_endog > 1`. If you want to try to support multivariate models, that's great too!
> 
> Thanks again!

Sure. Will work on that

ChadFulton:
**Note**: there was typo in my code computing the `sse`, but I have corrected it above.

BenjaminLiuPenrose:
@ChadFulton how about now?

ChadFulton:
That's great, thanks!  I will have a change to make a couple of minor comments on the code itself hopefully later today. The only major thing remaining is coming up with some unit tests.

BenjaminLiuPenrose:
> That's great, thanks! I will have a change to make a couple of minor comments on the code itself hopefully later today. The only major thing remaining is coming up with some unit tests.

sure;)

BenjaminLiuPenrose:
@ChadFulton  any updates?

statsmodels/tsa/statespace/mlemodel.py
```diff
@@ -2894,6 +2894,76 @@ def zvalues(self):
...
+                "To be Implemented for Multivariate Model")
+        if baseline == "rwdrift":
+            return self.rsquared_rwdrift
+        elif baseline == "mean":
+            return self.rsquared_mean
+        elif baseline == "seasonal":
+            raise self.rsquared_seasonal
+        else:
+            raise NotImplementedError(
+                "baseline must be in ['rwdrife', 'mean', 'seasonal']")
```
ChadFulton, line 2909: typo, should be rwdrift

statsmodels/tsa/statespace/mlemodel.py
```diff
@@ -2894,6 +2894,76 @@ def zvalues(self):
...
+
+    @cache_readonly
+    def sse_finite(self):
```
ChadFulton, line 2912: Let's move this to be a local variable within `get_rsquared` for now, rather than a property.

statsmodels/tsa/statespace/mlemodel.py
```diff
@@ -2894,6 +2894,76 @@ def zvalues(self):
...
+
+    @cache_readonly
+    def rsquared(self):
+        """
+        In state space models there is not a single comparison
+        model that will always be useful for rsquared computation.
+        However, we have implmented `rsquared_rwdrift` and `rsquared_mean`
+        It is recommended to use `rsquared_rwdrift`
+        """
+        raise NotImplementedError(
```
ChadFulton, line 2926: I think that the text you have as the docstring should be the message given in the `NotImplementedError`

ChadFulton:
I have a couple of comments, but the main thing now is to get some unit tests. Are you familiar with unit testing?

Do you know of any resources we could test this against, or have you thought about other ways that we could write some tests?

BenjaminLiuPenrose:
> I have a couple of comments, but the main thing now is to get some unit tests. Are you familiar with unit testing?
> 
> Do you know of any resources we could test this against, or have you thought about other ways that we could write some tests?

working on that. where should I put the unit test? i guess under some /test dir

statsmodels/tsa/statespace/mlemodel.py
BenjaminLiuPenrose: i am pulling all calc under get_rsquared(). All rsquared property do nothing but will just call get_rsquared()

statsmodels/tsa/statespace/mlemodel.py
```diff
@@ -2894,6 +2894,76 @@ def zvalues(self):
...
+            return f_T * srss
+        sse = sse_finite()
+        if baseline == "rwdrift":
+            ssdm = np.var(np.diff(self.endog)) * (len(self.endog) - 1)
+            return 1. - sse / ssdm
+        elif baseline == "mean":
+            ssm = np.var(self.endog) * len(self.endog)
+            return 1. - sse / ssm
+        elif baseline == "seasonal":
+            import statsmodels.api as sm
```
ChadFulton, line 2914: Best practice is to just import what you need - so in this case, maybe `from statsmodels.regression.linear_model import OLS`

statsmodels/tsa/statespace/mlemodel.py
```diff
@@ -2894,6 +2894,76 @@ def zvalues(self):
         """
         return self.params / self.bse
 
+    def get_rsquared(self, baseline="rwdrift", **kwargs):
+        # if self.model.k_endog > 1:
+        #     raise NotImplementedError(
+        #         "To be Implemented for Multivariate Model")
+        def sse_finite():
```
ChadFulton, line 2901: No need to make this a function anymore - it's probably a bit cleaner to compute `sse` instead of delegating that to a function.

statsmodels/tsa/statespace/mlemodel.py
```diff
@@ -2894,6 +2894,76 @@ def zvalues(self):
         """
         return self.params / self.bse
 
+    def get_rsquared(self, baseline="rwdrift", **kwargs):
```
ChadFulton, line 2897: I'd replace `kwargs` with just a keyword argument `seasonal` that defaults to `None`.

statsmodels/tsa/statespace/mlemodel.py
```diff
@@ -2894,6 +2894,76 @@ def zvalues(self):
...
+        sse = sse_finite()
+        if baseline == "rwdrift":
+            ssdm = np.var(np.diff(self.endog)) * (len(self.endog) - 1)
+            return 1. - sse / ssdm
+        elif baseline == "mean":
+            ssm = np.var(self.endog) * len(self.endog)
+            return 1. - sse / ssm
+        elif baseline == "seasonal":
+            import statsmodels.api as sm
+            seasonal = 12 if 'seasonal' not in kwargs else kwargs['seasonal']
```
ChadFulton, line 2915: Instead of a default, I suggest raising an error if `baseline == 'seasonal'` but they didn't provide the `seasonal` argument. 

In the future, we can expand this to take account of the model's seasonal term or the frequency of a Pandas index, if it's given.

ChadFulton:
Thanks for the edits, that's looking good.  I had a couple more comments.

> @ChadFulton guess I will add test_summary_rsquared in \statsmodels\statsmodels\tsa\statespace\tests\test_mlemodel.py,

That sounds good.

> but what I can do for now is to check r2 is displayed but I don't know the 'ground true' value for each of r2 wrt to the dummy_model

We should be able to compute some known result to test against, at least for special cases.  For example, the `SARIMAX` model supports a model that is equivalent to linear regression. 
Using that case, we could check that `rsquared_mean` is equivalent to the `rsquared` from `OLS` on the same dataset.

> also, can you also point me out how to extend r2 metric for multi-endog variables problem?

Essentially, this is just computing the value separately for each of the endogenous variable.

statsmodels/tsa/statespace/mlemodel.py
BenjaminLiuPenrose: sure

statsmodels/tsa/statespace/mlemodel.py
BenjaminLiuPenrose: maybe MLEResults users want to specify parameters for different type of baseline models, like 'order1or2 difference'? so still leave kwargs there. it looks a bit neat as well

BenjaminLiuPenrose:
@ChadFulton test case added, it is a trivial case

BenjaminLiuPenrose:
@ChadFulton not sure how to print R2 for multivariate case

ChadFulton:
This is looking good, thanks!

> not sure how to print R2 for multivariate case

I agree that this is a tough call. Currently for other output, like the test statistics, I just print the list, but that is not very pretty.  One option would be to create a table in the summary output that displays the R2 for each `endog` variable in the multivariate case.

statsmodels/tsa/statespace/mlemodel.py
```diff
@@ -4031,8 +4109,12 @@ def summary(self, alpha=.05, start=None, title=None, model_name=None,
             ('No. Observations:', [self.nobs]),
             ('Log Likelihood', ["%#5.3f" % self.llf]),
         ]
-        if hasattr(self, 'rsquared'):
-            top_right.append(('R-squared:', ["%#8.3f" % self.rsquared]))
+        if hasattr(self, 'rsquared_rwdrift'):
+            top_right.append(('R-squared:',
+                              ["%#8.3f" % self.get_rsquared('rwdrift')[0]]))
+        if hasattr(self, 'rsquared_mean'):
```
ChadFulton, line 4115: I think by default we should only show the `rwdrift` version

statsmodels/tsa/statespace/mlemodel.py
```diff
@@ -2894,6 +2894,84 @@ def zvalues(self):
...
+        model that will always be useful for rsquared computation.
+        However, we have implmented `rsquared_rwdrift` and `rsquared_mean`
+        It is recommended to use `rsquared_rwdrift`
+        """
+        return self.get_rsquared('')
+
+    @cache_readonly
+    def rsquared_mean(self):
+        """
+        (float) conventional R-squared, 1 - sse/ssm
```
ChadFulton, line 2953: Should this be `(float or array)` now that the multivariate case is supported?

statsmodels/tsa/statespace/mlemodel.py
BenjaminLiuPenrose: done

BenjaminLiuPenrose:
I'm now printing it as np array

BenjaminLiuPenrose:
any ideas why the ci/appveyor/pr test failed?

ChadFulton:
> any ideas why the ci/appveyor/pr test failed?

I'm not sure, but it seems like it's unrelated.

On a side note - this is looking good, and I'm sorry for the delay - it will probably take me a day or two to get back to this, based on other time commitments.

BenjaminLiuPenrose:
> > any ideas why the ci/appveyor/pr test failed?
> 
> I'm not sure, but it seems like it's unrelated.
> 
> On a side note - this is looking good, and I'm sorry for the delay - it will probably take me a day or two to get back to this, based on other time commitments.

sure, np

BenjaminLiuPenrose:
@ChadFulton updates?

statsmodels/tsa/statespace/tests/test_mlemodel.py
```diff
@@ -597,6 +597,23 @@ def test_summary():
         res.summary()
 
 
+def test_summary_rsquared():
+    from statsmodels.regression.linear_model import OLS
+    endog = np.array([1, 2, 4, 8, 16])
+    exog = np.array([1, 2, 4, 8, 16])
```
ChadFulton, line 603: This shouldn't be a perfect fitting model, so you should make exog not identical to endog.  Also, so that the OLS R^2 matches, exog needs to include a constant column.

statsmodels/tsa/statespace/tests/test_mlemodel.py
```diff
@@ -597,6 +597,23 @@ def test_summary():
...
+
+    mod = sarimax.SARIMAX(endog, exog, order=(0, 0, 0), trend='c')
```
ChadFulton, line 605: Add `concentrate_scale=True` so you don't have to estimate the variance parameter.

BenjaminLiuPenrose:
> cases

Yea, I agreed. What do you think will be a good test case for `rwdrift` and `seasonal`?

statsmodels/tsa/statespace/mlemodel.py
```diff
@@ -2894,6 +2894,85 @@ def zvalues(self):
...
+        # calc rsquared wrt different baseline models
+        if baseline == "rwdrift":
+            ssdm = np.var(np.diff(endog, axis=0), axis=0) * (len(endog) - 1)
+            rsquared = 1. - sse / ssdm
+        elif baseline == "mean":
+            ssm = np.var(endog, axis=0) * len(endog)
+            rsquared = 1. - sse / ssm
+        elif baseline == "seasonal":
+            from statsmodels.regression.linear_model import OLS
+            from statsmodels.tools.tools import add_constant
```
bashtage, line 2915: You could just use AutoReg which directly supports this model specification at a high level.

```
statsmodels.tsa.ar_model.AutoReg(endog, 0, trend='c', seasonal=True, periods=seasonal)
```

statsmodels/tsa/statespace/mlemodel.py
```diff
@@ -2894,6 +2894,85 @@ def zvalues(self):
...
+            return rsquared
+
+    def rsquared(self):
+        """
+        In state space models there is not a single comparison
+        model that will always be useful for rsquared computation.
+        However, we have implmented `rsquared_rwdrift` and `rsquared_mean`
+        It is recommended to use `rsquared_rwdrift`
+        """
+        return self.get_rsquared('')
```
bashtage, line 2949: Doesn't this produce an error?

bashtage: If the error is intentional, you should directly raise NotImplementedError here. You can pull the common message outside the class to share it.

statsmodels/tsa/statespace/mlemodel.py
```diff
@@ -2894,6 +2894,85 @@ def zvalues(self):
...
+
+    @cache_readonly
+    def rsquared_mean(self):
+        """
+        (float or array) conventional R-squared, 1 - sse/ssm
```
bashtage, line 2954: The docstring all have the wrong format.  Could you please update to NumPy style?

bashtage:
Overall pretty close and a useful contribution. It would be good to get this across the line.

PR #6631, opened 2020-04-07
DOC/TST: minor fixes for holtwinters simulate

s-scherrer:
Just two very small additions to #6560:

- I fixed the alignment of equations in the docstring of simulate
- I saw that one of the new tests was failing. I added a seed to the RNG to prevent it from failing due to RNG randomness.


ChadFulton:
Unrelated CI errors. This looks good, thanks!

PR #6654, opened 2020-04-21
Handle pathlib.Path objects

### LINKED ISSUE #6652: Handle pathlib paths with `get_file_obj`
I'm trying to read a statsmodels.tsa.statespace.sarimax.SARIMAXResults pickle file I've saved. I'd like to pass a pathlib.Path object as I find it easier to handle (I work with different machines, Mac and Windows).

The function `open` already supports pathlib objects. Why not using it?

You could replace `statsmodels.iolib.openfile.get_file_obj` with:

```python
if _is_string_like(fname):
    return _open(fname, mode, encoding)
try:  # try opening it with 'open'
    return open(fname, mode, encoding)
except TypeError:  # else, maybe return context manager (please check if it's always TypeError)
    try:
        # Make sure the object has the write methods
        if 'r' in mode:
            fname.read
        if 'w' in mode or 'a' in mode:
            fname.write
    except AttributeError:
        raise ValueError('fname must be a string or a file-like object')
    return EmptyContextManager(fname)
```

This is just a suggestion though :)

giuliobeseghi:
- [ ] closes #6652 
- [ ] tests added / passed.

### DIFF
--- a/statsmodels/iolib/tests/test_pickle.py
+++ b/statsmodels/iolib/tests/test_pickle.py
@@ -19,16 +19,17 @@ def test_pickle():
     assert_equal(a, b)
 
     # test with pathlib
-    path_pathlib = pathlib.Path(tmpdir) / "res.pkl"
+    path_pathlib = pathlib.Path(tmpdir) / "res2.pkl"
     save_pickle(a, path_pathlib)
-    b = load_pickle(path_pathlib)
-    assert_equal(a, b)
+    c = load_pickle(path_pathlib)
+    assert_equal(a, c)
 
     # cleanup, tested on Windows
     try:
         import os
 
-        os.remove(tmpdir + "/res.pkl")
+        os.remove(path_str)
+        os.remove(path_pathlib)
         os.rmdir(tmpdir)
     except (OSError, IOError):
         pass
@@ -38,6 +39,6 @@ def test_pickle():
     fh = BytesIO()
     save_pickle(a, fh)
     fh.seek(0, 0)
-    c = load_pickle(fh)
+    d = load_pickle(fh)
     fh.close()
-    assert_equal(a, c)
+    assert_equal(a, d)
--- a/statsmodels/iolib/tests/test_pickle.py
+++ b/statsmodels/iolib/tests/test_pickle.py
@@ -1,4 +1,5 @@
 from io import BytesIO
+import pathlib
 import tempfile
 
 from numpy.testing import assert_equal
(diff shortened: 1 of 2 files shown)


bashtage:
The failures are not yours.  This is working.

statsmodels/iolib/tests/test_pickle.py
```diff
@@ -8,16 +9,26 @@
...
-    b = load_pickle(tmpdir+'/res.pkl')
+
+    # test with str
+    path_str = tmpdir + "/res.pkl"
+    save_pickle(a, path_str)
+    b = load_pickle(path_str)
+    assert_equal(a, b)
+
+    # test with pathlib
+    path_pathlib = pathlib.Path(tmpdir) / "res.pkl"
```
bashtage, line 22: Maybe use a different file name just to be sure it is writing the file. res2.pkl.

statsmodels/iolib/tests/test_pickle.py
```diff
@@ -8,16 +9,26 @@
...
+    save_pickle(a, path_pathlib)
+    b = load_pickle(path_pathlib)
     assert_equal(a, b)
 
     # cleanup, tested on Windows
     try:
         import os
-        os.remove(tmpdir+'/res.pkl')
+
+        os.remove(tmpdir + "/res.pkl")
```
bashtage, line 31: Need to remove both here too then.

giuliobeseghi:
Also, we could split `test_pickle` to three different tests:

- test_pickle_from_string
- test_pickle_from_path_obj
- test_pickle_from_file_handler

bashtage:
I'm happy with the way it is; best would be to use `@pytest.mark.parameterize("path",[<a string>, <a Pathlib>, <a fh>])`.  Happy if you want to do this or not.

bashtage:
I missed one. Could you update the docstring here:

https://github.com/statsmodels/statsmodels/pull/6654/files#diff-bdb9c768dd6d4c03d15a5015c523a710R43

bashtage:
This is line 43, which is above your changes.

giuliobeseghi:
> I'm happy with the way it is; best would be to use `@pytest.mark.parameterize("path",[<a string>, <a Pathlib>, <a fh>])`. Happy if you want to do this or not.

We'd have to define the pathlib obj and the fh outside (maybe at the top of the module). Then we'd have to add an if/else block to handle either str and pathlib objs or fhs in two different ways. I think it's fine as it is now (although if the test fails, it fails for the three of them together).

bashtage:
Thanks for this.

PR #6664, opened 2020-04-25
Speed up HC2/HC3 standard error calculation, using less memory

pag:
This pull requests speeds up the calculation of HC2 and HC3 covariances in `statsmodels.regression.linear_model.OLS` for large N and uses much less memory when doing so: of the order of N rather than N^2 for a design matrix of size NxK. Without this improvement it's very impractical to estimate the HC2 covariance matrix for a problem with N=1e5 (would need >70gb of memory). Clearly HC2/3 covariances will be pretty similar to HC0 for large samples because both are consistent estimators, but it's important to be able to verify that sometimes.

The patch is tiny: instead of calculating the full hat matrix and then taking its diagonal with `h = diag(X C X')` where `C = inv(X'X)`, I produce the diagonal directly with `h(i) = Σi Σj Σk X(i,j)X(i,k)C(k,j)` via a call to `np.einsum`.

I've not included any additional tests because I think that the existing ones in `test_regression.py` suffice to check that the new implementation produces the same results as before. Please let me know if you disagree!

Performance improvement for various N with K=5: 

HC2 cov times

The timing code used was:

```python
import timeit
import numpy as np
import statsmodels.api as sm


times = {}

k = 5
for n in (10, 100, 1e3, 5e3, 1e4, 2.5e4, 5e4):
    n = int(n)
    print(n)
    
    y = np.arange(n)
    X = sm.add_constant(np.random.randn((n), k - 1))
    
    t = %timeit -o sm.OLS(y, X).fit(cov_type='HC2')
    times[n] = t.best
```

### DIFF
--- a/statsmodels/regression/linear_model.py
+++ b/statsmodels/regression/linear_model.py
@@ -1850,6 +1850,10 @@ class RegressionResults(base.LikelihoodModelResults):
                    scale[:, None] * self.model.pinv_wexog.T)
         return H
 
+    def _abat_diagonal(self, a, b):
+        # equivalent to np.diag(a @ b @ a.T)
+        return np.einsum('ij,ik,kj->i', a, a, b)
+
     @cache_readonly
     def cov_HC0(self):
         """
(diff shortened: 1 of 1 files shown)


bashtage:
Many thanks.

PR #6682, opened 2020-05-01
ENH/DOC: Notebook describing how to create state space custom models

cdagnino:
… models

I'm adding an example notebook that considers different versions of state space custom models.
I start with something relatively simple and then add complexity and explain the changes

- [ ] closes #xxxx
- [ ] tests added / passed. 

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>


cdagnino:
@ChadFulton Let me know if you have some time to comment on this :D

bashtage:
Some of the math seems to have issues:

https://github.com/statsmodels/statsmodels/blob/9f589231162aa7151065a15ec343690e5ba91c3f/examples/notebooks/statespace_custom_models.ipynb

ChadFulton:
> Some of the math seems to have issues:


Strange - the math doesn't render for me on Github, but when I download the notebook it rendered fine in Jupyter.  I wonder what will happen when it's rendered for the website?

ChadFulton:
Basically, this looks great to me.  We've had a number of people with questions about custom state space models and time-varying models in particular, and I think this will be really useful. Thanks for submitting this.

- There are two things I noticed in the notebook, but I'm having trouble with Github putting the comments inline in the file, so I'll put them below.
- We need to clear the output of the notebook before we merge it in (to e.g. avoid having megabytes of images from the notebooks in Github).  You can do that when you're running the notebook, using "Cell > All Output > Clear". Then commit that clean version of the notebook.
- Finally, we need to add it in the manifest file for the web page and create an image that goes along with the title. You can take a look at https://github.com/statsmodels/statsmodels/pull/6614/files for a recent notebook addition that did this. The image isn't too important, but it's usually nice to have a graph or something to catch the eye.  Maybe the figure from the first example showing the coefficients?

The two things I noticed were:

1. Just after the "Model 2" header, you have "explain how it's an extension", which I guess maybe is a placeholder that should be removed or filled in?
2. In the "Model 3" section, under "1) The __init__ function ...", you have "Note that we didn't have to specify `k_endog` anywhere. The initialization does this for us after checking the dimensions of the `exog` matrix."  but I think maybe you meant the "`endog` matrix" instead of "`exog` matrix"?

ChadFulton:
Two more things:

1. We haven't used type hints anywhere else, so it might be cleaner to remove those.
2. It looks like the actual definition of `TVRegressionExtended` doesn't currently take into account all of the improvements described above it (e.g. to transform/untransform).

bashtage:
FWIW I've started adding type hints in my most recent code. I've found it provides better quality code and improves intellisense in PyCharm.

cdagnino:
> Some of the math seems to have issues:
> 
> https://github.com/statsmodels/statsmodels/blob/9f589231162aa7151065a15ec343690e5ba91c3f/examples/notebooks/statespace_custom_models.ipynb

It looks good in my notebook! I guess it might be some edge case that Jupyter knows how to render but Github doesn't?

cdagnino:
Thanks for the comments!

> * We need to clear the output of the notebook before we merge it in (to e.g. avoid having megabytes of images from the notebooks in Github).  You can do that when you're running the notebook, using "Cell > All Output > Clear". Then commit that clean version of the notebook. 

[DONE]

> * Finally, we need to add it in the manifest file for the web page and create an image that goes along with the title. You can take a look at https://github.com/statsmodels/statsmodels/pull/6614/files for a recent notebook addition that did this. The image isn't too important, but it's usually nice to have a graph or something to catch the eye.  Maybe the figure from the first example showing the coefficients?

 [DONE, I think]


> 1. Just after the "Model 2" header, you have "explain how it's an extension", which I guess maybe is a placeholder that should be removed or filled in?

[Yep! Filled in]

> 2. In the "Model 3" section, under "1) The **init** function ...", you have "Note that we didn't have to specify `k_endog` anywhere. The initialization does this for us after checking the dimensions of the `exog` matrix."  but I think maybe you meant the "`endog` matrix" instead of "`exog` matrix"?

[Yep, Done]

ChadFulton:
Looks good to me, thanks for this addition! Notwithstanding my slow response, I hope you'll consider submitting additional notebooks if something occurs to you.

After merge, I'll check the website to see whether we need to resize the image (I'm not sure how it'll end up looking on the website, since it's 1000 x 300 resolution), and that the math shows up correctly.

cdagnino:
@ChadFulton Happy to contribute! I'd be happy to talk and see if there's something else I can add.

bashtage:
This notebook has broken the documentation build.  It is too slow and the documentation build is timing out.

bashtage:
Will  have to omit the PyMC3 cell. 14400 draws / 15-20 draws/second > 10 minutes for a single cell.  These PyMC3 notebooks are problems for the docbuild in general. We probably need to store them as rendered.

cdagnino:
@bashtage Do I need to change stuff on my end? The pull request says it's closed
(In any case, I can probably reduce the draws and write that they should do X number of draws to get something sensible)

bashtage:
I have disabled building it for now, so it will appear without being run.  We need to think about how we can accommodate more complex notebooks without boring down the overall build

Nothing for you to do, and thanks for submitting the excellent notebooks.

PR #6699, opened 2020-05-10
ENH: New exponential smoothing implementation

### LINKED ISSUE #6635: BUG: Inconsistencies in underlying state space model of holtwinters.ExponentialSmoothing
#### Short explanation

The statsmodels implementation of exponential smoothing in `holtwinters` implicitly uses the additive error state space models. Fits and calculations of information criteria are based on this and might be wrong when users assume the multiplicative error model.

#### Background

Up to recently, the statsmodels implementation of exponential smoothing in `holtwinters` did not use an explicit state space model. Fitting the model parameters and initial state was and is done by minimizing the sum of squared residuals, where `residuals = data - predicted`. This is currently implemented in the `SMOOTHERS`.

However, minimizing the sum of residuals with `residuals = data - predicted` really only makes sense when using an additive error model,
```
y_t = yhat_t + e_t,
```
 with `e_t = y_t - yhat_t ~ N(0, sigma**2)`. In this case, negative log-likelihood is
```
l = n * log(SSE/n),
```
so minimizing SSE is the same as minimizing negative log-likelihood.

When using a multiplicative error model, the state space model is 
```
y_t = yhat_t * (1 + e_t) = yhat_t + yhat_t * e_t.
```
In this case, `e_t = (y_t - yhat_t) / (yhat_t) ~ N(0, sigma**2)`. Therefore, the residuals should be `residuals = (data - prediction)/prediction`. In this case, negative log-likelihood is (see [Hyndman et al. 2002 (PDF)](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.468.7646&rep=rep1&type=pdf), equations (7) and (9)):
```
l = n * log(SSE/n) + 2 * sum(log(yhat))
```
with `SSE = sum((y - yhat) / yhat)`.

In the R package `forecast` (or the newer version, `fable`), the residuals are calculated as proposed above, that is, they are different for additive and multiplicative error models. Also, the default criterion used for optimizing the parameters is log likelihood.

#### How this affects the current implementation

With the introduction of state space simulations in #6560, it is now possible to choose between an additive and a multiplicative error model for simulations. However, fits are still performed implicitly using the additive model. Similar, information criteria are calculated using the likelihood of the additive model.

Users of `simulate` with `error="mul"` will therefore get imperfect fits and wrong information criteria for the model they assume to be using.

#### How this could be fixed

Fixing would require some changes behind the curtain, especially in the smoothers, but more importantly, it would require the user to specify the error model **before** the fit. An additional keyword argument in the `ExponentialSmoother` constructor with `"add"` as default value would be backwards compatible.

s-scherrer:
- [ ] closes #6635 
- [x] tests added / passed. 

This is work in progress on a new implementation of the ETS model family, as discussed in #6635. 

I implemented most of the essentials (fit, predict, simulate, smooth, loglike) and validated most of them vs results I got from the `forecast` package in R.
`smooth` and `loglike` are the same as in R up to 2 decimals, `simulate` up to 3, and `predict` up to 1 for the worst models, but often much better.

Currently missing/problems:

* ~The fit results are not as good as the ones using R. I am using bounded L-BFGS to minimize log-likelihood, with smoothing level, smoothing trend, and smoothing season between 0 and 1 (these correspond to alpha, beta*, gamma* in FPP2). For the initial values, I am using `_initialization_simple` in `statsmodels.tsa.exponential_smoothing.initialization`. I already tried to use the same initial states as in R, but this didn't help. Does someone (@ChadFulton) know a good solution to this?~
* ~I don't really get how the results wrappers work or what they are good for, so I didn't use this yet~
* There are probably still some things missing that a statsmodels model should provide
* `get_prediction` is still missing, could this be part of a follow-up PR?

I would appreciate some feedback and comments.


statsmodels/tsa/exponential_smoothing/ets.py
```diff
@@ -0,0 +1,1288 @@
...
+
+    @property
+    def _start_params(self):
+        lb = []
+        ub = []
+        for b in self.bounds:
+            lb.append(b[0])
+            ub.append(b[1])
+
+        # See https://github.com/robjhyndman/forecast/blob/master/R/ets.R
```
bashtage, line 461: Can you provide a different reference? forecast it GPL licensed and so code from it cannot be used or directly translated.

bashtage: If this appears in any written document that isn't code, it would be sufficient.

s-scherrer: I removed this reference and am now using the start params similar to the implementation in the state space module.
This was more of an experiment, I was unaware that GPL also forbids translations.

bashtage: It doesn't forbid them, but direct translations should result in code that is also GPL'd. The PyData econsystem uses BSD/MIT and not GPL.

s-scherrer:
I was investigating the fitting issues I have a bit more:

Here's a plot of the oil data, the fit using my implementation, the results of using my fit implementation but with initial states as in R, and the smoothing results using the parameters found in R.

R_vs_python_comparison_oil

The fits are all very close, but differ at the beginning. Is this something we can accept?

bashtage:
Yes. This is a good PR and would be great to get it in before it loses momentum. I think it is fine. One could point out in docs somewhere how an end-user could match R closer if they care.  We could also developed an alternative heuristic. 

As for initialization, you have known and estimated.  Can you also add the heuristic method that @ChadFulton has for additive models?

bashtage:
You might need to unpack `loglike_R` so that it is a float scalar, not an array, in test code.

bashtage:
This is what the logs show

```
 assert -149.06014828665928 <= (array([-149.89088944]) + 0.0001)
```

I'm not sure how sensitive you should be either given the starting value changes.

s-scherrer:
I thought that I could instead just compare the fitted values, without the first 20 values. This seems to work with an rtol of 0.05, except for the MMN and MMM models.

s-scherrer:
I implemented the heuristic initialization. I also found that 'known' was not working previously, that's why in the plot above the initial values were off, even when using the ones in R.

Now, there's very little difference between the heuristic and using the fit obtained with R.

R_vs_python_comparison_oil

So I assume that the heuristic might even work better as initial guess when using 'estimated'. Should I use the heuristic guesses as initial guess for 'estimated'?

bashtage:
Sounds like a good idea.

ChadFulton:
This is great, thanks very much @s-scherrer for putting in the work to create this PR. I'm sorry for the delay in reviewing it - I'll try to move as quickly as I can, since I agree with @bashtage that we should not waste time in getting it merged.

ChadFulton:
This is looking great, and I'm glad that we'll be able to improve our ETS models!

I am still going through it, but I have a couple of suggestions for now:

1. One of the reasons that the test precisions are low is that you're using versions of the datasets with only four decimal places. It should be easy to copy in a more precise version.
2. Rather than parameterizing the trend with beta^*, I would prefer to parameterize it with beta. This is what is used in the `ets` function from R's `forecast` package, which I think is in more use than the `fpp` package. Also, I used the beta parameterization in in the `statespace` linear exponential smoothing models.  Would this be a difficult change to make?
3. For compatibility with other tsa models, the `smooth` method should return a results object at the given parameters. I would rename your current method to be `_smooth`.

I'll have more comments later, but I wanted to give some feedback sooner rather than later. Thanks!

s-scherrer:
>     1. One of the reasons that the test precisions are low is that you're using versions of the datasets with only four decimal places. It should be easy to copy in a more precise version.

Good point, I'll try to do this.

>     2. Rather than parameterizing the trend with beta^*, I would prefer to parameterize it with beta. This is what is used in the `ets` function from R's `forecast` package, which I think is in more use than the `fpp` package. Also, I used the beta parameterization in in the `statespace` linear exponential smoothing models.  Would this be a difficult change to make?

I think it should be possible without too much work. My main motivation behind this was that I can then use fixed bounds instead of  nonlinear constraints for the optimization, I'm not sure whether this is possible with statsmodels optimizers, but I think it should work with scipy.minimize.

I think I should find some time this weekend to add these changes.

bashtage:
Could you estimate with transformed parameters beta** but then report with
beta?

s-scherrer:
> Could you estimate with transformed parameters beta** but then report with beta?

In principle yes, but when allowing user defined bounds or fixed parameter values, I would again need nonlinear constraints, so I believe its easier to just change to beta and gamma with nonlinear constraints.

statsmodels/tsa/exponential_smoothing/_ets_smooth.pyx
```diff
@@ -0,0 +1,153 @@
+#!python
+#cython: wraparound=False, boundscheck=False, cdivision=True, annotate=True
+
+import numpy as np
+cimport numpy as np
+
+np.import_array()
+
+cpdef _initialize_ets_smooth(np.float_t [:] params, Py_ssize_t n):
```
bashtage, line 9: We should use a fused type that supports both float32 and float64 here?  This would produce "free" version that let users use single or double precision data.

bashtage: https://cython.readthedocs.io/en/latest/src/userguide/fusedtypes.html

statsmodels/tsa/exponential_smoothing/_ets_smooth.pyx
```diff
@@ -0,0 +1,153 @@
...
+    cdef Py_ssize_t m
+
+    # get params
+    alpha, beta_star, gamma_star, phi = params[0:4]
+    m = len(params[6:])
+
+    # initialize states
+    # l = xhat[:,0], b = xhat[:,1], s = xhat[:,2:2+m]
+    # seasons are sorted such that xhat[:,2+m-1] contains s[-m]
+    xhat = np.zeros((n, 2 + m))
```
bashtage, line 23: Are these used in estimation?  If so, it might be worth considering passing chat in rather than reallocating it each iteration.

bashtage:
Close/reopen to trigger CI.

bashtage:
Quite a few lint issues.

```
tatsmodels/tsa/exponential_smoothing/ets.py:307:80: E501 line too long (80 > 79 characters)
statsmodels/tsa/exponential_smoothing/ets.py:324:5: E303 too many blank lines (2)
statsmodels/tsa/exponential_smoothing/ets.py:343:80: E501 line too long (81 > 79 characters)
statsmodels/tsa/exponential_smoothing/ets.py:344:80: E501 line too long (80 > 79 characters)
statsmodels/tsa/exponential_smoothing/ets.py:347:80: E501 line too long (80 > 79 characters)
statsmodels/tsa/exponential_smoothing/ets.py:349:80: E501 line too long (80 > 79 characters)
statsmodels/tsa/exponential_smoothing/ets.py:351:80: E501 line too long (80 > 79 characters)
statsmodels/tsa/exponential_smoothing/ets.py:564:80: E501 line too long (80 > 79 characters)
statsmodels/tsa/exponential_smoothing/ets.py:589:80: E501 line too long (81 > 79 characters)
statsmodels/tsa/exponential_smoothing/ets.py:737:80: E501 line too long (80 > 79 characters)
statsmodels/tsa/exponential_smoothing/ets.py:745:15: E222 multiple spaces after operator
statsmodels/tsa/exponential_smoothing/ets.py:872:80: E501 line too long (83 > 79 characters)
statsmodels/tsa/exponential_smoothing/ets.py:895:80: E501 line too long (80 > 79 characters)
statsmodels/tsa/exponential_smoothing/ets.py:1093:80: E501 line too long (80 > 79 characters)
statsmodels/tsa/exponential_smoothing/ets.py:1156:9: E303 too many blank lines (2)
statsmodels/tsa/exponential_smoothing/ets.py:1192:18: E225 missing whitespace around operator
statsmodels/tsa/exponential_smoothing/ets.py:1194:18: E225 missing whitespace around operator
statsmodels/tsa/exponential_smoothing/ets.py:1301:12: E261 at least two spaces before inline comment
statsmodels/tsa/exponential_smoothing/base.py:11:1: F401 'statsmodels.base.wrapper as wrap' imported but unused
statsmodels/tsa/exponential_smoothing/base.py:151:47: F821 undefined name 'StateSpaceMLEResultsWrapper'
Previously passing files failed linting.
Linting newly added files with strict rules
New files: statsmodels/tsa/exponential_smoothing/base.py statsmodels/tsa/exponential_smoothing/ets.py statsmodels/tsa/tests/test_exponential_smoothing.py
statsmodels/tsa/exponential_smoothing/base.py:11:1: F401 'statsmodels.base.wrapper as wrap' imported but unused
statsmodels/tsa/exponential_smoothing/base.py:151:47: F821 undefined name 'StateSpaceMLEResultsWrapper'
statsmodels/tsa/exponential_smoothing/ets.py:307:80: E501 line too long (80 > 79 characters)
statsmodels/tsa/exponential_smoothing/ets.py:324:5: E303 too many blank lines (2)
statsmodels/tsa/exponential_smoothing/ets.py:343:80: E501 line too long (81 > 79 characters)
statsmodels/tsa/exponential_smoothing/ets.py:344:80: E501 line too long (80 > 79 characters)
statsmodels/tsa/exponential_smoothing/ets.py:347:80: E501 line too long (80 > 79 characters)
statsmodels/tsa/exponential_smoothing/ets.py:349:80: E501 line too long (80 > 79 characters)
statsmodels/tsa/exponential_smoothing/ets.py:351:80: E501 line too long (80 > 79 characters)
statsmodels/tsa/exponential_smoothing/ets.py:564:80: E501 line too long (80 > 79 characters)
statsmodels/tsa/exponential_smoothing/ets.py:589:80: E501 line too long (81 > 79 characters)
statsmodels/tsa/exponential_smoothing/ets.py:737:80: E501 line too long (80 > 79 characters)
statsmodels/tsa/exponential_smoothing/ets.py:745:15: E222 multiple spaces after operator
statsmodels/tsa/exponential_smoothing/ets.py:872:80: E501 line too long (83 > 79 characters)
statsmodels/tsa/exponential_smoothing/ets.py:895:80: E501 line too long (80 > 79 characters)
statsmodels/tsa/exponential_smoothing/ets.py:1093:80: E501 line too long (80 > 79 characters)
statsmodels/tsa/exponential_smoothing/ets.py:1156:9: E303 too many blank lines (2)
statsmodels/tsa/exponential_smoothing/ets.py:1192:18: E225 missing whitespace around operator
statsmodels/tsa/exponential_smoothing/ets.py:1194:18: E225 missing whitespace around operator
statsmodels/tsa/exponential_smoothing/ets.py:1301:12: E261 at least two spaces before inline comment
statsmodels/tsa/tests/test_exponential_smoothing.py:36:1: E303 too many blank lines (3)
statsmodels/tsa/tests/test_exponential_smoothing.py:44:1: E302 expected 2 blank lines, found 3
statsmodels/tsa/tests/test_exponential_smoothing.py:50:1: E305 expected 2 blank lines after class or function definition, found 1
statsmodels/tsa/tests/test_exponential_smoothing.py:56:35: E128 continuation line under-indented for visual indent
statsmodels/tsa/tests/test_exponential_smoothing.py:58:35: E128 continuation line under-indented for visual indent
statsmodels/tsa/tests/test_exponential_smoothing.py:64:1: E302 expected 2 blank lines, found 1
statsmodels/tsa/tests/test_exponential_smoothing.py:66:31: E222 multiple spaces after operator
statsmodels/tsa/tests/test_exponential_smoothing.py:87:80: E501 line too long (81 > 79 characters)
statsmodels/tsa/tests/test_exponential_smoothing.py:121:41: E231 missing whitespace after ','
statsmodels/tsa/tests/test_exponential_smoothing.py:125:1: E302 expected 2 blank lines, found 1
statsmodels/tsa/tests/test_exponential_smoothing.py:150:1: E302 expected 2 blank lines, found 1
statsmodels/tsa/tests/test_exponential_smoothing.py:252:1: E302 expected 2 blank lines, found 1
statsmodels/tsa/tests/test_exponential_smoothing.py:324:1: E302 expected 2 blank lines, found 1
statsmodels/tsa/tests/test_exponential_smoothing.py:388:68: E231 missing whitespace after ','
```

These days I just do 

```
black -l 79 statsmodels/tsa/exponential_smoothing/ets.py statsmodels/tsa/tests/test_exponential_smoothing.py

sort statsmodels/tsa/exponential_smoothing/ets.py statsmodels/tsa/tests/test_exponential_smoothing.py
```

There are a couple in base that need manual fixes.

s-scherrer:
> Quite a few lint issues.

Should be fixed in the next commit. Just out of curiosity: Why do you use `sort` after `black`?

bashtage:
Should have been `isort`.  `isort` aranges imports by categories: Compat first, then stdlib, then known 3rd party, then unknown 3rd, then 1st party.

bashtage:
For example, https://github.com/statsmodels/statsmodels/pull/6699/files#diff-6640e5c34a0dcf373519f96f61e79dc0R1 is not quite right (at least according to these rules) since it has a blank line.

statsmodels/tsa/exponential_smoothing/_ets_smooth.pyx
s-scherrer: Done

statsmodels/tsa/exponential_smoothing/_ets_smooth.pyx
s-scherrer: Changed this. I added a `_loglike_internal` function that takes `yhat` and `xhat` as arguments, which is used during fitting, but Codacy complains about the replacement of `loglike` with `_loglike_internal`. Is there a better way to do this? I want to avoid duplicating the code from `statsmodels.base.model.fit`.

s-scherrer:
I will fix the linting issues with the next commit. Is it possible to run `flake8` with exactly the same settings as it is run in the CI? When I run `flake8 --isolated` or `lint.sh` on my machine, the errors don't show up.

bashtage:
You should be able to run `LINT=true ./lint.sh` from the root on any *nix system.  You may need to run

```
git checkout master
git rebase upstream/master
git push
git checkout <feature branch>
```

to make sure your origin/master matches upstream/master.

bashtage:
`tatsmodels/tsa/tests/test_exponential_smoothing.py:283:25: E226 missing whitespace around arithmetic operator` this is my fault.  black puts that space in but we don't use it, so you have to manually restore these back.  Sorry.

s-scherrer:
> You should be able to run LINT=true ./lint.sh from the root on any *nix system.

My version of `lint.sh` wasn't up to date, thanks for the help.

> this is my fault. black puts that space in but we don't use it, so you have to manually restore these back

Actually my fault, I didn't black `test_exponential_smoothing.py`. Did it now, everything is fine.

s-scherrer:
Sorry for that merge fuck-up, is there a way to correct this?

I tried using nonlinear constraints, but this didn't work as well as I hoped, so I went back to normal interval bounds.
The bounds are now given as a dictionary of parameter names and intervals (this is a bit different to how it's done in the statespace module).
If `"smoothing_level"` or `"smoothing_seasonal"` are in the bounds dictionary (or are fixed parameters), beta is used internally during fitting, otherwise beta-star is used internally during fitting.

bashtage:
Do you have any big issues remaining? Or can we hopefully do the final review this coming week?

s-scherrer:
> Do you have any big issues remaining? Or can we hopefully do the final review this coming week?

I think the major points of the general implementation of the math and basic functionality should now be addressed. What I still not fully understood is the wrapping mechanism, I wanted to look into this in the coming days.

Some methods that the statespace models implement are still missing (better covariance/hessian/score, some statistical tests, prediction results) - I think this could be part of follow up PRs?

Also, testing currently focusses mostly on comparisons of the implementation with R, I didn't check coverage yet and would probably add some more tests.

s-scherrer:
About that failing test in CI:
```
 >       assert loglike >= loglike_R - 1e-4
E       assert -150.19433085754278 >= (-148.77635253193012 - 0.0001)
...
FAILED statsmodels/tsa/tests/test_exponential_smoothing.py::test_fit_vs_R[setup_model23]
```

I can't reproduce this on my machine, the fit works here.What could be the reason for this?

ChadFulton:
> About that failing test in CI:
> I can't reproduce this on my machine, the fit works here.What could be the reason for this?

Sometimes there can be issues with different systems finding slightly different numerical solutions. In that case, though, usually it's only one or two.  It looks like most of the systems are running into a problem here.

One solution might be to modify the test to use better starting parameters (i.e. hand-pick the starting parameters to be close to the optimum).

statsmodels/tsa/exponential_smoothing/ets.py
```diff
@@ -0,0 +1,1679 @@
...
+        # Wrap data / squeeze where appropriate
+        if repetitions > 1:
+            names = ["simulation.%d" % num for num in range(repetitions)]
+        else:
+            names = "simulation"
+        return self.model._wrap_data(
+            y, start_idx, start_idx + nsimulations - 1, names=names
+        )
+
+    def forecast(self, steps=1, anchor=None):
```
ChadFulton, line 1510: I think it would be better to follow the typical Statsmodels API in `tsa` for `predict` and `forecast`, which would be:

- `forecast` only performs out-of-sample forecasting that starts immediately after the sample ends.
- `predict` is the more general function that allows selecting both a `start` / `end`

So I think we should remove the `anchor` argument here (I know that this means a little refactoring, sorry about that!)

I'm not necessarily opposed to an `anchor` argument for `forecast` in general, but I think it would be better for us to consider whether we want to add it across the entire API rather than just here.

s-scherrer: Ah, yes, this was a point where I wasn't sure whether I understood it correctly. I now renamed my original `forecast` to `_forecast`, and am calling this function from the new `forecast`.

ChadFulton:
Thanks for the updates! I’ll take another look over the next couple of days for anything that sticks out as must be changed before merging, but I would think that we can aim to merge this PR this week, unless there’s more that you want to do first?

ChadFulton:
This looks great to me, and I would like to merge it.  We can follow up with PRs with whatever small things that we find later.

@s-scherrer:

1. Do you agree that it's ready to merge, or is there more you'd like to do?
2. If there are any things that you'd still like to do but which aren't crucial, could you make new "ENH" issues for them?

Thanks again!

s-scherrer:
@ChadFulton 

In my opinion it's ready to merge. I will create some issues for enhancements in the next days.

Thanks a lot to both you and @bashtage for your help and comments, I learned a lot during this PR.

ChadFulton:
Thanks @s-scherrer!

GoldbergData:
Hi all,

I am comparing `statsmodels=0.11.1` with `statsmodels>0.11.1`. I get difference results (predictions) with `0.11.1` versus newer versions. Would the bug fix included here be the primary driver for different results of `from statsmodels.tsa.holtwinters import ExponentialSmoothing`? I don't have a reproducible example to share right now, but can provide one if necessary.

s-scherrer:
@GoldbergData are you using the `initialization_method` argument? According to the docs:

```
None defaults to the pre-0.12 behavior where initial values are passed as part of fit. If any of the other values are passed, then the initial values must also be set when constructing the model. If ‘known’ initialization is used, then initial_level must be passed, as well as initial_trend and initial_seasonal if applicable. Default is ‘estimated’. “legacy-heuristic” uses the same values that were used in statsmodels 0.11 and earlier.
```
https://www.statsmodels.org/stable/generated/statsmodels.tsa.holtwinters.ExponentialSmoothing.html#statsmodels.tsa.holtwinters.ExponentialSmoothing

bashtage:
There have been a lot of changes in `holtwinters` in the 0.12 round, and some further ones in the 0.13 round.  I'm not surprised that this changed things.  Specifically, estimation would silently fail to converge in earlier implementations. 

Are the new estimates "worse" than the old ones, in the sense that the SSE is larger using the new ones?

PR #6742, opened 2020-05-21
Update broken link

adrienpacifico:
https://statsmodels.github.io/dev/test_notes.html is broken.
Replace it with:
https://www.statsmodels.org/stable/dev/test_notes


bashtage:
Thanks

PR #6765, opened 2020-05-31
ENH: Decomposition of forecast updates in state space models due to the "news"

ChadFulton:
- [x] tests added / passed. 

This PR adds the ability to decompose updates to forecasts based on the "news" from updated datasets, in state space models. This is often used used in the context of "nowcasting", e.g. the following:

- "Nowcasting." Bańbura, Giannone, and Reichlin. (2011)
- "Now-casting and the real-time data flow." Bańbura, Giannone, Modugno, and Reichlin. (2013)
- "Maximum likelihood estimation of factor models on datasets with arbitrary pattern of missing data." Bańbura and Modugno. (2014)

Included is an example notebook, and a rendered version of it can be found at https://nbviewer.jupyter.org/gist/ChadFulton/a263ed5dac30ac3616f785dd80205026

This can be used with any state space model, but it will be particularly useful once I finish coding up their mixed-frequency dynamic factor model, which is used by the NY FRB, see https://www.newyorkfed.org/research/policy/nowcast.

TODO:

- [x] add notebook to `manifest.json`, create image.


bashtage:
@ChadFulton Missing an image for the notebook. just above 
https://www.statsmodels.org/devel/examples/index.html#state-space-models-technical-notes

PR #6774, opened 2020-06-03
Add Root Mean Square Percentage Error as rmspe()

### LINKED ISSUE #6773: Add Root Mean Square Percentage Error to evaluation measures
There are many cases in which a Percentage Error is a more appropriate measure of model fit than an Absolute Error.

I have added a simple Root Mean Square Percentage Error function, for which I am ready to make a pull request.

#### Additional context
The only complexity is handling zero values. My function includes a parameter in which the user is able to specify whether they would like the zero-value case to return `np.nan` (default) or some other value.

peterprescott:
- [ x] closes #6773
- [ ] tests added / passed. 
- [ x] code/documentation is well formatted.  
- [ ]x properly formatted commit message. See 

### DIFF
--- a/statsmodels/tools/eval_measures.py
+++ b/statsmodels/tools/eval_measures.py
@@ -71,7 +71,7 @@ def rmse(x1, x2, axis=0):
 def rmspe(y, y_hat, axis=0, zeros=np.nan):
     """
     root mean squared percentange error
-    
+
     Parameters
     ----------
     y, y_hat : array_like
--- a/statsmodels/tools/eval_measures.py
+++ b/statsmodels/tools/eval_measures.py
@@ -90,8 +90,8 @@ def rmspe(y, y_hat, axis=0, zeros=np.nan):
     y_hat = np.asanyarray(y_hat)
     y = np.asanyarray(y)
     error = y - y_hat
-    percentage_error = np.divide(error, y, out=np.full_like(error,zeros), where=y!=0)
-    return np.nanmean((percentage_error)**2, axis=axis) * 100
+    percentage_error = np.divide(error, y, out=np.full_like(error, zeros), where=y!=0)
+    return np.nanmean(percentage_error**2, axis=axis) * 100
 
 def maxabs(x1, x2, axis=0):
     """maximum absolute error
--- a/statsmodels/tools/eval_measures.py
+++ b/statsmodels/tools/eval_measures.py
@@ -2,8 +2,9 @@
 """some measures for evaluation of prediction, tests and model selection
 
 Created on Tue Nov 08 15:23:20 2011
+Updated on Wed Jun 03 10:42:20 2020
 
-Author: Josef Perktold
+Authors: Josef Perktold & Peter Prescott
 License: BSD-3
 
 """
@@ -67,6 +68,30 @@ def rmse(x1, x2, axis=0):
     x2 = np.asanyarray(x2)
     return np.sqrt(mse(x1, x2, axis=axis))
 
+def rmspe(y, y_hat, axis=0, zeros=np.nan):
+    """
+    root mean squared percentange error
+    
+    Parameters
+    ----------
+    y, y_hat : array_like
+       The performance measure depends on the difference between these two
+       arrays.
+    axis : int
+       axis along which the summary statistic is calculated
+    zeros : float
+       value to assign to error where actual value is zero
+
+    Returns
+    -------
+    rmse : ndarray or float
+       root mean squared error along given axis.
+    """
+    y_hat = np.asanyarray(y_hat)
+    y = np.asanyarray(y)
+    error = y - y_hat
+    percentage_error = np.divide(error, y, out=np.full_like(error,zeros), where=y!=0)
+    return np.nanmean((percentage_error)**2, axis=axis) * 100
 
 def maxabs(x1, x2, axis=0):
     """maximum absolute error
--- a/statsmodels/tools/eval_measures.py
+++ b/statsmodels/tools/eval_measures.py
@@ -70,25 +70,26 @@ def rmse(x1, x2, axis=0):
 
 def rmspe(y, y_hat, axis=0, zeros=np.nan):
     """
-    root mean squared percentange error
+    Root Mean Squared Percentage Error
 
     Parameters
     ----------
-    y, y_hat : array_like
-       The performance measure depends on the difference between these two
-       arrays.
+    y : array_like
+      The actual value.
+    y_hat : array_like
+       The predicted value.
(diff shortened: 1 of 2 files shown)


statsmodels/tools/eval_measures.py
```diff
@@ -67,6 +68,30 @@ def rmse(x1, x2, axis=0):
     x2 = np.asanyarray(x2)
     return np.sqrt(mse(x1, x2, axis=axis))
 
+def rmspe(y, y_hat, axis=0, zeros=np.nan):
+    """
+    root mean squared percentange error
```
bashtage, line 73: Capital letter to start please.

statsmodels/tools/eval_measures.py
```diff
@@ -67,6 +68,30 @@ def rmse(x1, x2, axis=0):
...
+    
+    Parameters
+    ----------
+    y, y_hat : array_like
```
bashtage, line 77: One parameter per entry, so you will need two lines here. Multiple entries do not render correctly in sphinx.

statsmodels/tools/eval_measures.py
```diff
@@ -67,6 +68,30 @@ def rmse(x1, x2, axis=0):
...
+       The performance measure depends on the difference between these two
+       arrays.
+    axis : int
+       axis along which the summary statistic is calculated
```
bashtage, line 81: Capitals here and below as well to start descriptions.

statsmodels/tools/eval_measures.py
```diff
@@ -67,6 +68,30 @@ def rmse(x1, x2, axis=0):
...
+    zeros : float
+       value to assign to error where actual value is zero
+
+    Returns
+    -------
+    rmse : ndarray or float
+       root mean squared error along given axis.
+    """
+    y_hat = np.asanyarray(y_hat)
```
bashtage, line 90: Either asarray (no subclasses please) or use array_like from statsmodels.tools.validation.

statsmodels/tools/eval_measures.py
```diff
@@ -67,6 +68,30 @@ def rmse(x1, x2, axis=0):
...
+    y = np.asanyarray(y)
+    error = y - y_hat
+    percentage_error = np.divide(error, y, out=np.full_like(error,zeros), where=y!=0)
```
bashtage, line 93: Space after ,

statsmodels/tools/eval_measures.py
```diff
@@ -67,6 +68,30 @@ def rmse(x1, x2, axis=0):
...
+    return np.nanmean((percentage_error)**2, axis=axis) * 100
```
bashtage, line 94: No need for ()

statsmodels/tools/eval_measures.py
peterprescott: I would,but the original script doesn't, so I believe [the proper stylistic guideline](https://google.github.io/styleguide/pyguide.html#4-parting-words) is that Local Style should trump Global Style, so as not to cause unnecessary distraction

statsmodels/tools/eval_measures.py
peterprescott: Again, here I'm just following the style of the rest of the script.

statsmodels/tools/eval_measures.py
peterprescott: And again, hre I'm just following the style of the rest of the script.

statsmodels/tools/eval_measures.py
bashtage: It doesn't.  Those old ones are just strings in need of change that haven't been fixed yet.  Feel free to change those as well.

statsmodels/tools/eval_measures.py
peterprescott: And again, I'm just aligning with the rest of the script.

There may be some obsolete conventions used in the script, but my intention here is simply to add a new feature.

statsmodels/tools/eval_measures.py
bashtage: New content needs to conform to current standards. There is a lot of old code that was developed with fewer standards in place and fewer tools to verify them.

statsmodels/tools/eval_measures.py
peterprescott: Thanks!
```suggestion
    percentage_error = np.divide(error, y, out=np.full_like(error, zeros), where=y!=0)
```

statsmodels/tools/eval_measures.py
bashtage: The old code may be buggy in some edge cases.  We cannot accept new code that maintains the same bad behavior. 

I'm not sure why you are putting your foot down for small changes. If you make these then this can be easily merged.

statsmodels/tools/eval_measures.py
peterprescott:
```suggestion
    return np.nanmean(percentage_error**2, axis=axis) * 100
```

statsmodels/tools/eval_measures.py
peterprescott: Haha, sorry. I can change it :)

statsmodels/tools/eval_measures.py
peterprescott: Done now

statsmodels/tools/eval_measures.py
peterprescott: Done now

statsmodels/tools/eval_measures.py
peterprescott: Done now

statsmodels/tools/eval_measures.py
peterprescott: Done now

statsmodels/tools/eval_measures.py
peterprescott: Done now

statsmodels/tools/eval_measures.py
peterprescott: Done now

bashtage:
Any chance you could add a test?  It could be very basic but will show that the code works for some reasonable data, and protect against any changes in upstream libraries like NumPy that would break it. 

It would be great to have one that tested the zeros option and one that did not.

peterprescott:
Hopefully that test does the trick -- I just adapted the one given for Root Mean Square Error.

bashtage:
Looks like you mised an import of  rmspe

bashtage:
@peterprescott Any idea why this is failing?  Seems to be quite far off.

PR #6780, opened 2020-06-05
ENH: Add pandas dataframe capability to variance_inflation_factor

o-P-o:
Proposing a small change to the variance_inflation_factor() method in the outliers_influence package, in order to allow `exog` input to be a pandas DataFrame as well as a numpy array.

In the test, the value computed for the VIF using my proposed code edit with a pandas dataframe input is 16.4394, which I compare to the value computed using the current state of the method, taking an array as input.

### DIFF
--- a/statsmodels/stats/outliers_influence.py
+++ b/statsmodels/stats/outliers_influence.py
@@ -9,6 +9,7 @@ License: BSD-3
 from collections import defaultdict
 
 import numpy as np
+import pandas as pd
 
 from statsmodels.compat.python import lzip
 from statsmodels.compat.pandas import Appender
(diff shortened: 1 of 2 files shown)


statsmodels/stats/outliers_influence.py
```diff
@@ -186,7 +187,11 @@ def variance_inflation_factor(exog, exog_idx):
     https://en.wikipedia.org/wiki/Variance_inflation_factor
     """
     k_vars = exog.shape[1]
-    x_i = exog[:, exog_idx]
+    x_i = None
+    if isinstance(exog, pd.DataFrame):
```
jseabold, line 191: Instead of this check, could you add `np.asarray` here as you've done in the test? That way, a DataFrame and an array will pass through.

statsmodels/stats/tests/test_outliers_influence.py
```diff
@@ -14,3 +17,8 @@ def test_reset_stata():
     stat = reset_ramsey(res, degree=4)
     assert_almost_equal(stat.fvalue[0, 0], 1.52, decimal=2)
     assert_almost_equal(stat.pvalue, 0.2221, decimal=4)
+
+    exog_idx = list(data.columns).index('urban')
+    X_arr = np.asarray(data)
+    vif = variance_inflation_factor(X_arr, exog_idx)
```
jseabold, line 23: You'll want `X_arr` here to be a DataFrame instead of an array to make sure your code changes are tested.

jseabold (changes requested):
@o-P-o Requested a few minor changes.

bashtage:
Thanks.

PR #6785, opened 2020-06-07
ENH: automatic lag selection for Box-Pierce, Ljung-Box #6645

w31ha0:
- [x] closes #xxxx
- [x] tests added / passed. 

With reference to issue [#6645](https://github.com/statsmodels/statsmodels/issues/6645)

Automatic lag selection according to the reference below for Box-Pierce, Ljung-Box. According to the reference at page8, it introduces a penalty term, which switches between AIC and BIC depending on the value of correlation. This penalty term penalises the test statistic and the optimal lag length is one which is able to maximise the test statistic after applying the penalty.

A q value of 2.4 is recommended in the reference for the penalty term according to other research results quoted.

[Reference](https://www.sciencedirect.com/science/article/abs/pii/S0304407609000773?via%3Dihub#!)

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>

### DIFF
--- a/statsmodels/stats/diagnostic.py
+++ b/statsmodels/stats/diagnostic.py
@@ -575,7 +575,7 @@ def acorr_ljungbox_automatic(x, boxpierce=False, model_df=0, return_df=None):
     nobs = x.shape[0]
     maxlag = nobs - 1
 
-    #Compute threshold metrics
+    # Compute threshold metrics
     sacf = acf(x, nlags=maxlag, fft=False)
     sacf2 = sacf[1:maxlag + 1] ** 2 / (nobs - np.arange(1, maxlag + 1))
     q = 2.4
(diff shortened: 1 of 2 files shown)


bashtage:
Looks good overall. Doc string needs to follow the standard format with Parameters and Returns.

bashtage:
Should this just be an option for acorr_lungbox? Seems very similar since it just adds a data driven method to select a lag for the test.

w31ha0:
> Should this just be an option for acorr_lungbox? Seems very similar since it just adds a data driven method to select a lag for the test.

Makes sense. I have removed the new method and added a new bool flag (auto_lag) into the existing method.

w31ha0:
@bashtage When does this usually get merged?

bashtage:
@w31ha0 Thanks for this contirbution!

PR #6877, opened 2020-07-10
BUG: fix raise exception on granger causality test

### LINKED ISSUE #6876: Error catching in grangercausalitytest
When passing a zero lag to `grangercausalitytests` in a list, such as in:

```python
granger = grangercausalitytests(df, maxlag=[0, 5, 10])
```

The expected behavior would be raising an exception (line 1197 of stattools.py):

```python
try:
        lags = np.array([int(lag) for lag in maxlag])
        maxlag = lags.max()
        if lags.min() <= 0 or lags.size == 0:
            raise ValueError('maxlag must be a non-empty list containing only '
                             'positive integers')
    except Exception:
        maxlag = int_like(maxlag, 'maxlag')
        if maxlag <= 0:
            raise ValueError('maxlag must a a positive integer')
        lags = np.arange(1, maxlag + 1)
```

However, this exception seems to be catched by the `except Exception` and pass `maxlag` (which now has the value of `lags.max()`) to `int_like`, as a result, the calculation is performed for all the lags from 1 to `maxlag.max()`. 

I think this could be fixed just by changing the `except Except` by `except TypeError`.

I can try and fix this myself, if that is ok.

thank you,

ppgaluzio:
Issue #6876: Fix for the appropriate behavior when passing a list
containing a zero lag to `maxlag` (e.g. `[0, 1, 2]`), now it raises
`ValueError`

- [x] closes #6876
- [ ] tests added / passed.

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>

### DIFF
--- a/statsmodels/tsa/stattools.py
+++ b/statsmodels/tsa/stattools.py
@@ -1336,18 +1336,17 @@ def grangercausalitytests(x, maxlag, addconst=True, verbose=True):
     addconst = bool_like(addconst, "addconst")
     verbose = bool_like(verbose, "verbose")
     try:
+        maxlag = int_like(maxlag, "maxlag")
+        if maxlag <= 0:
+            raise ValueError("maxlag must a a positive integer")
+        lags = np.arange(1, maxlag + 1)
+    except TypeError:
         lags = np.array([int(lag) for lag in maxlag])
         maxlag = lags.max()
         if lags.min() <= 0 or lags.size == 0:
             raise ValueError(
                 "maxlag must be a non-empty list containing only "
-                "positive integers"
-            )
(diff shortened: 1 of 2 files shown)


bashtage:
Code itself looks good.  Could you write a test?

bashtage:
Basically show that the ValueError is raised.

ppgaluzio:
Sure, I already implemented the test (it passed on my machine and the test on travis-ci also). Do I need to do something else here on github (update the PR or something like that? This is my first PR so I'm not really sure)

bashtage:
Great! Thanks.

PR #6882, opened 2020-07-12
ENH: get_prediction method for ETS

### LINKED ISSUE #6359: ENH: Add Prediction Intervals to Holt-Winters class

s-scherrer:
Added a method that returns a PredictionResults object, which contains
mean prediction and prediction intervals calculated from multiple
simulations.

This commit also contains some minor bug fixes regarding the anchor
value of simulations.

- [x] closes #6359
- [x] tests added / passed. 

I also would like to add and example of this (and other things) to the ETS example notebook in the following days.


ChadFulton:
This will be great, thanks for following up with this! Two initial comments:

1. How about adding a `method=None` argument to `get_prediction`, with options `'exact'` and `'simulated'`. The default value of `None` would then select `'exact'` when that is available (which is never right now, but we can add this later for some of the Gaussian cases) and will select `'simulated'` otherwise.

2. Have you thought about how to test this? One option I've been using recently is writing Monte Carlo tests with large sample sizes and seeds, but then marking them with skip test so that they don't have to be run on the CI each time we make a PR.  This way, we at least have a test available that checks the results.

statsmodels/tsa/exponential_smoothing/ets.py
```diff
@@ -1748,11 +1752,54 @@ def predict(self, start=None, end=None, dynamic=False, index=None):
...
+        y[nsmooth:] = self._forecast(nforecast, anchor_forecast)
 
         # Wrap data / squeeze where appropriate
         return self.model._wrap_data(y, start, end_forecast - 1)
 
+    def get_prediction(self, start=None, end=None, dynamic=False,
+                       index=None, simulate_repetitions=1000,
+                       **simulate_kwargs):
+        """
+        Returns a
```
bashtage, line 1764: You should not list the return type here.  Describe in a short sentence what this does.

statsmodels/tsa/exponential_smoothing/ets.py
```diff
@@ -1748,11 +1752,54 @@ def predict(self, start=None, end=None, dynamic=False, index=None):
...
+        index : pd.Index, optional
+            Optionally an index to associate the predicted results to. If None,
+            an attempt is made to create an index for the predicted results
+            from the model's index or model's row labels.
+        simulate_repetitions : int, optional
+            Number of simulation repetitions for calculating prediction
+            intervals.  Default is 1000.
+        **simulate_kwargs :
+            Additional arguments passed to the ``simulate`` method.
+        """
```
bashtage, line 1798: Return type shoudl go here as in

```
Returns
-------
PredictionResults
    Predicted mean values and prediction intervals
```

s-scherrer:
I added the `method` keyword as requested and implemented the formulas from FPP for the additive models.

There are still some issues I had:

For some models the exact formula and the simulations disagree strongly, e.g. when fitting an AAdA model to the austourists dataset. Below is a plot of data, mean, and prediction intervals using the exact and the simulation method.
Figure_austourists
In this case the prediction intervals using the exact method diverge much more quickly. I believe the reason for this is that the fit finds a damping rate (phi) of 0.98 (the upper bound), and the formula diverges for phi to 1. This itself doesn't make a lot of sense to me, since for phi to 1 the model is the same as the AAA model. With the AAA model both methods have very similar results (sorry for the missing legend, but there's just too many lines):

Figure_AAA

I also found some differences between the statespace exponential smoothing implementation and the ets implementation. With the same parameters, log-likelihood and scale are the same, but the information criteria are different. Is this due to the burn-in of the Kalman filter?
There are also some differences in the predicted values when using "dynamic" prediction. I understood dynamic prediction to be multi-step prediction, right? The next figure shows the difference between a (in-sample) prediction of both implementations. After some time, dynamic prediction starts - this is where they start to deviate.

Figure_ets_vs_statespace

Any ideas on these issues?

ChadFulton:
Great that you were able to add the analytic methods!  I'll take a closer look at this PR tonight, but obviously it's looking good.

ChadFulton:
Here are a couple of initial responses:

> I also found some differences between the statespace exponential smoothing implementation and the ets implementation. With the same parameters, log-likelihood and scale are the same, but the information criteria are different. Is this due to the burn-in of the Kalman filter?

This is because the variance of the error term is considered a parameter in state space models, while Hyndman doesn't consider it as a parameter for AIC computation in ETS models (see e.g. his "Forecasting with Exponential Smoothing" book, p. 194). So the `df_model` attribute is one larger in the state space models. I guess this is maybe because Hyndman is trying to maintain consistency with the basic exponential smoothing models that were heuristics and so didn't consider the distribution of an error term. 

For purposes of model comparison, it doesn't matter so much as long as we're consistent across the models within each class. If the variance is included in all of the ETS models, then it's just essentially an extra constant term in the AIC.

The error variable parameter is still estimated by MLE, it's just that it's estimate can be computed analytically and then concentrated out of the log-likelihood function. I notice that `statespace` is computing the scale (sigma^2) differently too. I have this noted in `statespace/tests/test_exponential_smoothing.py`, that R's `forecast` package is using a degree-of-freedom correction. I'll think a bit about both of these issues.

> There are also some differences in the predicted values when using "dynamic" prediction. I understood dynamic prediction to be multi-step prediction, right? The next figure shows the difference between a (in-sample) prediction of both implementations. After some time, dynamic prediction starts - this is where they start to deviate.

I think the difference here is just that in `statespace`, `dynamic` is an offset to `start`, while it looks like you currently have it as an absolute location.

s-scherrer:
> I think the difference here is just that in statespace, dynamic is an offset to start, while it looks like you currently have it as an absolute location.

Thanks, I somehow misread the documentation. Fixed now.

bashtage:
@s-scherrer What is the status of this PR.  Would really like this in for 0.12.

s-scherrer:
> @s-scherrer What is the status of this PR. Would really like this in for 0.12.

Sorry, I was a bit busy in the last weeks. I made some small fixes to the notebook and had to merge with master, if the checks are passed I would say it's ready.

bashtage:
I'm going to do a merge than a fixup.

PR #6884, opened 2020-07-13
TST: ets: test for simple exponential smoothing convergence

### LINKED ISSUE #6883: BUG: ETS_Model fails to converge for basic model.
#### Describe the bug

ETS_Model fails to converge -- or even move off starting value, in a simple model.

```python
from statsmodels.tsa.exponential_smoothing.ets import ETSModel
from statsmodels.tsa.statespace.exponential_smoothing import ExponentialSmoothing

import numpy as np
gen = np.random.default_rng(0)
e = gen.standard_normal(12000)
y = e.copy()
for i in range(1, e.shape[0]):
    y[i] = y[i-1] - 0.2 * e[i-1] + e[i]
y = y[200:]
res= ExponentialSmoothing(y).fit()
ets_res= ETSModel(y).fit()
```
<details>

</details>

#### Expected Output

Should converge like EstponentialSmoothing does. Instead see
```
C:\Anaconda\lib\site-packages\scipy\optimize\_numdiff.py:519: RuntimeWarning: invalid value encountered in true_divide
  J_transposed[i] = df / dx
c:\git\statsmodels\statsmodels\base\model.py:568: ConvergenceWarning: Maximum Likelihood optimization failed to converge. Check mle_retvals
  "Check mle_retvals", ConvergenceWarning)
```

and `summary()`

```
"""
                                 ETS Results                                  
Dep. Variable:                      y   No. Observations:                11800
Model:                       ETS(ANN)   Log Likelihood              -24532.560
Date:                Mon, 13 Jul 2020   AIC                          49069.121
Time:                        17:56:06   BIC                          49083.872
Sample:                             0   HQIC                         49074.073
                              - 11800                                         
Covariance Type:               approx                                         
-----------------------------------------------------------------------------------
smoothing_level     0.1000      0.001     79.183      0.000       0.098       0.102
initial_level       1.3786      0.843      1.635      0.102      -0.275       3.032
Ljung-Box (Q):                         nan   Jarque-Bera (JB):                  nan
Prob(Q):                               nan   Prob(JB):                          nan
Heteroskedasticity (H):                nan   Skew:                              nan
Prob(H) (two-sided):                   nan   Kurtosis:                          nan

Warnings:
[1] Covariance matrix calculated using numerical (complex-step) differentiation
```

<details>

Tested against master.

</details>

s-scherrer:
- [ ] closes #6883 
- [ ] tests added / passed. 

I added a test for the example in #6883. It works locally, PR to check if it is also working in CI.


bashtage:
Possible a SciPy bug?  What SciPy are you using?  I'm on 1.5.1 as are all of the Azure workers.

s-scherrer:
> Possible a SciPy bug? What SciPy are you using? I'm on 1.5.1 as are all of the Azure workers.

Seems likely. I've been on 1.4.1, after upgrading to 1.5.1 it also fails.

statsmodels/tsa/exponential_smoothing/_ets_smooth.pyx
```diff
@@ -17,16 +17,30 @@ ctypedef fused numeric:
 cpdef _initialize_ets_smooth(
     numeric [:] params,
     numeric[:,:] xhat,
+    np.uint8_t [:] is_fixed,
```
bashtage, line 20: Might as well use np.long here which then maps to `dtype=int` when creating an array.  While uint8 is space-efficient, it is not any faster than using the native int type, and this mask is never large.

statsmodels/tsa/tests/test_exponential_smoothing.py
```diff
@@ -618,3 +619,31 @@ def test_hessian(austourists_model):
     austourists_model.hessian(
         fit.params, approx_complex_step=False, approx_centered=True,
     )
+
+
+def test_convergence_simple():
+    # issue 6883
+    gen = np.random.default_rng(0)
```
bashtage, line 626: Need to use np.random.RandomState(0) here since we still support very old NumPy.

bashtage:
I have a feeling this patch will fix a lot of seemingly random CI failures that were allowed to fail.

statsmodels/tsa/exponential_smoothing/_ets_smooth.pyx
s-scherrer: I used `np.long_t` in the cython code and `dtype=np.long` now. Using `np.long_t` and `dtype=int` failed on the Windows machines.

s-scherrer: Still fails with `ValueError: Buffer dtype mismatch, expected 'long_t' but got 'long'` on the Windows machines. Any idea why?

bashtage:
Should be long_t in Cython and int in Numpy.

s-scherrer:
> Should be long_t in Cython and int in Numpy.

That's what I had in `cda196d`, which also failed with the same message.

bashtage:
If it keeps being a problem just use uint8_t or 32_t in both and don't worry about it.  Very much unnecessary optimization I've started you on I'm sure.

bashtage:
Just checked what worked for me

`cdef long[::1] _xi` and `np.ones(4,dtype=int)`

bashtage:
LGTM.  Let me know if you think it is ready.

s-scherrer:
I added an evaluation of `summary()` to the fixed parameters fit test, just to see if it runs - I thought this would probably be most affected by removing parameters. Seems to run fine here, if there's nothing else I would say it's ready.

bashtage:
Green. Merging.  Thanks.

PR #6895, opened 2020-07-20
Added label option to summary.to_latex()

Jimmy2027:
Proposing a label option to the statsmodels.iolib.summary2.Summary class such that the resulting latex table can be referenced in a latex document.

Below a small example showing the latex table with label:
```
import pandas as pd
import numpy as np

dependent_variable = 'var1'
expression = 'var3'
fit = model.fit()
summary = fit.summary()
summary.label = 'MyLabel'
print(latex_summary)
```
```
\begin{table}
\caption{Mixed Linear Model Regression Results}
\label{MyLabel}
\begin{center}
\begin{tabular}{llll}
\hline
Model:            & MixedLM & Dependent Variable: & Q("var1")  \\
No. Observations: & 10      & Method:             & REML       \\
No. Groups:       & 10      & Scale:              & 0.0616     \\
Min. group size:  & 1       & Likelihood:         & -4.1766    \\
Max. group size:  & 1       & Converged:          & Yes        \\
Mean group size:  & 1.0     &                     &            \\
\hline
\end{tabular}
\end{center}
\hline
\begin{center}
\begin{tabular}{lcccccc}
\hline
          & Coef.  & Std.Err. &   z    & P$> |$z$|$ & [0.025 & 0.975]  \\
\hline
\hline
\end{tabular}
\begin{tabular}{lllllll}
Intercept & 0.642  & 0.031    & 20.609 & 0.000       & 0.581  & 0.704   \\
var3      & -0.258 & 0.203    & -1.275 & 0.202       & -0.655 & 0.139   \\
var2 Var  & 0.062  &          &        &             &        &         \\
\hline
\end{tabular}
\end{center}
\end{table}
```

### DIFF
--- a/statsmodels/iolib/summary2.py
+++ b/statsmodels/iolib/summary2.py
@@ -199,8 +199,14 @@ class Summary(object):
 
         return tab
 
-    def as_latex(self):
+    def as_latex(self, label=''):
         """Generate LaTeX Summary Table
+
+        Parameters
+        ----------
+        label : str
+            Label of the summary table that can be referenced
+            in a latex document (optional)
         """
         tables = self.tables
         settings = self.settings
@@ -210,6 +216,8 @@ class Summary(object):
             title = '\\caption{' + title + '}'
         else:
             title = '\\caption{}'
+
+        label = '\\label{' + label + '}'
 
         simple_tables = _simple_tables(tables, settings)
         tab = [x.as_latex_tabular() for x in simple_tables]
@@ -222,7 +230,7 @@ class Summary(object):
             # create single tabular object for summary_col
             tab = re.sub(to_replace, r'\\midrule\n', tab)
 
-        out = '\\begin{table}', title, tab, '\\end{table}'
+        out = '\\begin{table}', title, label, tab, '\\end{table}'
         out = '\n'.join(out)
         return out
 
--- a/statsmodels/iolib/tests/test_summary2.py
+++ b/statsmodels/iolib/tests/test_summary2.py
@@ -17,6 +17,7 @@ class TestSummaryLatex(object):
         desired = r'''
 \begin{table}
 \caption{}
+\label{}
 \begin{center}
 \begin{tabular}{lll}
 \hline


bashtage:
Is there a doc string where this change can be seen?

Jimmy2027:
Maybe it would make more sense to make the label an option of the "to_latex()" method and consequently make the describing docstring there. A label is probably not needed for the other formats (text, HTML, ...). 
Should I make this change to the PR?

Preview:
```
def as_latex(self, label=None):
    """Generate LaTeX Summary Table

    Parameters
    ----------
    label : str
        Label of the summary table that can be referenced in a latex document (optional)
    """
    tables = self.tables
    settings = self.settings
    title = self.title

    if title is not None:
        title = '\\caption{' + title + '}'
    else:
        title = '\\caption{}'

    if label:
        label = '\\label{' + label + '}'
    else:
        label = '\\label{}'
```

statsmodels/iolib/summary2.py
```diff
@@ -211,6 +212,11 @@ def as_latex(self):
         else:
             title = '\\caption{}'
 
+        if self.label:
```
bashtage, line 215: if self.label is "" then this if then becomes unnecessary.

Jimmy2027:
I have made the label an option of the "as_latex()" method and removed the "if label" statement in my last commit 8d697937e598142338db425fe0d584475f62444d

bashtage:
@Jimmy2027 You need to correct the test now that it has a label.

Jimmy2027:
All tests are passed now :)

bashtage:
One last one needs PEP cleanup:

```
statsmodels/iolib/summary2.py:208:80: E501 line too long (92 > 79 characters)
Previously passing files failed linting.
```

statsmodels/iolib/summary2.py
```diff
@@ -199,8 +199,13 @@ def as_html(self):
...
         return tab
 
-    def as_latex(self):
+    def as_latex(self, label=''):
         """Generate LaTeX Summary Table
+
+        Parameters
+        ----------
+        label : str
+            Label of the summary table that can be referenced in a latex document (optional)
```
bashtage, line 208: Wrap after 79 chars

bashtage:
One last request: can you squash to one commit.

You can do this on the command line with:

```
git remote add upstream https://github.com/statsmodels/statsmodels.git
git fetch --all
git rebase -i upstream/master
<The chang all "pick" to "s" except the very first one, and save, and then edit the combined commit to a single message>
git push -f
```

Only do the git push -f if the rebase went correctly.  If not, you can abort it if you get stuck in the middle, or to just go back you can 

```
git reset --hard origin/Jimmy2027-label
```

bashtage:
Many thanks!

PR #6896, opened 2020-07-20
Fixed the shifted column names in summary.to_latex()

Jimmy2027:
Proposing a fix to an issue with shifted column names when creating a latex table with the "to_latex()" function of the statsmodels.iolib.summary2.Summary class.

Currently when creating a latex table of the summary class, the column names of the second table of the summary get placed in a separate tabular, which results in them not being aligned with the rest of the table.

My proposed fix could be applied directly by an user to the summary dataframes before applying the ".to_latex()" function, though it is very unintuitive. It would be much more user-friendly if the formatting was handled directly by the "to_latex()" function.

This fix might not be the cleanest solution, the intention of this PR is more to raise awareness of the issue while indicating where the issue comes from.

# Example:
```
import pandas as pd
import numpy as np

dependent_variable = 'var1'
expression = 'var3'
fit = model.fit()
summary = fit.summary()
print(latex_summary)
```
## Current behaviour:
```
\begin{table}
\caption{Mixed Linear Model Regression Results}
\begin{center}
\begin{tabular}{llll}
\hline
Model:            & MixedLM & Dependent Variable: & Q("var1")  \\
No. Observations: & 10      & Method:             & REML       \\
No. Groups:       & 10      & Scale:              & 0.0710     \\
Min. group size:  & 1       & Likelihood:         & -4.5306    \\
Max. group size:  & 1       & Converged:          & Yes        \\
Mean group size:  & 1.0     &                     &            \\
\hline
\end{tabular}
\end{center}
\hline
\begin{center}
\begin{tabular}{lcccccc}
\hline
          & Coef.  & Std.Err. &   z    & P$> |$z$|$ & [0.025 & 0.975]  \\
\hline
\hline
\end{tabular}
\begin{tabular}{lllllll}
Intercept & 0.475  & 0.290    & 1.638  & 0.101       & -0.093 & 1.043   \\
var3      & -0.024 & 0.444    & -0.055 & 0.956       & -0.895 & 0.847   \\
var2 Var  & 0.071  &          &        &             &        &         \\
\hline
\end{tabular}
\end{center}
\end{table}
```
Which results in:
image

## New behaviour:

```
\begin{table}
\caption{Mixed Linear Model Regression Results}
\begin{center}
\begin{tabular}{llll}
\hline
Model:            & MixedLM & Dependent Variable: & Q("var1")  \\
No. Observations: & 10      & Method:             & REML       \\
No. Groups:       & 10      & Scale:              & 0.0308     \\
Min. group size:  & 1       & Likelihood:         & -1.0323    \\
Max. group size:  & 1       & Converged:          & Yes        \\
Mean group size:  & 1.0     &                     &            \\
\hline
\end{tabular}
\end{center}
\hline
\begin{center}
\begin{tabular}{lllllll}
\hline
          & Coef.  & Std.Err. & z      & P$> |$z$|$ & [0.025 & 0.975]  \\
Intercept & 0.972  & 0.191    & 5.097  & 0.000       & 0.598  & 1.346   \\
var3      & -0.955 & 0.343    & -2.785 & 0.005       & -1.627 & -0.283  \\
var2 Var  & 0.031  &          &        &             &        &         \\
\hline
\end{tabular}
\end{center}
\end{table}
```
Which results in:

image

### DIFF
--- a/statsmodels/iolib/summary2.py
+++ b/statsmodels/iolib/summary2.py
@@ -206,6 +206,12 @@ class Summary(object):
         settings = self.settings
         title = self.title
 
+        table1 = tables[1].reset_index()
+        table1 = table1.T.reset_index().T.reset_index(drop=True)
+        table1.iloc[0, 0] = ''
+        tables[1] = table1
+        settings[1]['index'] = settings[1]['header'] = False
+
         if title is not None:
             title = '\\caption{' + title + '}'
         else:
--- a/statsmodels/iolib/summary2.py
+++ b/statsmodels/iolib/summary2.py
@@ -218,7 +218,14 @@ class Summary(object):
             title = '\\caption{}'
 
         simple_tables = _simple_tables(tables, settings)
-        tab = [x.as_latex_tabular() for x in simple_tables]
+
+        tab = []
+        for i, x in enumerate(simple_tables):
+            if i == 1:
+                tab.append(x.as_latex_tabular(cmidrule=True))
+            else:
+                tab.append(x.as_latex_tabular())
+
         tab = '\n\\hline\n'.join(tab)
 
         to_replace = ('\\\\hline\\n\\\\hline\\n\\\\'
--- a/statsmodels/iolib/table.py
+++ b/statsmodels/iolib/table.py
@@ -404,9 +404,14 @@ class SimpleTable(list):
         formatted_rows.append('</table>')
         return '\n'.join(formatted_rows)
 
-    def as_latex_tabular(self, center=True, **fmt_dict):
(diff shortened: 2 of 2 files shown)


Jimmy2027:
True, a cmidrule under the coefficient names makes the table look much nicer! I have added a cmidrule option to the "SimpleTable.as_latex_tabular" method in commit 1a9127a18463b7721d0641f2c2c92cee1d8b2c7a

This results in a table that looks like following with the code above:
image

bashtage:
This approach cannot work unfortunately.  I have an approach in #6900 that should.  It is quite convoluted inside these io functions so simple fixed often break something else. 

Thanks for raising this and the attempt!  Please do submit PRs for issues that need fixing.

PR #6905, opened 2020-07-21
removed additional hline between tabulars

Jimmy2027:
Proposing to remove the hline between the tabulars of the latex table created by the "to_latex()" function of the statsmodels.iolib.summary2.Summary class.

# Example:
```
import pandas as pd
import numpy as np
dependent_variable = 'var1'
expression = 'var3'
fit = model.fit()
summary = fit.summary()
print(latex_summary)
```

## Current behavior:
image

## New behavior:
image


bashtage:
Thanks!

PR #6919, opened 2020-07-24
Use scipy.linalg.svd over numpy

tmct:
- [ ] closes #xxxx
- [ ] tests added / passed. 

I've seen scipy's LAPACK wrapper give a significant performance increase compared to numpy, both in elapsed time and memory usage. It looks like others have come to this conclusion too: https://stackoverflow.com/a/58137088

**Notes**:

  needed for doc changes.
  ```
  ```

</details>

### DIFF
--- a/statsmodels/tools/tools.py
+++ b/statsmodels/tools/tools.py
@@ -400,7 +400,7 @@ def pinv_extended(x, rcond=1e-15):
     """
     x = np.asarray(x)
     x = x.conjugate()
-    u, s, vt = np.linalg.svd(x, False)
+    u, s, vt = scipy.linalg.svd(x, full_matrices=False)
     s_orig = np.copy(s)
     m = u.shape[0]
     n = vt.shape[1]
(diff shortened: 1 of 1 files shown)


josef-pkt:
That was a moving target.
When we started with this (10 years ago) numpy.linalg.pinv was much faster than scipy.linalg.
Many internals of scipy.linalg have changed.

can you do some timing for our common cases (nobs, k_vars) with nobs much larger than k_vars, e.g 1 Million by 100 or by 500

statsmodels/tools/tools.py
```diff
@@ -506,7 +506,7 @@ def fullrank(x, r=None):
     if r is None:
         r = np.linalg.matrix_rank(x)
```
bashtage, line 507: This call is wasteful since it is performing a second svd.

josef-pkt: this function is not used, old code

We keep some of the old linalg functions as recipes, until we have our own toolset for linalg

tmct:
Yes, definitely, I'll do some timing tests in a couple of days before declaring it ready - thanks for the suggestions.

bashtage:
Looks like there may be a bad test somewhere.

bashtage:
That SO post seems to be out-of-date.  I ran the notebook there and the NumPy and SciPy times are identical on both Windows and Linux, whether using OpenBLAS or MKL.

image

bashtage:
Thanks for looking into this.

tmct:
I have worked out why I was seeing different times for numpy - I was using numpy with MKL and scipy without MKL... Sorry about that!

No problem, will keep on the lookout for genuine improvements. Thanks for checking up on this.

PR #6937, opened 2020-07-31
ENH: Add dynamic factor model with EM algorithm, option for monthly/quarterly mixed frequency model

ChadFulton:
- [x] tests added / passed. 

This basically adds the dynamic factor model described in Bańbura et al (2011), along with the the EM algorithm from Bańbura and Modugno (2014) for parameter fitting.

Highlights:

- Unlike the current DFM model in Statsmodels, this can handle models with hundreds of variables
- It also supports allowing each variable to load on only a subset of the factors
- And it supports specifying "blocks" of factors that transition independently (according to a VAR process) from the other blocks of factors.
- It can handle the monthly / quarterly mixed frequency approach of these papers, and so this model supports nowcasting exercises (made possible by the addition of the "news" concept in #6765). In particular, it can replicate the [New York Fed Staff Nowcast](https://www.newyorkfed.org/research/policy/nowcast.html)

There is a somewhat long notebook that goes through many of the features, and ends in a nowcasting exercise similar to the NY Fed nowcasting model: https://nbviewer.jupyter.org/gist/ChadFulton/a9172cd6f41947333f47bd4842832619

- I haven't put the notebook into the PR, because it takes about 3 minutes to run, and I'm not sure if that's too long for the documentation builder. In any case, we have a couple of notebooks like that now, and I need to figure out a better solution before the v0.12 release.

------------

    Bańbura, Marta, and Michele Modugno.
    "Maximum likelihood estimation of factor models on datasets with
    arbitrary pattern of missing data."
    Journal of Applied Econometrics 29, no. 1 (2014): 133-160.

    Bańbura, Marta, Domenico Giannone, and Lucrezia Reichlin.
    "Nowcasting."
    The Oxford Handbook of Economic Forecasting. July 8, 2011.


ChadFulton:
This also makes state space models more robust by computing the log-determinant in a numerically safe way, rather than computing the determinant and then taking the log.

PR #6942, opened 2020-08-03
Make glm's predict function return numpy array even if exposure is a pandas series

kbrose:
- [ ] closes #xxxx
- [ ] tests added / passed. 
<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>

### DIFF
--- a/statsmodels/genmod/generalized_linear_model.py
+++ b/statsmodels/genmod/generalized_linear_model.py
@@ -870,7 +870,7 @@ class GLM(base.LikelihoodModel):
         elif exposure is None:
             exposure = 0.
         else:
-            exposure = np.log(exposure)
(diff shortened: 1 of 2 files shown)


bashtage:
Why? You can use as array in your own code to ensure numpy.

bashtage:
Taking another look it seems pretty reasonable. Exog should determine type and index in most cases.  This said, if exposure has an index and exog does not, it seems that the index should be preserved.

kbrose:
I passed in exposure as a pandas series with the name "exposure", and got out a pandas series with the name "exposure". That didn't seem correct to me and is why I submitted the PR. Thanks for taking a look at it!

bashtage:
This makes sense and is likely safer.  Thanks.

PR #6973, opened 2020-08-18
ENH/BUG: robust.scale.iqr does need centering, since quantiles are translation equivariant

esmucler:
This is regarding the recently added robust.scale.iqr, which computes the normalized interquartile range. Since quantiles are translation equivariant, the interquartile range as an estimate of scale does not need centering. Centering, as implemented now, does not hurt either, it is inconsequential, but there is no need for it. This PR simply removes centering from robust.scale.iqr.

- [Y] tests added / passed. 

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>

### DIFF
--- a/statsmodels/robust/scale.py
+++ b/statsmodels/robust/scale.py
@@ -72,7 +72,9 @@ def iqr(a, c=Gaussian.ppf(3/4) - Gaussian.ppf(1/4), axis=0):
     a = array_like(a, 'a', ndim=None)
     c = float_like(c, 'c')
 
-    if a.size == 0:
+    if a.ndim == 0:
+        raise ValueError("a should have at least one dimension")
+    elif a.size == 0:
         return np.nan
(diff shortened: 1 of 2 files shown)


statsmodels/robust/scale.py
```diff
@@ -79,11 +75,7 @@ def iqr(a, c=Gaussian.ppf(3/4) - Gaussian.ppf(1/4), axis=0, center=np.median):
     if a.size == 0:
         return np.nan
```
bashtage, line 76: a must have at least 1 dim since axis=0.  So we should add a check and raise ValueError if a.ndim == 0

bashtage:
Looks good, but would be nice to add ndim check.

bashtage:
Failing on linting.

bashtage:
Thanks.

PR #6977, opened 2020-08-21
ENH: Implementation of the robust Qn scale estimator

esmucler:
- [X] related to #6798
- [X] tests added / passed. tested against robustbase from R and also against a naive implementation

This is an implementation of the Qn robust estimator of scale. The estimator was first proposed in Rousseeuw and Croux (1993), Alternatives to the Median Absolute Deviation. The main appeal of the estimator is that it has maximal breakdown point (50%), a high asymptotic efficiency at the normal distribution (~80%) and that it does not require an auxiliary location estimator. The implementation here follows the one described in Croux and Rousseeuw (1992), Time-Efficient Algorithms for Two Highly Robust Estimators of Scale. The implementation is tested against a naive (and slow) implementation of the estimator and also against the implementation available in the robustbase R package for a couple of datasets. The implementation in robustbase also includes some finite sample corrections, which were derived (according to the documentation) from a simulation study that is not referenced. I did not include these finite sample corrections in my implementation, since I am not sure of how they were constructed, and in my opinion, they make the estimator less transparent.
<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>


statsmodels/robust/scale.py
```diff
@@ -81,6 +85,73 @@ def iqr(a, c=Gaussian.ppf(3/4) - Gaussian.ppf(1/4), axis=0):
         return np.squeeze(np.diff(quantiles, axis=0) / c)
 
 
+def qn_scale(a, c=1 / (np.sqrt(2) * Gaussian.ppf(5 / 8)), axis=0):
+    """
+    Computes the Qn robust estimator of scale, a more efficient alternative
```
bashtage, line 90: This needs to be split into a 1-line header, then a space, then an extended defintioin. For example,

```
Computes the Qn robust estimator of scale

The Qn scale estimator is a more efficient alternative to the MAD ...
```

statsmodels/robust/scale.py
```diff
@@ -81,6 +85,73 @@ def iqr(a, c=Gaussian.ppf(3/4) - Gaussian.ppf(1/4), axis=0):
...
+        Input array.
+    c : float, optional
+        The normalization constant. The default value is used to get consistent
+        estimates of the standard deviation at the normal distribution.
+    axis : int, optional
+        The default is 0.
+
+    Returns
+    -------
+    The Qn robust estimator of scale
```
bashtage, line 110: Should be
```
{float, ndarray}
    The Qn robust estimator of scale
```

Assuming it returns a float for a 1-d array.

statsmodels/robust/scale.py
```diff
@@ -81,6 +85,73 @@ def iqr(a, c=Gaussian.ppf(3/4) - Gaussian.ppf(1/4), axis=0):
...
+    """
+    a = array_like(a, 'a', ndim=None, dtype=np.float64, contiguous=True,
+                   order='C')
+    c = float_like(c, 'c')
+    if a.ndim == 0:
+        raise ValueError("a should have at least one dimension")
+    elif a.size == 0:
+        return np.nan
+    else:
+        return np.apply_along_axis(_qn, axis=axis, arr=a, c=c)
```
bashtage, line 120: Rather than returning firectly, set this to a variable, e.g., 

```
out = np.apply...
if out.ndim == 0:
    return float(out)
return out
```

So that the 1-d reduction returns a float.

statsmodels/robust/scale.py
esmucler: Done in b04b51cb2b83b3d9e1828d3b53956c1a7fd44629

statsmodels/robust/scale.py
esmucler: Done in b04b51cb2b83b3d9e1828d3b53956c1a7fd44629

statsmodels/robust/scale.py
esmucler: Done in b04b51cb2b83b3d9e1828d3b53956c1a7fd44629

statsmodels/robust/scale.py
```diff
@@ -81,6 +85,72 @@ def iqr(a, c=Gaussian.ppf(3/4) - Gaussian.ppf(1/4), axis=0):
...
+    Parameters
+    ----------
+    a : array_like
+        Input array.
+    c : float, optional
+        The normalization constant, used to get consistent estimates of the
+        standard deviation at the normal distribution.  Defined as
+        1/(np.sqrt(2) * scipy.stats.norm.ppf(5/8)), which is 2.219144.
+    axis : int, optional
+        The default is 0. Can also be None.
```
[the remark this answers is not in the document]
bashtage: Thanks.

bashtage:
Final request : can you squash to a more modest number of commits, say 5 (1 is also OK). After this and green, it is in.

esmucler:
> Final request : can you squash to a more modest number of commits, say 5 (1 is also OK). After this and green, it is in.

No problem. If I squash to 1 commit, should I keep the "squashed" commit messages or just leave a single one that says, for example, "add an implementation of the robust Qn scale estimator" ?

bashtage:
Thanks. I rebased and merged it in #6990.

PR #6997, opened 2020-08-26
ENH: Allow custom bandwidth functions in KDEUnivariate fit

### LINKED ISSUE #6988: ENH: Let users set KDE bandwidth use a user-defined bandwidth function
#### Describe the bug

Unexpected exception raised for KDE fit() when input data has exactly zero bandwidth.

#### Code Sample (copy-paste works)

```python
from statsmodels.nonparametric.kde import KDEUnivariate
import numpy as np

kde = KDEUnivariate(np.array([0.1, 0.1]))
kde.fit()
```
<details>

I did have checked wether there is a similar open bug and couldn't find one.
I did look at the source code on master, but did not run it and seems the behavior is there as well.

</details>

#### Expected Output

If running statsmodels in a pipeline then the user data might be anything, and would expect it handles this gracefully instead of raising an exception. This would allow the user to see a nice graph in eg seaborn instead of an error for a very simple case.

```python
In [1]: print(kde.density)
Out[1]: array([ ... ])
In [1]: print(kde.support)
Out[1]: array([ ... ])
```

Also behavior of `[0.1, 0.1]` should be very similar to `[0.1, 0.1000000000001]` (consider details points 1, 2, 3 below):
```python
from statsmodels.nonparametric.kde import KDEUnivariate
import numpy as np

kde = KDEUnivariate(np.array([0.1, 0.1000000000001]))
kde.fit()
```

<details>

1. The == 0 in `nonparametric/bandwidths.py:171 select_bandwidth()` comparing with a float is not kosher numerical methods practice, usually a comparison using a very small error term `eps` should be done.

2. The silverman bandwidth estimation might seem to be correct (== 0) in a situation where two observations that are the same. The bandwidth of a gaussian is actually not zero, but also not trivial to calculate.
Closest reference I could find is [this](https://math.stackexchange.com/questions/2693800/probability-that-two-consecutively-generated-integers-with-normal-distribution-a).
Net, in this case, the standard deviation estimator `std()` is not the best linear unbiased estimator for the standard deviation due to low number of samples and the uncertainty `eps`.
Based on this approach, I would suggest to use in this case `A == max(min(std(X), IQR/1.34), eps * scipy.stats.norm.ppf(q=1-eps/mu))` for input to the silverman or scott bandwidth estimators. In case `eps` is not specified just use 90% confidence directly, however the user should be allowed to specify it as well.
The logic is quite simple: the uncertainty `eps` when compared to the average `mu` gives a certain confidence level (90% in my example if we take `eps=0.01` ie one digit lower) which translates to an equivalent standard deviation based on that uncertainty ie `1.21 * eps`.
TL;DR The `eps` acts as a lower bound to standard deviation for cases with low number of observations.

3. While raising an exception in this case might seem pythonic, it's still not warranted, as one could still have a result out of this using a fitted Gaussian with `avg=0.1` and `std=0.012815515655446004` (based on point 2 above). Think about what a normal user has to do: handle exception by monkey patching the kde results so it can be used further down the code.

</details>

<details>

INSTALLED VERSIONS
------------------
Python: 3.8.5.final.0
OS: Linux 5.7.16-200.fc32.x86_64 #1 SMP Wed Aug 19 16:58:53 UTC 2020 x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8

statsmodels
===========

Installed: 0.11.0 (/usr/lib64/python3.8/site-packages/statsmodels)

Required Dependencies
=====================

cython: Not installed
numpy: 1.18.4 (/usr/lib64/python3.8/site-packages/numpy)
scipy: 1.4.1 (/usr/lib64/python3.8/site-packages/scipy)
pandas: 0.25.3 (/usr/lib64/python3.8/site-packages/pandas)
    dateutil: 2.8.0 (/usr/lib/python3.8/site-packages/dateutil)
patsy: 0.5.1 (/usr/lib/python3.8/site-packages/patsy)

Optional Dependencies
=====================

matplotlib: 3.2.2 (/usr/lib64/python3.8/site-packages/matplotlib)
    backend: Qt5Agg 
cvxopt: Not installed
joblib: 0.13.2 (/usr/lib/python3.8/site-packages/joblib)

Developer Tools
================

IPython: 7.12.0 (/usr/lib/python3.8/site-packages/IPython)
    jinja2: 2.11.2 (/usr/lib/python3.8/site-packages/jinja2)
sphinx: 2.2.2 (/usr/lib/python3.8/site-packages/sphinx)
    pygments: 2.4.2 (/usr/lib/python3.8/site-packages/pygments)
pytest: 4.6.11 (/usr/lib/python3.8/site-packages)
virtualenv: Not installed

</details>

dbivolaru:
- [x] closes #6988
- [x] tests added / passed. 

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>

### DIFF
--- a/statsmodels/nonparametric/kde.py
+++ b/statsmodels/nonparametric/kde.py
@@ -344,7 +344,10 @@ def kdensity(X, kernel="gau", bw="normal_reference", weights=None, gridsize=None
 
     # if bw is None, select optimal bandwidth for kernel
     try:
-        bw = float(bw)
+        if callable(bw):
+            bw = float(bw(X, kern))
+        else:
+            bw = float(bw)
     except:
         bw = bandwidths.select_bandwidth(X, bw, kern)
     bw *= adjust
@@ -454,7 +457,10 @@ def kdensityfft(X, kernel="gau", bw="normal_reference", weights=None, gridsize=N
     kern = kernel_switch[kernel]()
 
     try:
-        bw = float(bw)
+        if callable(bw):
+            bw = bw(X, kern) # user passed a callable custom bandwidth function
+        else:
+            bw = float(bw)
     except:
         bw = bandwidths.select_bandwidth(X, bw, kern) # will cross-val fit this pattern?
     bw *= adjust
--- a/statsmodels/nonparametric/kde.py
+++ b/statsmodels/nonparametric/kde.py
@@ -95,7 +95,7 @@ class KDEUnivariate(object):
             - "triw" for triweight
             - "uni" for uniform
 
-        bw : str, float
+        bw : str, float, callable
             The bandwidth to use. Choices are:
 
             - "scott" - 1.059 * A * nobs ** (-1/5.), where A is
@@ -106,6 +106,10 @@ class KDEUnivariate(object):
               calculated from the kernel. Equivalent (up to 2 dp) to the
               "scott" bandwidth for gaussian kernels. See bandwidths.py
             - If a float is given, it is the bandwidth.
+            - If a callable is given, it's return value is used.
+              The callable should take exactly two parameters ie. fn(X, kern)
+                X    - the clipped input data
+                kern - the kernel instance used
 
         fft : bool
             Whether or not to use FFT. FFT implementation is more
@@ -281,10 +285,14 @@ def kdensity(X, kernel="gau", bw="normal_reference", weights=None, gridsize=None
         - "tri" for triangular
         - "triw" for triweight
         - "uni" for uniform
-    bw : str, float
+    bw : str, float, callable
         "scott" - 1.059 * A * nobs ** (-1/5.), where A is min(std(X),IQR/1.34)
         "silverman" - .9 * A * nobs ** (-1/5.), where A is min(std(X),IQR/1.34)
         If a float is given, it is the bandwidth.
+        If a callable is given, it's return value is used.
+        The callable should take exactly two parameters ie. fn(X, kern)
+          X    - the clipped input data
+          kern - the kernel instance used
     weights : array or None
         Optional  weights. If the X value is clipped, then this weight is
         also dropped.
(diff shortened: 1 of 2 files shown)


bashtage:
The docstring needs to be updated to reflect that a callable is acceptable, and what the signature of the callable should be.

statsmodels/nonparametric/kde.py
```diff
@@ -454,7 +457,10 @@ def kdensityfft(X, kernel="gau", bw="normal_reference", weights=None, gridsize=N
     kern = kernel_switch[kernel]()
 
     try:
-        bw = float(bw)
+        if callable(bw):
```
bashtage, line 460: The if should be before the try. If the callable fails, we don't want to hit the except.

statsmodels/nonparametric/kde.py
```diff
@@ -344,7 +344,10 @@ def kdensity(X, kernel="gau", bw="normal_reference", weights=None, gridsize=None
 
     # if bw is None, select optimal bandwidth for kernel
     try:
-        bw = float(bw)
+        if callable(bw):
```
bashtage, line 347: Same here, should be 
```
if callable(bw):
    float(bw(X, kern))
else:
    try: 
        ...
```

statsmodels/nonparametric/tests/test_kde.py
```diff
@@ -348,3 +348,19 @@ def test_fit_self(reset_randomstate):
...
+
+class TestKDECustomBandwidth(object):
+
+    @classmethod
+    def setup_class(cls):
+        cls.kde = KDE(Xi)
+        cls.weights_200 = np.linspace(1, 100, 200)
+        cls.weights_100 = np.linspace(1, 100, 100)
+
+    def test_check_is_fit_ok_with_custom_bandwidth(self):
```
bashtage, line 361: Would be good to add a test using an existing callable bandwidth method.  This should produce an identical bandwidth whether using the callable or the string version.

bashtage (changes requested):
Some changes needed, especially to the docstring.

bashtage (approved):
LGTM

PR #7038, opened 2020-09-16
TST: Use unused test @parameterize variable

has2k1:
Pytest @parameterize variable `kernel` was unused in the body of
the test function.

Reference: #6547

- [ ] closes #xxxx
- [ ] tests added / passed. 

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>

### DIFF
--- a/statsmodels/nonparametric/tests/test_kernel_density.py
+++ b/statsmodels/nonparametric/tests/test_kernel_density.py
@@ -423,5 +423,5 @@ def test_all_kernels(kernel, reset_randomstate):
     data = np.random.normal(size=200)
     x_grid = np.linspace(min(data), max(data), 200)
     density = sm.nonparametric.KDEUnivariate(data)
-    density.fit(kernel="gau", fft=False)
+    density.fit(kernel=kernel, fft=False)
     assert isinstance(density.evaluate(x_grid), np.ndarray)


bashtage:
Seems there are bugs too. Thanks for picking up on this.  Do you want to try and fix to bugs too? If not I'll fix them using your commit as the first.

has2k1:
@bashtage, I see the failing tests but I am not sure what the bugs are.

1. Should the `density.evaluate` *never* return a singular `nan` (a non-ndarray) for any input?
2. The [in_domain](https://github.com/statsmodels/statsmodels/blob/df1c1a1bd17bfbab7c61bd903f06362a7bb5dc0b/statsmodels/sandbox/nonparametric/kernels.py#L163-L183) checking routine looks wrong to me, but this is really old code so I doubt myself.

If the domain calculation is indeed buggy, the solution I have is

```python
    def in_domain(self, xs, ys, x):
        """
        Returns the filtered (xs, ys) based on the Kernel domain centred on x
        """
        # Disable black-list functions: filter used for speed instead of
        # list-comprehension
        # pylint: disable-msg=W0141
        def isInDomain(xs):
            """Used for filter to check if point is in the domain"""
            u = (xs - x)/self.h
            return (u >= self.domain[0]) & (u <= self.domain[1])

        if self.domain is None:
            return (xs, ys)
        else:
            indomain_idx = isInDomain(xs)
            if any(indomain_idx):
                return (xs[indomain_idx], ys[indomain_idx])
            else:
                return ([], [])
```

This corrects an `n x n`  calculation where a point is in the domain only if all other points are in the domain! This is what looked wrong to me. If I am totally off, then please takeover.

bashtage:
> This corrects an `n x n` calculation where a point is in the domain only if all other points are in the domain! This is what looked wrong to me. If I am totally off, then please takeover.

I think this function is wrong.  But maybe not a simple fix.

has2k1:
> > This corrects an `n x n` calculation where a point is in the domain only if all other points are in the domain! This is what looked wrong to me. If I am totally off, then please takeover.
> 
> I think this function is wrong. But maybe not a simple fix.

Please take over.

bashtage:
> Please take over.

I didn't mean your function! Sorry about that.  I meant the original function.  It should be finding the in-domain points to estimate the density.  Basically the domain definition is just wrong.  Many of these kernels have infinite support they are just 0 outside of [-1,1].

PR #7040, opened 2020-09-17
Bug fix ets get prediction

### LINKED ISSUE #7018: ETSModel get_prediction bug if start date is greater than dataset
The get_prediction method encounters an error if the start date is after the end of the dataset.

import numpy as np
import pandas as pd
from statsmodels.tsa.api import ETSModel
values = np.arange(100)* (1+np.random.randn(100)/10)
dates = [np.datetime64('today') + np.timedelta64(-101+i) for i in range(100)]
x_train = pd.Series(values, index=dates)
x_train = x_train.asfreq(freq='D', method='bfill')
fit1 = ETSModel(x_train, seasonal_periods=7, trend='add').fit()
works = fit1.get_prediction(start=np.datetime64('today')- np.timedelta64(5), end=np.datetime64('today')+ np.timedelta64(5))
err = fit1.get_prediction(start=np.datetime64('today'), end=np.datetime64('today')+ np.timedelta64(5))

  File "\Anaconda3\lib\site-packages\statsmodels\tsa\exponential_smoothing\ets.py", line 2234, in __init__
    start : (end + 1)

ValueError: could not broadcast input array from shape (0) into shape (6)

pratyushsharan:
- [x ] closes #7018 
- [x ] tests added / passed. 
- [x ] code/documentation is well formatted.  
- [x ] properly formatted commit message. See 

<details>
Fixed the issue where the ETSModel get_prediction method fails when start is greater than the last index. Modified test case to reflect this use case, which would fail on master and would pass once this branch is merged.

**Notes**:

  needed for doc changes.
  ```
  ```

</details>

### DIFF
--- a/statsmodels/tsa/exponential_smoothing/ets.py
+++ b/statsmodels/tsa/exponential_smoothing/ets.py
@@ -2230,9 +2230,10 @@ class PredictionResults:
         )
         self.row_labels = self.predicted_mean.index
         self.endog = np.empty(nsmooth + ndynamic) * np.nan
-        self.endog[0 : (end - start + 1)] = results.data.endog[
-            start : (end + 1)
-        ]
+        if nsmooth > 0:
+            self.endog[0: (end - start + 1)] = results.data.endog[
+                start: (end + 1)
+            ]
         self.model = Bunch(
             data=results.model.data.__class__(
                 endog=self.endog, predict_dates=self.row_labels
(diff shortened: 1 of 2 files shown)


bashtage:
Could you parameterize the test using different start values including one in-sample, one at the end of the sample, one just out of sample, and one well out of sample?

Thanks,
Kevin

pratyushsharan:
> Could you parameterize the test using different start values including one in-sample, one at the end of the sample, one just out of sample, and one well out of sample?

Added the suggested tests. Have also added a test case where `start` == `end`, as a sanity check.

bashtage:
Many thanks.

PR #7042, opened 2020-09-18
Added the option `full_matrices=False` in the PCA method

omartrinidad:
Added the option full_matrices=False in PCA in order to deal fast with big matrices.


bashtage:
Need a test.  What will change? If the output dimension changes then this will need to become an option, at least at first, and possibly deprecation.

omartrinidad:
Hi, I tried to calculate PCA for a matrix with around 66 000 rows... the operation froze. I guess the SVD output, either full or truncated, both are valid for the purpose of the PCA calculation, but there is a huge gain using the SVD truncated matrices.  
I will create the test.

bashtage:
@omartrinidad Does this change any of the output dimensions?

bashtage:
Still need tests.  Looks good, subject to testing that the output dimensions aren't changing.

omartrinidad:
> Still need tests. Looks good, subject to testing that the output dimensions aren't changing.

Yes, I am trying to understand the already written tests to add a new one.

omartrinidad:
> Still need tests. Looks good, subject to testing that the output dimensions aren't changing.

I already wrote the test, it is passing.

In the `svd` function the parameter `full_matrices` is by default set as True.

I did use `assert_allclose` to compare the factors when `full_matrices` is set to True against when is set to False.

omartrinidad:
I did send again the pull-request.

bashtage:
Did you mean to close this?

omartrinidad:
> Did you mean to close this?

I sent again another pull-request #7329 using a different branch, rather than `master`. 

Not sure if it is good to keep this pull-request active.

PR #7117, opened 2020-10-27
Travis CI:  Use services: xvfb and $TRAVIS_PYTHON_VERSION

cclauss:
https://docs.travis-ci.com/user/gui-and-headless-browsers/#using-services
https://docs.travis-ci.com/user/environment-variables

- [ ] closes #xxxx
- [x] tests added / passed. 

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,6 +90,7 @@ jobs:
     # Latest pre-release packages
     - python: 3.9
       env:
+        - SCIPY=1.5.4
         - PIP_PRE=true
         - BUILD_INIT=tools/ci/travis_pip.sh
     - python: 3.8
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -57,7 +57,7 @@ jobs:
         - LINT=true
     - python: 3.9
       env:
-        - PYTHON=3.7
```
bashtage, line 52: Let's move the doc build to 3.8.  It has a long tail of deps and better for it to be stable.

bashtage:
I think we need to wait until SciPy has wheels for 3.9 before merging this, and then to move some of the allowed failures to required.

bashtage:
I also think pip-pre should stay 3.8 for now since it is probably the Python that most are still using.

bashtage:
LGTM.  I guess 3.9 isn't ready yet.

cclauss:
Python 3.9 is clearly ready but statsmodels still runs Python 3.8 in `allow_failures` mode.

bashtage:
The allowed failure is for pre release packages.  It helps pick up changes early.  3.9 should be a required pass.

PR #7141, opened 2020-11-03
ENH: Adding an if clause in regression.rolling.WLS._fit_single to avo…

### LINKED ISSUE #7140: Avoid np.linalg.inv call in regression.rolling.RollingWLS.fit if output is not used
#### Problem description

In the documentation of `statsmodels.regression.rolling.RollingWLS.fit` the parameter `method` is described as 

> method{‘inv’, ‘lstsq’, ‘pinv’}
> Method to use when computing the the model parameters.
> ‘inv’ - use moving windows inner-products and matrix inversion. This method is the fastest, but may be less accurate than the other methods.
> ‘lstsq’ - Use numpy.linalg.lstsq
> ‘pinv’ - Use numpy.linalg.pinv. This method matches the default estimator in non-moving regression estimators.

An obvious reason that 'inv' is fastest in the current implementation to be that `np.linalg.inv` is called also for 'lstsq' and 'pinv':
image

(ll. 222 in https://github.com/statsmodels/statsmodels/blob/master/statsmodels/regression/rolling.py)

If `params_only==True` then `wxpwxi` seems to be never used but has impact on performance. 

#### Proposed solution
Doing in the try-except block quoted above something like   
```python
    if method == "inv" or not params_only:
        wxspwxi = np.linalg.inv(wxpwx)
    [...]
```
should speed up the the method for `method=='lstsq'` and `method=='pinv'`.

mm842:
…id unnecessary calls of np.linalg.inv

- [x] closes #7140
- [x] tests added/passed. 

<details>

I ran the following code to compare the performance:
```python
import timeit
import numpy as np
from statsmodels.regression import rolling

data_x = np.random.normal(size=(500, 20))
data_y = np.random.normal(size=500)

def f():
    rolling.RollingWLS(data_y, data_x, window=450, min_nobs=450).fit(params_only=True, method='inv')
    
def g():
    rolling.RollingWLS(data_y, data_x, window=450, min_nobs=450).fit(params_only=True, method='pinv')
    
def h():
    rolling.RollingWLS(data_y, data_x, window=450, min_nobs=450).fit(params_only=True, method='lstsq')
    

print("Time for 'inv': ", timeit.timeit("f()", setup="from __main__ import f", number=1000))
print("Time for 'pinv': ", timeit.timeit("g()", setup="from __main__ import g", number=1000))
print("Time for 'lstsq': ", timeit.timeit("h()", setup="from __main__ import h", number=1000))
```

 __before__ the change:
```
Time for 'inv':  9.21082185799969
Time for 'pinv':  66.74583183600043
Time for 'lstsq':  45.32076134499948
```
__after__ the change:
```
Time for 'inv':  8.389769997997064
Time for 'pinv':  25.263389345000178
Time for 'lstsq':  16.581308099001035
```

**Notes**:

  needed for doc changes.
  ```
  ```

</details>

### DIFF
--- a/statsmodels/regression/rolling.py
+++ b/statsmodels/regression/rolling.py
@@ -223,7 +223,8 @@ class RollingWLS(object):
         if nobs < self._min_nobs:
             return
         try:
-            wxpwxi = np.linalg.inv(wxpwx)
+            if (method == "inv") or not params_only:
+                wxpwxi = np.linalg.inv(wxpwx)
             if method == "inv":
                 params = wxpwxi @ wxpwy
             else:
--- a/statsmodels/regression/tests/test_rolling.py
+++ b/statsmodels/regression/tests/test_rolling.py
@@ -286,7 +286,7 @@ def test_params_only(basic_data, method):
     res = mod.fit(method=method, params_only=False)
     res_params_only = mod.fit(method=method, params_only=True)
     assert_array_equal(res_params_only.params, res.params)
-    
+
 
 def test_min_nobs(basic_data):
     y, x, w = basic_data
--- a/statsmodels/regression/tests/test_rolling.py
+++ b/statsmodels/regression/tests/test_rolling.py
@@ -279,6 +279,15 @@ def test_methods(basic_data, params_only):
     assert_allclose(res_inv.params, res_pinv.params)
 
 
+@pytest.mark.parametrize("method", ["inv", "lstsq", "pinv"])
+def test_params_only(basic_data, method):
+    y, x, _ = basic_data
+    mod = RollingOLS(y, x, 150)
+    res = mod.fit(method=method, params_only=False)
+    res_params_only = mod.fit(method=method, params_only=True)
+    assert_array_equal(res_params_only.params, res.params)
+    
+
 def test_min_nobs(basic_data):
     y, x, w = basic_data
     if not np.any(np.isnan(np.asarray(x))):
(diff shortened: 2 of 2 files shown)


bashtage:
All we need now is a smoke test that uses all three configs with and without params_only.

mm842:
> All we need now is a smoke test that uses all three configs with and without params_only.

I'd think an effective way to incorporate this is to just parametrize test_methods by params_only (see last commit https://github.com/statsmodels/statsmodels/pull/7141/commits/f5a277ef5b8020c2691e9e437e6aba8517f47db0) 
Do you agree on that or think better to create a new test for that?

bashtage:
Happy with any approach that makes sure there is no bug introduced in the `if` (not that I think there is).

mm842:
> Happy with any approach that makes sure there is no bug introduced in the `if` (not that I think there is).

Actually, since the if clause checks 2 statements (method, tested in `test_methods`) and params_only, I added a test for the latter: test_params_only, which tests for each of the 3 methods if the results with `params_only=True` and `params_only=False` coincide.

mm842:
> Happy with any approach that makes sure there is no bug introduced in the `if` (not that I think there is).

That should be covered by the `test_methods` and `test_params_only` now. 
`test_params_only` worked on x64 as well with assert_array_equal but failed on the x86 AppVeyor setup with errors around 1e-14 and less - so I replaced it with `assert_allclose`.

bashtage:
GTM.  Many thanks.

PR #7147, opened 2020-11-07
Loosen tolerance in TestHoltWinters.test_holt_damp_fit

ginggs:
I see the following on s390x architecture in Ubuntu 21.04 development:
Arrays are not almost equal to 1 decimals
ACTUAL: 6082.159328812731
DESIRED: 6082.0


bashtage:
Probably OK.  We don't really target s390x, and now that Travis is getting hard to use, it will be nearly impossible to add.

bashtage:
Thanks.

PR #7150, opened 2020-11-10
CLN: common base class for statespace and ets

### LINKED ISSUE #6881: CLN: remove base class of new ETS implementation
The base class of the `ETSModel` class introduced in #6699 is in large parts only a copy of `statespace.mlemodel.MLEModel`, omitting the methods that are unnecessary for `ETSModel`.

It would probably be better to create a common base class for `ETSModel` and `MLEModel`.

s-scherrer:
- [x] closes #6881
- [x] tests added / passed. 

This summarizes all common methods of the statespace methods and the new ETS implementation in a single base class.


ChadFulton:
Thanks for following up on this @s-scherrer. This is a good idea, and I'll review as soon as I can. I anticipate that this will lead to some small issues popping up (as always happens when things change), but this is a good step forward, so I think that it will be good to merge this sooner, and then we can fix anything that does come up.

PR #7240, opened 2021-01-04
Clarify seasonal_decompose docs

jseabold:
Noticed a thinko in this old function then clarified the language a bit.


bashtage:
Thanks.

PR #7243, opened 2021-01-06
ENH: Add small sample df adjustment option to test_serial_correlation

miro-mlynarik:
The sample code below demonstrates the discrepancy of statsmodels output of Ljung-Box serial correlation test p-value due to faulty degrees of freedom used in the test. The correct version should include subtraction of estimated model's degrees of freedom. This change will significantly align the output with R's Arima() output given by checkresiduals(). The remaining differences are probably due to different state space representation used and other minor technical differences.

**Statsmodels version**: 0.12.1
```
import pandas as pd
import numpy as np
import statsmodels.tsa.api as tsa
import rpy2
import warnings
from scipy.stats import boxcox
from statsmodels.tsa.arima.model import ARIMA
warnings.filterwarnings("ignore")
%load_ext rpy2.ipython

passengers = pd.read_csv('https://raw.githubusercontent.com/jbrownlee/Datasets/master/airline-passengers.csv', 
                         index_col=0, header=0, names=['month', 'observed_raw'])
passengers.index = pd.date_range("1949-01", freq="MS", periods=144)
passengers['observed_transformed'], lmbda = boxcox(passengers['observed_raw'])

model = tsa.STLForecast(
    endog=passengers['observed_transformed'], model=ARIMA, seasonal=33, 
    model_kwargs={'order': (2,1,1), 'trend': 'n', 'freq': passengers.index.freq}, period=12
).fit()

residual_acf = pd.DataFrame({
    'residual_acf': tsa.acf(model.model_result.filter_results.standardized_forecasts_error[0], 
                            adjusted=False, nlags=10, qstat=True, fft=False, alpha=0.05)[0][1:], 
    'q_statistic': model.model_result.test_serial_correlation(method='ljungbox', lags=10)[0][0],
    'p_value': model.model_result.test_serial_correlation(method='ljungbox', lags=10)[0][1]}, index=np.arange(1, 11)).T

pasengers_seasonally_adjusted = model.result.resid + model.result.trend  
residual_acf
```
The R output:
```
%%R -i pasengers_seasonally_adjusted -w 600 -h 600
library(fpp2)
library(forecast)
fit <- Arima(pasengers_seasonally_adjusted, order=c(2,1,1), include.constant=FALSE)
checkresiduals(fit)
fit 
```
**R output**:
image

**Original statsmodels output**:
image

**Fixed statsmodels output**:
image

### DIFF
--- a/statsmodels/tsa/statespace/mlemodel.py
+++ b/statsmodels/tsa/statespace/mlemodel.py
@@ -3146,7 +3146,7 @@ class MLEResults(tsbase.TimeSeriesModelResults):
 
         return output
 
-    def test_serial_correlation(self, method, lags=None):
+    def test_serial_correlation(self, method, df_adjust=False, lags=None):
         """
         Ljung-Box test for no serial correlation of standardized residuals
 
@@ -3168,7 +3168,13 @@ class MLEResults(tsbase.TimeSeriesModelResults):
             After 0.12 the default maxlag will change to min(10, nobs // 5) for
             non-seasonal models and min(2*m, nobs // 5) for seasonal time
             series where m is the seasonal period.
-
+        df_adjust : bool, optional
+            If True, the degrees of freedom consumed by the model is subtracted
+            from the degrees-of-freedom used in the test so that the adjusted
+            dof for the statistics are lags - model_df. In an ARMA model, this
+            value is usually p+q where p is the AR order and q is the MA order.
+            When using df_adjust, it is not possible to use tests based on
+            fewer than model_df lags.
         Returns
         -------
         output : ndarray
@@ -3225,11 +3231,15 @@ class MLEResults(tsbase.TimeSeriesModelResults):
                     FutureWarning
                 )
 
+            model_df = 0
+            if df_adjust:
+                model_df = max(0, self.df_model - self.k_diffuse_states - 1)
+
             for i in range(self.model.k_endog):
                 results = acorr_ljungbox(
                     self.filter_results.standardized_forecasts_error[i][d:],
                     lags=lags, boxpierce=(method == 'boxpierce'),
-                    return_df=False)
+                    model_df=model_df, return_df=False)
                 if method == 'ljungbox':
                     output.append(results[0:2])
                 else:
--- a/statsmodels/tsa/statespace/tests/test_mlemodel.py
+++ b/statsmodels/tsa/statespace/tests/test_mlemodel.py
@@ -851,6 +851,27 @@ def test_diagnostics():
         res.test_serial_correlation(method='boxpierce')
 
 
+def test_small_sample_serial_correlation_test():
+    # Test the Ljung Box serial correlation test for small samples with df
+    # adjustment using the Nile dataset. Ljung-Box statistic and p-value
+    # are compared to R's Arima() and checkresiduals() functions in forecast
+    # package:
+    # library(forecast)
+    # fit <- Arima(y, order=c(1,0,1), include.constant=FALSE)
+    # checkresiduals(fit, lag=10)
+    from statsmodels.tsa.statespace.sarimax import SARIMAX
+    niledata = nile.data.load_pandas().data
+    niledata.index = pd.date_range('1871-01-01', '1970-01-01', freq='AS')
+    mod = SARIMAX(
+        endog=niledata['volume'], order=(1, 0, 1), trend='n',
+        freq=niledata.index.freq)
+    res = mod.fit()
+
+    actual = res.test_serial_correlation(
+        method='ljungbox', lags=10, df_adjust=True)[0, :, -1]
+    assert_allclose(actual, [14.116, 0.0788], atol=1e-3)
+
+
 def test_diagnostics_nile_eviews():
     # Test the diagnostic tests using the Nile dataset. Results are from
     # "Fitting State Space Models with EViews" (Van den Bossche 2011,


statsmodels/tsa/statespace/mlemodel.py
```diff
@@ -3229,7 +3229,7 @@ def test_serial_correlation(self, method, lags=None):
                 results = acorr_ljungbox(
                     self.filter_results.standardized_forecasts_error[i][d:],
                     lags=lags, boxpierce=(method == 'boxpierce'),
-                    model_df=self.df_model-1, return_df=False)
+                    model_df=max(self.df_model-1, 0), return_df=False)
```
bashtage, line 3232: This shouldn't be hardcoded.  It would need to be added as an option for test_serial correlation. Also, isn't df_model always at least 1?  I don't exactly see why this is needed.

miro-mlynarik: Because state space model test failed after the first commit. The model had 0 df_model obviously.:/

bashtage: In either case this cannot become the default option, and so must be explicitly passed by an end-user as an option to `test_serial_correlation`.

miro-mlynarik: So that means that even though the model instance contains the information on `df_model`, it nevertheless needs to be manually input by the end user?

bashtage: It can be automatic, but must be off by default.  Somehing like `df_adjust=False` as an additional keyword-only argument. If `df_adjust` is `False` then use `0`.  If `df_adjust` is `True`, then use an adjustment.  

I'm also not really sure if `df_model - 1` is always the correct choice.  Presumably, the `-1` is due to an assumption that the model contains an intercept-like term.  What if this isn't satisfied?

bashtage:
FWIW There is nothing "true" about the R implementation of this test.  The justification is asymptotic and so theoretically it makes no difference whether the df is adjusted or not. In moderately large samples there will be little difference between the current implementation and R.

statsmodels/tsa/statespace/mlemodel.py
miro-mlynarik: Now I understand, thanks, I will modify the solution. Regarding the -1 problem, as far as I can see my provided example, it is functional even though it does not include an intercept (`trend='n'`), so I assume the -1 is due to residual standard error estimated coefficient.

miro-mlynarik:
@bashtage Okay, the asymptotical property might be present, but frankly, I can see quite a big value in having the outputs from the Python library aligned with the R's most used time series function. It makes the whole work of the data scientists more predictable and transparent over time. In addition, what is in your opinion the moderately large sample? In my sample case there are 144 observations and the result with and without the df adjustment is opposite (significance/non-significance) and I don't think the 150 observations is that little amount. In any way, I think this little change can make the users option to use their preferred solution even for not that big samples.

statsmodels/tsa/statespace/mlemodel.py
miro-mlynarik: @bashtage I've updated the code according to your suggestion. If anything is not yet in order, please let me know.

statsmodels/tsa/statespace/mlemodel.py
```diff
@@ -3168,7 +3168,11 @@ def test_serial_correlation(self, method, lags=None):
             After 0.12 the default maxlag will change to min(10, nobs // 5) for
             non-seasonal models and min(2*m, nobs // 5) for seasonal time
             series where m is the seasonal period.
-
+        df_adjust : bool, optional
+            If True, the degrees of freedom consumed by the model is subtracted
+            from the degrees-of-freedom used in the test so that the adjusted
+            dof for the statistics are lags - model_df. In an ARMA model, this
+            value is usually p+q where p is the AR order and q is the MA order.
```
bashtage, line 3175: Should add the text:

When using df_adjust, it is not possible to use tests based on fewer than model_df lags. 

(This is to explain to users why they will get NaN values).

statsmodels/tsa/statespace/mlemodel.py
```diff
@@ -3225,11 +3229,16 @@ def test_serial_correlation(self, method, lags=None):
                     FutureWarning
                 )
 
+            if df_adjust:
+                df_model = self.df_model - 1
```
bashtage, line 3233: can you call this model_df to align with name in function.

statsmodels/tsa/statespace/mlemodel.py
```diff
@@ -3225,11 +3229,16 @@ def test_serial_correlation(self, method, lags=None):
...
+            else:
```
bashtage, line 3234: Skip the else by initially setting model_df = 0, and then only have the if clause.

bashtage:
The main issue is that you need to add a test that has df_adjustment=True.

statsmodels/tsa/statespace/mlemodel.py
miro-mlynarik: Done.

statsmodels/tsa/statespace/mlemodel.py
miro-mlynarik: Done

statsmodels/tsa/statespace/mlemodel.py
miro-mlynarik: Done.

bashtage:
LGTM.  Just needs a test.

Also, can you rebase on upstream/master? Not sure why but Azure isn't reporting.

ChadFulton:
Thanks @miro-mlynarik and @bashtage for noting this and getting the PR together!

Other than the unit test that @bashtage mentions, I think the other thing is to make sure that the degrees of freedom that we're using will be correct for all state space models.

The `df_model` attribute is currently set for use in the penalty term for information criteria, and it is of the form `k_params + k_diffuse_states` (where `k_params` includes the scale term, regardless of whether it is concentrated out of the likelihood or not).  This is appropriate because the initial value of any diffuse states is essentially an unknown (see e.g. Durbin and Koopman, section 7.4).

However, Harvey (1989) (section 5.4.2) and Commaneur and Koopman (2007) (section 8.5) both recommend setting the small-sample-adjusted degrees of freedom for the Ljung-Box test to be the number of estimated parameters excluding the scale term, and do not include the number of diffuse states.

So I think we probably should do:

```python
if df_adjust:
    df_model = max(0, self.df_model - self.k_diffuse_states - 1)
```

bashtage:
What @ChadFulton says seems right.  Basically you only really want p+q.  Not sure what to do about models with exogenous regressors either.  Seasonality could also be challenging to get right as well.  

Does this fix applied to SARIMAX as well as ARIMA?

ChadFulton:
> Does this fix applied to SARIMAX as well as ARIMA?

Currently, it's being put in `MLEModel`, so the new argument would be available for all state space models.

miro-mlynarik:
Thanks for the comments. Yes, I'm currently preparing the unit test. 

> Not sure what to do about models with exogenous regressors either. Seasonality could also be challenging to get right as well.
Good point, I have not thought about this option. I will check these generalizations in the R example and try to find a solution incorporating it. 

> Currently, it's being put in MLEResults, so the new argument would be available for all state space models.
Yes, I was thinking the same (even though I'm not an expert on this codebase as I've seen it for the first time only last week:)

ChadFulton:
> I was just wondering why `self.df_model` is zero for custom state space models

`self.df_model` is based on the number of parameters, but there's no way to tell for us to tell in general what the number of parameters in any custom state space model is. If the model is set up as we usually do it, then the size of the `params` vector yields the number of parameters.

> (see test_diagnostics_nile_eviews) (which is how I found about the negative problem)

That model was just set up in an ad hoc way to generate the test, but indeed as it is written there are no "parameters", since we just put values directly into the system matrices.  Rewriting it to handle parameters better is just a little more work:

```python
niledata = nile.data.load_pandas().data
niledata.index = pd.date_range('1871-01-01', '1970-01-01', freq='AS')

class TestModel(MLEModel):
    def __init__(self, endog):
        super().__init__(endog, k_states=1,
                         initialization='approximate_diffuse',
                         initial_variance=1e15, loglikelihood_burn=1)

        self.ssm['design', 0, 0] = 1
        self.ssm['transition', 0, 0] = 1
        self.ssm['selection', 0, 0] = 1

    def update(self, params, **kwargs):
        params = super().update(params, **kwargs)
        self.ssm['obs_cov', 0, 0] = params[0]
        self.ssm['state_cov', 0, 0] = params[1]

mod = TestModel(niledata['volume'])
res = mod.filter(np.exp([9.600350, 7.348705]))
```

Also, when this test was written, we didn't support exact diffuse initialization, but we do now. So you could use `super().__init__(endog, k_states=1, initialization='diffuse')` instead, which would probably help with testing that the `k_diffuse_states` part is dropping out correctly.

miro-mlynarik:
> > I was just wondering why `self.df_model` is zero for custom state space models
> 
> `self.df_model` is based on the number of parameters, but there's no way to tell for us to tell in general what the number of parameters in any custom state space model is. If the model is set up as we usually do it, then the size of the `params` vector yields the number of parameters.
> 
> > (see test_diagnostics_nile_eviews) (which is how I found about the negative problem)
> 
> That model was just set up in an ad hoc way to generate the test, but indeed as it is written there are no "parameters", since we just put values directly into the system matrices. Rewriting it to handle parameters better is just a little more work:
> 
> ```python
> niledata = nile.data.load_pandas().data
> niledata.index = pd.date_range('1871-01-01', '1970-01-01', freq='AS')
> 
> class TestModel(MLEModel):
>     def __init__(self, endog):
>         super().__init__(endog, k_states=1,
>                          initialization='approximate_diffuse',
>                          initial_variance=1e15, loglikelihood_burn=1)
> 
>         self.ssm['design', 0, 0] = 1
>         self.ssm['transition', 0, 0] = 1
>         self.ssm['selection', 0, 0] = 1
> 
>     def update(self, params, **kwargs):
>         params = super().update(params, **kwargs)
>         self.ssm['obs_cov', 0, 0] = params[0]
>         self.ssm['state_cov', 0, 0] = params[1]
> 
> mod = TestModel(niledata['volume'])
> res = mod.filter(np.exp([9.600350, 7.348705]))
> ```
> 
> Also, when this test was written, we didn't support exact diffuse initialization, but we do now. So you could use `super().__init__(endog, k_states=1, initialization='diffuse')` instead, which would probably help with testing that the `k_diffuse_states` part is dropping out correctly.

Hi, thank you for the sample code, but I must admit I don't actually have practical experience with state space models so I don't feel confident to write tests on this topic. If you consider it necessary, could you please add it on your own? 

Just FYI, in my last commit, I have:
- added a simple ARIMA model unit test which could cover basic usage, 
- added your updated version of `model_df` formula 
- evaluated whether there's a need to change the formula due to seasonal or exogenous variables. As I've found here [seasonal](https://otexts.com/fpp2/seasonal-arima.html) and here [exog](https://otexts.com/fpp2/regarima.html), in both cases the df should be counted so I let the formula as it was.

miro-mlynarik:
> LGTM. Just needs a test.
> 
> Also, can you rebase on upstream/master? Not sure why but Azure isn't reporting.

Please check if I've rebased it correctly.

miro-mlynarik:
@ChadFulton I wanted to finalize the unit test (and fix the import error which occurred) in a standard way by testing it first locally, but I'm receiving errors I don't understand. I suspect they might be related to specific configuration of this repo. Can you please check the message and maybe guess what could go wrong? I'm really helpless here as my personal repo does not have this problem. Thank you very much. 
image

bashtage (approved):
LGTM.  Still not sure that the DoF change is always right, but it should be in most cases that don't have exogenous regressors.

bashtage:
I'll leave it a bit to see if @ChadFulton has any thoughts.

bashtage:
Oh, can you square/rebase to a single commit (or maybe two)?

ChadFulton:
> LGTM. Still not sure that the DoF change is always right, but it should be in most cases that don't have exogenous regressors.

I agree that I am not certain about how many cases this d.of. fits, although since it is optional and in any case the distribution of the statistic seems to be an approximation in most cases I agree that it seems like a good addition.

> Oh, can you square/rebase to a single commit (or maybe two)?

I agree this would be good. I see there are several merge commits in the list of commits, but it doesn't seem problematic in the files changed.

ChadFulton:
> Hi, @ChadFulton, thank you for the sample code, but I must admit I don't actually have practical experience with state space models so I don't feel confident to write tests on this topic. I'm not sure if you consider it really necessary, but if so, could you please add it on your own (or post here the code so I can copy-paste it on my own)?

Sure, I'll follow up on this separately.

Thanks again for the PR!

miro-mlynarik:
> Oh, can you square/rebase to a single commit (or maybe two)?

Hi, I tried to execute `git rebase -i master` on my feature branch and squash all but two commits into one, but it kind of ended up differently as I intended. Maybe the wrong command was used or maybe in this state of the branch it cannot be easily squashed, given there are already two merge commits. Would you or @ChadFulton please be so kind and sort it out? Thanks. In the future PR's I will try to be more cautious about pushing too many commits and merging them.

bashtage:
It you haven't pushed than you can reset to the version on GitHub.  You can then add upstrea,

git add remote upstream https://github.com/statsmodels/statsmodels.git
git fetch --all
git rebase upstream/master -i

and then squash all but the first.

miro-mlynarik:
> If you haven't pushed than you can reset to the version on GitHub. You can then add upstream.
> 
> git add remote upstream https://github.com/statsmodels/statsmodels.git
> git fetch --all
> git rebase upstream/master -i
> 
> and then squash all but the first.

I've done what you proposed, but it didn't go smoothly. I tried to resolve the conflicts by skipping problematic commits and then pushed the changes to my origin, but I don't know if this is the desired state. If not, I can delete this branch a open a new one with just single commit containing all changes to those two files I've changed a open a new PR with copied content from this one. Sorry for the inconvenience, this is my first actual contribution to github projects.

miro-mlynarik:
> If you haven't pushed than you can reset to the version on GitHub. You can then add upstream.
> 
> git add remote upstream https://github.com/statsmodels/statsmodels.git
> git fetch --all
> git rebase upstream/master -i
> 
> and then squash all but the first.

Thank you. Since I already pushed, I had to also do force push to origin, but I've finally managed to clean the commits section. Thanks for assistance.

ChadFulton:
Looks great, thanks for working through all of this!

PR #7254, opened 2021-01-13
ENH: Copula and joint distribution

### LINKED ISSUE #4046: SUMM/ENH: resurrecting copulas

tupui:
- [ ] closes #4046
- [ ] tests added / passed. 

This PR's intent is to allow the construction of correlated multivariate distributions.


bashtage:
`pdf`, `logpdf`, `cdf`, and `plot` would all be valuable methods for any copula.

statsmodels/distributions/joint_distribution.py
```diff
@@ -0,0 +1,236 @@
...
+      Statistics and Applied Probability 134, 2015.
+
+    """
+
+    @abstractmethod
+    def __init__(self, seed=None):
+        self.rng = np.random.default_rng(seed)
+
+    @abstractmethod
+    def random(self, n):
```
bashtage, line 66: Needs to accept an argument `random_state` that can handle a `random_state` or `BitGenerator` object.

bashtage: This is like what scipy does.  

https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.rv_continuous.rvs.html

Seed shouldn't be part of the class initialization.

statsmodels/distributions/joint_distribution.py
```diff
@@ -0,0 +1,236 @@
...
+        self.density = stats.norm()
+        self.mv_density = stats.multivariate_normal(cov=cov)
+
+    def random(self, n):
+        x = self.mv_density.rvs(size=n, random_state=self.rng)
+        return self.density.cdf(x)
+
+
+class StudentCopula(Copula):
+    """Student copula."""
```
bashtage, line 112: docstrings should have a reasonable amount of mathematicaal detail similar to what scipy does.

tupui: ok I will see to add at least the definition of C. But I am not a mathematician and I would just copy paste. So if you know better, it's better if you do it.

statsmodels/distributions/joint_distribution.py
tupui: ok sure. I was not sure about the architecture here. If you think something else would be better in general let me know.

tupui: For the accepted values, I propose to not talk about `RandomState ` as we do in `stats.qmc` to strive people to not use it. Let me know if I should do otherwise.

statsmodels/distributions/joint_distribution.py
bashtage: > ok I will see to add at least the definition of C. But I am not a mathematician and I would just copy paste. So if you know better, it's better if you do it.

Copy-paste from Wiki is a good start.

statsmodels/distributions/joint_distribution.py
bashtage: It seems fine to me to prefer `Generator` objects.  The problem is that a `RandomState` has to be default until statsmodels requires NumPy 1.17.

tupui: Is there somewhere a function to check the random state? Something like `scipy._lib._util.check_random_state`. I could not find anything like this.

tupui:
> `pdf`, `logpdf`, `cdf`, and `plot` would all be valuable methods for any copula.

I added a plot function to the base class and the structure for the `pdf`, `logpdf`, `cdf`. But I am really not sure about these. I see some existing work in `sandbox/distributions/copula`. Can I use this? @josef-pkt is this correct, or you want to do things differently?

tupui:
@bashtage I added `pdf`, `cdf` to all copulas. Also I added a `plot_pdf` helper (I will overwrite the one for `IndependentCopula` as constant).

statsmodels/distributions/joint_distribution.py
```diff
@@ -0,0 +1,435 @@
...
+            cov = [[1., 0.], [0., 1.]]
+        self.density = stats.t(df=df)
+        self.mv_density = stats.multivariate_t(shape=cov, df=df)
+
+    def random(self, n=1, random_state=None):
+        x = self.mv_density.rvs(size=n, random_state=random_state)
+        return self.density.cdf(x)
+
+    def pdf(self, x):
+        return self.mv_density.pdf(x)
```
josef-pkt, line 252: This is not correct, we need the copula pdf and not the pdf of the multivariate t distribution.
I guess we can implement logpdf (loglike) and get pdf through exp(logpdf)

josef-pkt:
@tupui you need to add references about from where you took the formulas for individual copula pdf, cdf

tupui:
> @tupui you need to add references about from where you took the formulas for individual copula pdf, cdf

Of course sure, I will do this next time. I will soon push a fit function for theta. Still thinking a bit more about the interface before proposing you something.

josef-pkt:
I just found your references in the Copula class.

I will start with the Copula hierarchy, see #7257 now that I have the generator functions fixed and verified.
I will move things out of the sandbox to distributions.copulas and distributions.transforms_copulas, but not today.

The archimedean copulas will subclass the sandbox `CopulaArchimedean`, but the subclasses can override cdf and pdf with specific implementation, and other things like tau that will come later. Several of the extras can be computed generically using the generator transform.
`CopulaArchimedean` can subclass top level Copula class that might only have generic or general methods.

your plot functions for 2-dim. 
some methods and copulas will have restrictions to work only for bivariate cases. 
Extreme value copulas are currently all bivariate and I haven't looked at multivariate extensions.
pdf for multivariate (d>2) EV copulas look difficult in general.

archimedean and elliptical can be multivariate, but I haven't looked at explicit, non-generated cdf and pdf for those.

tupui:
I added more ref about `dev-copulas` in the docstring of the file. Also I adapted the theta from tau functions.

Seeing all your work on sandbox in the other issue, I guess I will wait for you to merge this and I will rebase onto that. If you want me to proceed differently let me know.

statsmodels/distributions/joint_distribution.py
```diff
@@ -0,0 +1,526 @@
...
+        ax.set_xlim(0, 1)
+        ax.set_ylim(0, 1)
+        ax.set_aspect('equal')
+        cbar = plt.colorbar(cs, ticks=vticks)
+        cbar.set_label('p')
+
+        return fig
+
+    @property
+    def tau(self):
```
josef-pkt, line 193: `tau` should be analytical or numerical integration

josef-pkt:
@tupui 

did you look at how to use quasi-random numbers for copula sampling?
e.g. https://cran.r-project.org/web/packages/copula/vignettes/qrng.html

I think in the past I only looked at univariate and multivariate normal quasi-random numbers based on halton sequences, where we only need univariate `ppf`.

tupui:
> @tupui
> 
> did you look at how to use quasi-random numbers for copula sampling?
> e.g. https://cran.r-project.org/web/packages/copula/vignettes/qrng.html
> 
> I think in the past I only looked at univariate and multivariate normal quasi-random numbers based on halton sequences, where we only need univariate `ppf`.

The only thing I've ever used is the `ppf` as what I have in the code here. Do we need to add something else?

Pinging some experts from OpenTURNS who might be interested to help/give a hand: @regislebrun @mbaudin47

regislebrun:
What you need is the conditional quantile function (quantile=ppf). It means, the inverse of the following function:
```
t->P(X_k<=t|X_1=x_1,...,X_{k-1}=x_{k-1})
```
You should have a look [here](https://github.com/openturns/openturns/blob/91e63496348500e775f84292d906cccf97dca044/lib/src/Uncertainty/Model/DistributionImplementation.cxx#L607) for a possible implementation. It is very easy to implement QMC sampling once you have the conditional quantile, which is not so easy to implement efficiently in a generic way. IMO you should consider to rely on OpenTURNS for this kind of computation, but in democracy everybody is free to reinvent the wheel, as we also did in many parts of OpenTURNS ;-)!

tupui:
> you should consider to rely on OpenTURNS for this kind of computation, but in democracy everybody is free to reinvent the wheel, as we also did in many parts of OpenTURNS ;-)!

You know I like the way it's done in OpenTURNS 😉 But the license is not compatible with statsmodels. So if we want to propose something here, it has to be done here (or from an OK license).

statsmodels/distributions/joint_distribution.py
tupui: I will have to do some research here. TBH my knowledge about all this is all here.

josef-pkt:
for elliptical copula we mainly need multivariate normal
multivariate t rvs is a chi2 rvs times multivariate normal.

copulas has some of it
https://github.com/sdv-dev/Copulas/blob/master/copulas/bivariate/base.py#L284
and the sampling function

AFAIR, some copulas have explicit versions.

interface
for sampling we need more random draws than dimensions in some cases.
So we need some way to get them from several halton sequences. 
I never looked at combining halton sequences for e.g. mixture or multi-part distributions.

regislebrun:
@tupui At least you will have a reference implementation to compare to! As the `getSampleByQMC` method is protected, it is not exposed in the Python interface. Here is an equivalent way to perform QMC using OT:
```
import openturns as ot
import openturns.viewer as otv

size = 2048
c = ot.GumbelCopula(5.5)
rng = ot.SobolSequence(2) # could be HaltonSequence, FaureSequence...
sample = [c.computeSequentialConditionalQuantile(q) for q in rng.generate(size)]
cloud = ot.Cloud(sample)
graph = ot.Graph(str(c) + " QMC sampling", r"$u_1$", r"$u_2$", True)
graph.add(cloud)
view = otv.View(graph, square_axes=True)
view.ShowAll()
view.close()
```
It gives you the following result:
Figure_1

josef-pkt:
also to this PR:
many pdf have a product structure. In those cases it is better to implement logpdf and exp them in the pdf instead of the other way. That makes loglikelihood also numerically more stable.

regislebrun:
The main point is that the sequential inversion of the conditional CDF will give you a valid sample of the copula even if it is fed by a low discrepancy sequence (ie the uniform discrete distribution supported by such a sample of size n will tend to the copula distribution when the size tends to infinity), while more evolved methods valid for random sampling may failed when the random generator is replaced by a low discrepancy sequence.

josef-pkt:
@regislebrun 
Do you know how well quasi random number behave when merging sequences?
eg. `rvs_t = rvs_chi2 * rvs_normal`
where we use ppf of chi2 and of normal for the two parts

(I'm not sure anymore whether it's chi or chi2 in this representation)

bashtage:
MV_t is MVN / sqrt(chi2(v)/v)

And the answer is badly for mixing. You really need to quantile function to use QMC values.

tupui:
@josef-pkt If we want to add the option to use QMC for sampling the copulas we might be better of waiting for the merge of `scipy.stats.qmc`. Then we have several QMC engines we can plug into the `self.copula.random` methods.

mbaudin47:
> 
> 
> > you should consider to rely on OpenTURNS for this kind of computation, but in democracy everybody is free to reinvent the wheel, as we also did in many parts of OpenTURNS ;-)!
> 
> You know I like the way it's done in OpenTURNS 😉 But the license is not compatible with statsmodels. So if we want to propose something here, it has to be done here (or from an OK license).

What is the constraint, since OT is LGPL? Would'nt it is be simpler to add OT as a dependence of statmodels?

mbaudin47:
> Do you know how well quasi random number behave when merging sequences?
> eg. `rvs_t = rvs_chi2 * rvs_normal`
> where we use ppf of chi2 and of normal for the two parts
> (I'm not sure anymore whether it's chi or chi2 in this representation)

The jointly bivariate sequence created by joining two QMC univariate sequences is not a bivariate QMC. To see this, remember that each dimension of e.g. the Sobol' sequence has its own primitive polynomial (but all dimensions have common basis 2). For the Halton sequence, this is perhaps clearer, since each dimension has its own prime number. Does this match your question?

josef-pkt:
> Does this match your question?

I haven't looked at halton/sobol sequences in a while, and didn't remember yet the bivariate part, i.e. in the conditional approach we need to create all uniform variables at once.

The more specific question was for mixture distributions. How good is QMC if for example we use the mixture representation for elliptical distributions, even if it's not a proper QMC? like in the multivariate t case above.
In the textbook definition for elliptical distributions we would multiply a radial random variable with a random variable on the unit circle, Similar applications would be scale mixtures of normal distributions.

My question is whether this destroys QMC properties too much, or if this is still useful to do and better than pure random draws.

mbaudin47:
> My question is whether this destroys QMC properties too much, or if this is still useful to do and better than pure random draws.

I am not sure of understanding what "too much" may mean since a QMC is a tms-net. This property ensures that all elementary volumes in base b contain the required number of points. More precisely, it is a set of b^m points such that any elementary interval in base b with volume b^{t-m} contains exactly b^t points. If just one point is not in the elementary interval, this is not a tms-net anymore. The property that is required is that the CDF of the joint random vector is a tms-net. This allows to use the Koksma-Hlawka theorem and get a quadrature rule. To ensure this, I cannot see how we may avoid the conditional quantile function (which is just the quantile function if marginals are independent). Otherwise, I think that nothing can be proved anymore and "convergence" is not guaranteed, isn'it ? Perhaps an experiment would show this more clearly.

josef-pkt:
about adding OpenTURNS as dependency:

At this stage I just want to get the basic copula features into statsmodels, not develop a full copula package or compete with specialized packages.
We are very restrictive in runtime dependencies for statsmodels. So, at this stage OpenTURNS would be too much to add.
We have cvxopt, GPL, as optional dependency as optional alternative to scipy.optimize in a few parts. But I don't think it's maintained much and we ran into version issues in the past.

So, optional dependencies for specific parts are not ruled out, but writing wrappers around other packages is too much work given our limited maintainer time.
E.g. I thought a few times of adding sklearn as dependency for some parts, but, because of the license compatibility, it is easier to copy and refactor code, or to just leave it up to the user to mix and match.

It will become clearer what the requirements for copulas in statsmodels are, when statsmodels itself becomes a user of it. This will happen when we add more non-gaussian, no-linear multivariate models. 
Currently we only have multinomial/ordinal type of multivariate models. The two areas that I was looking at specifically are multivariate extreme value statistics and multivariate survival or failure time models.

@mbaudin47 Thanks for the explanation in your last comment. I never looked at the theory behind QMC.

For most bivariate copulas we should be able to get the conditional cdf/quantile function. I still don't know yet how much we can provide for more than two dimensions.

regislebrun:
I made some experiments with OT, in order to compare different sampling schemes for the multivariate-t distribution. I played with a 2D student-t distribution with nu=5.0 and rho=0.5.

Here is what you get when you use conditional quantiles and a pseudo-random generator:
Student_qmc_False

and if you use the radial/angular decomposition and a pseudo-random generator:
Student_qmc2_False_False

They look the same, as expected.

Now, I use conditional quantiles based on a 2D low-discrepancy sequence (Sobol sequence):
Student_qmc_True

It looks like the quasi-random cloud, but with a more pronounced texture, typical of low discrepancy sequences. If I use the radial/angular decomposition with a 2D low-discrepancy sequence for the angular part (the normal distribution) and a separate 1D low-discrepancy sequence for the radial distribution I get:
Student_qmc2_True_True

It is clear that something is wrong here. If I use instead a 3D low-discrepancy sequence, then extract the two first components to feed the angular part and the third component to feed the radial part I get:
Student_qmc2_True_False

This time it looks right, but the low-discrepancy structure is lost.

It confirms that the use of low discrepancy sequences to sample multivariate distributions is much more evolved than simply replace the pseudo-random generator by a 1D low discrepancy sequence or by independent low discrepancy sequences. Only the conditional quantiles give you the guarantee that things will be correct.

Here is the quick and dirty script I used:
```
import openturns as ot
import openturns.viewer as otv

R = ot.CorrelationMatrix(2, [1.0, 0.5, 0.5, 1.0])
nu = 5.0
student_distribution = ot.Student(nu, [0.0]*2, [1.0]*2, R)
normal_distribution = ot.Normal([0.0]*2, [1.0]*2, R)
radial_distribution = ot.Chi(nu)

size = 4096

for QMC in [False, True]:

    ######################################
    # Sample using conditional quantiles #
    ######################################
    if QMC:
        inSample = ot.SobolSequence(2).generate(size)
    else:
        inSample = ot.ComposedDistribution([ot.Uniform(0.0, 1.0)]*2).getSample(size)
    outSample = ot.Sample([student_distribution.computeSequentialConditionalQuantile(x) for x in inSample])
    # Draw things
    g = ot.Graph()
    g.setAxes(True)
    g.setGrid(True)
    cloud = ot.Cloud(outSample)
    cloud.setPointStyle("dot")
    g.add(cloud)
    g.setBoundingBox(ot.Interval([-10.0]*2, [10.0]*2))
    view = otv.View(g)
    view.save("Student_qmc_" + str(QMC) + ".png")
    view.close()
    #############################################
    # Sample the radial part & the angular part #
    #############################################
    angular = ot.Sample([normal_distribution.computeSequentialConditionalQuantile(x) for x in inSample])
    for indep in [False, True]:
        if QMC:
            if indep:
                inSample1D = ot.SobolSequence(1).generate(size)
            else:
                inSample3D = ot.SobolSequence(3).generate(size)
                inSample = inSample3D.getMarginal([0, 1])
                inSample1D = inSample3D.getMarginal(2)
        else:
            inSample1D = ot.Uniform(0.0, 1.0).getSample(size)
        radial = radial_distribution.computeQuantile(inSample1D.asPoint()).asPoint() * (nu**(-0.5))
        outSample = ot.Sample([angular[i] / radial[i] for i in range(size)])
        # Draw things
        g = ot.Graph()
        g.setAxes(True)
        g.setGrid(True)
        cloud = ot.Cloud(outSample)
        cloud.setPointStyle("dot")
        g.add(cloud)
        g.setBoundingBox(ot.Interval([-10.0]*2, [10.0]*2))
        view = otv.View(g)
        view.save("Student_qmc2_" + str(QMC) + "_" + str(indep) + ".png")
        view.close()
```
Note that most of the lines are dedicated to graphics, and that the combination QMC=False, indep=False and QMC=False, indep=True give the same results (but I was too lazy to think about a clever code ;-))
```

**Comment by tupui** [comment]
@josef-pkt correct me if I am wrong. But the way I see this is that I can restrict the scope of this PR to only have deal with using the ppf to construct a joint distribution. This would rely on the copula defined in either #7292 and #7296. Shall I wait for you to merge one (both?) of these to rely on the copula you have there. Or you want to cherry pick something to put here and we can move on with this one? In the end this would be a small addition to the much larger work on copula.

**Comment by josef-pkt** [comment]
for random numbers, we will need a `method` argument or separate methods in the class, 
one for standard rvs with any available standard random number generation, and 
one that only uses ppf with or without quasi-random numbers.

E.g. multivariate t copula is easy with standard random numbers, but it requires work with quasi-random numbers to get the sequence of conditional univariate distributions.

Another point is to have unit tests for the random numbers.
We will not be able to verify that we have correct random numbers without some kind of goodness-of-fit measure or test. 
I started with grid computing cell probabilies for a 2-d or n-d grid.

I get easily distracted, but I should merge my moving PR soon.

**Comment by mbaudin47** [comment]
> You know I like the way it's done in OpenTURNS wink But the license is not compatible with statsmodels. So if we want to propose something here, it has to be done here (or from an OK license).

Thank you very much for this information. The license issue is crucial, but is off-topic. I posted a message on the statsmodels ML to clarify this particular point.

**Comment by tupui** [comment]
> for random numbers, we will need a `method` argument or separate methods in the class, 
> 
> one for standard rvs with any available standard random number generation, and 
> 
> one that only uses ppf with or without quasi-random numbers.
> 
> 
> 
> E.g. multivariate t copula is easy with standard random numbers, but it requires work with quasi-random numbers to get the sequence of conditional univariate distributions.
> 
> 
> 
> Another point is to have unit tests for the random numbers.
> 
> We will not be able to verify that we have correct random numbers without some kind of goodness-of-fit measure or test. 
> 
> I started with grid computing cell probabilies for a 2-d or n-d grid.
> 
> 
> 
> I get easily distracted, but I should merge my moving PR soon.
> 
> 

Ok sure no problem. I just want to be aligned with what you want and know what I can do 😀

**Comment by josef-pkt** [comment]
I haven't written a single line of code for elliptical copulas yet.
It would be great if you get pdf and rvs for multivariate normal and t copulas to work.
:)

**Comment by tupui** [comment]
> I haven't written a single line of code for elliptical copulas yet.
> It would be great if you get pdf and rvs for multivariate normal and t copulas to work.
> :)

Are you sure these are not correct? Because this is what is done in Copulas: https://github.com/sdv-dev/Copulas/blob/fa67784dd59d9d7154146b13e055cee877e75c4e/copulas/multivariate/gaussian.py#L121

**Comment by josef-pkt** [comment]
you have `self.mv_density.pdf(x)` which cannot work because x is uniform in the copula and the domain of the pdf is the real line (R^k), so this needs at least x=ppf(u)
I didn't look carefully at gaussian copula, but pdf for the t-copula also has an extra division in the text books versions.

Notation: I switched to `u` in copulas instead of `x`, so it's more obvious that it is a random variable with  uniform margin.
bivariate copulas often use u, v

**Comment by tupui** [comment]
> you have `self.mv_density.pdf(x)` which cannot work because x is uniform in the copula and the domain of the pdf is the real line (R^k), so this needs at least x=ppf(u)
> I didn't look carefully at gaussian copula, but pdf for the t-copula also has an extra division in the text books versions.
> 
> Notation: I switched to `u` in copulas instead of `x`, so it's more obvious that it is a random variable with uniform margin.
> bivariate copulas often use u, v

Oh yes indeed! Sorry I am just realizing what you mean. I will switch for u as well then. Thanks!

**Comment by tupui** [comment]
This looks better now  @josef-pkt 

Screenshot 2021-02-08 at 11 45 58

**Comment by josef-pkt** [comment]
needs unit test to verify, eg. against R `copula` package (that's what I used)

**Comment by tupui** [comment]
> needs unit test to verify, eg. against R `copula` package (that's what I used)

I found this: https://firsttimeprogrammer.blogspot.com/2016/03/how-to-fit-copula-model-in-r-heavily.html
Will get to this when I got time. I never touch at R... 😅

**Comment by josef-pkt** [comment]
you can also write unit tests with comparison numbers from OpenTURNS

**Comment by tupui** [comment]
@josef-pkt I added tests for the copulas. Everything is the same as with R now (fixed a few bugs and rechecked some equations from _H. Joe (2015), Dependence Modeling with Copulas_ (chapter 4 has everything).

I just commented out the CDF of student-t. I failed to use what you have in sandbox.

**Comment by regislebrun** [comment]
Just in case, if you want to build a Student copula in OT (as there is no dedicated class):
```
import openturns as ot
student_copula = ot.Student(5.0, [0.0]*2, [1.0]*2, ot.CorrelationMatrix(2,[1,0.5,0.5,1])).getCopula()
```
and then you can call its many methods.

**Comment by tupui** [comment]
@josef-pkt I merged master here. Now, how do you want to proceed? 

I see you have added an `Archimedian` class and `FrankCopula`. Shall I remove all the copula code I have and only have the `JointDistribution`? I am not sure here as it seems we took a different approach to write the equations.

I could also move the plot functions to the Archimedian class.

Let me know how I should evolve this. Also, feel free to commit on the branch directly 😃

**Comment by josef-pkt** [comment]
For archimedean I have mainly generic code based on generator functions with FrankCopula as explicit case.
We can add the main other copulas as explicit cases, cdf and pdf are available generically but can be overridden if there are simple expressions for it.
I didn't add any `rvs`, we can have the specific code as much as available. I didn't look into details for a generic archimedean rvs yet. The specific classes should also have explicit statistics like kendalltau or spearmanrho, but we can add the generic code that requires numerical integration, IIRC.

`elliptical` is empty in my merged code. You can pretty much copy what you have for t and gaussian to that module

The base `copulas` module needs a base class Copula which can have the plot methods. I have not added the base class.

I think we add independence copula to `other_copulas` plus some parametric ones that are not ev, archimedean or elliptical, like plackett.


For later: 
I'm not sure yet about nonparametric copulas, if and where we add those. 
There are also copula transformation like the power transformation that is in a basic function still in `copulas`. Those might also go into a separate module.

The `api.py` module should be the main access for users that has all the copula classes and functions in one path.

Design for args/params handling is still a bit undecided. We will need them as method arguments, when we are fitting distributions or models, but I didn't check the interface yet.

File: statsmodels/distributions/tests/test_joint_distribution.py [hunk 5]
```diff
@@ -0,0 +1,256 @@
...
+    [1] "Gumbel"
+     [1]  1.0391696  0.6539579  0.9878446  0.8679504 16.6030932  0.7542073
+     [7]  0.6668307  0.6275887  0.7477991  1.1564864
+     [1] 0.27194634 0.05484380 0.15668190 0.37098420 0.98176346 0.23422865
+     [7] 0.05188260 0.08659615 0.33086960 0.15803914
+
+
+
+
+"""
```
josef-pkt, line 131: put the above test in a separate file.
we can add a tests/results directory to store R scripts, explanations and result numbers. The latter if they don't fit easily inside a test class or function

tupui:
@josef-pkt I updated the PR. I moved the copulas in the corresponding files and slightly changed the inheritance of some classes.

I think this is a good start and we can cleanup things now.

statsmodels/distributions/copula/archimedean.py
```diff
@@ -72,12 +77,12 @@ def logpdf(self, u, args=()):
         return logpdfv
 
 
-class FrankCopula(ArchimedeanCopula):
+class FrankCopulaGeneratorBased(ArchimedeanCopula):
```
josef-pkt, line 80: No need to rename.
All archimedean copulas need to call super or add the transform.
We need the transform in the classes to have some things work generically

tupui: Ah ok I understand! I will merge both. And add the transform call to every class.

tupui:
@josef-pkt I am not a fan of passing the `theta` as args to all functions. I prefer to have it in the init or it should be explicit at least. I refactored in that direction.

I did not really look yet at `CopulaDistribution` and `CopulaDistributionBivariate` as here I am not sure what you want to do.

tupui:
@josef-pkt let me know if I can do something else. In any case, feel free to commit on the branch directly.

josef-pkt:
unit test fail
- matplotlib needs to be optional, we have a helper function for plot methods
- we need args in copula methods. We cannot require theta or other copula parameters in `__init__`  or we would have to create a new instance each time when we want to `fit` to estimate copula params, e.g. by MLE.

josef-pkt:
it's mainly the last commit, that changed things too much from my pattern
I need to look more closely to what we still want.

josef-pkt:
@tupui  from where did you get the default values for archimedean theta, e.g. `theta=2`?

I think there shouldn't be any default values.
We might want to use `theta=None` in `__init__` as sentinel for not having a fixed parameter.
In archimedean we could take specifying theta in init as fixing the parameter. In EV copulas we have several parameters and we would need a different pattern to fix just some of those parameters, e.g. fixing asymmetry parameters.

The removal of args in archimedean and in unit tests can be reversed by amending last commit.

josef-pkt:
still open question: args tuple versus single args, consistency versus convenience ?

specific archimedean classes don't have uniy test coverage, and according to codecov, test coverage for pdf, logpdf in Frank copula would be removed.

tupui:
@josef-pkt I can add back `*args` but IMO we should not. It's an anti-pattern and obfuscate code inspection. I am always dying when I have to do some matplotlib for this reason. Using unknown parameters is a famous complain about APIs on StackOverflow, famous python podcasts, etc. We should try to design an interface which does not require the use of `*args`, at least not on the public API used by final users. Plus, for now we just have `theta`. Even if we would need to add 2 or even 3 other parameters it would be cleaner than `*args`.

For the default I took an arbitrary default value. So you want the code to raise if we don't specify `theta`? I am not a fan of being able to instantiate a class which is in a "broken" state if you don't do some specific actions before. I would propose instead to make `theta` mandatory and for the case of `theta` fitted by MLE (or anything else) we can use a `classmethod`. So something like `Copula.from_mle(...)`.

josef-pkt:
We need arguments in the functions, otherwise it's messy to create a model that use copulas where the copula parameters are estimated.
My initial pattern was like the scipy stats distribution with (almost) stateless classes, i.e. each method needs all arguments. We can add args in `__init__` so we have a combination of frozen/semi-frozen and no-parameter instances.

My guess is that copulas are in many cases not the relevant class for user. Copula distributions and copula (regression) models.

Named parameters are not generic enough for library use. We need a simple common pattern for args/params, so we can build additional features that do not rely on the parameterization of specific copulas.
(I never ever used the named shape parameters in scipy.stats distributions)

The details are still open, we don't have enough models to have much experience with many different parameters, except for packing them in `params` which only works if we have one parameter value and not vectorized.

E.g. it's more important that we are able to write a multivariate extreme value model based on generic copula interface, than having a userfriendly but inconsistent interface to copulas.

We can allow scalar args in the case of copulas with only a single parameter, but the problem if we go away from tuples is unpacking if there are too many options for how to specify the parameters, and interpretation of broadcasting/shape if we have array parameters.

tupui:
OK then your call. I will re-introduce the `*args` parameter.

TBH I am not a fan of how `scipy.stats.distribution` works in general because of its two ways of working. It is confusing for the user. I mean, the whole pattern here is breaking 2 "recommendations" from the zen of python: `Explicit is better than implicit.` and `There should be one-- and preferably only one --obvious way to do it.`

josef-pkt:
> There should be one-- and preferably only one --obvious way to do it

statsmodels is built on dual use, library and interactive, so we have two obvious ways, one for each use case :)
api, formula, pandas are alternatives to plain numpy and direct imports.

(Pandas was for a long time focused on interactive use and a pain in library use. Most complaints about tidy verse are that it's a pain for code development/maintenance and long term use. The basic structure of statsmodels has lasted for over 10 year and still works well.)

josef-pkt:
possible interface for args:

if args is None and self.args is not None:
    args = self.args
elif not isinsatance(args, tuple):
    args = (args,)

The first part will need to be made more general when only some, but not all parameters of copula are fixed, needed for EV copulas

For elliptical copulas we have two possible cases 
- cov/corr is specified/unrestricted d * (d - 1) parameters, but usually given as cov matrix
- patterned cov/corr: e.g. banded, equal correlation, ... as in GEE options, and we have only a smaller set of parameters, that could be estimated by MLE.

Using full MLE for elliptical cov will not be coming soon, because there are too many tools missing to estimate patterned cov matrices outside of GEE.

tupui:
> if args is None and self.args is not None:
> 
>     args = self.args
> 
> elif not isinsatance(args, tuple):
> 
>     args = (args,)

You mean this is to be used for the methods right? If so I propose to have a common validation function for it so we avoid duplication.

> The first part will need to be made more general when only some, but not all parameters of copula are fixed, needed for EV copulas

No sure what this means but I will implement the validation part and you can add stuff there.

> - patterned cov/corr: e.g. banded, equal correlation, ... as in GEE options, and we have only a smaller set of parameters, that could be estimated by MLE.

I don't have this I guess as for now I coded something which needs the matrix. Or I misunderstand what you mean.

josef-pkt:
to the last point for elliptical cov:
With bivariate we only have one correlation, with many components in the multivariate case, cov will be large and have many parameters. If there is some structure like autocorrelation or equal correlation, then we can estimate a smaller set of parameters and construct the full cov based on those.
We don't have this yet for MLE models, but when we have it, then it will be useful to add the option to copula based models.

about partially fixed params:

The asymmetric EV copulas can have two asymmetry and one correlation parameter. When used for the standard symmetric/exchangable case, then the asymmetry parameters can be fixed to some values or restricted so that both asymmetry parameters are the same. So, partial fixing becomes necessary in fitting when we have a smaller number of free parameters.
If parameters are already given, then it doesn't matter (all params are fixed).

tupui:
@josef-pkt I put back `args`. There is just one test failing now. Not sure.

We still need to decide what to do about `CopulaDistribution`, `Copula` and `CopulaDistributionBivariate` (not used).
Most copulas subclass `Copula` (the general class I added). We could merge `CopulaDistribution` and `Copula` if needed. For the other class, not sure we need it.

Let me know what you think about all that.

tupui:
@josef-pkt friendly reminder. Let me know how you want to proceed 😃

josef-pkt:
Leave it open for now, we close it if and when the new PR gets merged.
I still need to compare some things.

I'm trying to write an introductory overview notebook to see how well things work.

tupui:
> I'm trying to write an introductory overview notebook to see how well things work.

If that can help, I had things like these in a notebook.

```python
import numpy as np
from scipy import stats
import seaborn as sns

from statsmodels.sandbox.distributions.copula import *
from statsmodels.distributions.joint_distribution import *

copula = GumbelCopula(2)
sample = copula.random(10000)

# some plot
h = sns.jointplot(sample[:, 0], sample[:, 1], kind='hex')
h.set_axis_labels('X1', 'X2', fontsize=16)


# another one
copula = ClaytonCopula(0.3)
_ = copula.plot_pdf(10)

```

josef-pkt:
do you have your notebook in a gist or somewhere online?

josef-pkt:
I have mostly notebooks for extreme value copulas from when I was working on those, and bernstein copula.

tupui:
I will close my PR in favor of #7408 as its now merged.

PR #7326, opened 2021-02-15
BUG: Fix libsturng issue #7324

### LINKED ISSUE #7324: libqsturng _psturng fails when degrees of freedom is between 2 and 2.5 and fminbound chooses a p value between 0.9 and 0.9175
#### Describe the bug

This bug is related, but different from the bug I first reported [last February](https://github.com/statsmodels/statsmodels/issues/6541). And might be easier to solve.  When the degrees of freedom (v) is >= 2 for and q not in one of the two boundary cases given in `psturng` we get to [line 839](https://github.com/statsmodels/statsmodels/blob/4e77b4b07545f848fe001c24197cf49214a8c2d1/statsmodels/stats/libqsturng/qsturng_.py#L839) and run `fminbound` on `optfunc` which computes `_qsturng`  for p from 0.1 to 0.999.  Depending on `v` and `p` the code needs to interpolate in both dimensions.  The problem is that the way the sets for `_select_vs` and `_select_ps` are lined up, when `2 < v < 2.5` and `0.9 < p < 0.9175`.  In this case, `_select_vs` returns the tuple `1, 2, 3` and `_select_ps` returns `0.850, 0.9, 0.95`.  But the pair `(0.85, 1)` is not in the table --- which it is not expected to be as previous noted, the papers only support `p >= 0.9` for `v < 2`.  

The easiest solution (which also seems correct) seems to be to shift [line 500](https://github.com/statsmodels/statsmodels/blob/4e77b4b07545f848fe001c24197cf49214a8c2d1/statsmodels/stats/libqsturng/qsturng_.py#L839) to be `elif p>= 0.900`.  This would jibe with the previous lines and mathematically, I see no reason why not to do this and then would jibe with the `v's` selected.  

Note the statsmodels code has a print value in it that actually helps.  From that we can see we have the case where `p` is just above `0.9` and with `v < 2.5` then as noted above, the code tries to look up `(0.85, 1)`, which as noted, is not in the list.  This is fixed by not having it look that up and keep interpolation of values for `p > 0.9` within p's select all over 0.9. 

```python
from statsmodels.stats.libqsturng import psturng
psturng(12.5872, 24, 2.02)
```
<details>

As the person who posted the other bug, I know this hasn't come up before and I know it happens both on release and master.  

</details>

#### Expected Output

I expected this call to complete with a value that is in between the values we get if we have fewer groups or more (e.g. 14 and 36) which shift the values of `p` tested in the `fminbound` function to fit within the current options without hitting this bug.  

<details>

INSTALLED VERSIONS
------------------
Python: 3.7.7.final.0
OS: Darwin 18.7.0 Darwin Kernel Version 18.7.0: Tue Nov 10 00:07:31 PST 2020; root:xnu-4903.278.51~1/RELEASE_X86_64 x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8

statsmodels
===========

Installed: 0.12.1 (/Users/ataylor/.pyenv/versions/3.7.7/envs/gameshowell_test/lib/python3.7/site-packages/statsmodels)

Required Dependencies
=====================

cython: Not installed
numpy: 1.18.4 (/Users/ataylor/.pyenv/versions/3.7.7/envs/gameshowell_test/lib/python3.7/site-packages/numpy)
scipy: 1.5.4 (/Users/ataylor/.pyenv/versions/3.7.7/envs/gameshowell_test/lib/python3.7/site-packages/scipy)
pandas: 1.1.4 (/Users/ataylor/.pyenv/versions/3.7.7/envs/gameshowell_test/lib/python3.7/site-packages/pandas)
    dateutil: 2.8.1 (/Users/ataylor/.pyenv/versions/3.7.7/envs/gameshowell_test/lib/python3.7/site-packages/dateutil)
patsy: 0.5.1 (/Users/ataylor/.pyenv/versions/3.7.7/envs/gameshowell_test/lib/python3.7/site-packages/patsy)

Optional Dependencies
=====================

matplotlib: Not installed
cvxopt: Not installed
joblib: 1.0.0 (/Users/ataylor/.pyenv/versions/3.7.7/envs/gameshowell_test/lib/python3.7/site-packages/joblib)

Developer Tools
================

IPython: 7.20.0 (/Users/ataylor/.pyenv/versions/3.7.7/envs/gameshowell_test/lib/python3.7/site-packages/IPython)
    jinja2: 2.11.2 (/Users/ataylor/.pyenv/versions/3.7.7/envs/gameshowell_test/lib/python3.7/site-packages/jinja2)
sphinx: Not installed
    pygments: 2.7.4 (/Users/ataylor/.pyenv/versions/3.7.7/envs/gameshowell_test/lib/python3.7/site-packages/pygments)
pytest: 6.1.2 (/Users/ataylor/.pyenv/versions/3.7.7/envs/gameshowell_test/lib/python3.7/site-packages/pytest)
virtualenv: 20.2.1 (/Users/ataylor/.pyenv/versions/3.7.7/envs/gameshowell_test/lib/python3.7/site-packages/virtualenv)

</details>

ameliajtaylor:
- [X] closes #7324
- [X] tests added / passed. 

Implements the bug fix proposed in #7324 and adds a test that will cause a failure on master and tests the change we made against a value from R.

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>


bashtage:
Thanks

PR #7329, opened 2021-02-17
Added the option `full_matrices=False` in the PCA method

omartrinidad:
- [x] Related to #7042
- [x] tests added / passed.


bashtage:
LGMT. Thanks for this.

PR #7355, opened 2021-02-27
Changed endog and exog if given any fixed parameters

madhushree14:
- [ ] closes #xxxx
- [ ] tests added / passed. 

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>


ChadFulton:
Thanks for posting this @madhushree14!

To better understand what you'v done, it would be helpful if you could post an example that includes:

1. How to call the function with the new argument `fixed_params`
2. The output of the function when calling it with `fixed_params`
3. How the output differs from calling it without `fixed_params`.

ChadFulton:
Thanks for your work on this.  How are things going on it?  Have you gotten to the point where you can give an example?  Do you think you'll have availability to work on this and, if not, do you have any objections to @jackzyliu moving this forward instead?

madhushree14:
Hi Chad, yes @jackzyliu can work further on this...

PR #7356, opened 2021-02-28
ENH: moved core logic of test_heteroskedasticity method of state space model to a new function

### LINKED ISSUE #7348: ENH: separate the core logic of test_heteroskedasticity in state space model as a function

[`test_heteroskedasticity`](https://www.statsmodels.org/stable/generated/statsmodels.tsa.statespace.mlemodel.MLEResults.test_heteroskedasticity.html) method of state space model (MLEModel) is not flexible.
- You cannot change the data range to apply the test (see also #7342 )
    - length of diffuse periods (d) is fixed
- You cannot apply it for other type of time series models.
    - it is only for state space models.

On the other hand, I don't feel frustration with other two diagnostic test methods in MLEModel, [test_serial_correlation](https://www.statsmodels.org/stable/generated/statsmodels.tsa.statespace.mlemodel.MLEResults.test_serial_correlation.html) and [test_normality](https://www.statsmodels.org/stable/generated/statsmodels.tsa.statespace.mlemodel.MLEResults.test_normality.html), 
because we can just use their core logic functions directly,  [`stats.diagnostic.acorr_ljungbox`](https://www.statsmodels.org/stable/generated/statsmodels.stats.diagnostic.acorr_ljungbox.html) and [`stats.stattools.jarque_bera`](https://www.statsmodels.org/stable/generated/statsmodels.stats.stattools.jarque_bera.html) respectively.

Create a new function and move [core logic of the `test_heteroskedasticity` method](https://github.com/statsmodels/statsmodels/blob/ad4b474b712517515400b0447998c6771897e02d/statsmodels/tsa/statespace/mlemodel.py#L3088-L3144) into it.
Then, you can both call this function in `test_heteroskedasticity` method and use it as an independent function for other time series models.

My idea:
`stats.diagnostic.het_timeseries_residual_breakvar(resid, alternatives="two-sided", use_f=True)`

Parameters

- resid: ndarray
residual of a time series model
- *alternative: str, ‘increasing’, ‘decreasing’ or ‘two-sided’
- *use_f: bool, optional

Returns
- *output: ndarray
an array with (test_statistic, pvalue)

marked with `*` ... same as in the [original method](https://www.statsmodels.org/stable/generated/statsmodels.tsa.statespace.mlemodel.MLEResults.test_heteroskedasticity.html)

Add optional arguments to the current `test_heteroskedasticity` method to change the diffuse periods.
In this case, however, you cannot apply the test to other time series models.

#### Additional context

Another merit of separation of the logic is that we can easily add new options of test logics in `test_heteroskedasticity` method in the future.

tatamiya:
I made a new function to test heteroskedasticity of time series residuals and 
replaced the core logic of `test_heteroskedasticity` method in state space model
by the new function.

- [ ] closes #7348 
- [ ] tests added / passed. 

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>

### DIFF
--- a/statsmodels/tsa/statespace/mlemodel.py
+++ b/statsmodels/tsa/statespace/mlemodel.py
@@ -27,6 +27,7 @@ import statsmodels.tsa.base.prediction as pred
 
 from statsmodels.base.data import PandasData
 import statsmodels.tsa.base.tsa_model as tsbase
+from statsmodels.tsa.stattools import breakvar_heteroskedasticity_test
 
 from .news import NewsResults
 from .simulation_smoother import SimulationSmoother
@@ -3077,71 +3078,20 @@ class MLEResults(tsbase.TimeSeriesModelResults):
 
         if method == 'breakvar':
             # Store some values
-            squared_resid = self.filter_results.standardized_forecasts_error**2
+            resid = self.filter_results.standardized_forecasts_error
             d = np.maximum(self.loglikelihood_burn, self.nobs_diffuse)
             # This differs from self.nobs_effective because here we want to
             # exclude exact diffuse periods, whereas self.nobs_effective only
             # excludes explicitly burned (usually approximate diffuse) periods.
             nobs_effective = self.nobs - d
+            h = int(np.round(nobs_effective / 3))
 
-            test_statistics = []
-            p_values = []
-            for i in range(self.model.k_endog):
-                h = int(np.round(nobs_effective / 3))
-                numer_resid = squared_resid[i, -h:]
-                numer_resid = numer_resid[~np.isnan(numer_resid)]
-                numer_dof = len(numer_resid)
-
-                denom_resid = squared_resid[i, d:d+h]
-                denom_resid = denom_resid[~np.isnan(denom_resid)]
-                denom_dof = len(denom_resid)
-
-                if numer_dof < 2:
-                    warnings.warn('Early subset of data for variable %d'
-                                  '  has too few non-missing observations to'
-                                  ' calculate test statistic.' % i)
-                    numer_resid = np.nan
-                if denom_dof < 2:
-                    warnings.warn('Later subset of data for variable %d'
-                                  '  has too few non-missing observations to'
-                                  ' calculate test statistic.' % i)
-                    denom_resid = np.nan
-
-                test_statistic = np.sum(numer_resid) / np.sum(denom_resid)
-
-                # Setup functions to calculate the p-values
-                if use_f:
-                    from scipy.stats import f
-                    pval_lower = lambda test_statistics: f.cdf(  # noqa:E731
-                        test_statistics, numer_dof, denom_dof)
-                    pval_upper = lambda test_statistics: f.sf(  # noqa:E731
-                        test_statistics, numer_dof, denom_dof)
-                else:
-                    from scipy.stats import chi2
-                    pval_lower = lambda test_statistics: chi2.cdf(  # noqa:E731
-                        numer_dof * test_statistics, denom_dof)
-                    pval_upper = lambda test_statistics: chi2.sf(  # noqa:E731
-                        numer_dof * test_statistics, denom_dof)
-
-                # Calculate the one- or two-sided p-values
-                alternative = alternative.lower()
-                if alternative in ['i', 'inc', 'increasing']:
-                    p_value = pval_upper(test_statistic)
-                elif alternative in ['d', 'dec', 'decreasing']:
-                    test_statistic = 1. / test_statistic
-                    p_value = pval_upper(test_statistic)
-                elif alternative in ['2', '2-sided', 'two-sided']:
-                    p_value = 2 * np.minimum(
-                        pval_lower(test_statistic),
-                        pval_upper(test_statistic)
-                    )
-                else:
-                    raise ValueError('Invalid alternative.')
-
-                test_statistics.append(test_statistic)
-                p_values.append(p_value)
-
-            output = np.c_[test_statistics, p_values]
+            output = breakvar_heteroskedasticity_test(
+                resid[:, d:],
+                subset_length=h,
+                alternative=alternative,
+                use_f=use_f
+                )
         else:
             raise NotImplementedError('Invalid heteroskedasticity test'
                                       ' method.')
--- a/statsmodels/tsa/stattools.py
+++ b/statsmodels/tsa/stattools.py
@@ -1295,6 +1295,149 @@ def levinson_durbin_pacf(pacf, nlags=None):
     return arcoefs, acf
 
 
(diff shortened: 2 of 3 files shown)


tatamiya:
This is now WIP because I'd like to 
- modify the docs of original `test_heteroskedasticity` method to reduce redundancy
- add tests for 
   - chi^2 test
   - alternative="increasing" / "decreasing" cases
   - multivariate input

statsmodels/tsa/stattools.py
```diff
@@ -1295,6 +1295,149 @@ def levinson_durbin_pacf(pacf, nlags=None):
...
+
+    Tests whether the sum-of-squares in the first subset of the sample is
+    significantly different than the sum-of-squares in the last subset
+    of the sample. Analogous to a Goldfeld-Quandt test. The null hypothesis
+    is of no heteroskedasticity.
+
+    Parameters
+    ----------
+    resid : array_like
+        Residuals of a time series model.
```
ChadFulton, line 1311: Can you add a note about the expected shape of `resid`?  (see below for a suggestion to use `(nobs, nvars)` rather than `(nvars, nobs)`)

statsmodels/tsa/stattools.py
```diff
@@ -1295,6 +1295,149 @@ def levinson_durbin_pacf(pacf, nlags=None):
...
+
+    See section 5.4 of [1]_ for the above formula and discussion, as well
+    as additional details.
+
+    References
+    ----------
+    .. [1] Harvey, Andrew C. 1990. *Forecasting, Structural Time Series*
+            *Models and the Kalman Filter.* Cambridge University Press.
+    """
+    squared_resid = np.atleast_2d(resid) ** 2
```
ChadFulton, line 1374: We typically don't have spaces around the `**` operator.

statsmodels/tsa/statespace/mlemodel.py
```diff
@@ -3077,71 +3078,20 @@ def test_heteroskedasticity(self, method, alternative='two-sided',
...
-                    )
-                else:
-                    raise ValueError('Invalid alternative.')
-
-                test_statistics.append(test_statistic)
-                p_values.append(p_value)
-
-            output = np.c_[test_statistics, p_values]
+            output = breakvar_heteroskedasticity_test(
+                resid[:, d:],
```
ChadFulton, line 3090: Will need a transpose after changing `breakvar_heteroskedasticity_test` to accept `(nobs, nvars)` arrays.

statsmodels/tsa/stattools.py
```diff
@@ -1295,6 +1295,149 @@ def levinson_durbin_pacf(pacf, nlags=None):
...
+    nvars, nobs = squared_resid.shape
```
ChadFulton, line 1375: Since this is now a general test outside of `statespace`, I will suggest that this should be `(nobs, nvars)` rather than `(nvars, nobs)`. Then you can pass `

ChadFulton (commented):
This looks great, thanks very much.  In addition to a few minor changes, can you please add the new function as an import in the file `stats/api.py`? This will make it available when doing:

```
import statsmodels.api as sm
sm.stats.breakvar_heteroskedasticity_test(...)
```

ChadFulton:
>  add tests for
> - [ ] chi^2 test
> - [ ] alternative="increasing" / "decreasing" cases
> - [ ] multivariate input

Can you elaborate on these? I think it already supports the chi^2 version, increasing/decreasing hypotheses, and multivariate input?

tatamiya:
Thank you for your comments!

> >add tests for
 chi^2 test
 alternative="increasing" / "decreasing" cases
 multivariate input

> Can you elaborate on these? I think it already supports the chi^2 version, increasing/decreasing hypotheses, and multivariate input?

Yes, as you say, the function already supports them
but they are not covered by tests.
I consider it would be safer to add tests to guarantee that
the code runs as we expect.

statsmodels/tsa/tests/test_stattools.py
```diff
@@ -336,6 +337,23 @@ def test_ld(self):
         assert_almost_equal(pacfyw, pacfld, DECIMAL_8)
 
 
+class TestBreakvarHeteroskedasticity(object):
```
bashtage, line 340: There is no need to use a class.  This should be a single function.  Classes are mostly vestiges of using nose/unittest.

bashtage:
Would be good to add the test to the docs.

statsmodels/tsa/stattools.py
tatamiya: I found it would be better to allow only 1-d input for this function and iterate over the `model.k_endog` in `test_heteroskedasticity` method as `test_normality` and `test_serial_correlation` do.

bashtage: Then `(nobs,)` would be the correct description.

statsmodels/tsa/tests/test_stattools.py
tatamiya: I used a class because I wrote more than one tests for one function.
Here, we can run tests in a class by a single command: 
```console
pytest tests/test_stattools::TestBreakvarHeteroskedasticity
```

statsmodels/tsa/stattools.py
tatamiya: Now, 2-d input is allowed by small modification.
I will add a note as `(nobs,) or (nobs, nvars)`

ChadFulton: > I will add a note as (nobs,) or (nobs, nvars)

This looks like it still needs to be done.

ChadFulton (commented):
This looks great, thank you! Just two things:

- It would be great to add the dimension to the `resid` argument docstring, as you suggested
- @bashtage, was the argument for keeping the test class persuasive to you (i.e. that it allowed running all tests using `pytest tests/test_stattools::TestBreakvarHeteroskedasticity`)? If it was not, then I would ask @tatamiya to refactor the test into functions.

I look forward to merging this after these two issues are completed. Thanks again!

statsmodels/tsa/stattools.py
```diff
@@ -1295,6 +1295,147 @@ def levinson_durbin_pacf(pacf, nlags=None):
...
+    Tests whether the sum-of-squares in the first subset of the sample is
+    significantly different than the sum-of-squares in the last subset
+    of the sample. Analogous to a Goldfeld-Quandt test. The null hypothesis
+    is of no heteroskedasticity.
+
+    Parameters
+    ----------
+    resid : array_like
+        Residuals of a time series model.
+        The shape is 1d (nobs,) or 2d (nobs, nvars).
```
tatamiya, line 1312: @ChadFulton Here I added the expected shape of input.

PR #7366, opened 2021-03-08
DOC: fix misleading ETSModel example

### LINKED ISSUE #7365: DOC: fix misleading ETSModel implementation of simple exponential smoothing example
#### Describe the bug

In the [example page of using `ETSModel`](https://www.statsmodels.org/dev/examples/notebooks/generated/ets.html#Simple-exponential-smoothing), the simple exponential smoothing model is actually a Holt's model (it includes an additive trend).

#### Suggested fix

This is the current "simple exponential smoothing" model:

```python
model = ETSModel(oil, error='add', trend='add', damped_trend=True)
```
The above includes an additive (damped) trend which effectively makes it a Holt's model. For a simple exponential smoothing model, there should be no trend I believe, so the code should simply be:

```python
model = ETSModel(oil, error='add')
```

This change also ensures that the results match those from the R `forecast` package (the plot in the current page is not identical):

plot

#### Next steps

I'm happy to make a PR with this small fix. As always, thanks to all for their hard work in developing, maintaining and improving this package.

TomasBeuzen:
- [x] closes #7365 
- [x] documentation is well formatted.  

### DIFF
--- a/examples/notebooks/ets.ipynb
+++ b/examples/notebooks/ets.ipynb
@@ -105,16 +105,16 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "model = ETSModel(oil, error='add', trend='add', damped_trend=True)\n",
+    "model = ETSModel(oil)\n",
     "fit = model.fit(maxiter=10000)\n",
-    "oil.plot(label='data')\n",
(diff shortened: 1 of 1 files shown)


bashtage:
Thanks.

PR #7379, opened 2021-03-13
BUG: `du` calculation in `qic` should have both means and endog sorted in the same order

### LINKED ISSUE #7376: BUG: QIC calculation in GEE is wrong?
I was reading through the QIC code for GEE [here](https://www.statsmodels.org/stable/_modules/statsmodels/genmod/generalized_estimating_equations.html#GEE.qic) and was wondering if `du = means - self.endog` should be `du = means - endog_li` instead. The `means` arrays generated by the below code is sorted by groups, shouldn't the endogenous variable be sorted by groups too?

```
  means = []
  omega = 0.0
  # omega^-1 is the model-based covariance assuming independence

  for i in range(self.num_group):
      expval, lpr = self.cached_means[i]
      means.append(expval)
      dmat = self.mean_deriv(self.exog_li[i], lpr)
      omega += np.dot(dmat.T, dmat) / scale

  means = np.concatenate(means)
```

If that's right then the corrected version should have

```
endog_li = np.concatenate(self.endog_li)
du = means - endog_li
nstep = 10000
qv = np.empty(nstep)
xv = np.linspace(-0.99999, 1, nstep)
for i, g in enumerate(xv):
    u = endog_li + (g + 1) * du / 2.0
    vu = varfunc(u)
    qv[i] = -np.sum(du**2 * (g + 1) / vu)
qv /= (4 * scale)
```

instead of

```
du = means - self.endog
nstep = 10000
qv = np.empty(nstep)
xv = np.linspace(-0.99999, 1, nstep)
for i, g in enumerate(xv):
    u = self.endog + (g + 1) * du / 2.0
    vu = varfunc(u)
    qv[i] = -np.sum(du**2 * (g + 1) / vu)
qv /= (4 * scale)
```

sanjay-rendu:
…ting du

- [ ] closes #7376 
- [ ] tests added / passed. 

<details>

**Notes**:

* Replaced `self.endog` with `np.concatenate(self.endog_li)` to make sure means are subtracted from the corresponding endogenous variable value

### DIFF
--- a/statsmodels/genmod/generalized_estimating_equations.py
+++ b/statsmodels/genmod/generalized_estimating_equations.py
@@ -1721,12 +1721,13 @@ class GEE(GLM):
 
         # The quasi-likelihood, use change of variables so the integration is
         # from -1 to 1.
-        du = means - self.endog
+        endog_li = np.concatenate(self.endog_li)
+        du = means - endog_li
         nstep = 10000
         qv = np.empty(nstep)
         xv = np.linspace(-0.99999, 1, nstep)
         for i, g in enumerate(xv):
-            u = self.endog + (g + 1) * du / 2.0
+            u = endog_li + (g + 1) * du / 2.0
             vu = varfunc(u)
             qv[i] = -np.sum(du**2 * (g + 1) / vu)
         qv /= (4 * scale)


kshedden:
Thanks, I think you are right.  I will merge and then add a test.  I was thinking that we should have an option where if the data are pre-sorted by group, then we can do everything through views.  But that can come later.

PR #7394, opened 2021-03-31
BUG: fix wrong axis label in qqplot_2samples

Jiacheng-Liu:
The label location is wrong in qqplot_2samples. When data1 and data2 are of equal sizes, x_label will correspond to data2, which is wrong, (it should correspond to data1). 

BTW, why permitting different data sizes here? To me, it doesn't make sense for a Q-Q plot to have different sample sizes on data1 and data2.

Not sure how to add a test for plot, but this code can produce results needed:

d1 = np.arange(3)
d2 = np.arange(0,30,10)
pp_x = sm.ProbPlot(d1)
pp_y = sm.ProbPlot(d2)
qqplot_2samples(pp_x, pp_y,xlabel="d1")
new_qqplot_2samples(pp_x, pp_y,xlabel="d1")

- [ ] closes #xxxx
- [x] tests added / passed. 

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>

### DIFF
--- a/statsmodels/graphics/gofplots.py
+++ b/statsmodels/graphics/gofplots.py
@@ -770,7 +770,7 @@ def qqplot_2samples(
         data2 = ProbPlot(data2)
     if data2.data.shape[0] >= data1.data.shape[0]:
         fig = data1.qqplot(
-            xlabel=xlabel, ylabel=ylabel, line=line, other=data2, ax=ax
+            xlabel=ylabel, ylabel=xlabel, line=line, other=data2, ax=ax
         )
     else:
         fig = data2.qqplot(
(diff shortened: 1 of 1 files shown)


bashtage:
> BTW, why permitting different data sizes here? To me, it doesn't make sense for a Q-Q plot to have different sample sizes on data1 and data2.

It just interpolates between points.  In principle it should work well except in the extreme tails.

bashtage:
The fix needs to be a bit deeper I think.

Jiacheng-Liu:
@bashtage, I think I know what caused this bug: qqplot plot by default plot the data on y-axis, but qqplot_2sample want to plot data1 to x-axis, causing a conflict. the changes proposed will fix this error.

bashtage:
I incorporated your branch and made some deeper changes in #7413.  Thanks.

PR #7397, opened 2021-04-02
BUG: Fix float index usage in IRF error bands

jseabold:
- [x] tests added / passed. 

Looking over #6769 I noticed that the IRF error bands don't run with default arguments because using floats as indices is an error now. Added smoke tests for this part of the code at least...


bashtage:
LGMT. Thanks.

PR #7422, opened 2021-04-22
BUG: Runs test numeric cutoff error

rnmourao:
- [ ] closes #xxxx
- [X] tests added / passed. 

When you try to execute the `runstest_1samp` with the parameter `cutoff` set as a number, it raises a `TypeError`:

> TypeError: '>=' not supported between instances of 'list' and 'int'

I added a conversion to `np.float64` when the `cutoff` is a number.

Here is a snippet that shows the problem:

```{python}
from statsmodels.sandbox.stats.runs import runstest_1samp
import numpy as np


x = [1] * 5 + [2] * 6 + [3] * 8
cutoff = 2
runstest_1samp(x, cutoff=cutoff, correction=False)
```

I didn't add the bugfix correction in the **release notes** because I didn't find a file for `statsmodels==0.12.2`.

**Notes**:

  needed for doc changes.
  ```
  ```

</details>

### DIFF
--- a/statsmodels/sandbox/stats/runs.py
+++ b/statsmodels/sandbox/stats/runs.py
@@ -132,6 +132,8 @@ def runstest_1samp(x, cutoff='mean', correction=True):
         cutoff = np.mean(x)
     elif cutoff == 'median':
         cutoff = np.median(x)
+    else:
+        cutoff = np.float64(cutoff)
     xindicator = (x >= cutoff).astype(int)
     return Runs(xindicator).runs_test(correction=correction)
 
--- a/statsmodels/sandbox/stats/runs.py
+++ b/statsmodels/sandbox/stats/runs.py
@@ -23,6 +23,7 @@ import numpy as np
 from scipy import stats
 from scipy.special import comb
 import warnings
+from statsmodels.tools.validation import array_like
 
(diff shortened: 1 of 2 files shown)


statsmodels/sandbox/stats/runs.py
```diff
@@ -132,6 +132,8 @@ def runstest_1samp(x, cutoff='mean', correction=True):
         cutoff = np.mean(x)
     elif cutoff == 'median':
         cutoff = np.median(x)
+    else:
```
bashtage, line 135: Could you add an `x = array_like(x,"x")` at the start of the function

statsmodels/sandbox/stats/runs.py
```diff
@@ -132,6 +132,8 @@ def runstest_1samp(x, cutoff='mean', correction=True):
...
+        cutoff = np.float64(cutoff)
```
bashtage, line 136: If x is an array, then cutoff can simply be `float(cutoff)`

bashtage:
Thanks.  I'll merge once green.

bashtage:
Thanks.

rnmourao:
> Thanks.

Don't mention it.

PR #7435, opened 2021-05-05
fix typo in ets error

ltsaprounis:
Small typo fix for the error raised when endog is not strictly positive and multiplicative components are chosen.

from: 
```python
ValueError: endog must be strictly positive when usingmultiplicative error, trend or seasonal components.
```

to:
```python
ValueError: endog must be strictly positive when using multiplicative error, trend or seasonal components.
```


bashtage:
Thanks.

PR #7450, opened 2021-05-17
Correct small typo in Theta model Notebook

juanitorduz:
This is just a small typo correction on a notebook example. No code component was modified.


ChadFulton:
Thanks!

PR #7458, opened 2021-05-21
DOC: typo, plats->plots

gdmcbain:
Just a little glitch I noticed while reading https://www.statsmodels.org/dev/examples/notebooks/generated/autoregressions.html.


bashtage:
Thanks

PR #7462, opened 2021-05-24
BUG: Prevent indent running on None

### LINKED ISSUE #7460: is class DocString modifying c-extension `tp_doc` attributes after PyTypeReady?
https://github.com/statsmodels/statsmodels/blob/77bb1d276c7d11bc8657497b4307aa7575c3e65c/statsmodels/tools/docstring.py#L534

It seems the DocString class is modifying `tp_doc` strings after `PyTypeReady` has been called, which does not work on PyPy. It works only by chance on CPython. Is it possible to modify this code to do that manipulation in a precompile step, write the docstrings out to a header file, and read them in before instantiating the type?

xref [PyPy issue 3471](https://foss.heptapod.net/pypy/pypy/-/issues/3471). Another option would be to (like NumPy) be aware that the `__doc__` attribute may be empty for optimized python (when using `python -OO`) or for PyPy.

bashtage:
Use non-protected indent to prevent indent on None

- [X] closes #7460 

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>


josef-pkt:
Why did you delete all the gee simulation files?

bashtage:
Accident -- they are all broken and when I was trying to recursively import all modules, these were raising.  I meant to move them to a sub-folder where they were no longer part of a module, but I think it doesn't really matter now.

stonebig:
Will there be a statsmodels-0.12.3 patch release in June ?

bashtage:
No. Next release is 0.13 which will likely be late summer.

Issue #6159, opened 2019-09-11
ENH: Improvements to new ARIMA-type estimators


ChadFulton:
Collection of follow-ups to #5827. These can/should be broken out into individual PRs. Many are relatively straightforward and would make a good first PR.

### GLS

- Add support for fixed parameters
- Improve "Returns" documentation for `other_results`.
- Add documentation for why we have e.g. `include_constant` but not other trend specifications (e.g. that it is to maintain consistency with estimation methods with assumptions that require demeaned series).
- Fix the following test and put it back into the GLS test suite:

```python
@pytest.mark.low_precision('Test against Example 6.6.3 in Brockwell and Davis'
                           ' (2016)')
# @pytest.mark.xfail(reason="Source appears to find suboptimal parameters")
def test_brockwell_davis_example_663():
    # TODO: the parameters described by BD appear to be suboptimal (based on
    # llf computed from state space form), so that this test fails. Should try
    # to confirm with ITSM2000 (i.e. see if we can get it to find better
    # parameters closer to what we find, or compare llf, or something).
    # TODO: quite a slow test, and xfail anyway due to finding better
    # parameters...
    # Get the data, perform seasonal differencing
    endog = sbl.diff(12).iloc[12:]

    exog = pd.Series((sbl.index > '1983-01-01').astype(int),
                     index=sbl.index).diff(12).iloc[12:]

    res, _ = gls(endog, exog, order=(0, 0, 12), max_iter=3)

    assert_allclose(res.exog_params, -328.45, atol=1e-2)
    assert_allclose(res.ma_params,
                    [.219, .098, .031, .064, .069, .111, .081,
                     .057, .092, -0.28, .183, -.672], atol=1e-3)
    assert_allclose(res.sigma2, 12581, atol=1)
```

### Hannan Rissanen

- Better warnings / errors when series are short relative to lag length
- Add support for fixed parameters
- Seems like we could add support for seasonal parameters in this model.

### Innovations algorithm

- Add support for ARMA models; see [1]_, p.154 and Example 5.1.6.  This estimator should be feasible given the Cython versions of the general innovations algorithm that introduced in #5947.

rajathpatel23:
@ChadFulton I would like to work on this enhancement, would like to request some help in guiding to start working on this enhancement.

Thank you

ChadFulton:
@rajathpatel23 that would be great, thanks!

There are a number of things here that I think should be pretty self-contained, including:

[...]

PR #7530, opened 2021-07-04
ENH: Add support for fixed parameters to SARIMAXSpecification and SARIMAXParams (GH7529)

### LINKED ISSUE #7529: ENH: Add support for fixed parameters to `SARIMAXSpecification` and `SARIMAXParams`
xref: #6159

-  `SARIMAXSpecification`:
   - move `hannan_rissanen._validate_fixed_params` to `SARIMAXSpecification.validate_fixed_params` 
   - add `SARIMAXSpecification.split_fixed_params` (similar to the existing `SARIMAXSpecification.split_params`)
- `SARIMAXParams`:
   - add function `SARIMAXParams.set_fixed_params` so that we can set fixed parameters as follows: 
   - implement **getters** such as `.is_exog_param_fixed`, `.is_ar_param_fixed`, ..., `.fixed_exog_params`, `.fixed_ar_params`, ...
   - implement free param **getters/setters** like `.free_exog_params`, `.free_ar_params`, ... 
- `hannen_rissanen`: applying these changes to vastly simplify support for fixed parameters in `hannen_rissanen`

For example, we could set and access fixed parameters as follows:
```
In[32]: from statsmodels.tsa.arima import specification, params
In[33]: spec = specification.SARIMAXSpecification(ar_order=2)
In[34]: fixed_params = {'ar.L1': 1}
In[35]: spec.validate_fixed_params(fixed_params)
In[36]: 
In[36]: p = params.SARIMAXParams(spec=spec)
In[37]: p.set_fixed_params(fixed_params)
In[38]: print(p.ar_names)
['ar.L1', 'ar.L2']
In[39]: print(p.ar_params)
[ 1. nan]
In[40]: print(p.is_ar_param_fixed)
[ True False]
In[41]: print(p.fixed_ar_params)
[1.]
In[42]: p.free_ar_params = [10]
In[43]: print(p.ar_params)
[ 1. 10.]
```

- Alternative to `set_fixed_params`: we could instead implement additional **setters** for `.is_exog_param_fixed`, `.fixed_exog_params`, etc., but 1) fixed parameters are usually passed around in the sparse form as a dictionary, and 2) this approach would unnecessarily increase the number of "functions" for `SARIMAXParams`

jackzyliu:
- [x] closes #7529 
- [x] xref #6159
- [x] tests added / passed. 

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>


jackzyliu:
Hi @ChadFulton, I was in the process of adding support for fixed params to GLS and thought it'd be much cleaner to add fixed-param support to SARIMAXSpecification and SARIMAXParams first like we previously discussed. 

This PR might be a bit dense, but most of the code changes come from either tests and identically structured param getters/setters in `SARIMAXParams`. Also, so for easier code review, I separated the changes into sequential but mostly self-contained commits, corresponding to the following components:

1. Changes to `SARIMAXSpecification`:
   - move `hannan_rissanen._validate_fixed_params` to `SARIMAXSpecification.validate_fixed_params` 
   - add `SARIMAXSpecification.split_fixed_params` (similar to the existing `SARIMAXSpecification.split_params`)
2. Changes to `SARIMAXParams`:
   - add function `SARIMAXParams.set_fixed_params` so that we can set fixed parameters as follows: 
   - implement _getters_ such as `.is_exog_param_fixed`, `.is_ar_param_fixed`, ..., `.fixed_exog_params`, `.fixed_ar_params`, ...
   - implement free param _getters/setters_ like `.free_exog_params`, `.free_ar_params`, ... 
3. Changes to `hannan_rissanen`: applying these changes to vastly simplify support for fixed parameters in `hannan_rissanen`

For example, we can now set and access fixed parameters as follows:
```
In[32]: from statsmodels.tsa.arima import specification, params
In[33]: spec = specification.SARIMAXSpecification(ar_order=2)
In[34]: fixed_params = {'ar.L1': 1}
In[35]: spec.validate_fixed_params(fixed_params)
In[36]: 
In[36]: p = params.SARIMAXParams(spec=spec)
In[37]: p.set_fixed_params(fixed_params)
In[38]: print(p.ar_names)
['ar.L1', 'ar.L2']
In[39]: print(p.ar_params)
[ 1. nan]
In[40]: print(p.is_ar_param_fixed)
[ True False]
In[41]: print(p.fixed_ar_params)
[1.]
In[42]: p.free_ar_params = [10]
In[43]: print(p.ar_params)
[ 1. 10.]
```

statsmodels/tsa/arima/params.py
```diff
@@ -394,3 +698,88 @@ def __repr__(self):
         if not self.spec.concentrate_scale:
             components.append('sigma2=%s' % self.sigma2)
         return 'SARIMAXParams(%s)' % ', '.join(components)
+
+    def _set_param_array(self, param_type, value, k_params, allow_infnan=False,
```
jackzyliu, line 702: Factored out this helper function from original param setters to handle setting partial param array with the `bool_indexer` arg. All free or full param setters use this help function except for `.params` and `.free_params`.

statsmodels/tsa/arima/params.py
```diff
@@ -351,6 +575,86 @@ def is_invertible(self):
...
+
+        for param_type in param_types:
+            # write to param_split
+            self._set_param_array(
+                param_type=param_type, value=split_fixed_params[param_type],
+                k_params=split_is_fixed_param[param_type].sum(),
+                title=None, bool_indexer=split_is_fixed_param[param_type]
+            )
+            # update is_param_fixed_split
+            self._is_param_fixed_split[param_type] = (
```
jackzyliu, line 611: Allows overwriting already fixed parameters to be consistent with other existing implementations, in which case a warning would be raised (see `_warn_fixed_overwrite` below).

statsmodels/tsa/arima/estimators/hannan_rissanen.py
```diff
@@ -130,15 +133,14 @@ def hannan_rissanen(endog, ar_order=0, ma_order=0, demean=True,
...
+        X_with_free_params = lagged_endog[:, free_ar_ix]
+        X_with_fixed_params = lagged_endog[:, fixed_ar_ix]
         y = endog[max_ar_order:]
         if X_with_fixed_params.shape[1] != 0:
-            y = y - X_with_fixed_params.dot(params_info.fixed_ar_params)
+            y = y - X_with_fixed_params.dot(p.fixed_ar_params)
 
         # no free ar params -> variance computation on the endog residual
         if X_with_free_params.shape[1] == 0:
-            p.ar_params = params_info.fixed_ar_params
```
jackzyliu, line 141: No need to set any fixed params again after calling `set_fixed_params` at the beginning once.

statsmodels/tsa/arima/estimators/hannan_rissanen.py
```diff
@@ -148,13 +150,7 @@ def hannan_rissanen(endog, ar_order=0, ma_order=0, demean=True,
...
             resid = res.resid
             p.sigma2 = res.scale
-            p.ar_params = _stitch_fixed_and_free_params(
-                fixed_ar_or_ma_lags=params_info.fixed_ar_lags,
-                fixed_ar_or_ma_params=params_info.fixed_ar_params,
-                free_ar_or_ma_lags=params_info.free_ar_lags,
-                free_ar_or_ma_params=res.params,
-                spec_ar_or_ma_lags=spec.ar_lags
-            )
+            p.free_ar_params = res.params
```
jackzyliu, line 153: We can now easily set free params like this. Fixed and free params can later together be accessed with, for example, `.ar_params`, so there is no need for this stitching function.

statsmodels/tsa/arima/estimators/hannan_rissanen.py
```diff
@@ -297,133 +278,3 @@ def hannan_rissanen(endog, ar_order=0, ma_order=0, demean=True,
         'resid': resid
     })
     return p, other_results
-
-
-def _validate_fixed_params(fixed_params, spec_param_names):
```
jackzyliu, line 302: Migrated to `SARIMAXSpecification`

statsmodels/tsa/arima/estimators/hannan_rissanen.py
```diff
@@ -297,133 +278,3 @@ def hannan_rissanen(endog, ar_order=0, ma_order=0, demean=True,
...
-        free_ar_lags=free_ar_lags, free_ma_lags=free_ma_lags,
-        # ixs
-        fixed_ar_ix=fixed_ar_ix, fixed_ma_ix=fixed_ma_ix,
-        free_ar_ix=free_ar_ix, free_ma_ix=free_ma_ix,
-        # fixed params
-        fixed_ar_params=fixed_ar_params, fixed_ma_params=fixed_ma_params,
-    )
-
-
-def _stitch_fixed_and_free_params(fixed_ar_or_ma_lags, fixed_ar_or_ma_params,
```
jackzyliu, line 394: "Folded" into `SARIMAXParams`.

ChadFulton:
Awesome, thanks for this PR and for including the helpful notes. I'll go through this soon (hopefully tonight and tomorrow, but certainly this week), and we can get this merged.

It's been great to have your help on this issue, so thanks again!

jackzyliu:
> Awesome, thanks for this PR and for including the helpful notes. I'll go through this soon (hopefully tonight and tomorrow, but certainly this week), and we can get this merged.
> 
> It's been great to have your help on this issue, so thanks again!

You are very welcome! I am just happy to help out. 

I think there are a few other ways to do this, like having separate fixed param setters instead of the `set_fixed_params` function. Would love to hear your thoughts!

statsmodels/tsa/arima/params.py
```diff
@@ -351,6 +575,86 @@ def is_invertible(self):
 
         return ma_stationary and seasonal_ma_stationary
 
+    def set_fixed_params(self, fixed_params, validate=False):
```
ChadFulton, line 578: Just curious: why default `validate=False`?

statsmodels/tsa/arima/params.py
```diff
@@ -351,6 +575,86 @@ def is_invertible(self):
...
+        """
+        Sets all fixed params at once, which can be accessed afterwards through
```
ChadFulton, line 580: Please add one-line description above this along with a line break.

statsmodels/tsa/arima/params.py
```diff
@@ -351,6 +575,86 @@ def is_invertible(self):
...
+            self._k_fixed_params_split[param_type] = (
+                self._is_param_fixed_split[param_type].sum()
+            )
+            self._k_free_params_split[param_type] = (
+                (~self._is_param_fixed_split[param_type]).sum()
+            )
+
+    def reset_fixed_params(self, keep_param_value=False):
+        """
+        - Set status of all fixed params to 'free';
```
ChadFulton, line 626: Please add one-line description.

statsmodels/tsa/arima/params.py
ChadFulton: This is great. And if I understand correctly, the warning does not appear when you overwrite previously fixed parameters with *new* fixed parameters using `set_fixed_params`, but it does appear when you try to set the values of free parameters over the top of previously fixed parameters, e.g. using the `ar_params` setter. Is that right?

ChadFulton (commented):
This is looking great, and sorry for the longer review time than I had hoped for.

I have a couple of individual comments, but there are three things I was hoping you could address:

1. In this general context, I think we need to allow for the fixing of the sigma2 parameter.  I give a longer description of why this is in an individual comment (but please let me know if I've misunderstood something).
2. There are a number of `assert`s used for validation in the public methods, which I think we need to remove and replace with e.g. `ValueError`s with informative error messages.
3. There are a number of places where the docstring does not include the short 1-line description as the first line.

These are pretty minor (although I think the sigma2 thing shows up in a number of places), so after taking care of them we should be ready to merge.

statsmodels/tsa/arima/specification.py
```diff
@@ -1047,6 +1047,161 @@ def construct_trend_names(self):
...
+        >>> spec = SARIMAXSpecification(ar_order=1)
+        >>> spec.validate_params({"ar.L1": 1})  # valid
+        >>> spec.validate_params({"ar.L2": 1})
+        ValueError: Invalid fixed parameter(s): ['ar.L2'].
+        >>> spec.validate_params({"sigma2": 1})
+        ValueError: Invalid fixed parameter(s): ['sigma2'].
+        >>> spec.validate_params({"ar.L1": np.nan})
+        ValueError
+        """
+        assert isinstance(fixed_params, dict)
```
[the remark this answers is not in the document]
jackzyliu: Got it! I think I will remove the 2 internal consistency checks at L1187-L1195, as well, which upon reading again seems a bit redundant.

statsmodels/tsa/arima/specification.py
```diff
@@ -1047,6 +1047,161 @@ def construct_trend_names(self):
...
+        if len(fixed_params) == 0:
+            return
+
+        validate_basic(
+            list(fixed_params.values()), len(fixed_params),
+            allow_infnan=False, title='fixed parameters'
+        )
+
+        fixed_param_names = set(fixed_params.keys())
+        valid_param_names = set(self.param_names) - {"sigma2"}
```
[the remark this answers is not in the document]
jackzyliu: > We need to allow sigma2 to be fixed in the general `SARIMAXSpecification`, because this parameter can be fixed in the statespace and innovations MLE approaches (and also in principle in the other approaches, although I don't think we have any way to do that right now - it does not affect the coefficient estimates in most cases but it can affect the prediction intervals). So we will generally need to do this check in the specific estimation functions.

Ah, I see. What do you think about adding "allow_fixed_sigma2" argument here? Also, it's very likely I didn't understand this correctly, but I kind of thought `concentrate_scale` had the effect of fixing the `sigma2` (to 1). So on a related note, I also wonder how fixing `sigma2` should interact with `concentrate_scale`? As of now, it seems like we should only allow fixing `sigma2` when `concentrate_scale=False`. 

> (More generally, we have other checks on fixed parameters that will need to be done in specific estimation functions. For example, the default statespace case and innovations MLE case require the parameters to lead to a stationary process, and they impose this via a transformation function that transforms all parameters. As a consequence, in those cases we cannot, for example, fix the value of `ar.L2` while leaving the value of `ar.L1` free.)

What do you think about directly borrowing (or even migrating) from `SARIMAX._validate_can_fix_params`, which would remove the need for "estimator"-specific checks for most (possibly all) use cases?

jackzyliu:
> This is great. And if I understand correctly, the warning does not appear when you overwrite previously fixed parameters with new fixed parameters using set_fixed_params, but it does appear when you try to set the values of free parameters over the top of previously fixed parameters, e.g. using the ar_params setter. Is that right?

@ChadFulton Actually, the warning appears in both cases as of now. This is something I actually wanted to check with you.  I let the warnings appear with `set_fixed_params` because I thought about what should happen when we have consecutive `set_fixed_params` with overwriting as follows:
```
spec = SARIMAXSpecification(ar_order=2)
p = SARIMAXParams(spec)
p.set_fixed_params({'ar.L1': 1, 'ar.L2': 2}) 
p.set_fixed_params({'ar.L1': 1})  # no warning since the param values did not change
p.set_fixed_params({'ar.L1': 10})  
# params now are {'ar.L1': 10, 'ar.L2': 2} with a warning in case overwriting is unintentional
```  
Do you think it makes more sense not to issue the warning when we overwrite with `set_fixed_params`?

statsmodels/tsa/arima/params.py
jackzyliu: Ah, I was imagining that, in most cases, we probably have separately checked `fixed_params` with `validate_fixed_params` along with any other checks before we call `set_fixed_params`, so usually there may not be the need to validate again.

jackzyliu:
Hi @ChadFulton, just giving this thread a bump in case it fell through the cracks. I had a few questions about your comments. Let me know what you think, and I will make changes accordingly. :)

ChadFulton:
> Hi @ChadFulton, just giving this thread a bump in case it fell through the cracks. I had a few questions about your comments. Let me know what you think, and I will make changes accordingly. :)

Thanks for the bump and I'm sorry for the long time in reviewing! Thank you for your responses, I will make sure to reply by tomorrow night.

Thanks again!

statsmodels/tsa/arima/specification.py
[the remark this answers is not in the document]
ChadFulton: > Ah, I see. What do you think about adding "allow_fixed_sigma2" argument here?

I agree that this seems like a good approach, and implementing the check here should prevent boilerplate in the individual methods.

> I also wonder how fixing sigma2 should interact with concentrate_scale? As of now, it seems like we should only allow fixing sigma2 when concentrate_scale=False.

Yes I agree.  Currently in `SARIMAX`, if `concentrate_scale=True`, then `'sigma2'` is removed from the list of parameters, and so if you try to do e.g. `mod.fix_params({'sigma2': 4.3})`, you get: `ValueError: Invalid parameter name passed: "sigma2".`, so this is one option here, or you could be more explicit too.

> I kind of thought concentrate_scale had the effect of fixing the sigma2 (to 1)

You're right that this is practically what happens when the Kalman filter is being run, but at the same time it also computes an estimate of sigma2, and then that estimate is plugged back in after the fact. So the end result is a model that does not generally have sigma2 = 1.

statsmodels/tsa/arima/params.py
ChadFulton: Sounds good to me, I'll leave the final decision up to you.

ChadFulton:
> > This is great. And if I understand correctly, the warning does not appear when you overwrite previously fixed parameters with new fixed parameters using set_fixed_params, but it does appear when you try to set the values of free parameters over the top of previously fixed parameters, e.g. using the ar_params setter. Is that right?
> @ChadFulton Actually, the warning appears in both cases as of now. This is something I actually wanted to check with you. I let the warnings appear with set_fixed_params because I thought about what should happen when we have consecutive set_fixed_params with overwriting as follows:
> 
> ```
> spec = SARIMAXSpecification(ar_order=2)
> p = SARIMAXParams(spec)
> p.set_fixed_params({'ar.L1': 1, 'ar.L2': 2}) 
> p.set_fixed_params({'ar.L1': 1})  # no warning since the param values did not change
> p.set_fixed_params({'ar.L1': 10})  
> # params now are {'ar.L1': 10, 'ar.L2': 2} with a warning in case overwriting is unintentional
> ```
> Do you think it makes more sense not to issue the warning when we overwrite with set_fixed_params?

I think that it makes more sense not to issue a warning when overwriting with `set_fixed_params`, because to me it seems like that function has a very explicit purpose. So to me, users doing this would be aware of what the consequences are.

By contrast, using e.g. the `ar_params` setter should be used for setting values for the free parameters, and so in my view overwrite the fixed parameters is probably not what the user wants to do. So I think the warning is good here.

Please let me know if I misunderstood anything or if you have a different view.

ChadFulton:
Sorry for the long time between responses.  I hope that clears things up, and I think we're in good shape to merge soon, after some (I think) small changes. Thanks!

jackzyliu (commented):
Hey @ChadFulton, thanks for clarifying! I just got around to making these changes:
- Adding support for `sigma2` in `SARIMAXSpecification` and `SARIMAXParams`
- Docstring fixes, such as adding one-line descriptions and making documentation more complete
- Replacing `asserts` with `ValueError`, and removing unnecessary ones
- Changing the overwriting warning behavior in `SARIMAXParams.set_fixed_params`

Please let me know if anything else needs changing!

jackzyliu:
Bump :) @ChadFulton

ChadFulton:
Thanks for the bump, I would really like to get this in soon but I’m won’t be able to do more for another week. But thanks again for your work on this!

bashtage:
There will be a 0.13.1 shortly after Python 3.10 comes out so if this doesn't make 0.13 it can be included in 0.13.1.

PR #7581, opened 2021-07-15
ENH: Add support for pickling for generic path-like objects

### LINKED ISSUE #7199: ENH: Allow both pathlib and string paths for file io operations such as load and save
Using an object which inherits pathlib.Path and overrides the open method breaks, as it tries to open it with the builtin open method and the given path is from the s3 (this behaviour is implemented in the custom object).

```python
import statsmodels.api as sm
LOGIT_PATH = URIPath(
    "some_path"
)
sm.load(LOGIT_PATH)


class URIPath(pathlib.Path):
    def open(...):
        ...
       handle_s3_case
```

Suggestion: use the .open() method of pathlib.Path

bashtage:
Add support for objects supporting open

closes #7199

- [X] closes #7199
- [X] tests added / passed. 

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>

### DIFF
--- a/statsmodels/iolib/tests/test_pickle.py
+++ b/statsmodels/iolib/tests/test_pickle.py
@@ -1,11 +1,13 @@
+from statsmodels.compat.python import lrange
+
 from io import BytesIO
+import os
 import pathlib
 import tempfile
 
 from numpy.testing import assert_equal
 
-from statsmodels.compat.python import lrange
-from statsmodels.iolib.smpickle import save_pickle, load_pickle
+from statsmodels.iolib.smpickle import load_pickle, save_pickle
 
 
 def test_pickle():
@@ -26,8 +28,6 @@ def test_pickle():
 
     # cleanup, tested on Windows
     try:
-        import os
-
         os.remove(path_str)
         os.remove(path_pathlib)
         os.rmdir(tmpdir)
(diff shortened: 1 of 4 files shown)


statsmodels/iolib/tests/test_pickle.py
```diff
@@ -42,3 +42,34 @@ def test_pickle():
...
+            errors=None,
+            newline=None,
+        ):
+            return self._path.open(
+                mode=mode,
+                buffering=buffering,
+                encoding=encoding,
+                errors=errors,
+                newline=newline,
+            )
```
josef-pkt, line 69: this looks really ugly, too much vertical space

bashtage: Embrace the 

image

Welcoming black into your code frees you from even looking at formatting.

bashtage: On a serious note, the vertical space has the long-run effect of minimizing diffs.

josef-pkt:
no Black, it was also rejected for scipy
too painful to read

bashtage:
> no Black, it was also rejected for scipy
> too painful to read

I don't really believe that. It is widely used in some very large orgs, e.g., Facebook and Dropbox.  I also don't agree with it.  I think it is a greybeard thing.

bashtage:
A lot of TSA is now black, albeit with the caveat to limit to 79 lines.

PR #7607, opened 2021-07-27
DOC: copula in user guide and examples

tupui:
Follow up of #7408 which adds Copula.

This PR adds some documentation in the user guide and an example notebook.

cc @josef-pkt


tupui:
@bashtage it would be great if you could have a look. Might be good to have another pair of eyes on this topic. Thanks 🙏

docs/source/distributions.rst
```diff
@@ -54,6 +54,31 @@ use potentially slow inherited generic methods.
    zinegbin
    zipoisson
 
+Copula
+------
+
+The `copula` sub-module provides classes to model the dependence between
+parameters. Copula are used to construct a multivariate joint distribution and
```
bashtage, line 61: Copulas or perhaps Copulae

bashtage (changes requested):
One small grammar mistake.  I will merge once fixed. Thanks.

docs/source/distributions.rst
tupui, line 61:
```suggestion
parameters. Copulae are used to construct a multivariate joint distribution and
```

tupui:
Thank you @bashtage

bashtage:
Thanks.

PR #7608, opened 2021-07-28
ENH: random number generation wrapper for rng, qrng

tupui:
This is another follow up of #7254 which adds a wrapper around `np.random.Generator` and alike.

It is similar to the one in SciPy for QMC, but it also allows to pass along an instance of `scipy.qmc.QMCEngine`. This way we can use QMC with copula.

c.f. discussion https://github.com/statsmodels/statsmodels/pull/7408#issuecomment-888266539

cc @josef-pkt

### DIFF
--- a/statsmodels/tools/rng_qrng.py
+++ b/statsmodels/tools/rng_qrng.py
@@ -29,10 +29,10 @@ def check_random_state(seed=None):
         Random number generator.
 
     """
-    if hasattr(stats.qmc, "QMCEngine"):
-        if isinstance(seed, stats.qmc.QMCEngine):
-            return seed
-    if isinstance(seed, np.random.RandomState):
+    if hasattr(stats, "qmc") and \
+            isinstance(seed, stats.qmc.QMCEngine):
         return seed
-
-    return np.random.default_rng(seed)
+    elif isinstance(seed, np.random.RandomState):
+        return seed
+    else:
+        return np.random.default_rng(seed)
--- a/statsmodels/tools/rng_qrng.py
+++ b/statsmodels/tools/rng_qrng.py
@@ -1,0 +1,50 @@
+import numbers
+
+import numpy as np
+import scipy.stats as stats
+
+
+def check_random_state(seed=None):
+    """Turn `seed` into a random number generator.
+
+    Parameters
+    ----------
+    seed : {None, int, `numpy.random.Generator`,
+            `numpy.random.RandomState`, `scipy.stats.qmc.QMCEngine`}, optional
+
+        If `seed` is None fresh, unpredictable entropy will be pulled
+        from the OS and `numpy.random.Generator` is used.
+        If `seed` is an int, a new ``Generator`` instance is used,
+        seeded with `seed`.
+        If `seed` is already a ``Generator``, ``RandomState`` or
+        `scipy.stats.qmc.QMCEngine` instance then
+        that instance is used.
+
+        `scipy.stats.qmc.QMCEngine` requires SciPy >=1.7. It also means
+        that the generator only have the method ``random``.
+
+    Returns
+    -------
+    seed : {`numpy.random.Generator`, `numpy.random.RandomState`}
+        Random number generator.
+
+    """
+    if seed is None or isinstance(seed, (numbers.Integral, np.integer)):
+        if not hasattr(np.random, 'Generator'):
+            # This can be removed once numpy 1.16 is dropped
+            msg = ("NumPy 1.16 doesn't have Generator, use either "
+                   "NumPy >= 1.17 or `seed=np.random.RandomState(seed)`")
+            raise ValueError(msg)
+        return np.random.default_rng(seed)
+    elif isinstance(seed, np.random.RandomState):
+        return seed
+    elif isinstance(seed, np.random.Generator):
+        # The two checks can be merged once numpy 1.16 is dropped
+        return seed
+    elif isinstance(seed, np.random.Generator):
+        return seed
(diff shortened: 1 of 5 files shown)


tupui:
I did not add tests for now as I want to agree first on what is wanted/needed.

tupui:
@bashtage do you have an opinion about this one?

statsmodels/tools/rng_qrng.py
```diff
@@ -0,0 +1,50 @@
...
+        If `seed` is already a ``Generator``, ``RandomState`` or
+        `scipy.stats.qmc.QMCEngine` instance then
+        that instance is used.
+
+        `scipy.stats.qmc.QMCEngine` requires SciPy >=1.7. It also means
+        that the generator only have the method ``random``.
+
+    Returns
+    -------
+    seed : {`numpy.random.Generator`, `numpy.random.RandomState`}
```
bashtage, line 28: Can it not return a QMCEngine?

statsmodels/tools/rng_qrng.py
```diff
@@ -0,0 +1,50 @@
...
+        Random number generator.
+
+    """
+    if seed is None or isinstance(seed, (numbers.Integral, np.integer)):
+        if not hasattr(np.random, 'Generator'):
+            # This can be removed once numpy 1.16 is dropped
+            msg = ("NumPy 1.16 doesn't have Generator, use either "
```
bashtage, line 35: Already 1.17+

statsmodels/tools/rng_qrng.py
```diff
@@ -0,0 +1,50 @@
...
+                   "NumPy >= 1.17 or `seed=np.random.RandomState(seed)`")
+            raise ValueError(msg)
+        return np.random.default_rng(seed)
+    elif isinstance(seed, np.random.RandomState):
+        return seed
+    elif isinstance(seed, np.random.Generator):
+        # The two checks can be merged once numpy 1.16 is dropped
```
bashtage, line 42: Already dropped.

statsmodels/tools/rng_qrng.py
```diff
@@ -0,0 +1,50 @@
...
+import numpy as np
+import scipy.stats as stats
+
+
+def check_random_state(seed=None):
+    """Turn `seed` into a random number generator.
+
+    Parameters
+    ----------
+    seed : {None, int, `numpy.random.Generator`,
```
bashtage, line 12: Generator also support Sequence[int].

bashtage (changes requested):
Some changes needed. Also appears that it needs to be used.

statsmodels/tools/rng_qrng.py
tupui: `Sequence[int]` did not work, only `Sequence`. But I guess it's fine as it will be caught by NumPy. Otherwise we could just put `np.random.default_rng(seed)` at the end to catch all errors.

statsmodels/tools/rng_qrng.py
tupui: Fixed.

statsmodels/tools/rng_qrng.py
tupui: Sorry, seemingly I have an issue with deps...

tupui (commented):
Thanks @bashtage, I updated.

statsmodels/tools/rng_qrng.py
```diff
@@ -0,0 +1,43 @@
...
+
+    Returns
+    -------
+    seed : {`numpy.random.Generator`, `numpy.random.RandomState`,
+            `scipy.stats.qmc.QMCEngine`}
+
+        Random number generator.
+
+    """
+    if seed is None or isinstance(seed, (numbers.Integral, np.integer, Sequence)):
```
bashtage, line 35: I think your idea to invert - specific instance first than pass remainder to `default_rng()` and let NumPy raise.

tupui: It is quite a simplification. Hence there is just a need to check for `RandomState` and `QMCEngine` as `Generator` can be passed to `default_rng`

tupui:
Might be for later.

A bit of a stretch, but not much: https://github.com/scipy/scipy/issues/13368. Do you remember this @bashtage? I proposed to wrap `QMCEngine` so that it behaves like a `Generator`. Also SciPy is adding UNU.RAN support and there is already `NumericalInverseHermite.qrvs`. Maybe we could take advantage of the latter at least to propose a real QMC like `Generator`. What do you think?

bashtage:
Looks like you have a test failure.

statsmodels/tools/rng_qrng.py
```diff
@@ -0,0 +1,38 @@
...
+        Random number generator.
+
+    """
+    if hasattr(stats.qmc, "QMCEngine"):
+        if isinstance(seed, stats.qmc.QMCEngine):
+            return seed
+    if isinstance(seed, np.random.RandomState):
+        return seed
+
+    return np.random.default_rng(seed)
```
tupui, line 38:
```suggestion
    if hasattr(stats, "qmc") and \
            isinstance(seed, stats.qmc.QMCEngine):
        return seed
    elif isinstance(seed, np.random.RandomState):
        return seed
    else:
        return np.random.default_rng(seed)
```

tupui:
Passing now. The seed is capricious. It can make the tests fails. Also I had to use a RandomState as passing a generator is not supported by SciPy functions... I will see to fix this upstream.

bashtage (approved):

bashtage:
LGTM. Thanks.

PR #7616, opened 2021-08-02
ENH: Add RUR stationarity test to statsmodels.tsa.stattools

mikylucky:
The RUR test implementation comes from the paper  "Range Unit-Root tests" by Aparicio et Al. [2006]

The critical values table has been recalculate with more iterations than the paper.

- [x] closes #xxxx
- [x] tests added / passed. 

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>

### DIFF
--- a/statsmodels/tsa/stattools.py
+++ b/statsmodels/tsa/stattools.py
@@ -2254,8 +2254,8 @@ look-up table. The actual p-value is {direction} than the p-value returned.
         rstore = ResultsStore()
         rstore.nobs = nobs
 
-        rstore.H0 = "The series is not {0} stationary"
-        rstore.HA = "The series is {0} stationary"
+        rstore.H0 = "The series is not stationary"
+        rstore.HA = "The series is stationary"
 
         return rur_stat, p_value, crit_dict, rstore
     else:
--- a/statsmodels/tsa/stattools.py
+++ b/statsmodels/tsa/stattools.py
@@ -12,6 +12,7 @@ import numpy as np
 from numpy.linalg import LinAlgError
 import pandas as pd
 from scipy import stats
+from scipy.interpolate import interp1d
 
 from statsmodels.regression.linear_model import OLS, yule_walker
 from statsmodels.tools.sm_exceptions import (
@@ -54,6 +55,7 @@ __all__ = [
     "levinson_durbin_pacf",
     "levinson_durbin",
     "zivot_andrews",
+    "rur",
 ]
 
 SQRTEPS = np.sqrt(np.finfo(np.double).eps)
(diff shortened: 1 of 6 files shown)


statsmodels/tsa/stattools.py
```diff
@@ -2134,6 +2136,141 @@ def _kpss_autolag(resids, nobs):
...
+        [1.0494, 1.1643, 1.2712, 1.4076, 3.0207, 3.3584],
+        [1.0846, 1.1959, 1.2988, 1.4344, 3.0653, 3.4073],
+        [1.1121, 1.2200, 1.3230, 1.4556, 3.0948, 3.4439],
+        [1.1204, 1.2295, 1.3318, 1.4656, 3.1054, 3.4632],
+        [1.1309, 1.2347, 1.3318, 1.4693, 3.1165, 3.4717],
+        [1.1377, 1.2402, 1.3408, 1.4729, 3.1252, 3.4807]
+    ])
+
+    # Interpolation for nobs
+    XX = np.array([nobs])
```
bashtage, line 2214: Why is the an array with a number value?  Seems unnecessary

statsmodels/tsa/stattools.py
```diff
@@ -2134,6 +2136,141 @@ def _kpss_autolag(resids, nobs):
...
+
+    inter_crit = np.zeros((len(XX), crit.shape[1]))
+    for i in range(crit.shape[1]):
+        f = interp1d(n, crit[:, i])
+        for j in range(len(XX)):
```
bashtage, line 2219: This loop seems unnecessary.

statsmodels/tsa/stattools.py
```diff
@@ -2134,6 +2136,141 @@ def _kpss_autolag(resids, nobs):
...
+            inter_crit[j, i] = f(XX[j])
+
+    # Calculate RUR stat
+    count = 0
+    max_p = x[0]
+    min_p = x[0]
+
+    for v in x[1:]:
```
bashtage, line 2227: This should be vectorizable with pandas `expanding`. Something like

xs = pd.Series(x)
exp_max = xs.expanding(1).max().shift(1)
exp_min = xs.expanding(1).min().shift(1)
count = (xs > exp_max).sum() + (xs < exp_min).sum()

statsmodels/tsa/stattools.py
```diff
@@ -2134,6 +2136,141 @@ def _kpss_autolag(resids, nobs):
...
+        else:
+            break
+
+    p_value = pvals[k]
+
+    warn_msg = """\
+The test statistic is outside of the range of p-values available in the
+look-up table. The actual p-value is {direction} than the p-value returned.
+"""
+    if p_value == pvals[-1]:
```
bashtage, line 2250: Just set `direction` to a string when needed then simplify the warning.

```python
direction = ""
if p_value == pvals[-1]:
    direction="smaller"
elif p_value == pvals[0]:
    direction="larger"
if direction:
    warnings.warn(
            warn_msg.format(direction=direction), InterpolationWarning
        )
```

statsmodels/tsa/stattools.py
```diff
@@ -2134,6 +2136,141 @@ def _kpss_autolag(resids, nobs):
     autolags = int(gamma_hat * np.power(nobs, pwr))
     return autolags
 
+def rur(x, store=False):
```
bashtage, line 2139: I think the name should change.  Possibilities are

```python
rur_unit_root_test
range_unit_root_test
```

There are probably some other sensible ones.  But `rur` is not discoverable -- it helps to have the test easy to find.

bashtage (changes requested):
Other requirements

1. Test.  Conisder using simulated values and `R` where the simulated values come from `np.random.RandomState(SOME_SEED_VALUE)`.  Alternatively, use one or more of the data sets available in statsmodels.datasets.
2. Add to docs in the appropriate section.

statsmodels/tsa/stattools.py
mikylucky: I copied it from the adfuller function, but I like your suggestion

statsmodels/tsa/stattools.py
```diff
@@ -2134,6 +2136,131 @@ def _kpss_autolag(resids, nobs):
...
+
+    crit_dict = {"10%": inter_crit[0,3], "5%": inter_crit[0,2], "2.5%": inter_crit[0,1], "1%": inter_crit[0,0]}
+
+    if store:
+        from statsmodels.stats.diagnostic import ResultsStore
+
+        rstore = ResultsStore()
+        rstore.nobs = nobs
+
+        rstore.H0 = "The series is not {0} stationary"
```
bashtage, line 2257: Why are there format characters `{0}` in these strings?

statsmodels/tsa/stattools.py
```diff
@@ -2134,6 +2136,131 @@ def _kpss_autolag(resids, nobs):
...
+look-up table. The actual p-value is {direction} than the p-value returned.
+"""
+    direction = ""
+    if p_value == pvals[-1]:
+        direction="smaller"
+    elif p_value == pvals[0]:
+        direction="larger"
+
+    if direction:
+        warnings.warn( warn_msg.format(direction=direction), InterpolationWarning )
```
bashtage, line 2247: PEP issues here.

statsmodels/tsa/stattools.py
mikylucky: it was a typo from a copy paste

statsmodels/tsa/stattools.py
mikylucky: Where can I see the error details?

bashtage:
Still  need tests.  Failure is unrelated.  I think it will go away if you rebase on main.

mikylucky:
I'm struggling to find an `R` implementation of the RUR test, how should I set up the testing?

bashtage:
If there is no reference, we can test the implementation against a very simple implementation, say one that uses a for loop.

You should probably also extend the test to cover https://www.tandfonline.com/doi/abs/10.1080/03610918.2012.732172?journalCode=lssp20 if possible.

mikylucky:
I will implement the EDRUR in a following version, I have to finish some work before.

Also, the failed test I think is being referring to a different part of the module

bashtage:
Failure is unrelated. A follow up with the improved test would be very helpful.

Thanks for the contribution.

PR #7791, opened 2021-10-11
BUG: incorrect HW predictions

### LINKED ISSUE #7790: BUG: incorrect HW predictions
Holt's model makes incorrect predictions for some types of inputs

```python
import pandas as pd
from statsmodels.tsa.holtwinters import Holt

# same inputs with two different types of freq (int and datetime)
array_input = [10, 20, 30, 40, 50]
pd_series_input = pd.Series([10, 20, 30, 40, 50], pd.date_range(start='1/1/2018', periods=5))

model_array = Holt(array_input).fit()
model_series = Holt(pd_series_input).fit()

print(model_array.predict()) # this model return 70
print(model_series.predict()) # this model return 60
```

Expected behavior that both models should return the same result.

I figured out the problem and fixed it in my [PR #7791](https://github.com/statsmodels/statsmodels/pull/7791).

statsmodels version: 0.13.0

Wooqo:
Resolve issue 7790. We don't need add freq to get correct start value.
- [x] closes #7790
- [x] tests added / passed. 

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>

### DIFF
--- a/statsmodels/tsa/holtwinters/tests/test_holtwinters.py
+++ b/statsmodels/tsa/holtwinters/tests/test_holtwinters.py
@@ -1025,6 +1025,38 @@ def test_damping_trend_zero():
     assert_allclose(pred1, pred2, atol=1e-10)
 
 
+def test_different_inputs():
+    array_input_add = [10, 20, 30, 40, 50]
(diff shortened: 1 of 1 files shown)


bashtage:
LGTM.  Thanks.

PR #7820, opened 2021-10-27
MAINT: clean up usage of private SciPy APIs as much as possible

rgommers:
xref the discussion in https://github.com/scipy/scipy/pull/14898, where @josef-pkt pointed out that `stats.kde.mvn` was used in statsmodels code. I went through the whole code base and tried to identify all usage of private SciPy APIs and fix them, or future-proof them for those APIs being moved around in SciPy 1.8.0.

It would be good to backport these changes and do a 0.13.1 release soon, because in the current state, 0.13.0 is broken when run against SciPy master.


bashtage:
LGTM. Thanks.

PR #7930, opened 2021-12-06
ENH: Option of initial values whe simulating VAR model

EC-AI:
- [ ] closes #xxxx
- [ ] tests added / passed. 

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>


statsmodels/tsa/vector_ar/util.py
```diff
@@ -215,10 +215,11 @@ def varsim(coefs, intercept, sig_u, steps=100, initvalues=None, seed=None):
         observations to start the autoregressive process.
         If offset is not None, then exog of the model are used if they were
         provided in the model
+    initvalues : None or ndarray (nlags, neqs)
```
bashtage, line 218: Type should be 

ndarray, optional

bashtage: After you make the other changes, the shape should move to the description.  It should be either (neqs,) or (nlags, neqs)

statsmodels/tsa/vector_ar/util.py
```diff
@@ -215,10 +215,11 @@ def varsim(coefs, intercept, sig_u, steps=100, initvalues=None, seed=None):
...
+        Initial values for use in the simulation.
     seed : {None, int}
         If seed is not None, then it will be used with for the random
         variables generated by numpy.random.
-
```
bashtage, line 221: Need a blank line here with no white space

statsmodels/tsa/vector_ar/util.py
```diff
@@ -241,6 +242,11 @@ def varsim(coefs, intercept, sig_u, steps=100, initvalues=None, seed=None):
         result[p:] += ugen[p:]
     else:
         result[p:] = ugen[p:]
+        
+    if initvalues is not None:
+        if not initvalues.shape == (p, k):
```
bashtage, line 247: Should handle case where initvalues are either (k,) or (p,k). In either case I think you only need to modify the docstring and broadcasting will take care of assignment expansion.

statsmodels/tsa/vector_ar/var_model.py
```diff
@@ -923,7 +923,7 @@ def is_stable(self, verbose=False):
         """
         return is_stable(self.coefs, verbose=verbose)
 
-    def simulate_var(self, steps=None, offset=None, seed=None):
+    def simulate_var(self, steps=None, offset=None, initial_state=None, seed=None):
```
bashtage, line 926: Keywords should match.  I would go with `initial_values` which is clearest.  This keyword must be added at the end since it is new and putting it in the middle will break any position only use. This should be keyword only

```
def simulate_var(self, steps=None, offset=None, seed=None, *, initial_values=None):
```

statsmodels/tsa/vector_ar/var_model.py
```diff
@@ -941,6 +941,8 @@ def simulate_var(self, steps=None, offset=None, seed=None):
             the linear predictor of those components will be used as offset.
             This should have the same number of rows as steps, and the same
             number of columns as endogenous variables (neqs).
+        initial_state : None or ndarray (nlags, neqs)
```
bashtage, line 944: Same constraints here. Either (nlags, neqs) or (neqs,).  Type should be array_like, optional.

statsmodels/tsa/vector_ar/util.py
bashtage: Sorry, should be array_like, optional

then you need to use asarray below, or better yet the function `array_like`

statsmodels/tsa/vector_ar/util.py
```diff
@@ -241,6 +242,11 @@ def varsim(coefs, intercept, sig_u, steps=100, initvalues=None, seed=None):
         result[p:] += ugen[p:]
     else:
         result[p:] = ugen[p:]
+
```
bashtage, line 245: add an 

```
initial_values = array_like(initial_values, "initial_values", optional=True)
```

statsmodels/tsa/vector_ar/util.py
bashtage: Also, please rename to `initial_values` for consistency with below.

bashtage (changes requested):
A few changes

* Rename to `initial_value` in both
* Accept either (neqs,) or (nlags, neqs)
* Please indicate the order in the doc string of the initial values. Is rows 0 the most recent value or the most distant value?
* Needs a minimal test with both shapes of inputs.

bashtage:
Changes look good.  If you could try and write a simple test with a both a VAR(1) and a VAR(2 (or more), and using both initial values that are (neqs,) and (negs, lags) then I'll merge.

EC-AI:
Wrote tests but any input will be appreciated. I don't have much experience writing tests as I'm more of a data scientist than a software developer.

bashtage:
Test looks simple and fine.  Need to fix the pep8 violations noted above.

bashtage:
Great! All passing and a test too.  Thanks for the great PR.

PR #8086, opened 2022-02-01
BUG: denominator needs to be a vector

xjcl:
multivariate_t_rvs function currently seems to fail with "cannot apply subscript to float" in df==np.inf case

- [ ] closes #xxxx
- [ ] tests added / passed. 

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>

### DIFF
--- a/statsmodels/sandbox/distributions/multivariate.py
+++ b/statsmodels/sandbox/distributions/multivariate.py
@@ -110,7 +110,7 @@ def multivariate_t_rvs(m, S, df=np.inf, n=1):
     m = np.asarray(m)
     d = len(m)
     if df == np.inf:
-        x = 1.
+        x = np.ones(n)
     else:
         x = np.random.chisquare(df, n)/df
     z = np.random.multivariate_normal(np.zeros(d),S,(n,))


xjcl:
Hope someone can take this off my hands. I wasn't able to find a bug report for this which surprised me. I couldn't find a test for multivariate_t_rvs. I guess a test for this would just call multivariate_t_rvs with df=np.inf and check if it doesn't throw, but I don't know how to add one.

bashtage:
What input are you using the produce the error?

bashtage:
Looks right to me.  Thanks.

bashtage:
## 💚 All backports created successfully

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|maintenance/0.13.x|[](https://github.com/statsmodels/statsmodels/pull/8097)|


### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport)

xjcl:
Glad I could contribute. I came across this function as it was posted here on StackOverflow and I needed it for work: https://stackoverflow.com/a/41967819/2111778. I think `scipy.stats.multivariate_t.rvs` provides similar functionality but we are stuck on a pre-1.6 version.

As for inputs reproducing the bug, **any** multidimensional input should produce the error as long as df (degrees of freedom) is `np.inf`. So for example

```
multivariate_t_rvs(
    np.array([0, 0]),
    np.array([[1, 0], [0, 1]]),
)
```

with the result

```
    return m + z/np.sqrt(x)[:,None]   # same output format as random.multivariate_normal
IndexError: invalid index to scalar variable.
```

as it tries to evaluate `1[:,None]`

josef-pkt:
@xjcl 
We also have a copy of scipy.stats multivariate_t version in  `statsmodels.compat._scipy_multivariate_t`.
We needed it for copulas, and it was easier to copy the scipy version than update our sandbox version.
https://github.com/statsmodels/statsmodels/blob/main/statsmodels/distributions/copula/elliptical.py#L13

Our copy scipy.stats multivariate_t will stay for a few versions, until our miminum scipy version includes it.

xjcl:
@josef-pkt So the code I patched isn't even used productively?

bashtage:
It is used, but will be retired eventually.

josef-pkt:
AFAIR, that code is not used in any public part of statsmodels.

I only needed some parts of multivariate_t when I wrote this, mainly the cdf.
The rvs was used for checking and testing.
e.g. my question when I got confused about independence versus uncorrelated:
https://stackoverflow.com/questions/4576256/why-do-my-numbers-not-match-multivariate-t-distribution-in-r-mvtnorm

(some properties like marginal and conditional distribution of multivariate elliptical distributions is going to show up again in copulas and similar. The sandbox has a few basics for elliptical distributions)

PR #8104, opened 2022-02-07
DOC: DEV Guide modify redundant text

pkaf:
Deleting/modifying some redundant texts from DEV guide to better reflect current workflow.


bashtage:
LGTM.  Thanks for this.

PR #8109, opened 2022-02-08
fix(setup): use correct setuptools backend

henryiii:
The default is `build-backend = "setuptools.build_meta:__legacy__"`, which is not ideal. This field really shouldn't have been optional, but it was.

The legacy backend injects the local path - that has to be done by hand with the non-legacy backend, since packages are discouraged from importing in setup.py. If that makes a difference, that will also have to be added here.

- [ ] closes #xxxx
- [ ] tests added / passed. 

Oops, didn't format the commit message. Can reformat after tests run.

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,13 +12,20 @@ from collections import defaultdict
 from distutils.command.clean import clean
 import fnmatch
 import os
-from os.path import abspath, join as pjoin, relpath, split
+from os.path import join as pjoin, relpath, split
+from pathlib import Path
 import shutil
 import sys
 
 import pkg_resources
 
+SETUP_DIR = Path(__file__).parent.resolve()
+
+sys.path.append(str(SETUP_DIR))
 import versioneer
+
+del sys.path[-1]
+
 
 try:
     # SM_FORCE_C is a testing shim to force setup to use C source files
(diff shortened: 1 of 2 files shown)


bashtage:
Doesn't seem to like this.  Having problems with versioneer.

setup.py
```diff
@@ -12,14 +12,21 @@
 from distutils.command.clean import clean
 import fnmatch
 import os
-from os.path import abspath, join as pjoin, relpath, split
+from os.path import join as pjoin, relpath, split
```
henryiii, line 15:
```suggestion
from os.path import join as pjoin, relpath
```

bashtage:
Fixed a minor flake issue.  Thanks.

bashtage:
## 💚 All backports created successfully

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|maintenance/0.13.x|[](https://github.com/statsmodels/statsmodels/pull/8111)|


### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport)

henryiii:
Personally, I'd use `setuptools_scm`, local importing into setup.py is discouraged, that's why it's only part of the `__legacy__` backend. Versioneer also doesn't support overiding the version via an environment variable, even though it's been asked for for years, while setuptools_scm does. But it was easy to fix.

PR #8118, opened 2022-02-13
BUG: Fix ValueError output in lagmat when using pandas

KishManani:
Fixes the error message in `lagmat` when `trim="backward"` and `use_pandas=True`. The current error message incorrectly states that `trim="forward"` and `use_pandas=True` are incompatible.


bashtage:
Thanks.

PR #8160, opened 2022-02-28
ENH: Add MSTL algorithm for multi-seasonal time series decomposition

KishManani:
This PR:

- Adds an implementation of the [MSTL algorithm](https://arxiv.org/pdf/2107.13462.pdf) and already [exists in the Forecast package in R](https://www.rdocumentation.org/packages/forecast/versions/8.16/topics/mstl). This provides a method to decompose a time series which has multiple seasonal components. 
- Adds an example notebook to show how to use it.
- Modifies the DecomposeResults class to allow plotting of multiple seasonal components. 

There are two deviations from the original algorithm:

1) Missing data handling is omitted and left to the user to do beforehand as a pre-processing step. This is because an equivalent of the R method used in the original paper to impute missing values does not have a simple implementation in Python. I also don't think this is core to the algorithm.

2) The original paper also handled a case when there is no seasonal component and uses Friedman’s Super Smoother
function to just extract a trend. This implementation assumes that there is at least one seasonal component otherwise a user would not use a decomposition method like this. Once again I don't think this is core to the algorithm.

I still need to compile the documentation and add that. I will do that if this feature is considered useful and the overall approach is accepted. 

Also any feedback regarding improving the tests would be appreciated.

- [ ] closes #xxxx
- [x] tests added / passed. 

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>


statsmodels/tsa/mstl.py
```diff
@@ -0,0 +1,240 @@
...
+from scipy.stats import boxcox
+
+from statsmodels.tools.typing import ArrayLike1D
+from statsmodels.tsa.seasonal import STL
+from statsmodels.tsa.tsatools import freq_to_period
+
+
+class MSTL:
+    """
+    MSTL(endog, periods=None, windows=None, lmbda=None, iterate=2, **kwargs)
```
bashtage, line 32: I don't like **kwargs.  It would be better to use a specific parameter `stl_args` or `stl_kwargs` as a `dict[str,bool | int | None]`. The alternative would be to list each of the STL arguments as keyword only arguments, although there are a lot.

statsmodels/tsa/mstl.py
```diff
@@ -0,0 +1,240 @@
...
+    >>> res.plot()
+    >>> plt.show()
+    """
+
+    def __init__(
+        self,
+        endog: ArrayLike1D,
+        periods: Optional[Union[int, Sequence[int]]] = None,
+        windows: Optional[Union[int, Sequence[int]]] = None,
+        lmbda: Optional[Union[float, str]] = None,
```
bashtage, line 99: I think many of these should be keyword only.  Not sure where to draw the line.  KW only make it a lot easier to make any future changes.

statsmodels/tsa/mstl.py
```diff
@@ -0,0 +1,240 @@
+"""
```
bashtage, line 1: Would be a good idea to reorg now with

tsa\
tsa\stl\
tsa\stl\__init__.py
tsa\stl\_stl.pyx
\tsa\stl\mstl.py

examples/notebooks/mstl_decomposition.ipynb
```diff
@@ -0,0 +1,222 @@
+{
```
bashtage, line 1: Can you add an image for the examples directory and add this to the yaml file that lists notebooks?

bashtage (changes requested):
Looks very close.  Some things that I think we should change, and then I'll give it another look.  Thanks.

ChadFulton:
Thanks for the contribution, this looks great!  Is it possible to write some unit tests that check the values against what you would get from the paper and/or the R package?

KishManani:
Hi @ChadFulton. 

The [paper](https://arxiv.org/pdf/2107.13462.pdf) used an electricity demand in Victoria dataset with a frequency of 30 mins. They then upsampled to hourly and limited their view to H1 of 2012.

I found the raw dataset here: https://github.com/tidyverts/tsibbledata/tree/master/data-raw/vic_elec/VIC2015. I applied the same transformations to get the data to the same time period and granularity. It looks pretty similar/identical to the paper. I've attached this dataset as `mstl_elec_vic.csv` file in this commit: [f42b54b](https://github.com/statsmodels/statsmodels/pull/8160/commits/f42b54b0d3319e46cc265829c83162a9a8c1dd03).

I then used this dataset in R with the following params: `res = mstl(x=y, iterate=2, s.window=c(11, 15), s.jump=1, t.jump=1, l.jump=1)` and saved this dataset as `mstl_test_results.csv` (naming convention chosen to follow what was done for STL). This dataset is also to be found in the same commit: [f42b54b](https://github.com/statsmodels/statsmodels/pull/8160/commits/f42b54b0d3319e46cc265829c83162a9a8c1dd03).

I've then added a test which shows that the results are pretty much identical. Let me know if you want me to add more!

I plan to update the example notebook with this more real world example and add that next.

KishManani:
I have now updated the example notebook to include generating the electricity demand dataset (mentioned (here)[https://github.com/statsmodels/statsmodels/pull/8160#issuecomment-1057590807]) that is now used in the unit test*. I have also updated the image file. See [a38930d](https://github.com/statsmodels/statsmodels/pull/8160/commits/a38930d3559668d83eb669ecb34a5c5caced88c2).

* The dataset added in [091c8b5](https://github.com/statsmodels/statsmodels/pull/8160/commits/091c8b5194e299ae1fbe5927d275036332de1780) has one additional day of data relative to the notebook. However, this has no impact on the unit test.

statsmodels/tsa/stl/mstl.py
```diff
@@ -0,0 +1,245 @@
...
+        """
+        Estimate multiple season components as well as trend and residuals
+        components.
+
+        Returns
+        -------
+        DecomposeResult
+            Estimation results.
+        """
+        periods, windows = self._sort_periods_and_windows()
```
bashtage, line 126: Somethign is wrong here. You first get `periods` and `windows`, and then right below you access `self.preiods`.  Should you be using these sorted periods below?  If so, could you add a test where you have periods out-of-order and also inorder and ensure the results are the same?

statsmodels/tsa/stl/mstl.py
```diff
@@ -0,0 +1,245 @@
...
+
+        # Remove long periods from decomposition
+        periods = tuple(
+            period for period in self.periods if period < self.nobs / 2
+        )
+        windows = self.windows[: len(periods)]
```
bashtage, line 132: Redefine without using the windows above.

statsmodels/tsa/stl/mstl.py
```diff
@@ -0,0 +1,245 @@
...
+        Returns
+        -------
+        DecomposeResult
+            Estimation results.
+        """
+        periods, windows = self._sort_periods_and_windows()
+
+        # Remove long periods from decomposition
+        periods = tuple(
+            period for period in self.periods if period < self.nobs / 2
```
bashtage, line 130: Redefine without using periods above.

statsmodels/tsa/stl/mstl.py
bashtage: If you really need to sort, should you not just do this in the `__init__` so that there is a consistent canonical representation of models that have the same periods/windows in any order?

bashtage (changes requested):
Probably a bug

statsmodels/tsa/stl/mstl.py
```diff
@@ -0,0 +1,245 @@
...
+https://arxiv.org/pdf/2107.13462.pdf
+"""
+from typing import Dict, Optional, Sequence, Tuple, Union
+
+import numpy as np
+import pandas as pd
+from scipy.stats import boxcox
+
+from statsmodels.tools.typing import ArrayLike1D
+from statsmodels.tsa.seasonal import STL
```
bashtage, line 26: You should directly import STL from where it is defined, not from `seasonal` which is just an API end-point. This can produce cyclic import problems.

statsmodels/tsa/stl/mstl.py
KishManani: The sorting logic is part of the algorithm which is why I was tempted to keep it in `.fit()`. But I think your suggestion also makes sense so I've refactored the logic for handling the various input types and the sorting of `windows` and `periods`, moved it to the `__init__`, and implemented  the suggested new test.

KishManani:
@bashtage once the review is completed would you like me to squash my commits or is that done automatically when you merge? Or is squashing not done for contributing to Statsmodels? Thanks!

bashtage:
@KishManani I'll leave it up to you.  We don't religiously squash before emerge.  I'm happy if you would like to go to 1 or leave it.  I think it looks good and I'll merge soon if there are no objections.

KishManani:
@bashtage I'll leave it as it is then. Thank you!

bashtage:
Thanks for this contribution.  It is great.

PR #8218, opened 2022-04-09
DOC: Change heading levels in MSTL notebook to fix docs

KishManani:
This PR changes the heading levels of the MSTL notebook to correct an issue with the docs (described below). Typos are also corrected and some parts are re-phrased for clarity.

The issue: the [Examples section](https://www.statsmodels.org/devel/examples/index.html) in the docs is currently displaying multiple headings from the MSTL notebook (see image below).

image


bashtage:
Thanks for the follow up

PR #8226, opened 2022-04-14
TST: Remove print statement in MSTL test fixture

KishManani:
This PR removes a print statement in a fixture in the MSTL tests.


bashtage:
Thanks.

PR #8228, opened 2022-04-15
Fix docs std_null twice instead of std_alternative

keesmulder:
Very minor change fixing argument names in the documentation


bashtage:
Thanks.

PR #8253, opened 2022-04-27
DOC: Updated duration.rst to ipython blocks

kryptec:
duration.rst was using the "code-block" tag instead of the "ipython" tag used elsewhere in the documentation. I think this is the reason the python output isn't seen in this document as it is in others.

### DIFF
--- a/docs/source/duration.rst
+++ b/docs/source/duration.rst
@@ -38,6 +38,7 @@ We fit the survival distribution only for the female subjects.
 
 
 .. ipython:: python
+   :okwarning:
 
    import statsmodels.api as sm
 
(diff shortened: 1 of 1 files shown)


bashtage:
The code has to run AFAIK.  Can you add imports and make sure it runs?

bashtage:
@kryptec The doc build failed since the code is missing some things:

https://github.com/statsmodels/statsmodels/runs/6201941766?check_suite_focus=true

kryptec:
Needed to make some extra changes to get all the code to output. I've added it in a separate commit, and going to open another pull request.

PR #8257, opened 2022-04-30
Bug when determine if all rows have same length

zhengkai2001:
Should be the length of each row in all rows (len(row)), but not the number of rows (nrow).

It's an obvious bug, and I don't have the time to actually setup a dev env and run tests. Feel free to close this and fix it yourself.


bashtage:
Looks right but need to check CI

bashtage:
Need a CI fix

bashtage:
Thanks.   Merging despite failure due to other issues in CI.

PR #8259, opened 2022-04-30
DOC: Updated duration.rst to display output

kryptec:
duration.rst was using the "code-block" tag instead of the "ipython" tag used elsewhere in the documentation, causing the python output to not appear. I switched the "code-block" tags to "ipython" tags, added okwarning/okexcept flags so it would run, and savefig flags so the figures would be saved. The code itself had to be slightly modified so sphinx would run it as well.

### DIFF
--- a/docs/source/duration.rst
+++ b/docs/source/duration.rst
@@ -38,6 +38,7 @@ We fit the survival distribution only for the female subjects.
 
 
 .. ipython:: python
+   :okwarning:
 
    import statsmodels.api as sm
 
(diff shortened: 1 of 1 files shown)


kryptec:
@bashtage I'm not sure why this is failing now since I only touched the documentation?

bashtage:
@kryptec The failure is unrelated to you.  Some upstream issues.

bashtage:
I'll merge if the doc build is OK.

bashtage:
DOc build passes, thanks.

PR #8296, opened 2022-05-30
MAINT: Update lint

bashtage:
Update lint list to be stricter

<details>

**Notes**:

  needed for doc changes.
  ```
  ```

</details>

### DIFF
--- a/lint.sh
+++ b/lint.sh
@@ -18,7 +18,7 @@ if [ "$LINT" == true ]; then
     echo "Linting known clean files with strict rules"
     # Default flake8 rules plus the additional rules from setup.cfg
     flake8 --isolated --ignore=E121,E123,E126,E226,E24,E704,W503,W504,E741,E203 \
-        examples/ \
+        examples \
         setup.py \
         statsmodels/__init__.py \
         statsmodels/_version.py \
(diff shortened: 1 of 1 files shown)


bashtage:
Created by

```
import glob
import subprocess 
dirs = glob.glob("statsmodels/**/",recursive=True)
bad = []
good = []
for d in dirs[1:]:
    print(d)
    has_py = glob.glob(d+"*.py")
    if not has_py:
        continue
    try:
        subprocess.run(["flake8","--isolated","--ignore=E121,E123,E126,E226,E24,E704,W503,W504,E741,E203",d],shell=True,check=True,capture_output=True)
        print(f"Good: {d}")
        good.append(d)
    except:
        print(f"Bad: {d}")
        bad.append(d)
    
import os 
def path_is_parent(parent_path, child_path):
    # Smooth out relative path names, note: if you are concerned about symbolic links, you should use os.path.realpath too
    parent_path = os.path.abspath(parent_path)
    child_path = os.path.abspath(child_path)

    # Compare the common path of the parent and child path with the common path of just the parent path. Using the commonpath method on just the parent path will regularise the path name in the same way as the comparison that deals with both paths, removing any trailing path separator
    return os.path.commonpath([parent_path]) == os.path.commonpath([parent_path, child_path])

parents = []
for d in good:
    print(d)
    remove = False
    for p in parents:
        remove |= path_is_parent(p, d)
    if not remove:
        parents.append(d)

all_files = glob.glob("statsmodels/**/*.py", recursive=True)
good_files = []
bad_files = []
for f in all_files:
    d = os.path.split(f)[0]
    good_parent = False
    for p in parents:
        good_parent = path_is_parent(p, d)
        if good_parent:
            print(f"Good file (skip due to parent): {f}")
            break
    if good_parent:
        continue
    try:
        subprocess.run(["flake8","--isolated","--ignore=E121,E123,E126,E226,E24,E704,W503,W504,E741,E203",f],shell=True,check=True,capture_output=True)
        print(f"Good file: {f}")
        good_files.append(f)
    except:
        print(f"Bad: {f}")
        bad_files.append(f)

import pathlib
good = ["examples/","tools/"] + good
good_files = ["setup.py"] + good_files
for l in sorted(set(good).union(set(good_files))):
    print(f"        {pathlib.Path(l).as_posix()} \\")

```

